PR #32, opened 2015-02-23
added overscanning to RasterizingLayer

smremde:
I've tested this with the ShapeFile samples on WPF.

Possibly need to think about adding the ability to disable the delayed update when it is outside of the existing render.


pauldendulk:
Sorry for the delay and thanks a lot!

PR #40, opened 2015-05-25
Dispatch WPF UI updates asynchronously instead of synchronously

scottdewald:
I had a custom ITileSource in which I was retrieving tiles from the local file system instead of remote HTTP requests. I was getting StackOverflowExceptions when the map data was updated until I dispatched UI updates asynchronously.


pauldendulk:
Thanks! I see no disadvantages.

I do not understand why it matters though. InvalidateVisual calls InvalidateArrange internally and  InvalidateArrange will trigger an asynchronously layout update:
https://msdn.microsoft.com/en-us/library/system.windows.uielement.invalidatearrange(v=vs.110).aspx
'Invalidates the arrange state (layout) for the element. After the invalidation, the element will have its layout updated, which will occur asynchronously unless subsequently forced by UpdateLayout.'

Issue #31, opened 2015-02-19
Rotating viewport changes position of points displayed on map


pbanka-rl:
**Steps to reproduce**

1. Run Mapsui.Samples.Wpf
2. Select "Projected Point" demo
3. Zoom in to max level and move the map so that the WGS84 point is visible:
![mapsui_before](https://cloud.githubusercontent.com/assets/10595505/6266765/e73fedbe-b83d-11e4-8bc4-fa174518e7b0.png)
4. Rotate map using slider in the lower left corner

**Actual outcome**
The point changes its position, relative to map (i.e. streets, houses etc) - see the picture:
![mapsui_after](https://cloud.githubusercontent.com/assets/10595505/6266792/1cc7ec8e-b83e-11e4-932b-6bca1bf7ea76.png)

**Expected outcome**
The point should stay in the same place, relative to map.

**Comments**
I noticed that after performing full rotation, the point is again at the same point as before rotationg. I tried finding the cause of this behaviour, but without success...

Map settings are created with:
```
MapControl.Map.Transformation = new MinimalTransformation();
MapControl.Map.CRS = "EPSG:3857";
```
and point layer is created by this code:
```
return new Layer
            {
                DataSource = new MemoryProvider(new Point(4.643331, 52.433489)) { CRS = "EPSG:4326" },
                Name = "WGS84 Point"
            };
```

pauldendulk:
Thanks for your excellent bug report. I can reproduce the issue. Your sample uses projection but it also occurs without projection. Once the map is rotated resizing the map also causes displacement of the points.

pauldendulk:
https://github.com/pauldendulk/Mapsui/pull/46

PR #46, opened 2015-06-15
Resolved issue #31 - Rotation in Xaml Renderer should be working fine now.

kwesolowski:
Fixed transformation stacking bug, causing displacement of rotated symbols.


pauldendulk:
Thanks! I will look at this tomorrow evening.

pauldendulk:
Impressive. These transformations are very hard to get right in this complicated code.

PR #48, opened 2015-07-23
Reduced code duplication and fixed NullReferenceException

kwesolowski:
BoundingBox(IEnumerable<BoundingBox> ) overload had mistake (min/max were not assigned after check). I also cleaned file a little.


pauldendulk:
Thanks a lot @kwesolowski

PR #80, opened 2016-07-22
Update Client.cs

tebben:
According to the WMS spec the width and height are not mandatory parameters for legend style. Previous code was expecting a width and height property for a legend causing some WMS layers to fail during parsing capabilities.


pauldendulk:
Thanks. Nice small pull requests make it easy to merge.

PR #86, opened 2016-08-23
Added ArcGIS token support 

tebben:


tebben:
Added - Refactored brush caching in Xaml rendering and is now also applied to polygons

PR #88, opened 2016-08-25
Update MapRenderer.cs

asad4237:
If exception happens inside RenderFeature function after at-least one child has been added to the canvas,   returning new canvas will leave the previously created (but not yet added to parent canvas) canvas abandoned, that will cause the exception when resuing RenderedGeometry object, because at-least one RenderedGeometry was attached to that abandoned canvas. Returning the same canvas will solve this error, as it will clear this canva's children on next render call.


pauldendulk:
Thanks for your analysis. Sorry for the late response.

PR #99, opened 2016-11-03
Keyboard interaction when WPF MapControl isn't focused

grudolf:
KeyDown and KeyUp events aren't fired when MapControl doesn't have focus. The workaround checks the key status on start of the drag in MapControlMouseLeftButtonDown.


pauldendulk:
Excellent. Thanks for that. I think the original code was written for Silverlight a long time ago and never worked in WPF.

Issue #52, opened 2015-08-18
Mapsui fetcher goes into a loop when tiles are missing


pauldendulk:
reported by @kwesolowski
 https://github.com/BruTile/BruTile/issues/23

pauldendulk:
High priority

pauldendulk:
Added some unit tests. The fetcher does not go into a loop. 

Room for improvement: When null is returned the fetcher will try again after viewport changes. The cache could store something indicating the tile is not available. This is something that should be tested carefully in various scenarios.

PR #102, opened 2016-11-18
Stability for XAML MapControl

### LINKED ISSUE #97: Exception in Xaml.MapControl.StartZoomAnimation
Hello,

using mousewheel to zoom on a map with a very small Viewport/BoundingBox (Height and Width are both 0 because I somehow managed to zoom the map to a single point) causes an ArgumentException:

```
System.ArgumentException: 'NaN' is not a valid value for property 'From'.
   at System.Windows.DependencyObject.SetValueCommon(DependencyProperty dp, Object value, PropertyMetadata metadata, Boolean coerceWithDeferredReference, Boolean coerceWithCurrentValue, OperationType operationType, Boolean isInternal)
   at System.Windows.DependencyObject.SetValueInternal(DependencyProperty dp, Object value)
   at Mapsui.UI.Xaml.MapControl.StartZoomAnimation(Double begin, Double end)
   at Mapsui.UI.Xaml.MapControl.MapControlMouseWheel(Object sender, MouseWheelEventArgs e)
   at System.Windows.Input.MouseWheelEventArgs.InvokeEventHandler(Delegate genericHandler, Object genericTarget)
   ...
```

Tested on Mapsui 0.8.0, current version source is identical.

drwatson1:
I found some more problems and fix them all. 
All these are about stability of desktop XAML MapControl. More over, I think these fixes #97.
I can't reproduce #52, may be it fixes too? Can do you check this?

### DIFF
--- a/Mapsui.Rendering.Xaml/MapRenderer.cs
+++ b/Mapsui.Rendering.Xaml/MapRenderer.cs
@@ -176,7 +176,7 @@ namespace Mapsui.Rendering.Xaml
             }
             catch (Exception ex)
             {
-                Logger.LogDelegate(LogLevel.Error, "Unexpected error in renderer", ex);
+                Logger.Log(LogLevel.Error, "Unexpected error in renderer", ex);
                 return canvas;
                 // If exception happens inside RenderFeature function after 
                 // at -least one child has been added to the canvas,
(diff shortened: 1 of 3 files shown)


pauldendulk:
Thanks drwatson1, but could you create an issue first so we can discuss possible solutions. The fix itself should contain the minimal changes for that issue. I am looking at #97 now and it seems it was already fixed by my earlier changes related to the empty map.

PR #112, opened 2016-12-26
Added Mercator Projection

Kibnet:
Added a class to calculate the coordinates in the Mercator projection.


pauldendulk:
Wonderful!

So now I have to think about the next steps....

pauldendulk:
@Kibnet I may be a tiny bit late with this question 😳 but, which epsg code is this?

PR #131, opened 2017-02-22
Change BackColor, so that RefreshGraphics is called

charlenni:
If you change BackColor of Map, than the map isn't refreshed. This is perhaps a bug.

PS: Sorry, see while creating PR that the idents are not correct. How much ident do you use? And how many spaces are one tab?


pauldendulk:
oh hey, I had not seen this PR. Great, thank you!

PR #142, opened 2017-03-08
Added PropertyName to OnViewportChanged

charlenni:
This could be usefull for handling the event.

### DIFF
--- a/Mapsui.Rendering.Skia-PCL/BitmapHelper.cs
+++ b/Mapsui.Rendering.Skia-PCL/BitmapHelper.cs
@@ -57,12 +57,13 @@ namespace Mapsui.Rendering.Skia
 
         public static void RenderRaster(SKCanvas canvas, SKBitmap bitmap, SKRect rect, float opacity = 1f)
         {
-            // Better for quality. Helps to compare to WPF
-            var color = new SKColor(255, 255, 255, (byte)(255 * opacity));
-            var paint = new SKPaint { Color = color, FilterQuality = SKFilterQuality.High };
-            canvas.DrawBitmap(bitmap, rect, paint);
+            // todo: Add some way to select one method or the other.
+            // Method 1) Better for quality. Helps to compare to WPF
+            //var color = new SKColor(255, 255, 255, (byte)(255 * opacity));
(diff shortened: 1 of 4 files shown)


pauldendulk:
Hiya, Some comments just by looking at your PR.

I removed all the Viewport properties that were propagated through the Map because you had to check a number of specific properties to know if the Viewport had changed. In most cases you just want to know if the viewport has changed. By adding the specific properties in the current setup you get the best of both worlds. You can ignore the specific property and but still only respond to viewport changes (because the are not mingled with other Map properties anymore), or you could listen to a specific property. So, great!

The [CallerMemberNameAttribute](https://msdn.microsoft.com/en-us/library/system.runtime.compilerservices.callermembernameattribute(v=vs.110).aspx) 

> Allows you to obtain the method or property name of the caller to the method.
PropertyChanged is now also called from Transform. So it will by called with the Transform 'property' (perhaps you could check this). This is odd. Within that method there is a property change of Resolution (which will be triggered when set in Transform) and the Center.X/Y is set. This will trigger the property change through OnViewportChanged(). Perhaps this can be resolved by setting the Center, and not just the X and Y separately. Perhaps you can take a look at that.

charlenni:
Thanks for your fast answer.

The CallerMemberName was only for the update of Center a problem. I got always ".ctor" for this, because it was an inline function. So I added here the PropertyName by hand (nameof(Center)). It would be possible to add the property name of which direction of center is changed (X, Y). But I don't think, that this is usefull.

I deleted the call of PropertyChanged in Transform, because it is useless. It is called while changing Resolution and Center. But nothing else changes. Btw. Transform isn't a good name. Translation would be much better to understand :)

pauldendulk:
About the name Translate instead of Transform. In most Matrix libraries Translate, Rotate and Scale are used in a consistent manner. Translate always means moving but not scaling. Here we do both. That is why I did not use Translate. There might be a better name than Transform.

pauldendulk:
Nice way to solve the Center property notification. The solution that I had in mind was to assign the Center property Center = new Center(newX, newY). I think in that case you will have one notification less in Transform. The name(Center) notification is still useful for when X/Y will be set separately.

Issue #206, opened 2017-07-01
LabelStyle.Halo not supported in Skia-PCL or WPF implementations


spaddlewit:
Added a Halo property to the sample project in LabelsSample.CreateLayer():

    featureWithDefaultStyle.Styles.Add(new LabelStyle { Text = "Default Label", Halo = new Pen(Color.Green, 4)});

No effect.

pauldendulk:
The effect between Skia and WPF is quite different. Is the Skia effect what you try to obtain? It is a bit blocky. I like the smoothness of WPF. 

Are you using other libraries with a halo effect? Which?

spaddlewit:
Yes, the Skia effect is what I'm trying to obtain. Essentially a wider stroke of the font. I also fill it, in case you have the alpha set on the font's actual color to ensure there are no gaps.

Both DotSpatial and ArcMap have this halo effect, and I've found that it greatly increases the readability of labels.

pauldendulk:
The difference in how it looks is now quite big. Perhaps there is an other way to implement it in WPF. What if you draw the same text with a bigger font behind the label text?

spaddlewit:
Drawing the font bigger wouldn't work, you'l get something like this:
![image](https://user-images.githubusercontent.com/29378443/27993988-ec785f1c-6481-11e7-9072-bebaa629ccb0.png)

I did some Googling and it seems there are other folks complaining that WPF doesn't contain the ability to stroke a font, and that it's a glaring omission. This stackoverflow has an implementation, but holy crap, it's involved. I don't use WPF, either.
https://stackoverflow.com/questions/93650/apply-stroke-to-a-textblock-in-wpf

pauldendulk:
This issue was fixed by accepting spaddlewit's PR

PR #208, opened 2017-07-04
Hotfix/issue206

spaddlewit:
This if a fix for Issue #206 in Skia, and a partial fix for #206 in WPF.

WPF doesn't have a halo capability built-in, so instead, the text has an outer glow effect to it.


spaddlewit:
The label render tests fail to compare to 'labels.png'. This is expected due to the changes.

PR #210, opened 2017-07-04
Hotfix/issue209

spaddlewit:
Ignores the Left/Bottom attributes from SKPaint.MeasureText. These were generating invalid values for the purpose of how the labels get renderered. It fixed my test strings, and the existing strings appear to be unaffected.

### DIFF
--- a/Mapsui.Rendering.Skia-PCL/LabelRenderer.cs
+++ b/Mapsui.Rendering.Skia-PCL/LabelRenderer.cs
@@ -88,6 +88,23 @@ namespace Mapsui.Rendering.Skia
             backRect.Inflate(3, 3);
             DrawBackground(style, backRect, target);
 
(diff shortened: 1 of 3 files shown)


pauldendulk:
I decided to just include this PR as is, although the visual difference between WPF and Skia is big. I understand this is a useful feature. A better halo in WPF is something for another issue. 

Your fork also contains some of the changes related to font positioning. Could you remove those? Or better, create a new branch with just the halo changes? Then it will show as one clean commit.

spaddlewit:
I think you're looking for the other pull request, #208 ? This one (#210) should be rejected since it's the font positioning one.

pauldendulk:
Not necessary anymore. This is now fixed by updating SkiaSharp and fixing a sign error  (removed minus in offset) in Mapsui.

PR #213, opened 2017-07-10
Remove stylus restrictions in Wpf MapControl (fixes #200)

### LINKED ISSUE #200: [Wpf] No Info event fired with a stylus (and touch)
Hi,

I am currently using the WPF library to develop a map with clickable features on it, using the Info event. However, our software will be deployed on Windows tablets equipped with stylus.
As it is, the WPF lib ignores events if they come from a stylus. This feature seems to be hard-coded on purpose in **Mapsui.UI.Wpf/MapControl.cs**, since `MapControlMouseLeftButtonDown(...)` and `MapControlMouseLeftButtonUp(...)` start with `if (e.StylusDevice != null) return;`.

We traced these lines back to commit b37b211ec3767da71edf68139a2372c50d2689f9, but couldn't understand the reason behind it.
So I tried simply commenting this line in both handlers, while preserving it in `MapControlMouseMove(...)`, and I couldn't notice any side effect.

As the Info event is unusable on tablet right now, and unless we are overlooking a major issue, would you accept a pull request removing this limitation?

PS: since `if (e.StylusDevice != null) return;` is in `MapControlMouseMove(...)` as well, some "CanMoveMapWithStylus" parameter could also be added to MapControl for the developer to decide whether he needs this.

Skymirrh:
See #200 for details.

### DIFF
--- a/Mapsui.UI.Wpf/MapControl.cs
+++ b/Mapsui.UI.Wpf/MapControl.cs
@@ -409,8 +409,6 @@ namespace Mapsui.UI.Wpf
 
         private void MapControlMouseLeftButtonDown(object sender, MouseButtonEventArgs e)
         {
-            if (e.StylusDevice != null) return;
-
             _previousMousePosition = e.GetPosition(this);
             _downMousePosition = e.GetPosition(this);
             _mouseDown = true;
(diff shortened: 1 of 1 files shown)


pauldendulk:
Thanks!

PR #214, opened 2017-07-17
Skia renderer rotation & iOS rotation control

JoeCooper:
I'm using a matrix to control the positioning of tiles.

This is the only way to do this but unfortunately means the world-to-screen algorithm is implemented twice; once in the viewport and again in the Skia raster renderer.

If one is changed, the other will have to be changed to match.

### DIFF
--- a/Mapsui.Rendering.Skia-PCL/RasterRenderer.cs
+++ b/Mapsui.Rendering.Skia-PCL/RasterRenderer.cs
@@ -10,48 +10,74 @@ namespace Mapsui.Rendering.Skia
 {
     public static class RasterRenderer
     {
-        public static void Draw(SKCanvas canvas, IViewport viewport, IStyle style, IFeature feature,
-            IDictionary<object, BitmapInfo> tileCache, long currentIteration)
-        {
-            try
-            {
-                var raster = (IRaster)feature.Geometry;
+		public static void Draw (SKCanvas canvas, IViewport viewport, IStyle style, IFeature feature,
+			IDictionary<object, BitmapInfo> tileCache, long currentIteration)
+		{
+			try
+			{
+				var raster = (IRaster)feature.Geometry;
 
-                BitmapInfo bitmapInfo;
+				BitmapInfo bitmapInfo;
 
-                if (!tileCache.Keys.Contains(raster))
-                {
-                    bitmapInfo = BitmapHelper.LoadBitmap(raster.Data);
-                    tileCache[raster] = bitmapInfo;
-                }
-                else
-                {
-                    bitmapInfo = tileCache[raster];
-                }
+				if (!tileCache.Keys.Contains (raster))
+				{
+					bitmapInfo = BitmapHelper.LoadBitmap (raster.Data);
+					tileCache [raster] = bitmapInfo;
+				}
+				else
+				{
+					bitmapInfo = tileCache [raster];
+				}
 
-                bitmapInfo.IterationUsed = currentIteration;
-                tileCache[raster] = bitmapInfo;
-                var destination = WorldToScreen(viewport, feature.Geometry.GetBoundingBox());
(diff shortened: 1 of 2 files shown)


pauldendulk:
Rotation in WPF also works correctly for rasters. Is this implementation different from yours?

(sorry, but I am slow in replying during the summer months)

pauldendulk:
Hey, rotation works in the Skia mode Mapsui.Samples.Wpf as well. That makes it much easier for me to work with it. I can now also compare them.

This is the Shapefile sample in WPF
image

This is the Shapefile sample in Skia:
image

JoeCooper:
I'm a little confused. Are you saying that rotation is already implemented? Because it wasn't working and there was no code for it. Why does it work in a sample?

pauldendulk:
It was not working in for Skia but when I checked out your fork it actually did work for Skia in WPF. There is a rotation slider in that sample. I was surprised by this because I assumed you had only worked on iOS. This is probably the result of your changes to the raster renderer.

JoeCooper:
"There is a rotation slider in that sample. ... I assumed you had only worked on iOS"

I can explain.

The core renderer needed to be updated. As you can see, this update applies to all platforms.

What I did specifically on iOS was amend the MapControl class to push state onto the rotation member based on user touch.

I will do the same to Android at some point.

I have a newer version, in a different branch, which I will send you later, which significantly improves performance of the iOS MapControl. If necessary, again, I will make a corresponding change on Android when I come to it. (At the moment I'm doing something with Itinero so I'm a tad blocked.)

pauldendulk:
Android can be done later. I can now use the WPF sample to take a look.

There is a slight displacement if I compare WPF with Skia in the Shapefile sample in Mapsui.Samples.Wpf. Looking at that now.

pauldendulk:
The displacement become bigger (in pixels) if I zoom in. What I do is:
- Start Mapsui.Samples.Wpf
- Select the Shapefile option
- Use the slighter at the top right to rotate
- Then alternate between 'wpf renderer' and 'skia renderer' in the dropdown at the top left

pauldendulk:
About the RasterRenderer. If I remove the bufferOne and bufferTwo matrices and just use the single matrix defined at the top the graphical behavior is the same. Concat takes a ref which suggests it uses the input value but it does not, so the value of the ref matrix does not matter.

JoeCooper:
"If I remove the bufferOne and bufferTwo matrices and just use the single matrix defined at the top the graphical behavior is the same ..."

Noted. I played it safe because the documentation was very sparse.

"There is a slight displacement if I compare WPF with Skia ..."

I have no idea. In my project I'm using other rendering code which is also using the Viewport and it correctly lines up with what MapsUI is rendering. Is the displacement there without my commits?

What kind of displacement is it? Does _jitter_ at all? (Could be due to the 32-bit pipeline ...?)

pauldendulk:
The problem ~~is~~ was only with the raster layers. The displacement was only visible when you zoomed in. It turned out to be the RoundToPixel method. It makes no sense in a transformed canvas.

JoeCooper:
You're welcome! I'm accustomed to dealing with the matrices due to some experience in game development.

I will be sending more PRs in the future. In a separate branch I've rewritten the MapControl class to better use auto layouts and use an OpenGL backend with significantly better performance. But I need to finalize some things before I bug you with it.

charlenni:
Do you say, you want to remove Skia and replace it with an OpenGL view on all platforms?

pauldendulk:
I assumed he meant the OpenGL implementation of Skia. It there is the [SKGLView](https://developer.xamarin.com/api/type/SkiaSharp.Views.iOS.SKGLView/) and the [SKGLLayer](https://developer.xamarin.com/api/type/SkiaSharp.Views.iOS.SKGLLayer/). But maybe I'm wrong.

JoeCooper:
"SKGLView" is correct. Really significant effect. Should work on everything.

PR #216, opened 2017-07-27
Make InvokeInfo work with Touch

gklamm:
As @Skymirrh said in #200, we had an issue with touch events which doesn't fire Info event for map item (IFeature).

We figured out that when manipulations are enabled, the manipulation processor handles the TouchDown event and captures the touch device. The touch event flow is never promoted to mouse events, so the Info event is not fire.

To fix that, we just enabled TouchUp to fire Info event on features. And making sure it doesn't fire after a manipulation gesture.


pauldendulk:
Thank you very much. Sorry for the long delay. I will have low activity during the summer months.

PR #222, opened 2017-08-30
Reorganized iOS MapControl using automatic layout

JoeCooper:
I was having problems with the MapControl failing to resize correctly, and other issues.

I've refactored MapControl and the associated AttributionView using UIKit's constraint based automatic layout system.

Also, I'm using the OpenGL-based Skia view which is significantly faster on device.


pauldendulk:
Absolutely fabulous!

PR #227, opened 2017-09-14
Map Disposal

JoeCooper:
This might be a little more controversial than the last PRs.

The code was written to dispose of Maps when they're replaced.

This caused a problem in a scenario where, using MvvmCross data bindings, sometimes (a race hazard) the same Map is assigned to the MapControl twice. Because of the call to Dispose, the map is ruined in the course of the second assignment.

Another solution would be to check if the incoming new map is the same as the extent map, however, I don't think it makes sense for the MapControl to permanently break an object it doesn't own (the Map). What do you think?

Another change here, in Map.cs, is I've removed the layer clear from Dispose(). The purpose of dispose is to release resources not governed by garbage collection. If you dereference the Map, the layers will be dereferenced when the Map is collected. If the Map is not in fact dereferenced than the client is not done with it.

One quirk is that AbortFetch is no longer called, either. I tried putting this in there, but this seems to permanently break TileLayers.


pauldendulk:
ah, this is complicated stuff. I am hesitant to change this it like this because I know everything is working in the projects I work on. Not calling AbortFetch has caused a nasty memory leak in an earlier project. 

I am aware that this is not perfect but my thoughts were to create a centralized fetcher that is not part of the layer. This way the Layers do not need to be disposed.

An alternative it to stop the fetcher thread when it is done. The waithandle there is what is preventing the dispose.

pauldendulk:
Perhaps I should start by reproducing your problem. You replace a Map then put the original Map back on the MapControl. Is that how it works?

JoeCooper:
Something like this:

var map = new Map();
// add some layers
mapControl.Map = map;
mapControl.Map = map;

The second assignment will break the map.

That code looks stupid but if I'm using MvvmCross data bindings than it does sometimes perform multiple assignments (never mind the details) and I've never noticed because I've never had a control which breaks the input.

"my thoughts were to create a centralized fetcher that is not part of the layer. This way the Layers do not need to be disposed."

It sounds like the fetch event does need proper disposal. I'm going to look into that and bring this up with you again.

JoeCooper:
Are you still supporting CF?

pauldendulk:
Compact framework? No. Why, are there still relics in the code base?

JoeCooper:
Yes, I've run into a relic. I'll factor it out.

JoeCooper:
New commit.

- MapControl asks the outgoing map to abort fetch.
- TileLayer does not release the fetcher when asked to abort. Rather, it only asks the fetcher to abort.
- Map no longer implements IDispose.

I thought the changes would be more substantial but it appears that TileFetcher is already designed to reboot if you use it after you shut it down.

JoeCooper:
I'm thinking about the zombie thread problem.

As-is, the Map is only disposed if a new Map is assigned. If for any reason the MapControl is destroyed, you can get a zombie thread.

Given that TileFetch can launch its thread whenever it needs to, how about having the thread shut its own thread down whenever it doesn't have any work to do?

This way a dereferenced TileFetcher is guaranteed to finish up. WDYT?

pauldendulk:
> Given that TileFetch can launch its thread whenever it needs to, how about having the thread shut itself down whenever it doesn't have any work to do?

Yes, that is what I was hinting at with my 'alternative' mentioned above. An early version actually worked like that but I rewrote it (and introduced a bug) because I did not want to create a thread again and again. Perhaps there are more efficient ways nowadays using async.

JoeCooper:
"Perhaps there are more efficient ways nowadays ..."

Yes. I'll do this tomorrow and explain.

JoeCooper:
OK. I've made the loop cancel itself.

You're already using TPL to launch the methods ("Task.Run(...)"). TPL enqueues operations on the thread pool. No thread is created. The ThreadPool has its own notion of maximum threads, so none of the code relating the maximum thread counts is necessary.

"Perhaps there are more efficient ways nowadays using async."

Async/await is syntactic sugar over TPL. So with respect to the cost of starting threads, it is irrelevant.

It _is_ relevant when dealing with IO bound operations like fetching tiles from a local source. Asynchronous IO is implemented at the _driver_ level. If properly utilized than an operating waiting on IO from storage or the network will not occupy any space in the thread pool.

To use it, both BRUTile _and its dependencies_ need to be updated. I will look at BRUTile shortly because I think there might be a thread safety issue in the MBTiles implementation and I'll see if it can offer TPL interfaces.

JoeCooper:
Quick followup; looks like the SQLite thread safety issue is already accounted for in BRUTile and it looks like SQLite.NET doesn't properly support asynchronous IO operations so never mind all that.

JoeCooper:
Nevermind w/r/t BRUTile thread safety & async. I see it's already accounted for and that SQLLite doesn't properly support asynchronous IO.

JoeCooper:
I've added another commit. A bug fix in MapControl; incoming maps need to have their viewport size set, because if the map is assigned _after_ a resize event, the maps never have their size set and do not render.

JoeCooper:
New commit; the previous changes caused problems on iPhone 7 Plus because the OS has both a "logical" and "native" scale factor and the correct one must be used or else the visuals will not scale correctly and will not match the touch interface.

JoeCooper:
Thanks!

pauldendulk:
Thank you :)

Now we are on it I am trying to make it simpler. I had not touched this code in a long time. I now removed FetchOnThread class and are looking into some other things.

One small question on this like:
```
if (_fetchLoopTask == null || _fetchLoopTask.IsCompleted || _fetchLoopCancellationTokenSource.IsCancellationRequested)
```
About the check on cancellation request. This means that if cancellation is requested the new thread is started. Should it not be the other way around?

JoeCooper:
"This means that if cancellation is requested the new thread is started. Should it not be the other way around?"

Nope.

The extant thread carries a reference to the old cancellation token. If cancellation is requested on that token, than the thread will, eventually, die. It is either already completed or it is about to complete. Therefore, a new thread with a new cancellation token is warranted.

pauldendulk:
Ah thanks, I understand now. 

I am looking into some further improvements and will be thinking a bit more out loud here. Feel free to comment or ignore :). 

The current implementation fires all the requests on the first iteration of the loop. The loop will continue to recalculate while those tiles are fetched. This is not efficient in the scenario where the viewport does not change because the calculation was complete on the first iteration. To prevent this you could leave the loop right away (or don't start a loop at all). This works well for step wise navigation (e.g. if you click a + or - to zoom in). But for continuous navigation (e.g. dragging, or mouse wheel with animation) this is not efficient because a new loop will be started on every change. 

What I am thinking of now is to use some kind of EventHandle.Wait inside the loop which is broken if all tiles have actually arrived and after that the loop should end. This solution will allow garbage collection although with a delay that lasts as long as the tile fetch lasts.

pauldendulk:
Another issue with the current implementation is that all tiles are requested at once. In many situations most of those tiles will not be relevant anymore because the user has zoomed further. This can also cause flooding of the server in some cases. I would like to restrict the number of outstanding requests at each moment to prevent this

JoeCooper:
"This may impact some of the changes I now made in master."

Why is the cancellation token now called "globalCancellationToken" in its parameters?

This isn't exactly my business but I find the code very confusing because it's broken into separate methods which are only called once, so reading through a procedure involving jumping up and down the page a lot. I know people like to "break up long functions" but it's like reading a book with all the pages shuffled. In this case it makes it particularly hard to reason about because some methods are called in the context of a worker thread and some will be on the Event Dispatch thread and you have to flip up and down to figure out which. On what principle is the fetch tiles procedure broken into these fragments?

I wouldn't bring it up but you put "[MethodImpl(MethodImplOptions.AggressiveInlining)]" on everything. Seems like if we did some smart inlining by hand, the code would be more linear.

"because a new loop thread will be started on every change."

No threads are started. The reference to the loop method is given to the thread pool which will act on it when it comes to it.

A new job is enqueued _if and only if_ the last job _is finished or finishing_. If there is an ongoing fetch loop, and another change is accepted, the currently running loop is going to see the changes on the next iteration.

Unless, of course:

"all tiles are requested at once"

If all the tiles are immediately queued for loading, the fetch loop will exit pretty much immediately; it might as well not be a loop at all.

"This can also cause flooding of the server in some cases"

In practically all cases, launching all tile fetches simultaneously should _harm_ performance.

In the case of a network, all the requests are going over the exact same path to the same endpoint. In the case of local IO, you're clobbering the same storage device with multiple requests on different physical locations. On devices with significant seek time (hard drives) this can make it disproportionately slow. It's not uncommon for the server to be using a hard drive.

Call for one tile at a time. Get tile; reassess; get tile; reassess; ... What do you think? Shall I implement this?

By the way. The tile source could be amended to take cancellation tokens. I think the SQLite-based MbTiles implementation will not be able to respect them but sources working over HTTP will definitely be able to respect them.

pauldendulk:
I am working on tests to clarify some things but am running into some unexpected issues, so it is taking a bit longer. Hold on.

pauldendulk:
The unexpected issue I ran into was [this](#232). It is not related to what was discussed above. This issue was already present in my original code, although not as likely to occur because the loop had a _Wait_ in it. It took some time because it was hard to grasp what was going on.

pauldendulk:
The first issue I was trying to point out is this:
https://github.com/pauldendulk/Mapsui/blob/master/Tests/Mapsui.Tests/Fetcher/TileFetcherTests.cs#L105

The loop logic has now changed a bit because of my fix above but the problem is similar. In the first iteration all tiles are fired up with Task.Run. Whether these threads are actually active depends on the threadpool as you describe but what matters is that they are in progress and the main loop is not waiting for them. In the mean time the main loop keeps iterating. In the unit test above the main loop iterated 18902 times. This made GetTilesNotInCache take 30% of all CPU according to the Visual Studio performance profiler without adding any value.

I am now looking into a solution where the calculation for missing tiles is only done when needed.

JoeCooper:
Just pull one at a time. Pull a time and check again what’s needed. Do the tile fetch synchronously in the loop; don’t check about the cache again until a tile is loaded.

There probably isn’t any benefit to asking the same storage source to provide multiple different tiles simultaneously. In fact it’s probably harmful on most.

pauldendulk:
An update. I turned everything upside down with the latest commit. 

I created a FetchDispatcher. The responsibility of this class is to control the order in the which the tiles are fetched. The TakeFetchOrder takes the next item to fetch from the queue. The CompleteFetchOrder method is there to return the tile that is fetched. The class has no threading at all.  It needs to be thread save though because it is called from background threads. 

There now is a FetchMachine with a number of FetchWorkers that are responsible for the tile requests on worker thread.  It needs a pointer to the FetchDispatcher. It starts its fetching when Start is called and stops if there are no more items returned from FetchDispatcher (or Stop is called, which is not often). It manages the worker threads but other than that it is simple.

The advantages of this solution. 
- The responsibility of the classes is more clear.
- TileFetcher had a thread loop that itself created threads. There is now just one level of threads.
- There is no more tracking of the number of threads that are running. There is just the initial set of FetchWorkers.

pauldendulk:
Some further changes. FetchWorker now just receives and Action and is now more generic. The FetchOrder class can now be removed. CompleteFetchOrder can now be made private.

JoeCooper:
If it's generic, isn't it just a clone of the ThreadPool that .NET already offers?

JoeCooper:
In fact, it explicitly running on top of the .NET thread pool. So it's tying up a .NET worker thread permanently in order to provide a worker thread. What am I missing?

pauldendulk:
This is something I had in mind. Making the FetchMachine more generic step by step might eventually make it replaceable with some standard .net component, or some component offered as nuget package. This would simplify the code. The current refactoring makes this an easier step.

However, the ThreadPool underlying Task.Run()  is not a sufficient replacement. 

An important aspect of the FetchMachine is that it restricts the number of threads active on this specific task. This is a requirement. This is important because the number of threads should not only be constrained by the capability of the local machine (this is something the ThreadPool can manage) but also by the tile service that is being queried. 

We have been working with tile services for quite a while and what we have seen many time is that firing an unlimited number of requests causes performance problems. Not just a linear delay (more tiles requests amounts to more delay) but a full collapse. There are many points in the line that could cause this collapse: threadpool starvation, http traffic, some OS limitation, throttling on the local network, or just a blockade of the tile service itself as a form a flood protection. This is the reason I want to control this at the source.

The previous TileFetcher limited the number of concurrent requests to mitigate this problem. These problems still occurred however when many layers were used that each fired a number of requests. This is the reason issue #1 was introduced. 

Another reason I what to control the number of threads is the fetcher should not take too much resources from the UI. I want the UI to remain fully responsive even if huge amounts of data are requested in the background.

Yet another is that I do not want to cancel requests for tiles that are already in progress. Canceling a thread is one thing, but canceling a request to a tile service is something else. The tile service is already working on the request and returning the tile, while the thread cold ignores this response because it is cancelled. When you drag the map around the Viewport changes at every move, this would trigger cancellation all the time.

The bottom line is, this is not just about threads, it is about requests to a tile service.

JoeCooper:
Is there any reason at all to give more than one request to the tile service at a time? Is there an empirical or theoretical advantage?

JoeCooper:
I should clarify, the code I wrote was definitely not designed to perform cancellation on map move. This was to be done in response to the Map getting discarded.

pauldendulk:
About cancellation. That comment was not specifically aimed at your code but more general at a threadpool mechanism that is regulated with cancel. I know there are alternative to do this. The FetchMachine is such an alternative. 

My rewrite was primarily a rewrite of my own code. One that I had planned in 2014 (see #1). I do these refactorings by moving code into smaller methods and then moving these methods around between classes. The threading mechanism you introduced still lives on in the FetchWorker.

JoeCooper:
Of course, of course. By all means, my code is not sacred. Your head is deeper into this than mine is. I was just slightly embarrassed cause I thought you thought I had it cancelling a lot. That’s not a technical argument so I shouldn’t have polluted the thread with that. Sorry.

About hammering the server, threads, etc.

Are you familiar with the details of async & await?

Do you know that when you “await” on a compatible IO process, the thread pool isn’t blocked?

From what I read, SQLite.NET doesn’t support this but the HTTP operations and file streams definitely support this.

You would need to amend the BRUTile interfaces to support this but it’s not too hard.

Sorry if it’s a patronizing question. I’m not clear on which parts of the framework you’re familiar with.

– Joe

pauldendulk:
Yes I am familiar with async/await and use it in other projects. It does indeed not block the calling thread but Task.Run does not block the calling thread either. 

The problem I try to prevent is that many heavy background threads cause a general overload of the entire system. This will be noticeable in the UI. It will become slower and sluggish. 

BruTile/Mapsui is older than async/await, but actually I don't see much need to rewrite to async/await. Right now the worker threads are in a background thread so there is no direct need for it. End users will not be confronted with it. They will just use the layers that do the asynchronous calls for them.

pauldendulk:
Some more clarification about the need for this mechanism. Geodan's [Phoenix](https://www.geodan.nl/solutions/phoenix/) is based on Mapsui. In Phoenix users can add layers as they wish. When adding too many layers the user interface becomes sluggish. This impacts the touch experience. If the map is responsive the users' experience is really that of touching. When the delay between what they do and the response of the application becomes too big it is not pleasant to work with anymore. My aim is to let user interaction always have priority.

PR #257, opened 2017-12-15
Add Triangle to available SymbolTypes

Skymirrh:
Hi Paul! Ellipses and Rectangles are cool and all, but we hit a wall at some point due to having so many objects for different things on the map in our app. So we thought we'd add Triangles to spice things up a bit.

2 things to note:
- Since we only use Xaml and Skia, I only could add and test rendering on WPF apps.
- I couldn't figure out how to update the `vector_symbol_symboltype.png` for the `RenderPointsWithDifferentSymbolTypes` unit test.

### DIFF
--- a/Mapsui.Rendering.Skia-PCL/PointRenderer.cs
+++ b/Mapsui.Rendering.Skia-PCL/PointRenderer.cs
@@ -67,23 +67,28 @@ namespace Mapsui.Rendering.Skia
         private static void DrawPointWithVectorStyle(SKCanvas canvas, VectorStyle vectorStyle,
             SymbolType symbolType = SymbolType.Ellipse)
         {
-
-            var halfWidth = (float)SymbolStyle.DefaultWidth / 2;
+            var width = (float)SymbolStyle.DefaultWidth;
+            var halfWidth = width / 2;
             var halfHeight = (float)SymbolStyle.DefaultHeight / 2;
 
             var fillPaint = CreateFillPaint(vectorStyle.Fill);
 
             var linePaint = CreateLinePaint(vectorStyle.Outline);
 
-            if (symbolType == SymbolType.Rectangle)
+            switch (symbolType)
             {
-                var rect = new SKRect(-halfWidth, -halfHeight, halfWidth, halfHeight);
-                DrawRect(canvas, rect, fillPaint, linePaint);
-            }
-            else if (symbolType == SymbolType.Ellipse)
-            {
-
-                DrawCircle(canvas, 0, 0, halfWidth, fillPaint, linePaint);
+                case SymbolType.Ellipse:
+                    DrawCircle(canvas, 0, 0, halfWidth, fillPaint, linePaint);
+                    break;
+                case SymbolType.Rectangle:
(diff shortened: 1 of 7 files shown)


Skymirrh:
As a side note, while testing things on the WPF sample, I noticed a strange behavior: Rendering was systematically done twice. This is easy to reproduce:

- Launch WPF sample in Debug mode.
- Switch to "Test samples" combo box.
- Put a breakpoint at [line 528 (`RenderWpf`) in MapControl.cs](https://github.com/pauldendulk/Mapsui/blob/a4f1600d19982f9ba6340eba8dc097215ee98cc1/Mapsui.UI.Wpf/MapControl.cs#L528).
- Click on sample "2".
- `RenderWpf` is called twice.

Not sure if that's normal or not, just wanted to let you know.

pauldendulk:
Fabulous!

I created an issue for WPF rendering done twice:
https://github.com/pauldendulk/Mapsui/issues/258

pauldendulk:
You can view the test samples if you start Mapsui.Samples.Wpf and select 'Test Samples' from the second dropbox. The triangle had a circle background. This was because the layer itself has a default symbol. This is a common thing that confuses users. I removed the layer symbol.

pauldendulk:
About fixing the unit tests. The rendering tests compare images. If there are no unwanted changes the images should be the same. This allows merciless refactoring with a safety net. 

If your intent is to change the rendering output the image generated by the test should be copied over the original (after you established the output is what you wanted). 

There is a description here:
https://github.com/pauldendulk/Mapsui/wiki/Rendering-Tests#1-unit-tests

Skymirrh:
>The triangle had a circle background. This was because the layer itself has a default symbol. This is a common thing that confuses users. I removed the layer symbol.

Cool, I wasn't sure if it was intended or not. In our app, we indeed reset the layer style to avoid this issue. If the goal is to avoid confusion on the user's part, may I suggest directly editing [`BaseLayer`](https://github.com/pauldendulk/Mapsui/blob/39f871a5729d44b5517841996eacd720cad995ff/Mapsui/Layers/BaseLayer.cs#L32) so that it uses `null` instead of a `new VectorStyle()` by default?

>If your intent is to change the rendering output the image generated by the test should be copied over the original (after you established the output is what you wanted).

Indeed, that was what I wanted to do. I forgot about checking the wiki: had I seen it, I would have updated the images as well... Welp, I'll know for the next time 😄

Skymirrh:
In 919043ebd2602c145cc7e35b590bf59c8e1b1437, you remove unused `x` and `y` parameters: this is a mistake on my end.

Lines 142-144 are:
```csharp
            var top = new Point(0, -circumradius);
            var left = new Point(sideLength * -0.5, inradius);
            var right = new Point(sideLength * 0.5, inradius);
```

They should be:
```csharp
            var top = new Point(x, y - circumradius);
            var left = new Point(x + sideLength * -0.5, y + inradius);
            var right = new Point(x + sideLength * 0.5, y + inradius);
```
That way, `DrawTriangle` could be called at an alternative point of your choice, like `DrawCircle` and `DrawRect`.

Not utterly important since you don't seem to need to do that for now, but just for conscience sake I wanted to mention it 🤐

PR #263, opened 2017-12-19
Fix ios rotation and add rotation snapping feature

radderz:
Fix iOS rotation, it used to not reset the last rotation causing rotation jumps on mutliple rotates.

Added new feature to enable rotation snapping. when the current rotation < UnSnapRotationDegrees it will not rotate, once gone passed this point it will jump to the rotation it would have been then start rotating as normal. ReSnapRotationDegrees allows a point for where the rotation will resnap to the 0 degree rotation point where rotation transititons to < ReSnapRotationDegrees.


radderz:
For testing the rotation snapping, I'd recommend something like this where it has a min rotation of 30 degrees before it rotates and doesn't snap till it gets back within 5 degrees of 0.

 new MapControl()
            {
                AllowPinchRotation = true,
                UnSnapRotationDegrees = 30,
                ReSnapRotationDegrees = 5
            };

radderz:
@pauldendulk Just did another fix in this pull request for the two finger zooming for android which was broken from your last checkin, missing brackets.

pauldendulk:
Thanks for the PR! I have some requests and questions. 

Could you do the iOS rotation fix in a separate PR? (and could it be it has been fixed in the meantime?)

I think you are developing on VS on Windows which defaults to different formatting than VS for Mac. If so, could you redo the PR on the most recent version of the code, it now is using the windows formatting so it is easier to see the relevant changes.

Rotation Snapping is now implemented for Android, but not for the other platforms. Is this correct?

For what purpose do you use rotation snapping? 

Do you know other map components that have this rotation snapping feature?

Could you give a brief explanation of the UnSnapRotationDegrees and ReSnapRotationDegrees?

radderz:
Hi Paul,

Google maps and windows 10 uwp default map app both do it to differing degrees. The purpose is to stop unwanted rotation when pinch zooming.

I implemented on both iOS and android. They were the only two with rotation implemented that I have used.

I am on holiday in Korea at the moment so will have to wait a week or two. The iOS rotation and Android zooming changes should be incorporated asap though as they are quite broken.

Cheers
Marm

radderz:
Ugh,

Mark*

pauldendulk:
The iOS rotation issue was indeed serious. I fixed it. See this issue. https://github.com/pauldendulk/Mapsui/issues/269

radderz:
Hi Paul, great, did you see the missing brackets change on the last check in of this pull request? Android pinch zooming was broken after a recent change. It would zoom but was incorrectly centred.

pauldendulk:
ah, What a useful remark! I was just now looking at the android centering issue and could not figure it out. I though I missed out skia scale somewhere.

pauldendulk:
One more question.  Do you develop on Windows or Mac? Because they format differently which messes up the pull requests.

pauldendulk:
Fixed Android centering https://github.com/pauldendulk/Mapsui/issues/270

radderz:
Hi Paul, I develop on windows visual studio

pauldendulk:
Okay, no problem then. I check if the iOS and Android MapControls.cs files are now fully in line with Windows formatting. Now, have a nice holiday!

radderz:
Hi @pauldendulk,

I think the way you are using double.Nan doesn't seem quite right, as it means you skip the first rotation tick, as you'll see on the way I had done it, I get the angle of the two presses when the second press occurs as the touches begin where the number of touches is now 2, so that becomes the reference angle before any "move" event has fired. 

Whereas you have this:

`if (!double.IsNaN(_previousRotation)) // only rotate relative to the previous rotation 



 
+				        { 



 
+				            _map.Viewport.Rotation += rotation - _previousRotation; 



 
+				        } `

Which whould cause the first move to be skipped. This wouldn't be that obvious without a more involved view where there are less frequent callbacks due to the cpu being loaded.

pauldendulk:
I see. Your solution takes into account the very first move. That's better. I made a quick commit to deal with this (did not test, got to run off again). Perhaps you can see if this is what you mean. https://github.com/pauldendulk/Mapsui/commit/8b53915978a410ff357fd0257f19914947e28546

radderz:
@pauldendulk the same is true for the previousRadius and previousCenter if you want to keep it consistant. You could then also remove the previousTouchCount checks.

I'll look to do the RotationSnapping Feature based on the upstream master soon. Do you understand the feature based one what I said above? I can enable it on the samples so that you can see it if you'd like.

pauldendulk:
UnSnapRotationDegrees: I understand this as a kind of threshold before the rotation is actually taken into account. Perhaps threshold could be in the name. ThresholdRotationDegrees.

ReSnapRotationDegrees: This is an angle to snap to so that that it is not fully continuous. Specifically handy if you want to go back to the un-rotated situation. But I would assume it snaps to any rotation that is a multiple of this setting, right? In that case I would call it just SnapRotationDegrees. Is there a name for this in the Google Maps API?

radderz:
Hi Paul, I have created a new pull request for this #273.

Thats not quite right, the resnap is the threshold of degrees off the 0 rotation point that a snap back to 0 degrees would occur. i.e if resnap was 5 degrees, and the user rotated from +10 degrees to +4 degrees, it would transition back to 0 degrees of rotation, and then it wouldn't unsnap until the UnSnap threshold is met again.

PR #273, opened 2018-01-08
Reimplement RotationSnap off upstream master, fix map crash from invo…

radderz:
…keInfo and build errors for VS Windows


radderz:
This feature makes it so that rotation must be be more significant in degrees than the UnSnapRotationDegrees property on the MapControl class (iOS and Android Only) before rotation will occur. This stops unintended rotation when just trying to pinch zoom. ReSnapRotationDegrees allows for a catch of up to ReSnapRotationDegrees away from the north direction to resnap back to 0. 

If these are left as 0 they will not make any changes.

This is a reasonably common feature on other map applications like the built in map application in windows 10.

radderz:
@pauldendulk - This is the newly written code, the samples show the thresholds now, if you'd like the name changed to something else let me know. The threshold is Asymetrical with Unsnap going away from the 0 rotation and Resnap going towards the 0 rotation.

Issue #277, opened 2018-01-11
Disparity between InvokeInfo on iOS/Droid/Uwp


radderz:
On MapControl.cs, iOS and Droid invoke Map.InvokeInfo on TouchEnd and MotionEventActions.Up which could have been a drag/pinch guesture rather than a tap.

It looks to me that this is a tapped like function to enable the map to show info etc on tap of a feature. However if you tap on a featre and drag the map it also invokes the info. On UWP it doesn't do this as it has a proper tapped event.

This means on a screen with features that you want to be clickable, just doing a screen drag has a chance of clicking the feature causing unintended consequences if it is being used as a selection mechanism that then centres on the clicked feature.

Either we need a Tapped hander for a feature or a InvokeInfo to only trigger if the guesture is a tap.

pauldendulk:
Yes, I agree. There are some options to resolve this. Perhaps we can use platform specific tab gesture. Gesture can be very performance consuming though. So perhaps we should add some checks ourselves. We should perhaps also share the code for this over all MapControls.

What do you mean with?
> a Tapped hander for a feature

radderz:
Sorry that is a typo, I was meaning if the InvokeInfo is working the way it is supposed to then we could have an additional "Event Handler" for Tapped specifically. 

I think the guesture recognizers shouldn't be too bad if we do some initial checks around if the finger has moved much and the time between down and up and if there is ever 2 touches to ignore it.

pauldendulk:
I was thinking about some of the built-in gesture recognizer of the platforms. I know for WPF it used to take a lot of processing for little effect. If we do those checks ourselves it won't be such a problem. 

btw, The MapControls are gradually growing to become more complex. I have to start thinking about refactoring and reusing more code in common components. As a consequence things will become more aligned.

radderz:
Yes there is a lot in them that could be sub classed as an abstract class with common parts shared. I have done this with a lot of functionality on top of the map and if you create a generic interactions conversion that then just references the base class it can help reduce the code a lot.

i.e. having the mouse down/up/drag etc all done in the sub class with all the relavent scalings pre done should reduce the duplication of code.

[...]

PR #289, opened 2018-01-18
iOS + Droid - Rewrite pinch to zoom and rotation code, add tap handlers for infoHandler

radderz:
Pretty big rewrite of touch handling as there were loads of edge cases where random moves and rotations would occur depending on which finger was released. iOS Simulator didn't show these errors much but on a real device they were many. I have done a fair bit of testing and the tap and touch interactions work without any errors that I can find now.

The only thing that isn't handled properly is a third finger's rotation only looks at the first 2 fingers, if you don't like the use of the new C# features with the tuples I can change this into a struct. The rotation, center and radius are all now using identical code. Only the touch interactions and how to get the previous is different.

@pauldendulk - Should the iOS control have the viewport limiter in the touches moved as it is there in the android one? I have done this but I don't use the limiter in the project.

### DIFF
--- a/Mapsui.UI.Android/MapControl.cs
+++ b/Mapsui.UI.Android/MapControl.cs
@@ -64,14 +64,8 @@ namespace Mapsui.UI.Android
             
             _gestureDetector = new GestureDetector(Context, new GestureDetector.SimpleOnGestureListener());
             _gestureDetector.SingleTapConfirmed += TapGestureHandler;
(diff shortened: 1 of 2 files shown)


radderz:
This fixes #277

pauldendulk:
Thanks you very much. This is great work and it is a step towards further refactoring the MapControls.

pauldendulk:
I love that new tuple syntax. It allows writing proper (pure) functions without having to use those out or ref parameters.

Issue #292, opened 2018-01-22
Rotation occurs based on the center point of the map rather than the center point of the current pinch that initiates the rotation


radderz:
This is less obvious but the rotation property rotates about the center of the Map so when pinch zooming and rotation away from the center point of the Map the location of the touches gets further from the actual map points that originally were touched.

This causes weird pinch location error as you move around in one big motion

pauldendulk:
Thanks for reporting. That could be a tricky one.

radderz:
I have done it in the new pull request #293 , I also added UWP rotation.

PR #293, opened 2018-01-22
Add rotation to Viewport.Transform

radderz:
Add rotation to Viewport.Transform. This will make the rotation occur around the centerpoint of the tranformation.

Added UWP rotation capability.


radderz:
This fixes #292 and Adds rotation to UWP

radderz:
@pauldendulk - Would be good if we can have this as part of 1.3.2-beta too and I'll see if the change breaks anything major.

pauldendulk:
The distance check was removed. This prevented a sudden jump of the center of pinch. Why is this not necessary anymore?

In the past it showed when you were panning with two fingers and then lifted one of those fingers.

radderz:
Hi Paul, its not need because I am not using the method that was used before which wasn't correct. This method looks at the translation x and y which doesn't matter if a single finger or multiple are down as its the relative movement that is respected. I have tested it on both mobile and pc and it works with all touch scenarios. 

I got the rotation about the right point, pan and zoom all working on the 3 platforms with these changes. And again managed to share the logic.

pauldendulk:
@radderz I released 1.3.2-beta2 with your changes. https://www.nuget.org/packages/Mapsui/1.3.2-beta2

PR #294, opened 2018-01-23
WPF Touch Standardisation + Rotation

radderz:
WPF now uses the same zoom, pan, rotation logic/code.


radderz:
This is the final changes to match #289 and #293 so that all platforms work the same from a touch interaction perspective.

PR #303, opened 2018-01-31
ScalebarWidget and ZoomInOutWidget

charlenni:
Added a ScaleBarWidget and a ZoomInOutWidget. Refactored Widgets drawing. Added WidgetTouched event handling.


pauldendulk:
I intend to merge this and do some rework on it. Could you summarize what the current issues are? What alternative choices you are considering?

charlenni:
If you have static Widgets like your attribution, you don't need any information from Map or Viewport. But if you have Widgets like zoom buttons, you need a reference to Map or Viewport. I encountered, that Viewport isn't enough (doesn't redraw map, if resolution is changed), so I changed to use Map. 

Perhaps Widgets should always know the Map, they belong to. Could made while adding them to Map.Widgets.

Another question is, how to handle platform specific handling of WidgetTouched events. You attribution needs for every platform a special handler, which is now hardcoded in MapControl. This is ok, but perhaps there is a better solution.

radderz:
If you are going to have zoom buttons then shouldn't they zoom while being held in rather than on tap so a widget may not work?

charlenni:
@radderz Checked this, but couldn't find an app, that this does. If you use maps.google.com on desktop, there is only a touched event and no hold down for zooming. Google app on Android hasn't any zoom buttons by default and maps of iOS doesn't have them too.

Widget get only called by WidgetTouched, so I haven't any other events. If we like to have this, we should make a redesign of Widgets.

Btw. the zoom buttons were only a example of how this could be done. Like Paul said, this could be done by normal controls on each platform.

@pauldendulk A system, where users could inject their own Widgets would be very interesting.

pauldendulk:
I am not sure I understand your remark @radderz

pauldendulk:
@charlenni What is the Alignment for? Could HorizontalAligment be reused?

charlenni:
With Alignment you could place the text relative to the Scalebar. There is only a Left, Right and Center, but no Position or Strech option. So it is more an TextAlignment

pauldendulk:
Thanks a lot @charlenni!

I will do some further rework. Perhaps you can respond to the things mentioned in the review.

charlenni:
Sorry for the big PR. I'm not very firm in Git, so it is each time a tricky thing. Try to think about this next time.

I don't have ReSharper, but try to use your style.

I don't know anything about WPF. It was the first time I ever worked with it. And this only, because you said, you want to have a solution for all renderers :). Perhaps it is possible to position controls in a Grid at any position. I didn't found it. So for me Grid seems to more limiting positioning than it helps. If you don't like it, you could remove the free positioning system.

The refresh issue of ZoomInOut should be solved. Problem was, that if you change Viewport.Resolution, the map isn't redrawn. I changed it by using Map.NavigateTo() to change the resolution.

I used OutlineTextBlock for the Halo effect. I didn't find the possibility to draw text with different stroke and fill colors. This is easy on Skia, but isn't possible in WPF. So I added a ready made control for this. If you have a better idee, something other is possible too.

Any changes to system files are not work from myself. I use VS 2017 and if there are changes, than VS made them. Sorry for that.

Try to write something in the Wiki.

charlenni:
In the Wiki, should I make a new page?

pauldendulk:
I have to think about that.

pauldendulk:
About the wiki. Everyone is allowed to edit the wiki, so you could add a new page yourself.

pauldendulk:
About OutlineTextBlock. Ah, I already noticed your Halo looked nicer than the Halo we have on the WPF label.  See discussion here: https://github.com/pauldendulk/Mapsui/issues/280. Perhaps we could use the OutlineTextBlock there too.

pauldendulk:
I'll continue discussion in the issue you created since the PR is now closed.

Issue #306, opened 2018-02-04
Zoom/Pan animation and Fling/Flick guestures


radderz:
Hi Guys,

This is an enhancement that I'd like to know if anyone would like to have as a setting/option.

I have built zoom/pan animation with acceleration UWP/Droid/iOS, I have also added "fling/flick" type deceleration. I did this in xamarin.forms using their Animation system but it shouldn't be a big deal to port that back to native since they have open source code which could be used as a guide.

I guess the question would be, is this a desired feature or would others prefer to do this on their own? Most Map tools include things like double click zoom/animation and fling type guestures. Assuming we leave it as an option feature of the map controls then I can't see a negative in including this in the Mapsui project. The way I wrote the fling/double click zoom it is similar to how I did the MapControl interations where it converts to non-native values and then uses shared code. Double click zooms about the point that you double click.

This could then be animated on 16ms ticks with a variety of different types of acceleration, i.e sin in sin out which is good for center point changes where it will accelerate till the half way point then decelerate to stop at the new center. Zoom is just be linear (but could be a setting), similar to the WPF scroll zoom. I have found this makes the map interaction feel much more natural, especially with the fling/deceleration. It just gets complicated to make them all settings rather than just included as it'd be a fair bit of if/else ifs etc to get the right combination based on the settings. I have found a weakness to the animation when MANY (i.e 1000+) bitmap symbols are on the map, but having this enabled/disabled will give the user the option.

There are some rules with the interactions that I have followed
- New touches cancel all animations
- All new animations cancel all other animations (i.e selecting a new center while a zoom or new center animation is going will cancel the other animations and start anew as the new animation.)

Cheers

charlenni:
I would like it. All things, that Mapsui bring in direction of a complete map api would be very good.

pauldendulk:
Sorry for my late reply.

Yes this is a great feature to have. I have some experience with this myself and it is something that can have a lot of nasty edge cases. The two rules you mention at the end are exactly the kind of rules to can help to keep it simple and prevent these edge cases.

My experience was that it better to do the animation at the level of the Mapsui renderer, and not at the level of the UI framework (which might seem attractive because it often has some animation functionality built in).

I used a cubiceaseout for feature animations:
https://github.com/pauldendulk/Mapsui/blob/41a20c5e80ea038efec47a0f19154a282cde1821/Mapsui/Layers/AnimatedFeatures.cs#L13
[...]

Issue #305, opened 2018-02-03
1.3.2 breaks Android taps on feature icons


Sebastian1989101:
It seems like the latest beta updates are breaking the ability to click on icons on Android. Many users report that they are not able to open the popup dialogs correctly since the last update which include 1.3.2-beta2 on Android (they appear after a double tap on a feature). I notice the same problem in the debugger that the click is often not registered correctly. 

It works on iOS pretty well with 1.3.2-beta2 but Android seems to be broken with it.

radderz:
I'll have a look as I used the guesture recognizer for it, its supposed to use a confirmed single tap for the info. As I was planning to add double tap to zoom which I have implemented for my own project.

radderz:
@Sebastian1989101 - Info is invoked on a guaranteed single tap in android, on ios, if you double tap it will call the invoke info twice.

Sebastian1989101:
@radderz that is exactly what I expect and what happens with Mapsui 1.3.1. But with 1.3.2 on Android the Info event is not invoked properly. Only after a few action on the map it might be invoked. It starts working again as soon as I downgrade to 1.3.1 without changing anything else. So it is diffenetly a problem with 1.3.2 of Mapsui.

radderz:
We use 1.3.2 beta 2 and touch interactions are reliable. Which beta are you using?

Sebastian1989101:
@radderz Beta 2 and it is unuseable on Android until you pan around or zoom. Than it works for a few clicks. As already mentioned it works as soon as I downgrade to 1.3.1 without any problems. Also 1.3.2 works on iOS without any problems.

Sebastian1989101:
The double tap is implemented within in my app. But the event is not even called once (don't matter how often I tap the feature icon). My guess is that the hitbox for the features is broken because it works sometimes (in my case ~20%). 

After the first call I store the feature identifier and if the same comes a 2nd time I open a Popup. That is what I mean with double tap. But currently the event is not invoked with 1.3.2. Maybe @pauldendulk has an idea what changed between 1.3.1 and 1.3.2 what could explain this different behaviour.

radderz:
[...]

PR #307, opened 2018-02-05
Enable invoke info on double tap

radderz:
Fix for #305 

Enabled double taps to also trigger info/widgets.

I think info/widgets needs to have a setting way to mention how many taps occured and a way for the info to return if it is handled. As double tap on the map is a common way to zoom which I may be implementing for issue #306. So if we have to have double tap info handling, then it should enable zoom if no item or widget was clicked, and some items may only be single or double click subscribed rather than both.

Any ideas on how we can improve the way the widget/info works to enable the above?


charlenni:
I would prefer a solution, where each call of an event handler return a bool, if this event is handled or not. If it is handled, than no other event handler is called.

This should be done by a variable in the event arguments. If there is more than one handler attached to an event and we solve this by a function with return value, than only the last handler could return this value.

pauldendulk:
@charlenni Is this based on other software where this is used?

pauldendulk:
How badly do we need a beta of this?

charlenni:
I find the problem when painting zoom buttons over map attribution. If you than touched on the buttons, the hyperlink was called, even it was drawn under the buttons.

With this in mind and radderz question, I searched for solutions. There are two: event handler with returning value or event argument mit return value. And if you don’t know, if more than one handler is attached to the event, the second is the better one.

radderz:
@charlenni - Yes I was thinknig the same, I was thinking of adding both the number of taps that caused the info event, and also a handled boolean. Then if I implement double tap zoom, if it is not handled by the info then it can still do the zoom instead. I'll do this shortly and create a new pull request for this and you can see if it matches what you were thinking.

@pauldendulk - are we able to make it easier to create listed pre-release versions? Is there any reason not to create an auto publish mechanism off a trigger?

pauldendulk:
I created an issue to improve pre-releases:
https://github.com/pauldendulk/Mapsui/issues/310

PR #318, opened 2018-02-16
Extending Skia renderer to respect PenStrokeCap and PenStyle

charlenni:
Respect PenStrokeCap and PenStyle in Skia render. Perhaps it is a fine adjustment for line width in PenStyleExtension.

Styles should be refactored. They are not very consistent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ project.lock.json
 /Tools/VersionUpdater.pdb
 /Tools/VersionUpdater.exe.config
 /Tools/VersionUpdater.exe
+/Mapsui.UI.Forms/Images/Pin.svg.bak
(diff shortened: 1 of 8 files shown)


pauldendulk:
Thank you very much!

charlenni:
Yes, more or less. You could delete it.

What do we with the Opacity. @Kortdag is right, when he said, that ithis is inconsistent. Rename Opacity to BitmapOpacity will break old apps. Perhaps use Opacity when Color.A is 255?

pauldendulk:
Let's discuss this further in the issue.

radderz:
I mentioned what to do re opacity in the issue, its pretty easy, you just apply it to the color that you are using to draw the non bitmap symbol vs the color specified, i.e multiple the alpha of the color by the opactity.

charlenni:
If I‘m right, Paul made a PR.

PR #320, opened 2018-02-18
Added Geometry Circle

charlenni:
Added Circle to Skia and WPF renderer, but only Skia works. There are problems with the transformation on WPF, that I couldn't solve.


radderz:
What problems are you having with the transformations?

radderz:
This looks good, I can have a look at the transformation once you explain the issue you are having, 

NOTE: Depending on the order of acceptance with #321 there will need to be an opacity added to this or I'll have to add it in my pull request for enabling opacity on the drawing of the circle.

charlenni:
The problem is, that the transformation in CircleRenderer

    path.RenderTransform = new XamlMedia.MatrixTransform { Matrix = GeometryRenderer.CreateTransformMatrix1(viewport) };

let the circles "vanish" :) The commented transformation works, but isn't used when you move the map.

And be carefull: the circles have all same radius. The different sizes is a projection result.

charlenni:
@radderz Added your changes

pauldendulk:
Thanks for adding the opacity changes. I can look at this tonight.

pauldendulk:
Got this to work with WPF. They have different size though, I can't explain this yet.

Issues I had:
- In determining the circle size I do not correct for resolution because this is done by the transformation.
- I do not set path.StrokeDashArray because this made the solution extremely slow. There is bug in there.

image

charlenni:
Different size to Skia or different size at different places?

charlenni:
And thank you for looking in that. I‘m happy to hear, that this now work on WPF too.

pauldendulk:
A more serious issue is that the solution assumes a spherical mercator projection. Mapsui should work for all projections.

pauldendulk:
The size in skia differs from the size in wpf.

charlenni:
Ok, I take the radius in meters. So I had to convert this meters to a length in world coordinates. Is there a way to do this without knowing the projection? Or should I use another point on the circle instead of a distance?

pauldendulk:
I'm afraid it needs projections and we don't have this implemented to work out of the box. There is a ProjectionSample. bIt explains how to use projections but users need to configure this to work with projections at all. Perhaps we should set this up to work out of the box. It will only work for spherical mercator. For other projections circles won't work.

pauldendulk:
btw, your branch also include the svg code. This makes it hard to review.

pauldendulk:
Ideally the circle would be drawn as separate points that are each of them transformed individually. They won't be circular anymore in the picture. 
http://planet.qgis.org/planet/tag/projection/

charlenni:
Sorry for the Svg code. I thought cherry-picked the Svg code out of the circle code. Git isn't my friend ;) I try to get better.

pauldendulk:
This is the code that generated the image above

```

using Mapsui.Styles;
using System;
using XamlMedia = System.Windows.Media;
using XamlShapes = System.Windows.Shapes;
using XamlPoint = System.Windows.Point;
using XamlColors = System.Windows.Media.Colors;
using Mapsui.Geometries;

namespace Mapsui.Rendering.Xaml
{
    public static class CircleRenderer
    {
        public static XamlShapes.Shape RenderCircle(Circle circle, IStyle style, IViewport viewport,
            SymbolCache symbolCache)
        {
            var path = CreateCirclePath(circle, style as VectorStyle ?? new VectorStyle());
            var matrixTransform = new XamlMedia.MatrixTransform { Matrix = GeometryRenderer.CreateTransformMatrix1(viewport) };
            path.RenderTransform = matrixTransform;
            
            if (path.Fill != null)
                path.Fill.Transform = matrixTransform.Inverse as XamlMedia.MatrixTransform;

            path.UseLayoutRounding = true;
            path.IsHitTestVisible = false;

            return path;
        }

        private static XamlShapes.Shape CreateCirclePath(Circle circle, VectorStyle style)
        {
            // The SL StrokeThickness default is 1 which causes blurry bitmaps
            var path = new XamlShapes.Path
            {
                StrokeThickness = 2,
                Fill = ToXaml(style.Fill)
            };

            if (style.Outline != null)
            {
                path.Stroke = new XamlMedia.SolidColorBrush(style.Outline.Color.ToXaml());
                path.StrokeThickness = style.Outline.Width;
                //!!!path.StrokeDashArray = style.Outline.PenStyle.ToXaml();
                path.Tag = style.Outline.Width; // see #linewidthhack
            }

            // Get current position
            var position = Projection.SphericalMercator.ToLonLat(circle.X, circle.Y);

            // Calc ground resolution in meters per pixel of viewport for this latitude
            double groundResolution = Math.Cos(position.Y / 180.0 * Math.PI);

            // Now we can calc the radius of circle
            var radius = circle.Radius / groundResolution;

            path.Data = new XamlMedia.EllipseGeometry
            {
                Center = new XamlPoint(circle.X, circle.Y),
                RadiusX = radius,
                RadiusY = radius
            };

            path.Opacity = 1;

            return path;
        }

        private static XamlMedia.Brush ToXaml(Brush brush)
        {
            return brush != null && (brush.Color != null || brush.BitmapId != -1) ?
                brush.ToXaml() : new XamlMedia.SolidColorBrush(XamlColors.Transparent);
        }
    }
}

```

pauldendulk:
In WPF I use matrixes for positioning.  This makes it harder to understand (especially the counter scaling needed for lines). In skia we might also need to used matrices to improve performance.

pauldendulk:
oh, I have pushed my changes to your repository. Did not know I had access. https://github.com/charlenni/Mapsui/commits/master

charlenni:
I now removed the Svg update from code by reverting the commit. And than changed the files with conflicts.

The different sizes with Skia and WPF were because of a bug. I assumed, that the radius is the diameter and divided it by 2. Corrected this now.

I added again the PenStyle, but removed the dash in the samples. It should now work correct.

Yes, circles aren't correct for bigger sizes. Should be ellipses with 3 different radii (one to west-east, one to south and one to north). But I assume, that circles are used for much smaller distances, so that this isn't relevant. and if someone would like to make bigger circles, she/he should do it by lines.

pauldendulk:
I have put some thought into this PR and I am afraid there is an serious issue with the functionality that can not be easily resolved. 

If you specify a circle in meters you have to calculate from the current coordinate system to meters. To be able to do that you need transform the coordinates. The Distance and Contains methods of the Circle Geometry are not correct in it's current form. To fix this the Circles' Distance and Contains methods will need to do some transformation. So the Geometry library should take a dependency on a projection library. This is possible, there are some geometry libraries that work like this, but it not what I want to do. The main reason is that NTS does not work like this and I intent to migrate to NTS at some point. In the NTS there is no circle geometry. See also [this](https://github.com/NetTopologySuite/NetTopologySuite/issues/212) question.

It is still possible to allow some kind of support for circles but it should be done without a Circle geometry. Instead there could be some helper methods that create a circular polygon, with some parameters that define the precision. There have been some posts in the issues here with such a solution.

charlenni:
Ok, I now removed the transformation from code. With this, users have to calculate correct radius by themselfs. If you move the circle on the map, you have to change the radius (if it is mercator projection).

And I'm sorry to say this, we have the same problem when using the ScaleBarWidget. There I use the same conversion. So we have to think about this again.

Btw. are there any other maps that we use, that don't use mercator or spherical mercator projection?

So the next step should be, that we should think about using transformations. Where should they take place? Is it possible to set a transformation for a layer and all features are transfered into Mapsui coordinate system automagically?

charlenni:
Ok, there isn't a Map.Transformation and it couldn't be one, because the transformation is layer dependent, isn't it? So each layer has its own transformation and all features on/in this layer should be in coordinates of this coordinate system. E.g. if I have a sherical mercator map, than I could create a layer above it with spherical mercator projection and than features could added with spherical coordinates. Is this correct? Or should the native coordinate system of Mapsui (1 on the aquator is 1 m in real, but not above or below) be used for all things and users had to convert all points to this coordinate system by themself?

pauldendulk:
I am going to close this PR. I think there should not a circle geometry in Mapsui for the reasons mentioned above.

Perhaps it will be possible in a later stage to add custom Geometries that are accompanied by their custom renders. I will built something like that for Widgets. So, once that is done it might make sense to add something similar for geometries. Such a custom Geometry could choose to support just SphericalMercator and nothing else, or other features that are only suitable for the specific needs of a project.

charlenni:
Thank you for looking into this PR. I try to implement this as polygon in Forms.

PR #321, opened 2018-02-18
Extend IStyle to include an opacity for all styles

radderz:
Moved the opacity of styles all the way down to the IStyle interface. Then implemented the opacity for all the things. If you think I missed one let me know, I added a basic sample, that could probably have more types of things added to it so that it proves the opacity model better.


radderz:
NOTE: Depending on the order of acceptance with #320 there will need to be an opacity added to this pull request or be added to #320 for enabling opacity on the drawing of the circle.

PR #322, opened 2018-02-19
Added Svg support for SymbolStyle

charlenni:
Added support for Svg instead of bitmap for SymbolStyle. Is implemented for Skia fully working and drawing in vector style. No conversion to bitmap. For WPF, only the functions are capable of managing this. It's missing a function to convert Svg images in bitmap images in SymbolCache.cs, line 23. Could be made by Skia. If WPF should draw Svg in vector style too, than there must be more coded changed, because SymbolCache.GetOrCreate() has return type BitmapImage.
Added a relative option to Offset. Up to now, Offset was absolut in pixel. With relative option, Offset could be from -0.5 to +0.5 in both directions and is multiplied with the image width/heigth when using. This should work on Skia and WPF. Hope I found all occurences ;)


charlenni:
Ok, there could be Svgs out there, that don't work. But some work :D

screenshot

radderz:
Hi @charlenni, have you played around with doing a performance test? I.e have 200+ svgs on the map during a zoom/pan operation compared with first converting the images to bitmaps and using the cache? More interested in how performant the SVG renderering is as it would be nice if it has similar performance for outside of mapsui too.

charlenni:
I added a sample with 200 tigers 🐯. On WPF it is horrible slow. On Skia you could use it with a little bit of patience. If you zoom in and have under 30 tigers, it works normal. Smaller Svgs like a Pin is ok.

Didn‘t check with conversion to bitmap.

charlenni:
I try to add the possibility to use Svg or convert the Svgs to bitmaps with an extension function (.ToBitmap() ), so the user could decide, which way to go. Do you think, this is a good way? Or should there be a flag in SymbolStyle (like UseBitmap) to do this?

pauldendulk:
I was thinking about some flag but with a flag there would still be a need for a conversion method. So it may be a good first step.

pauldendulk:
Wrt eris . We should use hardware accelerationh4ardwarematige is

charlenni:
Äh, why did you closed it? Any problems with this PR? Did you take a look on this PR?

pauldendulk:
Sorry, I was pocket typing.

pauldendulk:
What I was trying to say when the train suddenly arrived: We should use hardware acceleration on all platforms and do our performance testing with this as a requirement.

charlenni:
Ok, I expanded the sample with the tiger as png file and a new line in SvgSample. There you could decide, if the tiger should be used as bitmap or as Svg. The rest is the same: 200 tigers on a OSM map.

When I move the map with bitmap tigers, my Android device need 10 s to render the new map. If I do the same with Svg, the device needs 2 s to do the same. Even if I change the size of the png to a quarter of the origanal size (to 512x529) then it needs 5 s. Perhaps bitmaps are faster, if the image is much smaller. For that, I used the pin images. 2000 on a OSM map. Ok, you see nothing of the map but ;). The Png takes 2 s to move, the Svg 1 s.

Over all, it seems that Svg is faster than bitmaps. But try it by your own.

radderz:
@charlenni I really want to give this a good test but I have been really busy with work, if this is still pending next week I'll have a good go at testing it.

radderz:
I have tested this on droid, wpf and ios. Looks good perf wise, made a couple minor adjustments.

radderz:
This is good work @charlenni, I have noticed the performance in WPF is worse compared with Skia but a large margin, but this may be a limitation of the WPF renderer for Xaml rather than an issue in this implementatoin.

charlenni:
Thank you for your work. Perhaps the Svg2Xaml converter is very slow. But I'm not interested in the WPF renderer ;), so I don't want to look deeper into this problem.

PR #328, opened 2018-02-28
Added polygon fill patterns to Skia renderer

charlenni:
Btw. Wpf renderer doesn't respect opacity for bitmap fill style.


charlenni:
Thank you for looking into this PR and for accepting it.

PR #331, opened 2018-03-02
Improved Shapefiles demo

GeoDirk:
Added in ThemeStyle example to show how to color individual features based upon their name.
Added in a hover points layer example to show to dynamically add in points with features
Added in an example info layer that points to the countries shapefile that works on the mouse down


charlenni:
Did you your changes in Branch vectortiles? If yes, you should change it in master.

GeoDirk:
@charlenni & @pauldendulk 

Ah...yes, I'll move them over to the Master Branch and create a new pull request.

PR #335, opened 2018-03-07
Cleared some parts of Map.CRS and its usage

charlenni:
I worked on the projection problems I see. If there are even more, I can do this too.

What I didn't changed, was, that Map.Info() results (and surely more others) aren't converted back to coordinate system of layer, which would be correct.


charlenni:
It would be nice to discuss a little bit about projection in Mapsui as this isn't very clear. Or I don't understand it ;)

We should get a line in this, so users understand, what's going on.

pauldendulk:
Closing this. The projection parts needs a bigger rewrite than can be handled with this pull request.

PR #339, opened 2018-03-12
Add Rotate method to Polygon and LinearRing

dirkdokter:


charlenni:
Wouldn't it be better to rotate around the center of the geometry instead of Point(0, 0), if the center isn't given? Something like

    public LinearRing Rotate(double degrees)
    {
        return this.Rotate(degrees, this.GetBoundingBox().GetCentroid());
    }

pauldendulk:
I agree, that would more sense. Although that could be seen as added functionality. When this is added a following logical step would be to add Rotate to IGeometry and implement it for MultiPolygon and MultiLineString as well. 

I think in NTS and most other spatial libraries the rotation is not a member of geometry but an operation done on some geometry. Since we already have Rotation and Quad, Point and BoundingBox I don't see a big problem in adding it to IGeometry.

At some point Mapsui will migrate to NTS (it will be a major release) so this will be replaced. It will take a while though.

Just curious, why did you need this functionality?

dirkdokter:
True, rotating about the bouding box center / polygon centroid / outer-ring-vertex-centroid would make more sense. I'll probably change that this weekend.

As for why we need this functionality: we are using Mapsui for aerial photography flight planning software. The end-user draws a polygon region-of-interest. Our software performs some operations on this polygon, to generate some other geometry. Those operations require the polygon to be rotated to another frame (direction-of-flight frame).

PR #340, opened 2018-03-12
Add option to disable panning with mouse

dirkdokter:
This could be useful for allowing dragging of features (points, polygons, ...).

Related to https://github.com/pauldendulk/Mapsui/issues/148#issuecomment-287816100

### DIFF
--- a/Mapsui.UI.Wpf/MapControl.cs
+++ b/Mapsui.UI.Wpf/MapControl.cs
@@ -228,6 +228,7 @@ namespace Mapsui.UI.Wpf
             _map.ViewChanged(true);
         }
 
+        public bool AllowMousePanning { get; set; } = true;
         public bool AllowPinchRotation { get; set; }
         public double UnSnapRotationDegrees { get; set; }
         public double ReSnapRotationDegrees { get; set; }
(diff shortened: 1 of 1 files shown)


pauldendulk:
Thanks for your PR! Is it the intention to make a distinction between panning with mouse and panning with fingers?

There currently is a ZoomLock which has a similar purpose but only for Zoom.

radderz:
I'd say a PanLock is a pretty standard feature that woudl be good for both mouse and touch related maps. 

This would be best if it was done in a standard way with the three interactions of pan/zoom/rotate.

dirkdokter:
At this moment actually I don't need to distinguish between mouse panning and touch panning. So a PanLock would make sense.

pauldendulk:
I added this manually as everything had changes after charlennis PR. I also renamed AllowPinchRotation to PinchLock and ZoomLocked to ZoomLock and moved that last one to the shared project. ZoomLock and PanLock are now only implemented in WPF.

charlenni:
But @pauldendulk, aren‘t this things, that belong to the Viewport? That‘s nothing, that have to do with the MapControl. It says, if the Viewport could be changed or not. Or does it mean, that zoom and pinch events should be send any longer?

pauldendulk:
Well it should still be possible to change the Viewport from code. You could image an application where the map zooms to a certain point after an item it selected, while panning is not allowed.

On the other hand I do think this kind of functionality could be shared in some component outside of the MapControl. The same is true for the ViewportLimiter. This should perhaps be a state full object and a public member of the MapControl. It could also all be part of the Viewport, although the viewport was intended as a simple thing to indicate the viewport state, similar to the POV in games. So just, center, resolution, rotation.

PR #348, opened 2018-03-16
Moved MapControl to shared project

charlenni:
Moved all platform dependent MapControls to a shared project basis, so that same code could be reused on all platforms. No changes for namespaces.

Perhaps it could be moved more code to shared project. This wouldn't tested.

ReSharper has no errors ;)


charlenni:
No luck today :( First Paul changed things while I was creating the PR and than AppVeyor doesn't respect the conditional compiler symbols from the Wpf project.

Ok, ok, it isn't correct. I set the conditional compiler symbols only for debug mode, not for release.

pauldendulk:
This is a huge PR. Thanks a lot!!

The number of changes really scare me though. I going to merge anyway because if I postpone it it will only be harder. 

Some questions?
- Did you test on all platforms?
- Why did you introduce the new PCLTimer? Why not the timer in Mapsui.Utilities?
- You added new gestures? I see swipe and fling? Could you elaborate?
- The new event args are in the shared project. I would prefer to keep anything that is not partial in Mapsui core. Is there a reason to put them there?

charlenni:
First of all, I tried to change as less as possible on existing MapControls. I added the shared project, added it to all Mapsui.UI projects and removed functions and properties, that are the same on all platforms. It is much code, but was easy.

>     Did you test on all platforms?

I tested it on all platforms except iOS. The Mac isn't running and it takes much time. Normally each time I start it, I had to do much updates ;)

> Why did you introduce the new PCLTimer? Why not the timer in Mapsui.Utilities?

Argh, I didn't see it. So I will change my code to yours and than we can remove PCLTimer.

> You added new gestures? I see swipe and fling? Could you elaborate?

No, I added only functions to handle such events common for all platforms. As I said, I didn't change anything in the native MapControls. But if @radderz want to make such events, than there are handlers for it.

> The new event args are in the shared project. I would prefer to keep anything that is not partial in Mapsui core. Is there a reason to put them there?

No. The only good reason is, that they belong to MapControl and that I didn't want to change things all over the whole solution. They could transfered to Mapsui, no problem.

I think, there shouldn't any problems, but up to now, there are no advantages. We should now check each MapControl to see, what we could move to the shared project, because it is used on all platforms.

charlenni:
Ok, removed PCLTimer from Forms implementation. Now you could delete it with any of the next releases.

dirkdokter:
Great work! I started to change the WPF MapControl to use the methods of the shared MapControl, see https://github.com/dirkdokter/Mapsui/commits/mapcontrol_shared_wpf_wip (I did not create a PR yet, as it's still work-in-progress).

I noticed that you put OnTouchEnd/OnTouchStart/... methods in the shared MapControl, with the positions of the touch points as input. However, the current WPF MapControl uses the Translation/Rotation/Scale as available in the WPF OnManipulationDelta event arguments. 

The OnManipulationDelta event arguments also contain "Manipulators", which are basically touch points. For now, I just used those as arguments for OnTouchMove. My reasoning for this is that it should lead to consistent behaviour between WPF/UWP/Xamarin, if the OnTouchMove functionality would be changed in the future. An alternative would be to create a OnTouchMove overload, which accepts Translation/Rotation/Scale instead of touch points.

The disadvantage of using those manipulators/'touch points' (instead of Translation/Rotation/Scale) is that the 'inertia' features of WPF don't work any more: when OnManipulationDelta is fired due to inertia, the Manipulators property is empty. Which makes sense, as no manipulators are touching the screen at that point anymore.

charlenni:
The OnManipulationDelta event only exists on WPF. The other platforms (iOS, Android, UWP or Forms) don't have such an event. I started with this changes on Forms. There I used the native events of Skia Forms View. All events in the shared project are only proposals. They could be changed. But if you add high level functions like OnManipulationDelta, you should create versions for the other platforms too ;)

I see, that you use RefreshGraphics() and InvalidateCanvas(). Both draw the screen new. So, if you call RefreshGraphics, remove InvalidateCanvas.

PR #354, opened 2018-03-20
Changed drawing of lines in Skia renderer

charlenni:
Added a clipping function to check, if a part of a LineString should be drawn or not. It's much faster than without this (around 10-15 times).


pauldendulk:
Thanks again @charlenni. 

In what ways did you test performance? Just run the sample?

At some point we need to start using https://github.com/dotnet/BenchmarkDotNet for our graphics.

Clipping could also decrease performance since it is an extra step and Skia probably already does some clipping itself. Did you look at that?

Could we clip depending on some criteria?

charlenni:
I used the samples and checked, how long LineString.Draw/Polygon.Draw needs. It is not a very good test, be I think it's enough.

Yes, clipping needs allways some more time than only drawing. But I assume, that this is here not the problem. We draw only this features, that are within the viewport. And the Skia clipping did use much more time than the hand made. Don't know why, but found this.

I try to make a sample, where we draw many random lines on the map. Than we can test it.

charlenni:
Ok, did some more test.

It seems, that Skia has problems with dashed lines longer than 2^32 pixels. If you have longer lines or polygons, than the lines aren't dashed any longer. They are stroked solid. If they are shorter, than you they need horrible much time to draw. Because of this, the polygon sample was so slow. If you zoom in, the lines get stroked instead of dashed and than the whole is much faster.

I asked a question in Skia forum. Lets see, if there any good new ideas.

charlenni:
So, now I have an answer. See [here](https://groups.google.com/forum/#!forum/skia-discuss).

This in mind, the problem is, that Skia dashed lines are drawn by a two level  function:
1. All dashed lines and holes are calculated, even if the line is never inside the clip rect. If you use {30, 10} instead of {10, 10} this process needs only half the time. That's what I get.
2. Dashed lines are drawn respecting the clipped rect.

Now there are some speed tweaks for this:
1. Accept this PR and live with wrong dashes and speed reduction for many features are inside of clipt rect (e.g. vector tiles)
2. Say, that dashed lines for such huge features are special cases and dropp this PR

Any comment?

@AliFlux Did you tested the performance problem with Features without dashes? Does it still persits?

pauldendulk:
Ah, great to see that you got proper feedback from skia itself. How are you c++ skills? I did a lot of c++ ones but that was long ago.

Top priority for me is to avoid performance problems. So I prefer clipping. I don't mind the dash phase problem. Perhaps there are other solutions as well, like don't use dashes at all when the polygon is for some percentage outside of the bbox. Or maybe it should not be a percentage but an absolute distance... whatever generates a lot of dashes. The main argument for another strategy would be better performance.

charlenni:
Yes, performance is everything :)

I have to look into this again, when I'm further with the VectorTiles project. In the moment it is to slow. It delivers only around 3 fps. Should be faster. Perhaps I should made an own renderer for this. Or I could change the existing Skia renderer, so that it is fast enough. I will see. In any case, I will try to improve the Skia renderer. But because I don't know much about Wpf, I don't improve this renderer.

And sorry, but there will some more PR you have to check. Mostly more than one line PRs ;)

pauldendulk:
I think the best way to improve performance here is to use matrix transform. This is what I have used in WPF. In skia this is done with SKCanvas.SetMatrix. Or you could use Translate, Scale and Rotate, like in the skia PointRenderer:
```c#
private static void DrawPointWithSymbolStyle(SKCanvas canvas, SymbolStyle style,
       Point destination, float opacity, SymbolType symbolType = SymbolType.Ellipse)
{
            canvas.Save();
            canvas.Translate((float)destination.X, (float)destination.Y);
            canvas.Scale((float)style.SymbolScale, (float)style.SymbolScale);
            if (style.SymbolOffset.IsRelative)
                canvas.Translate((float)(SymbolStyle.DefaultWidth * style.SymbolOffset.X), (float)(-SymbolStyle.DefaultWidth * style.SymbolOffset.Y));
            else
                canvas.Translate((float) style.SymbolOffset.X, (float) -style.SymbolOffset.Y);
            DrawPointWithVectorStyle(canvas, style, opacity, symbolType);
            canvas.Restore();
}
```

pauldendulk:
In my recent commits I added IViewport.ToSKMatrix extension method and applied this skia polygon rendering. It should have a noticeable effect on performance. It should be relative straight forward to implement this for lines. I am not sure about the situation with points, it is more complicated, partially uses matrices and will benefit less (because drawing the bitmap is the most time consuming). 

Creating the transform matrix is just a few lines of code but it is hard to get this right. I'm happy it works correctly now. When applying the matrix I need to multiply the lineWidth with the IViewport.Resolution.

Another big performance gain will be to store the SKPath that is returned from ToSkia in the RenderedGeometry. The RenderedGeometry is meant to store items by style, but for skia I think this should be ignored. Styles should have their own cache.

pauldendulk:
Thanks for the gifs. The jitter seems to move the lines up and down. Strange. You think this is a rounding error? How does this behave if you are move zoomed-out?

charlenni:
That was my idea. I could create the SKPath when loading the vector tile and than only transform this path. Would be the fastes method. For LineStrings this is allready made. Looks like this

    var matrix = SKMatrix.MakeTranslation(-centerX, centerY);
    SKMatrix.PostConcat(ref matrix, SKMatrix.MakeScale(resolution, resolution));
    SKMatrix.PostConcat(ref matrix, SKMatrix.MakeRotation((float)(viewport.Rotation/180f*Math.PI)));
    SKMatrix.PostConcat(ref matrix, SKMatrix.MakeTranslation(screenCenterX, screenCenterY));

    path.Transform(matrix);

Matrix must created only once. 

But than the jitter comes into play. It is only if you zoom in. See here
zoomin

So I assume, that Skia uses float for this operations and what we see are rounding problems. And you could see it, when you make an animated zoom at the end of animation. Or when you move around. Look at the road in the upper right corner. They jumps up and down while moving.
moving

Perhaps I'm doing something wrong. Or the number of coordinates are to big.

charlenni:
It is the same, if you transform by hand and than create path or create path and than transform by matrices. The later is a little bit faster. The time consuming part is to create the path each time.

And yes, that could help. I assume, that the big coordinates (up to 10 millions) and the short parts like sin and cos (up to 1) is to much for the short float. So, if we could bring the closer together, it could work.

Transforming in a global matrix has the disadvantage, that you not only have to adjust line width, but other things too. Fill patterns and so on.

pauldendulk:
> It is the same, if you transform by hand and than create path or create path and than transform by matrices. The later is a little bit faster. The time consuming part is to create the path each time.

Did you put the matrix on the canvas with matrix.SetCanvas? A 'little bit faster' might be significantly faster once we improve performance in other aspect. Eventually skia rendering should be not much more than drawing SKPath.

pauldendulk:
The pre-transform step (let's call it that) should make some choice about what the maximum values are it allows. If the Map.Envelope exceeds this value it should calculate it's mutiplier.

charlenni:
No, I didn't put it as a canvas matrix. I only transformed the newly created path with transform. Times are (run each conversion 3 times, one type after other type): 211 ms for 3 runs for hand made transformation, 203 ms for Skia matrices. From this, it uses 73 ms for the constructor of SKPath, 28 ms for SKPath.MoveTo and 46 ms for SKPath.LineTo(). So 147 ms of 211/203 ms are used only for creating the path. With this numbers in mind, you know where to work :). For vector tiles the ready made SKPath should be saved.

charlenni:
Regarding the matrix problem: if you look at [Wikipedia](https://en.wikipedia.org/wiki/IEEE_754-1985), float has only 7 deciamal digits. And if we have allready a big number for coordinates (up to 20 millions), than there are not enough digits for a good calculation. For this, the whole coordinate system should be transformed to -0.2 to +0.2 or so. But to do this, I should create a new Mapsui :).

pauldendulk:
When you transform the skpath itself a copy needs to be made. If you set the matrix to the canvas this is no necessity (although I don't know exactly how it works).  Since the constructor takes 73 ms this could mean 73 (or more) ms less.

Yes, caching the skpath is the biggest gain, but once this is done any other small gain can become significant. Also, caching is only possible if you store the unmodified skpath, not the transformed one.  This makes transformation itself interesting.

pauldendulk:
Regarding the matrix problem: I don't think the pre-transform has to be a big change. You have to calculate the multiplier (or we could even start out with an arbitrary chosen one). When we create the skpath we multiply all the coordinates. In the matrix we take this multiplier into account. We store the multiplied skpath in the geometry cache and use that to draw. For feature info and Envelope we use the unmodified geometries.

pauldendulk:
I really think it is important to try canvas.SetMatrix because that could have a different implementation. Inside the canvas it might be multiplied with other matrices and thus end up with another result. It is indeed the most likely that it will have the same problems but we should not start work on a workaround before we test this.

Issue #346, opened 2018-03-15
Redesign of HoverLayers/InfoLayers, and handling user input events


dirkdokter:
My company currently is using Mapsui in an application which requires the end-user to be able to drag features (points) around in a Mapsui layer. The ‘feature dragging’ functionality seems similar to what is described in issues #155 and #148.

The code that we wrote to make the ‘feature dragging’ possible is a bit cluttered at the moment and requires some rewrites.  

While planning for this rewrite, the following ideas came to mind:

# Hover Layers and Info Layers
How about removing the HoverLayers and InfoLayers, and adding hover/click events to ILayer instead?  That would allow the users of Mapsui to attach their own events to a Layer, for example for allowing dragging of geometry (e.g. points, polygons).  Also, I think it could give more flexibility for handling different layers in a different way.

Partial backwards compatibility with the HoverLayers and InfoLayers could be kept making a ‘dummy’ IList<ILayer>, of which the Add and Remove methods simply add or remove some DefaultClickHandler or DefaultHoverHandler. Those default handlers should then mimic the behaviour of the current HoverLayers/InfoLayers.

The Map could invoke the relevant events on the Layers, in order of ‘z-index’/layer order. When an event is handled, some ‘Handled’ property should be set on the event arguments (similar to WPF’s Routed Events). If the ‘Handled’ property is set, other event handlers may (but don’t have to) ignore the event.

# User input handler
For functionality like dragging geometry, someone has to keep track of the ‘state’ of user input events (mouse-ups, mouse-downs, touches, mouse-enters, …). 

Currently, most user input logic is implemented in the (platform-specific) MapControl. I am considering writing a simple UserInputHandler class, which would basically be a state machine which keeps track of user input events. The UserInputHandler would be in the common Mapsui project, i.e. it would be the same for WPF/UWP/Xamarin. The MapControl will send the (platform-specific) events to the UserInputHandler. 

The UserInputHandler’s state could be included in the EventArgs of the aforementioned layer click/hover/drag handlers. 

Maybe the UserInputHandler could also fire clicked/dragged/… events, if anyone needs that. Essentially, the UserInputHandler serves as a platform-independent abstraction layer.

What are your thoughts on this?

charlenni:
What would than the function of MapControl?

pauldendulk:
[...]

PR #355, opened 2018-03-22
Several Geometry additions/changes

dirkdokter:
Non-breaking:
- Added Algorithms.Distance overload for two Geometries.Points (instead of 4 doubles).   
- Added method to get the segments (individual two-point lines) of a LineString.
- Added method to create a (closed) LineString from a LinearRing.

Breaking:
- LinearRing.IsClosed always returns true now, as a LinearRing is closed by definition, even if the StartPoint and EndPoint are not geometrically equal. **This might be breaking.**  Just remove this change if you think this can be a problem for anyone.

### DIFF
--- a/Mapsui.Geometries/LineString.cs
+++ b/Mapsui.Geometries/LineString.cs
@@ -241,5 +241,23 @@ namespace Mapsui.Geometries
             if (lineString == null) return false;
             return Equals(lineString);
         }
+
+
+        /// <summary>
+        ///     Returns a list of line string segments
+        /// </summary>
+        /// <returns>List of LineString</returns>
+        public List<LineString> GetSegments()
+        {
+            List<LineString> segments = new List<LineString>();
+            for (int i = 0; i < this.Vertices.Count - 1; i++)
+            {
+                LineString tmp = new LineString();
+                tmp.Vertices.Add(Vertices[i]);
+                tmp.Vertices.Add(Vertices[i + 1]);
+                segments.Add(tmp);
+            }
+            return segments;
+        }
     }
 }
--- a/Mapsui.Geometries/LinearRing.cs
+++ b/Mapsui.Geometries/LinearRing.cs
@@ -17,6 +17,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Mapsui.Geometries
 {
@@ -69,6 +70,8 @@ namespace Mapsui.Geometries
                 return Math.Abs(-sum/2);
             }
         }
+
+        public new static bool IsClosed => true; // LinearRing is closed by definition
 
         /// <summary>
         ///     Return a copy of this geometry
@@ -161,6 +164,27 @@ namespace Mapsui.Geometries
             return c;
         }
 
+        /// <summary>
+        ///     Returns a clone of the LinearRing as LineString
+        /// </summary>
+        /// <returns>LineString</returns>
+        public LineString GetLineString()
+        {
+            // Make deep copy
+            var tmpLineString = Clone();
+
+            // Check if first vertex is approximately equal to last vertex
+            if (Math.Abs(tmpLineString.StartPoint.X - tmpLineString.EndPoint.X) > Double.Epsilon ||
+                Math.Abs(tmpLineString.StartPoint.Y - tmpLineString.EndPoint.Y) > Double.Epsilon)
+            {
+                tmpLineString.Vertices.Add(tmpLineString.Vertices.First());
+            }
+
+            return tmpLineString;
+        }
+
+
+        
         public LinearRing Rotate(double degrees, Point center)
         {
             var rotatedLinearRing = this.Clone();
--- a/Mapsui.Geometries/Utilities/Algorithms.cs
+++ b/Mapsui.Geometries/Utilities/Algorithms.cs
@@ -35,6 +35,11 @@ namespace Mapsui.Geometries.Utilities
             return Math.Sqrt(Math.Pow(x1 - x2, 2.0) + Math.Pow(y1 - y2, 2.0));
         }
 
+        public static double Distance(Point a, Point b)
+        {
+            return Math.Sqrt(Math.Pow(a.X - b.X, 2.0) + Math.Pow(a.Y - b.Y, 2.0));
+        }
+
         /// <summary>
         ///     Converts the specified angle from degrees to radians
         /// </summary>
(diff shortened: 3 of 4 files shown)


charlenni:
Why do you deliver corners of BoundingBox in CCW? Is there a special reason for this?

dirkdokter:
No specific reason, other than just being accustomed to it because we use CCW by default in several internal applications.

Is the convention in Mapsui CW?

charlenni:
I don‘t know this. But Skia uses CW polygons for outer rings and CCW for inner rings. And Mapbox vector format do the same.

dirkdokter:
Makes sense to change it to CW then. Done.

pauldendulk:
About CW versus CCW. In my code there is this comment, that according to git history is written by me (can't remember). 

``` C#
// note: For Skia inner rings need to be clockwise and outer rings
// need to be counter clockwise (if this is the other way around it also
// seems to work)
// this is not a requirement of the OGC polygon.
```
So it think this means that the outer ring can be CCW if the inner ring is CW. Mapsui has no specific preference, but perhaps we should have some consistency in our interfaces. At this point I do not know what that should be. The OGC Simple Geometry specification also has no specific preference.

pauldendulk:
The LinearRing.IsClosed should return false if begin and end are not the same. Why did you change this? For editing is also handy to have the same begin and end point. 

From your other functionality I can see you are working on editing, because you are adding exactly the feature I need (I am also working on editing).

pauldendulk:
For what situation do you need a to transform a LinearRing to LineString?

pauldendulk:
General remark. Note, that at some point we will start using NTS. This will impact these changes. In some situations you could check if NTS has the implementation you want to add. If not, then you could perhaps better add it as an extension method instead of a extra method to the geometry itself.

charlenni:
Mapbox uses CW for exterior rings and CCW for interior rings. See [here](https://www.mapbox.com/vector-tiles/specification/#winding-order).

pauldendulk:
Ah. Lets follow that.

dirkdokter:
Thanks for merging.

> The LinearRing.IsClosed should return false if begin and end are not the same. Why did you change this? For editing is also handy to have the same begin and end point.

I changed this to always return True, as the LinearRing is simple and closed by definition:
https://github.com/pauldendulk/Mapsui/blob/222edcafc0197d3893426ecc1b868377561d106b/Mapsui.Geometries/LinearRing.cs#L25

I think my reasoning was as follows. Note that i'm not a mathematician, so please correct me if i'm wrong:

If a LinearRing is described/defined in that way, it's IsClosed property should always be true. Code that is using a LinearRing, or methods of LinearRing, should take into account that the first and last vertex are not necessarily the same.

I'm not sure what would be the most mathematically/topologically correct method. 
[Wikipedia ](https://en.wikipedia.org/wiki/Simple_polygon) states the following about Simple Polygons / Jordan Polygons:

> - A polygon encloses a region (called its interior) which always has a measurable area.
> - The line segments that make-up a polygon (called sides or edges) meet only at their endpoints, called vertices (singular: vertex) or less formally "corners".
> - Exactly two edges meet at each vertex.
> - The number of edges always equals the number of vertices.

However, the [lecture notes from an Illinois Computational Topology course](http://jeffe.cs.illinois.edu/teaching/comptop/chapters/01-simple-polygons.pdf) state the following:

> (...) A polygonal chain is closed if it has at least one edge and its first and last vertices coincide (that is, if
p_0 = p_n) and open otherwise.   (...)  A polygonal chain is simple if its vertices are distinct (except possibly
p_0 = p_n) and its edges intersect only at endpoints, or equivalently, if the corresponding path or
cycle is simple.

At a first thought, this seems to be in conflict with Wikipedia's definition. However, one should also realize that a  polygonal chain is not necessarily a polygon (I think only a closed polygonal chain is).

What does a LinearRing actually represent? A (closed?) polygonal chain?

Anyway, the correctness of topological definitions and class descriptions is probably not really important, it's just interesting.  

For now, the code should just be practical.

dirkdokter:
> General remark. Note, that at some point we will start using NTS. This will impact these changes. In some situations you could check if NTS has the implementation you want to add. If not, then you could perhaps better add it as an extension method instead of a extra method to the geometry itself.

I agree that it would be good to switch to NTS at some point. We're just re-inventing the wheel at the moment.

Actually, in NTS they added a comment to LinearRing, which makes the definition more clear:
 https://github.com/NetTopologySuite/NetTopologySuite/blob/master/NetTopologySuite/Geometries/LinearRing.cs

> A LinearRing is a LineString which is both closed and simple. In other words, the first and last coordinate in the ring must be equal, and the interior of the ring must not self-intersect. Either orientation of the ring is allowed.
> - A ring must have either 0 or 4 or more points.
> - The first and last points must be equal (in 2D).
> - If these conditions are not met, the constructors throw an ArgumentException

dirkdokter:
> From your other functionality I can see you are working on editing, because you are adding exactly the feature I need (I am also working on editing).

Yes, we already implemented editing functionality, and are using it successfully now. I still want to do some clean-up work and refactoring before pushing it.

It also uses a different structure for events (see #346), which makes is very #breaking.

infolayer

interctive_infolayer

charlenni:
Very nice.

pauldendulk:
@dirkdokter Thanks for the research :). About LinearRing.IsClosed. The guiding document for our geometry library and NTS is the Simple Feature specification by the OGC http://www.opengeospatial.org/standards/sfa.

It states at page 22:
> ⎯ IsClosed ( ): Integer — Returns 1 (TRUE) if this Curve is closed [StartPoint ( ) = EndPoint ( )]. 

This is about Curve, but LineString is a subclass of Curve and LinearRing is a subclass of LineString.

image

NTS and its java twin JTS have been developed to closely follow the OGC specifications.

pauldendulk:
About rotation (it was in your other PR). It is not directly part of the geometry in the specification and in NTS, so for my own purpose I added an external extension method for rotation of any geometry:

```C#
        public static void Rotate(IGeometry geometry, double degrees, Point center)
        {
            foreach (var vertex in geometry.AllVertices())
            {
                Rotate(vertex, degrees, center);
            }
        }

        private static void Rotate(Point vertex, double degrees, Point center)
        {
            // translate this point back to the center
            var newX = vertex.X - center.X;
            var newY = vertex.Y - center.Y;

            // rotate the values
            var p = Algorithms.RotateClockwiseDegrees(newX, newY, degrees);

            // translate back to original reference frame
            vertex.X = p.X + center.X;
            vertex.Y = p.Y + center.Y;
        }
```

pauldendulk:
About editing. Wow, I see you already made great progress. I have been working on very similar functionality. I will contact you for closer cooperation. There will probably be many edge cases to discuss. Also the user interaction design has many different options to choose from. Ideally I would like to have an interface that is UI independent on top of which different UIs can be implemented.

PR #356, opened 2018-03-22
Fix label rendering unit test

dirkdokter:
Fix label rendering unit test for Xaml and Skia. 

'Original' image contained red halo instead of black halo. Replaced with correct image.


pauldendulk:
When I run the current master are the tests are green. How come? Were they not green on your machine? Is the test you fixed not run correctly?

dirkdokter:
They failed for me (on the current master)
mapsuicapture
afbeelding

cas2406:
I have compared the generated image with the original image, and they don't match. However they do pass the tests on a windows 10 pc but not on a windows 7 pc.

This might be related to the palette flags of bitmap2 as these are invalid

debugscreenshot
debugscreenshot2
labels

cas2406:
Because there are a lot of pixels in the original image and the rendered image that are just white or transparent. it might be better to check if an pixel is transparent or equals the background color, and when this is the case ignore the pixel. 
Beceause otherwise you would still have the problem on very large images, because the percentage wrong pixels would become less.

pauldendulk:
@cas2406 okay, one idea is to check if a pixel is transparent. So, i could check if the alpha channel is 0 and if so, ignore the other channels. Is this what you mean?

I don't think comparing to the background color will work. If in one image a pixel is equal to the background color and in the other it is not, I would say it is just wrong. How did you see that?

pauldendulk:
I looked at some individual pixels that were different. A large percentage had RGB set to 0, 0, 0 and different alphas. I would say they are still different since I could draw those images on top of other images.

PR #361, opened 2018-03-29
Added atlas for bitmaps

charlenni:
There are now 3 different types of bitmaps for symbols: bitmaps, Svgs and an atlas. Bitmaps is the old one. Svg was there allready, but now it works with autodetection and caching on Wpf. New is the using of an atlas. One big atlas is uploaded as bitmap and than users could give a position and size on this atlas, and than this part of the atlas is drawn on map. Up to now it is with high quality filter a little bit slow. Have to look into this.

Handling could be seen in the AtlasSample. Tests are included too. Works on both renderer (Wpf and Skia).

One small change in the API: BitmapRegistry.Get() now returns object instead of Stream.


pauldendulk:
Thanks again! Some remarks.

- Where are the icons from?
- The new icons are bitmaps like the old ones. Could we perhaps treat them in a similar way? With some extra parameters to indicate which one to select.
- Why do we need to return an object in this case?

charlenni:
* The icons are from [this](https://github.com/lukasmartinelli/osm-liberty) project. 
* We do this. The atlas bitmap is handled as normal bitmap. The sprites/icons are handled as Atlas objects, which point to the BitmapId of the atlas bitmap and the place, which to use. Svgs handled as usual. An Atlas needs less space than copy the icon to a new bitmap.
* In this case, the result could be more than a Stream. It could be a Stream, a string (for Svgs) and an Atlas. To handle this, I had to convert from Stream to object as result of BitmapRegistry.Get().

pauldendulk:
You say it is slow, is it slower than if you would have used 100 bitmaps?

charlenni:
Yes. It is fast when you zoom in. It is a number problem. Many icons, many bitmaps to draw.

charlenni:
@pauldendulk While waiting for the review, is there a possibility to create an own NuGet package, so that I could use it be my own projects? Or could I copy the DLLs to my own projects in the package folder?

pauldendulk:
Yes, I add a few lines in the wiki about that:
https://github.com/pauldendulk/Mapsui/wiki/NuGet-of-latest-build#build-your-own-nuget-package

pauldendulk:
Speaking about performance. Now that I understand that FilterQuality.Low is better than default we can also use this when drawing individual bitmaps. It is now set to High. This could improve performance here too.

charlenni:
Drawing of 400 sprites on map (canvas.DrawImage()) takes same time than get SKTypeface :). Was around 50 ms. Difference DrawImage in low and high was 50 ms to 65 ms. So not my main fight against slowness ;)

pauldendulk:
Can you also measure the impact of using FilterQuality.Low versus None for raster tile layers?

charlenni:
So, did some more profiling. Set number of sprites to 10000 and started the profiler. It tokes around 50 ms for DrawImage, regardless of the quality. None to High nearly the same. It seems, that drawing isn't the problem here. Could be, because the images are very small (21 x 21). Even if you have the 10000 on map, the map is movable.

Than I thought, that it is a problem with bigger bitmaps. So I zoomed with OSM map into greater London and increased the zoom one step while the profiler was runing. For all calls of DrawImage at this operation I got 40 ms for filter at high, around 10 ms for medium, 2 ms for low and 1 ms for none.

I looked into BenchmarkDotNet, but didn't found anything about GUI testing. For me it is a library to test functions. So the profiler is the only thing to test with.

charlenni:
Did thought about this a little bit. Perhaps the naming isn't correct. Perhaps the object Atlas should be better called Sprite, because the big image is the atlas and the small ones are better called Sprite. Is this correct?

pauldendulk:
Op ma 9 apr. 2018 16:09 schreef charlenni <notifications@github.com>:

charlenni:
Any comment?

charlenni:
If we do it this way, than we change the complete API. That way I did it, there is only a very small change in a function, that is likely not used from outside of Mapsui.

charlenni:
Should I not change the name from Atlas to Sprite? create a new PR for this?

charlenni:
Could you do this? Would be much easier.

pauldendulk:
I will do it.

charlenni:
Thank you very much.

Issue #352, opened 2018-03-18
Massive performance issues on zoom


AliFlux:
Hi.

I have been experimenting with the control for some time. And I face performance issues when I zoom in on a map with polylines/LineString.

The problem can be replicated using `Simple LineStrings` demo. Upon zooming to street level (max zoom) in an area near the line, the map stutters and gets visibly slow on drag. Both Skia and WPF renderers show this issue.

My guess is that when we zoom in, the renderer still draws the line even if its out of the view port. And upon maximum zoom, the line becomes just way too long to render. I've considered rasterizing the line, but it doesn't suit my requirements.

Thanks.

pauldendulk:
Is this replicable with some of the samples? The sample polygons are
smaller but perhaps some perfomance decrease is visible. If not what could
make the difference? If you have example polygons it would speed op the fix.

Paul

Op 18 mrt. 2018 14:45 schreef "Ali Ashraf" <notifications@github.com>:

pauldendulk:
Thanks for reporting. I can replicate this.

pauldendulk:
In the case of the 'Simple LineStrings' sample it seems to be resolved by using PenStyle.Solid:
```
        public static IStyle CreateLineStringStyle()
        {
            return new VectorStyle
            {
                Fill = null,
                Outline = null,
                Line = {Color = Color.Red, Width = 4, PenStyle = PenStyle.Solid }
            };
        }
```
Dashed pen styles are a new feature. It might need some rework.

pauldendulk:
@AliFlux Can you confirm this solves the problem in your scenario?

pauldendulk:
I see a more noticeable effect in Skia than in WPF. I would suggest we do not try to fix the WPF issue. 

For me this issue does not have high priority (compared to the many other issues we are working on). @charlenni, some suggestions if you want to dig into this:
- Try to remove the creation of the SKPaint to see if this is an important factor. 
- If it is not try to create a Mapsui-less skia sample to reproduce. 
- If you don't see a way to fix it then post an issue in the SkiaSharp project.

charlenni:
[...]

PR #363, opened 2018-03-31
Fixed Skia renderer pixelated scaling issue

AliFlux:
Skia render shows pixelated tiles as the user zooms in #352. This is probably due to the nearest neighbor scaling algorithm.

pixelated3

The fix in this pull request uses an anti-aliased paint to draw the tiles. The quality is set to low, and the paint is cached to make sure that the performance impact is minimal and the tile isn't pixelated either.


pauldendulk:
This does indeed feel a lot faster than FilterQuality.High. I would like to measure the performance difference somehow though. The unit tests all run with data directly available in memory and generate a bitmap. Such a run should be measure with something like https://github.com/dotnet/BenchmarkDotNet

pauldendulk:
Thanks for the PR!

PR #364, opened 2018-03-31
Added a cache for SKTypeface

charlenni:
While working with the profiler for vector tiles, I encounter, that LabelRenderer calls each time SKTypeface.FromFamilyName(), which is a very expensive function (see e.g. [here](https://codereview.chromium.org/838743002/)). So I created a cache for this.

### DIFF
--- a/Mapsui.Rendering.Skia-PCL/LabelRenderer.cs
+++ b/Mapsui.Rendering.Skia-PCL/LabelRenderer.cs
@@ -146,14 +146,22 @@ namespace Mapsui.Rendering.Skia
             }
         }
 
+        private static readonly Dictionary<string, SKTypeface> CacheTypeface = new Dictionary<string, SKTypeface>();
+
         private static SKPaint CreatePaint(LabelStyle style, float layerOpacity)
         {
+            if (!CacheTypeface.TryGetValue(style.Font.FontFamily, out SKTypeface typeface))
+            {
+                typeface = SKTypeface.FromFamilyName(style.Font.FontFamily);
+                CacheTypeface[style.Font.FontFamily] = typeface;
+            }
+
             return new SKPaint
             {
                 TextSize = (float) style.Font.Size,
                 IsAntialias = true,
                 Color = style.ForeColor.ToSkia(layerOpacity),
-                Typeface = SKTypeface.FromFamilyName(style.Font.FontFamily),
+                Typeface = typeface,
                 IsStroke = false,
                 FakeBoldText = false,
                 IsEmbeddedBitmapText = true


pauldendulk:
Thanks for this find. If was thinking to add a Cache field to the IStyle for this kind of thing. The entire SKPaint object could by added to this. The feature could hold a cache of the geometry.

Currently the WPF renderer uses a cache for the style+geometry. This is because in WPF these two are formed into a single object. The problem here is that it the geometry needs a cache for each individual style. To make this work correctly with the dictionary a style comparison needs to be implemented, which is hard and you could easily introduce errors. 

A disadvantage of a static cache is that it could keep on growing so you would need to clean it up. This cleaning up can also become complicated. You need to choose when to clean up. 

(Thinking while I type...) 

Apart from the cache on the IStyle a separate font cache might stil be useful. When using a ThemeStyle one might create a style again and again. For bitmaps we also have the BitmapRegistry, which is similar. I like the BitmapRegistry solution because it is under the control of the user. I don't like the static cache solution used in other places because in some scenarios it created a huge increase in memory use which was impossible to understand for a user without looking at the Mapsui code.

charlenni:
I thought about this too, but there is one problem. When you look into modern styling themas, than they are zoom dependent. They produce for zoom=12.536 other styles then for zoom=12.537. With this in mind, the cache idea isn't so good :). Perhaps it is possible, when getting the list of IStyles to the list of Features while styling features for a given zoom.

I love the BitmapRegistry method too. It is wonderful and easy to use. With this, the atlas implementation was very simple. Good work.

But this said, this PR is a first step. It is easy, there are not much connections or problems to other parts of the code. And there shouldn't so much different typefaces, so that the cache isn't getting to hugh.

pauldendulk:
> I thought about this too, but there is one problem. When you look into modern styling themas, than they are zoom dependent. They produce for zoom=12.536 other styles then for zoom=12.537. With this in mind, the cache idea isn't so good :). Perhaps it is possible, when getting the list of IStyles to the list of Features while styling features for a given zoom.

The zoom dependent style is possible with ThemeStyle. The ThemeStyle method will return specific styles for the zoom levels. Those specific instances are passed to the renderer and will be used for caching. So this could work although it depends on the implementation. The level dependent style should should not be created new in each method call.

pauldendulk:
> But this said, this PR is a first step. It is easy, there are not much connections or problems to other parts of the code. 

That is true and an important argument. It does not complicate things.

> And there shouldn't so much different typefaces, so that the cache isn't getting to huge.

This is also true, although anything that can go wrong will go wrong at some point. 

I'll accept it as is. The above discussing could be input for another iteration.

charlenni:
Thanks.

Regarding zoom: There aren‘t anymore fixed zoom levels. Zooms in styling are floats. So it is difficult to create a cache.

Do you think, someone uses Fields or this[] of Feature? It is part of the IFeature implementation, but isn’t used for drawing.

pauldendulk:
About continuous zoom levels. Mapsui has continuous zoom levels. The list of resolutions are just something that come with the TileLayers and their underlying tiles services (it is part of their spec). The zoom controls can use these levels so that the bitmaps can be drawn pixel perfect, but this is not necessary. On mobile where pinch zoom is used the list of levels is almost irrelevant. Could you point me to examples of continuous zooming that could be problematic?

pauldendulk:
About Feature this[]. Yes this is used. 
- It can be used to store data that can be shown an a feature info request, or combined with other data. 
- It can be used for drawing in combination with the ThemeStyle. This a simple solution that is very powerful. Many intricate user requests can be implemented with this.

charlenni:
See [here](https://122e4e-mapbox.global.ssl.fastly.net/mapbox-gl-style-spec/#types-function-zoom-property).

Mapbox GL JS Style uses so called stopps. Nothing else than functions to get e.g. circle radius. The given example says: up to zoom level 5, the radius is 1 and above 10 it is 2. Between it is a linear function, so that radius is at zoom level 7.5 1.5. Could be the same with line width and other things. With this, you have a new style for a feature for every zoom level, where zoom level is a float. So each time the map is drawn, the style could be new.

Regarding this[]: ok, than it is a tags implementation, which isn't called with this name :). Nice to have in Feature, but not in IFeature. You don't need this there. I would create a Tags property for vector tiles, but have to create this[] too, because it is in IFeature. Perhaps I should add Tags property, which use the same Dictionary, but than we could get problems with others, that use this[] for special cases.

It would be nice, if IFeature is a very light interface, that only contains things, that are important for Mapsui.

PR #365, opened 2018-04-01
Added usefull functions to class Color

charlenni:
Added some tests for class Color

### DIFF
--- a/Mapsui/Styles/Color.cs
+++ b/Mapsui/Styles/Color.cs
@@ -1,4 +1,9 @@
 // ReSharper disable NonReadonlyMemberInGetHashCode // todo: Fix this real issue
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+
 namespace Mapsui.Styles
 {
     public class Color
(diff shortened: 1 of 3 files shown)


pauldendulk:
Where did you base the list of html colors on? This will be useful to many people that are used to the WPF and WinForms (System.Drawing) colors. We probably need to remove the list of hard coded colors and replace it with this.

pauldendulk:
User could use syntax like this:

```C#
var color = Color.FromString(nameof(System.Drawing.Color.Violet));
```

charlenni:
The list is the list of known Html colors. See [here](https://en.wikipedia.org/wiki/Web_colors) or [here](https://www.w3schools.com/colors/colors_names.asp).

I know, System.Drawing use some other color names. But it should be possible to add the missing ones to the list.

pauldendulk:
[This](https://www.w3schools.com/colors/colors_names.asp) list contains 140 color, but in the code 148 colors are defined. What are the extra 8 colors?

charlenni:
Belief not all, what you find in internet :) A short copy & paste action from this side (was the side I get them from) delivers this:
grafik
So, there are 140 different colors, but there doubles like Gray and Grey. So we get 140 colors with 148 different names.

pauldendulk:
ah okay.

Perhaps we could also create the actual color instances like in wpf. It might not be that much more code than we already have. Not sure how wpf colors are defined. What kind of thing is this:

```
System.Windows.Media.Colors.Gray
```

charlenni:
It seems, that System.Drawing.Color has the same names than Html (see [here](https://msdn.microsoft.com/de-de/library/system.drawing.color(v=vs.110).aspx)). And System.Windows.Media.Color too (see [here](https://msdn.microsoft.com/de-de/library/system.windows.media.colors(v=vs.110).aspx)). So I assume, the predefined colors should work for every color system.

Perhaps it is so, because the names are very old (defined for X11 systems).

PR #368, opened 2018-04-06
Word wrap

charlenni:
Long time since last PR ;)

This PR add a multi line label to Skia renderer (and only for this renderer, no code changes for Wpf renderer). For this, LabelStyle is expanded about the following fields:
* MaxWidth: Maximum width of text in em. If text is wider than this, text is shorten or word wraped regarding WordWrap.
* WordWrap: Line break mode for text. If width is bigger than MaxWidth, text is wraped regarding this flag as enum LineBreakMode with 
  - NoWrap: Do not wrap text
  - CharacterWrap: Wrap at character boundaries
  - HeadTruncation: Truncate the head of text and replace with "..."
  - MiddleTruncation: Truncate the middle of text and replace it with "..."
  - TailTruncation: Truncate the tail of text and replace it with "..."
  - WordWrap: Wrap at word boundaries
* LineHeight: Space from one text line to next text line in Paint.FontSpacing

Added some code to samples. Btw. is the drawing of the two lines example in Wpf correct placed? It should be placed Bottom, so that the bottom of text is in line with center point, but in Wpf the top is in line with center.

Added a test for 2 line text to tests.

This code isn't optimized for speed. Multi line text is allways a slow thing, so use it only, if you desperatly need it.


charlenni:
Short example image
grafik

pauldendulk:
> Long time since last PR ;)
hahaha :)

Were did you get all these line break modes from? Which other frameworks do you know that use them?

You are right, the bottom alignment is wrong. I think it already was like this in SharpMap. I understand it is confusing though because you see it actually drawn near the bottom. You have to understand the meaning of alignment to know it is wrong. Perhaps it should be renamed to 'Placement', to Placement.Bottom, Placement.Left. .... but then again, this would change the current left and right alignment which are correct. Usually alignment is used when positioning in a box, which is not confusing. Do you know anything else that uses alignment to a point like Mapsui does?

charlenni:
Found it [here](https://developer.xamarin.com/api/type/Xamarin.Forms.LineBreakMode/).

Yes, each map library should made it the same way.

pauldendulk:
Thanks for that reference to the Xamarin LineBreakModes. I am a big fan of copying good ideas from other frameworks.

About positioning markers and labels, it seems google and openlayers use pixels for positioning. Mapsui does as well for symbols, but I was thinking to use this relative indicators there as well. I would like to have a good example though, so if you see one let me know.

charlenni:
It could be used, but you could although use Offset to move it by pixels or relative to size. All three things are possible. Normally I would use an offset.

pauldendulk:
The disadvantage of offset is that you need to know the size of the marker, label. I was thinking about an positioning mode to allow all.

PR #376, opened 2018-04-12
ScaleBarWidget update

charlenni:
Added some checks for ScaleBarWidget for transformation


pauldendulk:
@charlenni I pushed some more changes to this. Perhaps you can give your opinion.
https://github.com/pauldendulk/Mapsui/commit/13deda7f79e104f5dcd8d83ee51ae691232aa138

- There was a transformation of the viewport center which changed the actual center of the Viewport so that panning became impossible. I now close the center before transforming.
- I removed the check on the number of layers. If there are no layers there still could be a theoretical scale. There is an empty map sample. It is there because that actually happens during some usage scenarios. Also during runtime all layers could be removed to be replaced later on. 
- I moved the checks done in the constructor to the renderer. This is because most of these values could change during runtime. In the renderer I don't throw but just return.

charlenni:
Perfect! I couldn't have done it better.

The only you missed are the checks for the Xaml renderer. Perhaps you could add this too.

PR #377, opened 2018-04-12
Added handling of StyleCollection

charlenni:
With this PR, StyleCollection could be used as IStyle for Label, Point, LineString and Polygon.

### DIFF
--- a/Mapsui.Rendering.Skia-PCL/LabelRenderer.cs
+++ b/Mapsui.Rendering.Skia-PCL/LabelRenderer.cs
@@ -40,12 +40,23 @@ namespace Mapsui.Rendering.Skia
                 horizontalAlignment: style.HorizontalAlignment, verticalAlignment: style.VerticalAlignment);
         }
 
-        public static void Draw(SKCanvas canvas, LabelStyle style, IFeature feature, float x, float y,
+        public static void Draw(SKCanvas canvas, IStyle style, IFeature feature, float x, float y,
             float layerOpacity)
         {
-            var text = style.GetLabelText(feature);
+            if (style is StyleCollection styleCollection)
+            {
+                foreach (var s in styleCollection)
+                {
+                    Draw(canvas, s, feature, x, y, layerOpacity);
+                }
+            }
+            else if (!(style is LabelStyle))
+            {
+                return;
+            }
+            var text = ((LabelStyle)style).GetLabelText(feature);
             if (string.IsNullOrEmpty(text)) return;
-            DrawLabel(canvas, x, y, style, text, layerOpacity);
+            DrawLabel(canvas, x, y, (LabelStyle)style, text, layerOpacity);
         }
 
         private static SKImage CreateLabelAsBitmap(LabelStyle style, string text, float layerOpacity)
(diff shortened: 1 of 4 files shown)


pauldendulk:
I just pushed a commit that I think solves this in a simpler way:
https://github.com/pauldendulk/Mapsui/commit/e7b77e5094b6263972ff66b0685e13690018bcc5
It takes into account the StyleCollection early on in the iterator so you don't have to do several checks later on. Could you check if this indeed solves your problem?

charlenni:
No, that wouldn't work. I return SytleCollections from GetStyle() of IThemeStyle. This isn't correct handled. Should be something like

            var layerStyles = ToArray(layer);
            foreach (var layerStyle in layerStyles)
            {
                var styles = layerStyle; // This is the default that could be overridden by an IThemeStyle

                {
                    if (layerStyle is IThemeStyle) styles = (layerStyle as IThemeStyle).GetStyle(feature);
                    if (styles == null || styles.Enabled == false || styles.MinVisible > viewport.Resolution || styles.MaxVisible < viewport.Resolution)
                        continue;

                    foreach (var style in ToArray(styles))
                    {
                        callback(viewport, style, feature, (float) layer.Opacity);
                    }
                }
            }

But I'm not sure, if this is good handled in such a time critical function. To much conversions. First we convert a normal style to an array and than go through this array to call (in a bad case) only once a draw function.

Wouldn't it be better to check, if style is a collection and than decide, what to do?

                {
                    if (layerStyle is IThemeStyle) style = (layerStyle as IThemeStyle).GetStyle(feature);
                    if (style == null || style.Enabled == false || style.MinVisible > viewport.Resolution || style.MaxVisible < viewport.Resolution)
                        continue;

                    if (style is StyleCollection styles)
                    {
                        foreach (var s in styles)
                        {
                            callback(viewport, s, feature, (float) layer.Opacity);
                        }
                    }
                    else
                    {
                        callback(viewport, style, feature, (float)layer.Opacity);
                    }
                }

pauldendulk:
oh, so it is a layer style, not a feature style?

pauldendulk:
Committed you second suggestion
https://github.com/pauldendulk/Mapsui/commit/dc2dc73ac1dc7c229e3f41039c2d10b5c901b31d

charlenni:
It could be both. StyleCollection is a normal IStyle, so it could be used at all this places, where IStyle is used. And yes, I used it as result of a GetStyle(). But someone else could use it as feature style. So the second foreach loop was correct in your first PR.

pauldendulk:
Hopefully the latest commit fixes your issue. 

The next improvement will be to remove the StyleCollection altogether and replace it with an ICollection<IStyle> everywhere.

charlenni:
What is, if style is a StyleCollection, which contains a StyleCollection? Perhaps this shouldn't be possible by definition.

pauldendulk:
These possibly recursive StyleCollections are a reason to remove them.

So, does the last commit fix your issue?

charlenni:
As mentioned above, you removed the handling from feature iteration. You should check it at both places: for layer styles and for feature styles. First you checked only for feature styles, than you removed this and now check only for layer styles.

If you want to use ICollection, than this is an API breaker. StyleCollection isn't. There could be a check implemented, so that you couldn't add StyleCollections to StyleCollections. But I'm also ok with ICollection<IStyle> for Style.

pauldendulk:
Do you actually need the StyleCollection on a feature in your case? If not then I will rather leave it out, to prepare for the removal of StyleCollection.  

A feature has no use for ThemeStyle because the feature attributes are known and it already consists of ICollection<IStyle>so you can already add multiple styles.

charlenni:
And no, I don't need this in the moment.

pauldendulk:
I assume this is fixed by taking StyleCollection into account in the visual feature iterator.

PR #382, opened 2018-04-17
Work on PolygonRenderer of Skia

charlenni:
Removed some bugs
Added some functionality
Worked on performance


pauldendulk:
Why is the StyleCollection cast necessary?
```c#
else if (style is StyleCollection styleCollection)
```

charlenni:
I would say, because style is of type IStyle.

PR #427, opened 2018-05-29
Comments

charlenni:
Started with some work to document Map, Viewport and parts of Layers.
Refactored Map and Viewport, so that it is better to understand, what Map and what Viewport does. Moved for this functions NavigateTo() and RotateTo() from Map to Viewport. At the end it should be possible to have one Map with many Viewports/MapControls, which show different parts of the map.

@pauldendulk Please check, and say, if this is ok. If not, drop the PR.


pauldendulk:
Thanks again! Could you give a summary of the code changes? Most is comments I see, but the diff gets a bit confused so it is sometimes hard to see where the code is changed and where it is just comments.

Changes:
- I see ViewChanged has move to viewport.
- You mentioned NavigateTo() and RotateTo() has moved to viewport.
- You mentioned "At the end it should be possible to have one Map with many Viewports/MapControls". What related changes are in the MR?

btw, if code comments and code changes were in separate PRs it would be easier to merge.

charlenni:
Oh, sorry for doing this again :(

The biggest part of the PR are comments and a reordering of the source code, so that the order is constructor, properties, public function and than private functions. I could try to redo the ordering of the source code, than it would be better viewable. Should I?

I added ViewChanged to Viewport, because I think, that this is the right place. The old function exists further, but is marked as obsolet. Same to NavigateTo() and RotateTo().

There aren't any changes for your last point. But I assume, that this should be the goal for future changes.

pauldendulk:
About ordering of methods/fields. I gave this no attention at all so far. Resharper has some tools for this. I have used it in the past. Then it was not fully to my liking. Perhaps it has improved. Perhaps there are alternatives. The reason I gave it no attention is that I hope for a fully automated solution that places every single method/fields in it's right place. 

## public/private sorting
I don't really like the public/private distinction as the most important one because it separates methods that deal with the same topic. A public method, has a public signature, but all the code within a public method is private to external users. When I move part of the private code to a separate method (which is something I try to promote) the code will suddenly move to the bottom of my code page. So I hope there will be a better automated process to order the methods, for instance by call order. If there is any fully automated process I will use it anyway, even if it sort in public/private. Perhaps we should do some googling for that.

## Local functions
One way to prevent the public/private sorting is local functions c# 7. Perhaps we should use that more:
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions

pauldendulk:
About ViewChanged. The plan for Mapsui 2.0 was to rename this to RefreshData. The name 'ViewChanged' indicates when it is called not what it is doing and currently it is also called in other situations, for instance when the data in the provider has changed. The implementation calls the Layers so I think it is more appropriate in Map than in Viewport. I understand the confusion caused by the bad naming.

pauldendulk:
You added a remark about SharpMap. This is the original license. I will keep those licenses there until I go and move every thing piece by piece to an MIT repository. 

In this case Viewport is my own original code (added when it was still part of SharpMap) so I will be able to just move it to MIT.

pauldendulk:
I moved Viewport.ViewChanged back to Map and renamed it to RefreshData.

I saw the Map was also added as a member to the viewport. This is something I really want to avoid. Map and Viewport get a circular dependency this way. In general I want to pass the minimal information needed to a method or class. The NavigateTo methods no use OnViewportChanged but I am not sure if this is enough. I have to leave it at this for now. Time for the weekend. 

More later!

Issue #429, opened 2018-05-30
UWP high CPU/GPU while doing *NOTHING*


jasells:
The subject says it all, really.  I think I do have a solution, but my fork is a little our of date.  Going to get your latest updates and issue a pull req in the next few days.   The current fix is for UWP only, but likely is similar on WPF, and takes CPU ~ 5-6% and GPU of 12-15% @ idle on a Surface Pro (core i7 7660U with Intel Iris Plus 640) to 0% CPU and 0% GPU *unless* the screen actually changes (zoom, pan, etc) @ 2.5-3% CPU and 6% max GPU.  

Debugger attached while making those measurements.

I have a couple of Task Man screen shots I'll post ASAP.

jasells:
Here is [link ](https://1drv.ms/u/s!AkJlaLtUNZiihLMSk4nN3McLsXR0Yw) to TaskManager screen-shot with current build. <br>

Here is [link](https://1drv.ms/u/s!AkJlaLtUNZiihLMRaKwrhXMsRkpPbA) to TaskManager screen-shot after fix.

charlenni:
Would be nice to know, where the problem is. Any hint?

jasells:
It is in the UWP [MapControl.cs](https://github.com/jasells/Mapsui/blob/1.4/Mapsui.UI.Uwp/MapControl.cs) mostly.  The current impl relies on CompositionTarget.Rendering event to make sure screen re-draws occur during drags.  This event is *not* on-demand, so it ends up doing a *lot* of redraws, even when nothing on screen has changed.
<br><br>
There a couple of ways to make it more on-demand/event driven vs. constant re-draw.  Looking into which is best and how to branch/make pull request.  Been a while since I did that on GitHub.

<br><br>


I also found an issue when removing a layer didn't refresh the map until the map was moved or NavigateTo was called (i.e. some other event caused a re-draw, not layer remove).

pauldendulk:
Wrapping this up. What I did not see at first is how the GPU was high while the Mapsui renderer was not called. This was because _renderTarget.Invalidate() not only triggers the paint event but also does a repaint of the current state of the skia canvas, even if not changed. The reason GPU was not high in WPF is that there was an _invalid check in CompositionTarget_Rendering itself (I was only focused on the paint method.

In UWP:
```
        private void CompositionTarget_Rendering(object sender, object e)
        {
            _renderTarget.Invalidate();
        }
```

In WPF:
```
[...]

PR #430, opened 2018-05-30
Uwp performance enhancement

jasells:
This pull request appears to fix the issue: #429

### DIFF
--- a/Mapsui.UI.Uwp/MapControl.cs
+++ b/Mapsui.UI.Uwp/MapControl.cs
@@ -79,6 +79,7 @@ namespace Mapsui.UI.Uwp
             ManipulationMode = ManipulationModes.Scale | ManipulationModes.TranslateX | ManipulationModes.TranslateY | ManipulationModes.Rotate;
             ManipulationDelta += OnManipulationDelta;
             ManipulationCompleted += OnManipulationCompleted;
+            //see event handler for how to force re-draw while manipulation is occuring
             ManipulationStarted += OnManipulationStarted;
             ManipulationInertiaStarting += OnManipulationInertiaStarting;
 
--- a/Mapsui.UI.Uwp/MapControl.cs
+++ b/Mapsui.UI.Uwp/MapControl.cs
@@ -272,6 +272,13 @@ namespace Mapsui.UI.Uwp
             _renderTarget.Invalidate();
         }
 
+
+        private void CompositionTarget_Rendering(object sender, object e)
+        {
+            //force Skia to re-draw
+            InvalidateCanvas();
+        }
+
         public void Clear()
         {
             _map?.ClearCache();
@@ -363,12 +370,6 @@ namespace Mapsui.UI.Uwp
                     Logger.Log(LogLevel.Warning, "Unexpected exception in MapDataChanged", exception);
                 }
             }
-        }
-
-        private void CompositionTarget_Rendering(object sender, object e)
-        {
-            //force Skia to re-draw
-            _renderTarget.Invalidate();
         }
 
         private void _renderTarget_PaintSurface(object sender, SKPaintSurfaceEventArgs e)
--- a/Mapsui.UI.Uwp/MapControl.cs
+++ b/Mapsui.UI.Uwp/MapControl.cs
@@ -68,7 +68,8 @@ namespace Mapsui.UI.Uwp
             Loaded += MapControlLoaded;
 
             SizeChanged += MapControlSizeChanged;
-            CompositionTarget.Rendering += CompositionTarget_Rendering;
+            //This event will fire @ +30Hz, so it causes lots of uneccesary re-draws when the app should be "idle"
+            //CompositionTarget.Rendering += CompositionTarget_Rendering;
 
             _renderer = new MapRenderer();
             _scale = GetSkiaScale();
@@ -78,6 +79,7 @@ namespace Mapsui.UI.Uwp
             ManipulationMode = ManipulationModes.Scale | ManipulationModes.TranslateX | ManipulationModes.TranslateY | ManipulationModes.Rotate;
             ManipulationDelta += OnManipulationDelta;
             ManipulationCompleted += OnManipulationCompleted;
+            ManipulationStarted += OnManipulationStarted;
             ManipulationInertiaStarting += OnManipulationInertiaStarting;
 
             Tapped += OnSingleTapped;
@@ -263,6 +265,9 @@ namespace Mapsui.UI.Uwp
             InvalidateMeasure();
             _renderTarget.InvalidateArrange();
             _renderTarget.InvalidateMeasure();
+
+            //make sure skia re-draws
+            _renderTarget.Invalidate();
         }
 
         public void Clear()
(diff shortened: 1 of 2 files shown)


jasells:
So... the 1.4.x impl forces Skia to be refreshed everytime CompositionTarget.Rendering event is fired.  I don't know a lot about it under the covers, but here is the [documentation](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Media.CompositionTarget).<br>
The abridged version is that this event is not tied to any state of your app, or control, but to the XAML rendering engine itself (and I would guess it is more like a timer or simply as-fast-as-possible loop).  So, the MapControl is constantly being re-drawn, regardless of the fact that nothing may have changed.  You can verify by starting the UWP sample with debugger and get everything loaded, wait 10-20 seconds and then place a breakpoint or debug print on the CompositionTarget.Rendering event handler.  You'll see it is called constantly, even if no interaction with the app.  So, in my poking around, I simply removed that event handler and found that the map only updated @ the *end* of a drag.  The adding/removing of the CompositionTarget.Rendering event was to make sure that the map visibly updates during a drag (manipulation), but then doesn't re-draw if the map is static/idle.<br>
As the comments point out, forcing Skia to re-draw from OnManipulationDelta event will also maintain map updates during a drag without using the  CompositionTarget.Rendering event at all, but in my admittedly limited testing, the impl in the pull request performs visually identically, with lower CPU/GPU usage.  Not sure why.  Maybe ManipulationDelta event fires more often than  CompositionTarget.Rendering event?  Maybe other code called from OnManipulationDelta() is being done more than once?

pauldendulk:
In the current master gpu is still high while the render method is not called. I don't understand.

image

pauldendulk:
In the new implementation the gpu is also high. And it is higher if you make the application window larger. The SKXamlCanvas has it's own render loop?

charlenni:
Normally (if I remember this right) Skia has two types of rendering methods: you could create a static image and let Skia draw this or you could use a function/event, that is called each time, when Skia canvas needs a redraw.

pauldendulk:
Compared them again. I barely see a difference. Perhaps the cpu is 2 instead 1.5. Do you see something else @jasells? Could it be your initial test was with an old version of uwp that did not check the _invalid boolean?

pauldendulk:
@charlenni Googling for skiasharp rendering loop did not give any answers. If you have some pointers where to look for I will try again.

jasells:
From #429 <br>

Here is TaskManager screen-shot with current build (1.4). mapsui-doingnothing-highcpu-gpu
 <br>

Here is TaskManager screen-shot after fix. mapsui-uwp-good  <br>

Both are @ idle (no map interaction/changes happening). On my Surface Pro (i7 7660U, Intel Iris Plus 640 GPU)<br>

My built in screen on the Surface is higher res than typical display (2736x1824), so likely is amplifying the issue somewhat, *but* the point is if no changes to the map are made, there is *no* reason to call SKXamlCanvas.Invalidate() and force a redraw.  The app sitting idle shouldn't be using any cycles CPU or GPU.<br>

A breakpoint (or debug.Writeline) in `
 private void CompositionTarget_Rendering(object sender, object e)

        {

            _renderTarget.Invalidate();

        }` 
will hit (a lot), even when app is idle and nothing on screen changed.
<br>
@pauldendulk, your break point in above screen cap is in _renderTarget_PaintSurface() event handler.  These are different events.   Skia doesn't have a render loop, the app/runtime does.  I'm not sure exactly what triggers the PaintSurface event, but it is not being raised if nothing is changing in the app.  A quick test with a Debug.Writeline() shows that PaintSurface is called several times on start up, or if the window size changes, or zoom, etc, and then stops.  
<br>

pauldendulk:
Oh, I see now that my test was not correct. Will test again tonight.

jasells:
I pushed up another fix that originally thought was unrelated, but now realize it is.  Adding/Removing Layers wasn't visible on screen w/o some other trigger for a screen refresh.  Added a trigger in [Mapsui.Map](https://github.com/jasells/Mapsui/blob/UWP-Perf/Mapsui/Map.cs) in LayersLayerRemoved/LayersLayerAdded to make sure the changes show on screen.  The reason is because the CompositionTarget.Rendering event *was* telling Skia to redraw @ ~60 HZ, so the changes got shown on a periodic basis (vs. event-driven).

pauldendulk:
I can reproduce your fix now. Looks great. Still not sure why there was such a high gpu because the render update was not actually called. The bind to CompositionTargetRendering trigger a render update somehow. In WPF a similar setup has no high gpu/cpu usage.

You now attach and detach to the render event. This was because of performance. I now removed the handler all together and call _renderTarget.Invalidate(). This seems to perform fine. What did you try?

pauldendulk:
btw, Refresh() updates both data and graphics. 

```c#
        public void Refresh()
        {
            RefreshData();
            RefreshGraphics();
        }
```

jasells:
-btw, Refresh() updates both data and graphics. <Br>
 
Yah, I didn't quite follow why that code was written that way.... most of the code seems to be event driven, but that appears to be more appropriate for a constant loop update (like from the CompositionTarget.Rendering event), or maybe from a input/touch event. 
<br><br>

-You now attach and detach to the render event. This was because of performance. I now removed the handler all together and call _renderTarget.Invalidate(). This seems to perform fine. What did you try?<br>
<br>

I don't quite follow without seeing your code...   Where are you calling Invalidate()?  
Subscription model of events is very useful for this kind of dynamic sub/unsub, though I get that is not the most commonly seen usage of events in .NET.  Nothing inherently bad (nor inefficient) about doing it, but it can be a bit difficult to follow if done lots of places.  That's why I put OnManipulationStarted/Ended methods together in code.  Could also move CompositionTarget_Rendering closer to those to help too.
<br><br>

- In WPF a similar setup has no high gpu/cpu usage <br>
Not sure, I haven't looked @ that code much.  A quick look shows some extra logic to check for need to Invalidate() not in the UWP version... I did try something similar, but the _invalid flag in UWP code was used many places and I couldn't figure out how to duplicate it.  *Slightly* more efficient to just unsubscribe from the event vs. doing a if(flag) return check @ 60+ Hz.... most time is in un-sub state anyway. WPF version:<br>

`
        private void CompositionTargetRendering(object sender, EventArgs e)

        {

            if (!_invalid) return; // Don't render when nothing has changed



            if (RenderMode == RenderMode.Wpf) RenderWpf();

            else SkiaCanvas.InvalidateVisual();

        }`

pauldendulk:
About Refresh(). The Refresh(), RefreshData() and RefreshGraphics() methods are public methods intended for users. Perhaps they are used now on some points where we should use _renderTarget.Invalidate().

I suspect that there will be more places where we will have to subscribe and unsubscribe which will complicate the code. Calling _renderTarget.Invalidate() from OnManipulationDelta is somewhat simpler. I will accept your merge, and apply that change so you can take a look. 

With the render loop totally removed it is perhaps also possible to remove the _invalid flag. Further simplifying the code.

What still confuses me is that in the old code the render method was called, but it exited right away, there was no actual rendering, yet the gpu was high. Somehow the skia view updated itself for some reason.

pauldendulk:
This is with the render loop removed and _invalid removed. In the manipulation delta method I call RefreshGraphics.

https://github.com/Mapsui/Mapsui/commit/45a027b5216203bdcd89f60bd9e42ad41fc2179d

PR #454, opened 2018-06-20
Replaced Mapsui.Utilities.Timer with System.Threading.Timer

charlenni:
Regarding the AnimatedPointsSample: Shouldn't the points be animated between the timer ticks? I cant reproduce this. Not with the new timer, not with the old timer.


pauldendulk:
It think this is an issue with the random generator logic. Introduces when doing some refactoring some months ago. They do animate initially don't they?

charlenni:
They start in the center, than go animated to their places. After this initial animation, only one feature vanish and another pops up.

pauldendulk:
Yes. I think the random generator restarts with same seed.

charlenni:
You are right. That is the problem.

PR #472, opened 2018-07-19
Add ArrowLineString model and renderings

kirrianp:
A new feature made up of a LineString whose scaling behavior is like LineString's one, and an arrow whose scaling behavior is like a Symbol point. 
Can be used to build oriented lines.

We did not implement any test or factory


pauldendulk:
Hi @kirrianp, Thanks for you PR.

I don't have a lot of time during the summer so it could take a while before I can look at this in depth. A general problem I see is that it introduces a new geometry type. The geometry types in Mapsui are based on the OGC simple feature specification which does not contain an arrow geometry. Furthermore we intend to replace the Mapsui geometries with those of NTS.

Some alternatives.
- Add a way to add custom geometries and custom renderers of those geometries. We are working on something similar for widgets.
- Change Mapsui symbology to allow arrows. I have already been thinking about use of symbols for the vertices of a geometry. They are useful for editing for example. For arrows the symbology should also be able to take into account end or begin nodes, and the orientation of the line to that node.
- Add something that allows an easy combination of points symbols and line symbols.

pauldendulk:
From glancing at the code it seems that most of the extra fields in the ArrowLineString are for drawing the arrow head. Is this correct? If so an arrow symbol style applied to a regular LineString seem not that far removed.

kirrianp:
Yes, you are right. I thought dealing with the inclination was easier in the model, but maybe I can manage the arrow's position with the rotation instead.

pauldendulk:
Closing this. What we could do instead:
- Allow custom renderers
- Add arrow as a kind of symbology 
- Allow arrows by adding symbols who's angle is determined by the angle of the line to the vertex
- Allow custom symbol renderers.

PR #514, opened 2018-09-18
Making Mapsui MVVM compatible

sanderswart-atmos:
In order to make Mapsui MVVM compatible, I've exposed the Map and Navigator properties of the WPF MapControl as DependencyProperties. This way, a viewmodel can bind to those properties in the xaml and manipulate the map layers and viewport. 

I've also added a simple example which adds a Bing Aerial map via a viewmodel and can manipulate the location of the Viewport's center from the viewmodel using the Navigator's methods.

Known bugs:
- When the map is first loaded in, it starts at the default resolution of 1. I haven't found a way to fix this, since setting the zoom level (using Navigator.ZoomTo(resolution)) when adding the map layer results in a memory leak. Setting the zoom level in the OnStartAnimations method works like it's supposed to.


pauldendulk:
Thanks sanders. I checked out your MVVM branch and started Mapsui.Samples.Wpf.Mvvm I don't see a map. Are there some specific instructions to get this running?

pauldendulk:
The problem was with incompatible versions of the skiasharp nugets. This problem was perhaps in the repo at the time you branched it. I updated all to 1.60.0 to get it to work.

pauldendulk:
When I start to see the mvvm sample:
- It starts with an empty map
- I press add bing layer and see the missing tile images.
- I zoom out to see the whole world.
- I press start animation. The map zooms in to a point in the ocean. The button goes to 'stop animation'.
- I don't see a smooth animation.
- I if start panning/zooming the map randomly jumps to other positions.

sanderswart-atmos:
I've just edited the example so it starts with OpenStreetsMap as initial map. This can be any map and is editable in the ViewModel (I can make a dropdown menu where you can select a map to elaborate more on that flow).

The example was just to show that you can set the map layers and navigate on the map using an MVVM pattern instead of using code-behind.

The issue that the Viewport starts fully zoomed in is something I can't seem to fix, since trying to set the resolution in the constructor or setter results in a memory leak where it seems like the entire map is being filled with tiles.

There is no animation since there is no animation yet in the navigator or MapControl. In my opinion, that would be the place where animations should be handled. They can be platform independent by just using native c# functions that animate the zoom level or center of the Viewport.

If you start panning when the animation is still running, the animation will override your position, thus making it jump to other positions. If you stop the animation by clicking on the button again, it will stop the animation and panning should be the same as the other examples.

pauldendulk:
The DependencyProperties are wpf specific. I am trying to work towards more similarity between all four platforms. I assume NotifyPropertyChanged could also be used instead. Do you see a reason to use DependencyProperties specifically?

pauldendulk:
This is what I see when I run the animation.

out6

pauldendulk:
When I run after a clean checkout the map shows nothing because of an exception. The SkiaSharp.Svg dll can not be loaded. You can see the exception in the debugger I you break on all common language runtime exception by checking them in the Exception Settings.

image

If downgrade to SkiaSharp 1.60.0 (not sure if this is necessary) and add SkiaSharp.Svg to the MVVM project it works again.

sanderswart-atmos:
> The DependencyProperties are wpf specific. I am trying to work towards more similarity between all four platforms. I assume NotifyPropertyChanged could also be used instead. Do you see a reason to use DependencyProperties specifically?

This is done so that you can bind to the map and navigator in the xaml, removing the need for code-behind.

The Skia error is one that I'm struggling with. Apparently, that is a skia error which should be fixed in the newest version.

sanderswart-atmos:
I could add a slider which controls the zoom level of the animation if that would make the example more intuitive to use

pauldendulk:
I am going to close this. The project has moved on in the meantime and there are too many changes. If we look at this a second time there should be a discussion first. Get clear what the limiting factor is. Discuss how it could be fixed.

PR #537, opened 2018-11-10
Added INotifyPropertyChanged to MapControl

charlenni:
Implement the INotifyPropertyChanged interface in MapControl
Added some comments
Removed some empty lines


charlenni:
The #if is, because the Forms implementation derives from Xamarin.Forms.View, which already contains a INotifyPropertyChanged implementation. I didn't find something to avoid this.

PR #538, opened 2018-11-13
Native Xamarin.Forms MapControl

charlenni:
As @pauldendulk suggested, I PR the new Xamarin.Forms MapControl. It is far from ready and there are many things, that should be done. Perhaps a new branch would be a better place.

It would be nice to have a package for testing this.


charlenni:
Thank you for looking into this PR. I try to answer some of your questions.

>     There is a MapView and MapControl. How do they relate?

I put this at the start, because I hope some other questions than answered more easily.

MapControl has the same API as all other MapControls from Mapsui, but is a Xamarin.Forms View. So it is possible to use MapControl as any other MapControl. MapView add some functionality on top, which users know from the original MapView, e.g. objects for Positions in lat/long, Layers for Pins, Polylines and Circles, Callouts. All this things are special for XF. I don't want to create another project for MapView, because they are close related to each other.

> There is an Objects folder. Is there a better way to describe the purpose of this folder? Should it perhaps be split up to allow a more purposeful name?

All the afore mentioned objects are located in the Objects folder. If you have a better name for this folder, we could change it.

> Both are very large, they might need to be split up more, or more reuse of shared code.

MapControl.Forms has around 650 lines, MapControl.WPF around 550 lines and MapControl.Android around 350 lines. I added some more events to MapControl than MapControl.Android has and I used more comments ;)

MapView has 900 lines. The easiest to do, is to delete the comments ;). No, sorry, here I have no idea, what to do. Perhaps we could later transfer parts out of MapView or someone has a good idea, what to change.

> I see there is a ValueTuple package dependency. We we still need this with the framework versions we are using?

I looked through the project, but couldn't find a referenve to ValueTuple. Where did you see it?

> Mapsui.UI.Android has two target framework versions. Is this is conscious choice? v8.1
    v7.1

Removed this.

> There are some files in a Backup folder. I think they can be removed.

Yes, I removed them.

> Why the changes in the unrelated project like in uwp and ios
> ios package.config has new references like xamarinios10 and microsoft.netcore.platforms, why?
> mapsui.ios.sln especially has many new files. If this is unintended then the best option is perhaps to do a clean checkt of mapsui master and copy only the forms related files into it, and PR that.

I had problems with my PC. Couldn't find the reason for over 4 month. So installed Win and VS new. Perhaps some changes are from the problems and testing. Tried to revert most of the changes.

> There is a license.md, there can be only one. You keep the copythings of your own code of course. But this license can not be part of the mapsui repository.

Oh, I didn't know this. I added the license, because some parts of the code are from another project ([Xamarin.Forms.GoogleMaps](https://github.com/amay077/Xamarin.Forms.GoogleMaps)). So, where should I put this credentials?

pauldendulk:
I am sorry but you can not include code in the pull request if you do not own the copyright. That is part of the CLA you signed https://cla-assistant.io/Mapsui/Mapsui. It needs to be removed. What parts does it concern?

pauldendulk:
You mention "original MapView". Do you mean the first one you created?

pauldendulk:
>  The easiest to do, is to delete the comments 

hahaha

pauldendulk:
There are a lot of non-forms files that have been changed. Probably because of merging your changes with the ones from master. To avoid this you could create a new branch from mapsui/master and copy your forms code into it. I just did this in my local copy your project and it was fairly easy to do.

Here is what I did. 

make a copy of your mapsui fork with the forms branch checked out.
`git checkout -b forms2 origin/master`
`git checkout forms2`
Than manually copy all to forms files in the copy i made into the forms2 branch I have checked out.
In Visual Studio add those csprojs to the sln.
In the shared MapControl in the #if Change Mapsui.UI to Mapsui.UI.Forms.
I had to change the version of Universal Windows to get it to compile on my machine.

Perhaps I forgot something, but not much.

I was thinking about making a PR from this code to your fork but that would mean I would be the author of all this code according to git history.

charlenni:
> You mention "original MapView". Do you mean the first one you created?

I‘m talking about the Xamarin.Forms.Maps.MapView, the original map component from Xamarin.Forms.

charlenni:
I changed the most of this. But, ok, I could make a new PR. Takes some time.

charlenni:
Should I close this PR?

charlenni:
I thought a little bit more about your suggestion. If I do it this way, I loose the complete history for the branch.

charlenni:
I changed many of the above mentioned things. To see this better, I close the this PR and create a new one.

PR #541, opened 2018-11-17
Native Xamarin.Forms.MapControl

charlenni:
As @pauldendulk suggested, I PR the new Xamarin.Forms MapControl. It is far from ready and there are many things, that should be done. Perhaps a new branch would be a better place.

I revert 8 files to their original state. Hope, this helps, so that you could read the code much better.

It would be nice to have a package for testing this.


gverbake:
Great work!!!! I just started to use this in a Xamarin Forms project, but still with version 1.4.8 with MapsuiForms.
I really would like to see a BETA nuget package with these changes in it to test it!

pauldendulk:
Went through above files. The .sln still has a many configs after re-adding. I will live with it. Let's clean this up once (it's not on the roadmap i heard) Xamarin will use the new csproj format.

I still get an error on mbtiles on android. I will create an issue for follow up work on forms.

Issue #545, opened 2018-11-28
When Widget is clicked, result is null


charlenni:
When you click on a Widget, the click is handled, but than there is returned null instead of a MapInfoEventArgs with Handled = true (see [here](https://github.com/Mapsui/Mapsui/blob/master/Mapsui.UI.Shared/MapControl.cs#L362)). 

For this, the widget callback should return a value, if the Widget handled this event or not. This said, the Widget callback should be ````Func<IWidget, Point, bool>```` instead of ````Action<IWidget, Point>````

pauldendulk:
Okay, let's do it like that.

charlenni:
Should I make this and the create a PR?

charlenni:
See PR #547

PR #546, opened 2018-11-28
Forms

charlenni:
Removed some more bugs and tested it with Android, iOS, UWP and Mac OS. All should now work.


Mapsui.UI.Forms/MapControl.cs
```diff
@@ -537,10 +533,13 @@ private bool OnDoubleTapped(Geometries.Point screenPosition, int numOfTaps)
 
             var eventReturn = InvokeInfo(Map.Layers, Map.Widgets, Viewport, screenPosition, screenPosition, _renderer.SymbolCache, WidgetTouched, numOfTaps);
 
-            if (!eventReturn.Handled)
+            if (eventReturn != null)
```
pauldendulk: Could this be more compact by using ? as the null check?

charlenni:
The problem with the null check could be dropped, if ````InvokeInfo()```` returns always a bool. See #545.

PR #547, opened 2018-11-30
Added a bool to return of WidgetTouched event

charlenni:
Added a bool to return of WidgetTouched event, to see, if Widget had handled the event or not.

Removed a bug in the samples

### DIFF
--- a/Mapsui/Widgets/Hyperlink.cs
+++ b/Mapsui/Widgets/Hyperlink.cs
@@ -7,8 +7,11 @@ namespace Mapsui.Widgets
     {
         public string Url { get; set; }
 
-        public override void HandleWidgetTouched(INavigator navigator, Point position)
+        public override bool HandleWidgetTouched(INavigator navigator, Point position)
         {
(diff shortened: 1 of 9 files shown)


Mapsui/Widgets/Hyperlink.cs
```diff
@@ -7,8 +7,11 @@ public class Hyperlink : TextBox
     {
         public string Url { get; set; }
 
-        public override void HandleWidgetTouched(INavigator navigator, Point position)
+        public override bool HandleWidgetTouched(INavigator navigator, Point position)
         {
+            // Because OpenURL is called from MapControl by default
+            // TODO: Shouldn't OpenURL() called from here instead of from MapControl?
```
pauldendulk: You will need to take into account the platform dependent implementation. Right now this part is in the MapControl. An alternative is to inject a IOpenInBrowser  interface to the Hyperlink and use platform dependent implementations of the interface. Side note: There might be an issue with Android's Context.

PR #550, opened 2018-12-03
Removed a bug, where Attribution isn't any longer clickable

charlenni:
Added handling of more than one widget at the same position. Now each widget is checked and if it doesn't handle the touch event, the next widget is checked.
Hyperlinks get a new event "Touched". With this, the user could handle touched events for the Hyperlink by themself.
Removed a small bug in Android MainActivity.

### DIFF
--- a/Mapsui.UI.Android/MapControl.cs
+++ b/Mapsui.UI.Android/MapControl.cs
@@ -65,14 +65,16 @@ namespace Mapsui.UI.Android
         private void OnDoubleTapped(object sender, GestureDetector.DoubleTapEventArgs e)
         {
             var position = GetScreenPosition(e.Event, this);
-            OnInfo(InvokeInfo(Map.Layers.Where(l => l.IsMapInfoLayer), Map.Widgets, Viewport, 
+            var widgets = Map.Layers.Select(l => l.Attribution).Where(a => a != null).Concat(Map.Widgets).ToList();
+            OnInfo(InvokeInfo(Map.Layers.Where(l => l.IsMapInfoLayer), widgets, Viewport, 
                 position, position, Renderer.SymbolCache, WidgetTouched, 2));
         }
 
         private void OnSingleTapped(object sender, GestureDetector.SingleTapConfirmedEventArgs e)
         {
             var position = GetScreenPosition(e.Event, this);
-            OnInfo(InvokeInfo(Map.Layers.Where(l => l.IsMapInfoLayer), Map.Widgets, Viewport, 
+            var widgets = Map.Layers.Select(l => l.Attribution).Where(a => a != null).Concat(Map.Widgets).ToList();
+            OnInfo(InvokeInfo(Map.Layers.Where(l => l.IsMapInfoLayer), widgets, Viewport, 
                 position, position, Renderer.SymbolCache, WidgetTouched, 1));
         }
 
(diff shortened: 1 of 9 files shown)


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -351,21 +355,20 @@ public MapInfo GetMapInfo(IEnumerable<ILayer> layers, Point screenPosition, int
...
+            if (touchedWidgets.Count() > 0)
             {
-                // todo:
-                // How should widgetCallback have a handled type thing?
-                // Widgets should be iterated through rather than getting a single widget, 
-                // based on Z index and then called until handled = true; Ordered By highest Z
-                var result = widgetCallback(widget, screenPosition);
-
-                if (result)
+                foreach (var widget in touchedWidgets)
```
pauldendulk: In general you never need to check on count before iterating over a list. The performance gain will be minimal and it maybe detrimental if there are more then zero items. The code is cleaner without.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -309,12 +309,16 @@ private void OnInfo(MapInfoEventArgs mapInfoEventArgs)
 
         private bool WidgetTouched(IWidget widget, Point screenPosition)
         {
-            if (widget is Hyperlink hyperlink)
+            var result = widget.HandleWidgetTouched(Navigator, screenPosition);
+
+            if (!result && widget is Hyperlink hyperlink && !string.IsNullOrWhiteSpace(hyperlink.Url))
             {
                 OpenBrowser(hyperlink.Url);
```
pauldendulk: I think you already mentioned this. I guess the OpenBrowser event should be inside the widget. The problem we need to solve is that OpenBrowser is platform dependent. We should use some kind dependency injection to fix this.

Mapsui.UI.Android/MapControl.cs
```diff
@@ -65,14 +65,16 @@ public void Initialize()
         private void OnDoubleTapped(object sender, GestureDetector.DoubleTapEventArgs e)
         {
             var position = GetScreenPosition(e.Event, this);
-            OnInfo(InvokeInfo(Map.Layers.Where(l => l.IsMapInfoLayer), Map.Widgets, Viewport, 
+            var widgets = Map.Layers.Select(l => l.Attribution).Where(a => a != null).Concat(Map.Widgets).ToList();
```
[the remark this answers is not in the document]
pauldendulk: I just added a Map.GetWidgetsOfMapAndLayers() to fix this.

Mapsui.UI.Shared/MapControl.cs
charlenni: I don‘t think so. The new concept is: first check, if the user has its own handling for a click on the attribution (show the attribution in an app own screen). If it is handled, than Mapsui does nothing. If it isn‘t handled, than Mapsui does it (by OpenBrowser).

Mapsui.UI.Shared/MapControl.cs
charlenni: That‘s ok. Do you do the changes or should I?

Mapsui.UI.Shared/MapControl.cs
pauldendulk: It's nicer if the MapControl does not need to know anything about the kind of control. The 'is handled' logic will still be relevant.

Mapsui.UI.Shared/MapControl.cs
pauldendulk: I will fix this now.

Mapsui.UI.Shared/MapControl.cs
pauldendulk: Since this is not something we need to fix now I will resolve this conversation (just noticed the resolve conversation button and guess I should use it).

Issue #583, opened 2019-01-22
System.NotSupportedException: Could not activate JNI Handle


johnpudd:
Hi, 

I've been banging my head against the wall on this while - I'm not sure if this is my environment... 

Running the Sample.Droid throws the following error in MainActivity 
`SetContentView(Resource.Layout.activity_main);`

`System.NotSupportedException: Could not activate JNI Handle 0x7ffa0c3f60 (key_handle 0x2d21799) of Java type 'md5c9c6bce245d1b7c0019d6ddca5bc6a77/MapControl' as managed type 'Mapsui.UI.Android.MapControl'.`

pauldendulk:
This is current master?

johnpudd:
In the immortal words of Alan Partridge (and ABBA before him)... AHAA!!!

[MapControl.cs#L235](https://github.com/Mapsui/Mapsui/blob/6d04f8d76b3aba385abcf4d8cf0cf65b7fc6d689/Mapsui.UI.Shared/MapControl.cs#L235)
`            if (!_map.Initialized && _map != null && _viewport.HasSize && _map?.Envelope != null)`

should be
`            if (_map != null && !_map.Initialized && _viewport.HasSize && _map?.Envelope != null)`

johnpudd:
THanks @pauldendulk

PR #584, opened 2019-01-22
 #583 - reorder _map != null check on CallHomeIfNeeded

johnpudd:


pauldendulk:
oh! let me merge this quickly before anyone else sees my stupid mistake.

PR #587, opened 2019-01-28
Update Mapsui.Samples.Forms.sln

bertt:


charlenni:
I assume, that it should be removed. It is very old and not used in this project.

PR #588, opened 2019-01-30
Added ctors for MemoryLayer

charlenni:


charlenni:
But the problem still exists. All users use Layer, because they think, this is the right layer. Layer should be called FetchLayer to make this clear.

pauldendulk:
Yes perhaps.

PR #591, opened 2019-02-03
Added a NeedsRefresh flag to DataChangedEventArgs

charlenni:
Added a NeedsRefresh flag to DataChangedEventArgs. With this, it is possible, that MemoryLayer isn't refreshed all the time.
Removed a debug WriteLine for Forms implementation.

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -150,8 +150,6 @@ namespace Mapsui.UI.Forms
                 // Is this a fling or swipe?
                 if (velocityX > 10000 || velocityY > 10000)
                 {
-                    System.Diagnostics.Debug.WriteLine($"Velocity X = {velocityX}, Velocity Y = {velocityY}");
-
                     e.Handled = OnFlinged(velocityX, velocityY);
                 }
 
@@ -428,7 +426,6 @@ namespace Mapsui.UI.Forms
             // Last touch released
             if (touchPoints.Count == 0)
             {
-                RefreshGraphics();
                 _mode = TouchMode.None;
                 _map.RefreshData(_viewport.Extent, _viewport.Resolution, true);
             }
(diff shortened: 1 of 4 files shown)


pauldendulk:
Perhaps MemoryLayer should not call DataChanged at all. What purpose would it have? 

The fact that it is called probably means something depends on the event. Perhaps that should be solved in a different way. What breaks if we remove the OnDataChanged on RefreshData.

charlenni:
But even the data of a MemoryLayer could change. Or isn't this possible?

pauldendulk:
Yes, but then it should call DataChanged when it actually changes. Not just when RefreshData is called and nothing changes. 

I now think I can remember I used RefreshData to trigger a refresh from the Navigator. After recent changes Refresh is evoked through a callback. So, I think we should just remove the MemoryLayer OnDataChanged and fix any issues that might follow from this, if any.

The WritableLayer has a SignalDataChanged to trigger OnDataChanged. It seemed like a workaround when I added it. A better solution might be that a Layer encapsulates the data and triggers OnDataChanged itself. But lets just add DataHasChanged() methods to the ILayer and IProvider interfaces so a refresh could be triggered from anywhere.

pauldendulk:
I plan to add the DataHasChanged() method soon, and remove the DataChanged event in the MemoryLayer. Because of this the NeedsRefresh field in this PR will not be useful anymore, but perhaps there are other changes in this PR you would like to commit. Perhaps you could create a new PR for that.

pauldendulk:
Also important. The fact that iOS gets performance problems because of too many invalidate call is a bug in itself. Perhaps in SkiaSharp.Views or Xamarin.

pauldendulk:
I'll close this one. If there is anything else in this MR that needs merging please add it to another MR.

charlenni:
Thank you for reviewing this.

Perhaps @Sebastian1989101 could test the next beta and report, if the screen update is fast enough.

Issue #603, opened 2019-02-19
SelectedPin NullReferenceException in HandlerPinsOnCollectionChanged


lumberjack4:
If removing Pins from the collection when there is no SelectedPin, a null reference exception will occur.

To Reproduce:
1. Run Pin Sample
2. Place a few pins on the map without selecting any of them
3. Remove pins in code
4. Exception

I modified the triple click to do:
```
// Yes I know you would normally call mapView.Pins.Clear(), this is for illustrative purposes
foreach (var existingPin in mapView.Pins.ToList())
{
    mapView.Pins.Remove(existingPin); 
}
```

I have fixed it in my repo (4a92e2688821807d2e5df74ef11920a5384268b2)  A pull request will be forthcoming.

pauldendulk:
Thanks, clean simple fix. I merged it.

pauldendulk:
Your fix is in the new release https://github.com/Mapsui/Mapsui/releases/tag/2.0.0-beta.19

PR #604, opened 2019-02-19
SelectedPin NullReferenceException

lumberjack4:
#603 
Added null check to SelectedPin


pauldendulk:
Thanks, under what circumstance does this happen?

Issue #608, opened 2019-02-21
Pin visibility not honored on initial creation in Mapsui.Forms


lumberjack4:
In my app I have a fixed set of pins that I can show on the map.  During map initialization, I want to go ahead and load the pins onto a map with the visibility set to false, so the user can place, show, and hide them at will.

When a pin is created the visibility is treated as true regardless of what the user specifies.  Once a pin is created, the visibility is correctly handled when changing the `IsVisible` property.  This is due to not setting the enabled flag of the underlying style on pin creation (it is handled on the property changed event).

The change is located here (7d14fafbb141883f6fba38ad3ecc00eb36b700c3).

pauldendulk:
Thanks, merged it.

PR #609, opened 2019-02-21
Apply Pin visibility on creation

lumberjack4:
#608 
Sets the pin's style to be enabled/disabled based on the `IsVisible` flag.


charlenni:
@pauldendulk I thought, I approved this PR. But it seems, that this didn‘t worked. Something I did wrong?

pauldendulk:
The review and the merge are separate. First you have to do a review and approve (as you did) and then you have to press the merge button below, and then confirm merge.

This is so you could setup a process where multiple reviewers have to approve before you can merge.

charlenni:
Ah, ok, I didn‘t see this. Perhaps next time 😉

Issue #607, opened 2019-02-21
SymbolRotation not rotating with SkiaSharp renderer


wjax:
As simple as it sounds.

If you create a symbol, add a style with a rectangle or square and try to rotate it wit SymbolRotation, it only works with WPF renderer but not with default Skia Renderer

charlenni:
You used a VectorStyle or a SymbolStyle? And if SymbolStyle, which type of symbol do you use? 

Made a PR (#612) for the later one.

wjax:
Use SymbolStyle created with this method:


private IFeature CreatePointWithStackedStyles()
        {
            point = new Feature { Geometry = SphericalMercator.FromLonLat(-6, 37) };

            point.Styles.Add(new SymbolStyle
            {
                SymbolScale = 2.0f,
                SymbolType = SymbolType.Triangle,
                Fill = null,
                Outline = new Pen { Color = Color.Yellow }
            });

            point.Styles.Add(new SymbolStyle
            {
                SymbolScale = 0.8f,
                Fill = new Brush { Color = Color.Red }
            });

            point.Styles.Add(new SymbolStyle
            {
                SymbolScale = 0.5f,
                Fill = new Brush { Color = Color.Black }
            });

            point.Styles.Add(new SymbolStyle
            {
                SymbolType = SymbolType.Rectangle,
                SymbolScale = 0.2,
                SymbolRotation = 0,
                Fill = new Brush { Color = Color.Blue }
            });

            point.Styles.Add(new LabelStyle
            {
                Text = "Stacked Styles",
                HorizontalAlignment = LabelStyle.HorizontalAlignmentEnum.Left,
            });

            return point;
        }

pauldendulk:
@wjax There are some issues but it seems to me that in beta 20 you can use orientation of vehicles if you use bitmap symbols instead of SymbolType.Triangle, providing you don' t use map rotation. The issues will be fixed before the 2.0 release.

pauldendulk:
With the above commit vehicle orientation should now work properly for bitmap symbols and default symbols. When map rotation is used SymbolStyle.RotateWithMap should be set to true.

PR #612, opened 2019-02-22
Symbol rotation

charlenni:
Should solve #607.


charlenni:
@pauldendulk We should discuss this theme further. There are more questions like Symbols should have two states: rotate with map (show always with the SymbolRotation angle to north) or stay with SymbolRotation to the viewport. I saw, that you calculate the rotation angle by MapRotation minus SymbolRotation. That's the first case. I now implemented the second case, because the Skia renderer doesn't have the map rotation. Or better, I didn't find it.

pauldendulk:
Did not dig into this but just one remark that might be related. 

There is a rendering unit test for rotation called RenderRotatedBitmapSymbolWithOffset which initialized a map with BitmapSymbolWithRotationAndOffsetSample.CreateMap(). SymbolRotation is set on a SymbolStyle with a bitmap. The test output shows rotation as expected. There is probably something else going wrong here but the test might be good for reference.

pauldendulk:
I understand your remark now. I counter rotate for map rotation but this is not always what you want. When you use pins this is correct. Even if you rotate the map the pins should be straight up. When you use vehicles this is not correct. You want them to point in the direction they move, so rotating with the map. Perhaps a RotationType field is needed.

PR #622, opened 2019-03-03
Forms MapControl Fixes

radderz:
@charlenni I finally have some time to play around with the forms side of mapsui, just updated to the beta to get started converting our mapsui usage over to v2. Have found a few issues that I have had to get around in hacky ways that we could discuss?

First is just a bug:

- Add math.abs around velocity decision for flight as it would only allow fling in specific directions. Only certain fling directions were allowed based on the logic.

Ideas:

- I have implemented double tap zoom using the xamarin.forms animation library that I could bring into this. When double tapping somewhere on the screen it will zoom about that point smootly over a short time.

- I am also reimplimenting fling also using the xamarin.forms library (although I am having issues with this compared with the native guesture recognisers). This could also be brought in to enable natual fling movements on the map if I can get it working. I am finding the flings are pretty long

I am only using the MapControl as the other one is pretty limiting. Have you seen the tearing issues when playing around with the mapcontrol?

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -150,7 +150,7 @@ namespace Mapsui.UI.Forms
                 // Is this a fling or swipe?
                 if (_touches.Count == 0)
                 {
-                    if (velocityX > 10000 || velocityY > 10000)
+                    if (Math.Abs(velocityX) > 10000 || Math.Abs(velocityY) > 10000)
                     {
                         // This was the last finger on screen, so this is a fling
                         e.Handled = OnFlinged(velocityX, velocityY);


radderz:
I think the fling should need a minimum total movement across the screen before fling is initiated, as on samsung devices with sensitive touch, it can do big flings just from tiny movements

pauldendulk:
"tearing issues" ? Is that the white edges you see in the map while dragging the map around? Is saw this yesterday. Is this is MapView as well as MapControl?

charlenni:
Hi Mark,

happy to hear from you.

Yes, the missing math.abs is really a bug. I didn’t find it, because I don’t use it up to now. And I have problems with this part of the code. If you look into the code, you will see, that you could get the velocity for each finger you use. I changed the code for the Samsung bug, but didn’t updated this part. I should do this. Than it would be useful. Perhaps I could add a value for the “length” of the touch.

The double tap is already there and waits for a default action if the user doesn’t handle it by herself/himself. So it would be easy to add a call to the standard zoom function of Mapsui. This should be a one line change. We could add a flag, if the user like to have this functionality. 

The next is the animation. I would like to see any form of animation in Mapsui shared code. Then it would be possible to use it for all MapControls, not only the Forms one. Than there should be a optional flag in all Navigator functions, so that all this movements could be animated. Perhaps with different „easings“ (like the Goggle Earth movement when changing location, which is getting faster and from the half slowing down). See Xamarin.Forms [documentation.](https://docs.microsoft.com/de-de/xamarin/xamarin-forms/user-interface/animation/easing).

So the question is, who wants to implement this? Any volunteers? 🤔

@pauldendulk Perhaps I should explain the differences of MapControl and MapView, because it could be a little bit confusing. MapControl should have the same functionality as all other MapControls. It uses the same coordinates. It interprets gestures and holds the Map object. MapView adds some functionality to mimic the Xamarin.Forms MapView like coordinates in WGS84, pins and a my location layer. So you normally use MapView in a Forms environment, but handling problems are often in MapControl. And users often mention problems with e.g. MapControl on Android. Than I have first to ask, if they use the native MapControl for Android or Forms and, if it is Forms, if they use a renderer around the native MapControl or the native Forms MapControl. Very confusing 😎

Ok, I see this “tearing” too. I thought, that it is my Android device, because I didn’t saw them on the iPhone. I see white horizontal or vertical lines when moving fast around. Do both of you see the same?

radderz:
> "tearing issues" ? Is that the white edges you see in the map while dragging the map around? Is saw this yesterday. Is this is MapView as well as MapControl?

@pauldendulk yea that's right, seems to be a rendering issues where your see maybe a partially rendered canvas, never had that issue before updating to v2. You can see it easily when doing a zoom animation on android.

It can also happen when doing agressive pinch zoom rotate move types motions

radderz:
@charlenni for the animations, I am just trying to get fling to work, but the animations are pretty easy in forms as it has a built in animations library. I had also implemented the same in the old mapsui for ios, uwp and android native views but since the project was using forms I used that libraries animation tools. 

Perhaps we could use the Xamarin.Forms.Animation code in mapsui if its got the right licence? It works well and is pretty easy to use.

pauldendulk:
I have seen the tearing in a sample on a callout issue. It was quite noticeable, since it was not mentioned before I assumed it was a specific issue for the app. 

Related to this. I should upgrade to the latest SkiaSharp version. This to be sure our current version does not cause this issue.

charlenni:
@radderz I looked today into the code and found, that the Forms MapControl allready has a double tap zoom implementation. But the zoom in and zoom out functions are not implemented correctly. If I do this and you set `UseDoubleTap` to `true`, than you could use double tap.

I implemented the zoom functions not, because I didn't have an idea, on which amount the resolution should changed? Next resolution? Or doubled/halfed? But now I found, that the new Navigator implementation got a `ZoomIn` and `ZoomOut` function. So I could use them.

I try to make a PR for this.

charlenni:
Made #625 for double tap. Could be tested with MyLocation sample.

charlenni:
@radderz I thought a while about the animation. I'm not sure, that we need such a big solution than Xamarin.Forms. We only want to animate movements of the map. This all happens in the Navigator. There we need two animations: one for the position and one for the resolution. If you want a movement like Google Earth, than you start two things: an animation to the new center and a zoom out/zoom in combination. Double tap with zoom in would need only a zoom in with animation. There could be standard animations, which could be used, or you create your own animation by providing a function, which calculates coordinates or resolution by a given value between 0 and 1. Easing could be made by adding an extra function, which change the 0-1 value over time before it is given to the user function for calculation.

Do you think, that would work and is enough for this?

charlenni:
I thought about the flicker. I assume, that it is an OpenGL problem. I saw it while testing the animation on UWP too. We didn't saw it on Mapsui 1.4.8, because it uses the CPU backend.

pauldendulk:
I will update to the new SkiaSharp soon. If the problem is still there we might need to create a minimal sample to reproduce it and submit a SkiaSharp issue.

PR #625, opened 2019-03-06
Added zoom in for double tap on Forms MapControl

charlenni:


charlenni:
@pauldendulk Do you have a suggestion, why AppVeyor build fails? I used always the newest master and changed only a few lines.

pauldendulk:
@charlenni I assume it is the itinero sample I added. Have to check.

pauldendulk:
@charlenni  Removing the Itinero sample fixed the buildserver fail.

pauldendulk:
@charlenni  Removing the Itinero sample fixed the buildserver fail.

Issue #598, opened 2019-02-08
How to draw layers with ordering priority?


leo-costa:
Hi 

I know Mapsui draw layers based on the order in which they were added to the map layer collection.
I have a user case now where some layers must be dynamically inserted into the map. Let's say there are 5 layers that needs to be inserted at a specific location. This is my current logic 

1. Have a 'dummy / marker' layer with a name inserted ahead of time.
2. When a dynamic layer comes, I need to find where this dummy layer is using ```FindLayer``` method
3. I also need to find the index of this dummy layer
4. Using the information above I insert my dynamic layer after the dummy one using ```Insert``` method from ```Map.Layers```

Is there a better way of doing this ? Perhaps a layer ordering could be implemented instead of just iterating though layers when drawing ?

pauldendulk:
There is no real easy way to do that right now. We could try to think of something better.

pauldendulk:
@leo-costa This should be improved. I am trying to get a better picture of the user's perspective. The LayerCollection also has Insert and Move methods but these are now sufficient. What kind of method would help you? 

For instance:
```InsertBefore(ILayer referenceLayer, ILayer layerToInsert)```
```InsertAfter(ILayer referenceLayer, ILayer layerToInsert)```

pauldendulk:
We also need to find a proper solution to reference a layer. We could reference to an instance, like above, it would do reference compare. But it the layers are read from disk this would not help. So we need some other identifier. 
- BaseLayer has an Id which is unique as long as layers are derived from BaseLayer, but a user could implement it's own base layer. 
- We could enforce 'Name' to be unique. But if data is read from disk or imported this might be problematic. It would not be possible to import certain data. Perhaps we could also add a Title which is not required to be unique.

pauldendulk:
We could also add an Order field managed by the user, but this is not always easy. If you want to insert something you need to update the Order fields of the layers following the inserted layer.

We could also manage the order field from the LayerCollection. Basically that is the layer index. Not sure if this would help. If we can add some extra methods so that the user does not need to bother with the Order or Index that would be best.

pauldendulk:
It would also be interesting to know how other libraries solve this.

charlenni:
I looked into this. 

Google Maps API for JavaScript does it [this](https://developers.google.com/maps/documentation/javascript/customoverlays) way. There are more than one layer type. 

[...]

PR #626, opened 2019-03-06
Added a property Level to each layer

charlenni:
Layers with higher level are drawn over layers with lower levels. When two or more layers have the same level, than the later added one is drawn on top of the earlier added one.

Not tested very well. Is only to show, how it could be implemented (see #598).


pauldendulk:
Thanks for the sample. It helps in the discussion but it looks like we are now going another route.

PR #627, opened 2019-03-07
Added an animation to Navigator and a FlyTo() function

charlenni:
Cave: this is not ready. It is only a first step for #622 and fling.

Added an animation to Navigator and a FlyTo() function for testing this animation.

@radderz Perhaps you could take a look into this. Perhaps this helps. I used the easing functions from Xamarin.Forms, but did the rest by my own. For fling, it should be possible to do it by a NavigateTo() function with a special easing. Do you know, which one to use? And how do you calculate the destination point by the velocity?


charlenni:
I thought about it. I have to add an animation for rotation too. So I have to do some rework on this. But for a test it would work.

charlenni:
Ok, now I added a class for Animation. This class is used by an AnimatedNavigator.

pauldendulk:
Lets do this in v3 beta, which could be very soon. I want to publish the current version as a release candidate. I am now updating the SkiaSharp nuget packages and we need to deal with the last real bugs

radderz:
@charlenni - fling is pretty wrong anyway, I have a good fling logic I can share for android and ios native applications. In forms I may have to go back to the native renderer logic I had as the fling speed numbers are bogus currently in mapsui.forms. if we can ge the mapsui fling speed to be the same as in the android one, then my logic would work for all platforms and gives a natural deceleration.

radderz:
@pauldendulk @charlenni 

This is the current issue with the velocity tracker that makes it nonsensical,

Using a linear regression causes an issue where if the finger was decelerating the last 200 ms, then the velocity gets regressed into a positive velocity which isn't real.

image

If you see the above image you'll see the 7 samples of velocities between the touch movements, and yet result is a net positive velocity, even though all of the sample velocities are negative values. I am not so clued up on the way velocity trackers work, but maybe we can find the Android source code for the built in tracker as this one doesn't work as intended.

For now I am going to have to revert back to the native ones with renderers till the velocity tracker is fixed as I don't think I'll be able to dedicate the time to understand it well enough to fix it.

If someone can fix the velocity tracker to spit out a realistic value that is similar to the Android and iOS native implementations I'll be able to implement fling, zoom to, navigate to as I already have done so.

Mark

charlenni:
@radderz Ok, that is my VelocityTracker. It seems, that there are some bugs ;)

Perhaps I could find some better implementation [here](https://github.com/mozilla/newtab-dev/blob/master/widget/gonk/libui/VelocityTracker.cpp#L393). But if I get it right, only the sign (positive or negative) is wrong. Or do you say, also the value is wrong?

radderz:
@charlenni - It isn't just the sign, in that example above, I did a load of flicks in the "-" direction, which worked fine, until I hit that specific case where the direct flips way due to the deceleration. I also think the value may be wrong as I had to hack and scale things really differently to the native one to get it feel remotely right and it wasn't consistant so I guess the approach may not be right. I think a good idea would be to have a renderer, that ouput the native values for now that you can use to compare vs your velocity tracker?

Is there source code for the Android native ones?

charlenni:
Ok, we have two different problems:
1. Get the correct velocity for the different touch events
2. Animate the movement of the map

Here we discuss only the first point. It is difficult to get the native values, because the velocity tracker only exists in the Forms implementation. So I have to create an own tracker with its own logic to calculate the correct velocity. Any ideas are welcome.

The second point should already work with this PR.

radderz:
@charlenni 
I have managed to get a renderer to generate valid fling values. You could use this to compare the values you have had up till now. We could do this using a dependance injection that overrides the existing fling code but it should be doable in c# code assuming you can find the right way of doing it.
[TileMapViewRenderer.zip](https://github.com/Mapsui/Mapsui/files/3036129/TileMapViewRenderer.zip)

TileMapView is just a content view holding the map control. I recreate the touch events in the android native type, and pass through to the built in guesture detector. From there you can output the X, Y velocities (in native points not scaled to skia scale).

charlenni:
Thank you for the code.

I debugged my code, and it seems, that the velocities aren’t calculated correct. I try to find the bug.

radderz:
@charlenni - looking at the c++ code within android, it basically just accepts the most recent speed. I have tested this on android and ios and it seems ok. I have changed VelocityTracker to a FlingTracker which just calculates the velocity between the last two tracked mouse events. Feels natural. Check out my pull request.

#646

charlenni:
@radderz Thank you for your help. In the moment I’m a little bit busy. I will try it in the next days.

@pauldendulk Do you think it is possible to change the Navigator, so that it is possible to add its own? Than we could test the animated Navigator in 2.0 too.

pauldendulk:
@charlenni You mean it should be possible to set a custom Navigator?

charlenni:
@pauldendulk Yes, that would be perfect. With this, someone could take the AnimatedNavigator.cs and use it with Mapsui 2.0 to test it and use it.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -277,7 +277,7 @@ public Map Map
                 if (_map != null)
                 {
                     SubscribeToMapEvents(_map);
-                    Navigator = new Navigator(_map, _viewport);
+                    Navigator = new AnimatedNavigator(_map, _viewport);
```
radderz: @charlenni can you change this back to Navigator? If you do we should be able to merge this in as then the default behaviour wouldn't change. Now with that other pull request finished you'll be able to set the navigator to be an animated one if desired.

charlenni:
I assume, that I looked on different implementations and than created an own version. A small animation with only the needed parts.

Issue #606, opened 2019-02-20
Accessing Pin.Callout leads into endless loop with latest Xamarin.Forms


mmichelbb:
I have compiled Mapsui.Forms against the latest version of Xamarin.Forms. 
Whenever I tried accessing the Callout on a pin (doing it the way as described in Forms.Shared PinSample), my app got stuck in an infinity loop and therefore doing nothing anymore.

I suspect that this is caused by line 457 in Mapsui.UI.Forms.MapView.cs:
`while (_callout == null) ;`

For now my solution is to create and populate the callout manually using reflection and injecting it into Pin._callout afterwards, so that MapView._callout is never called in the first place.

Here is the code I am using to populate the pins:

`
		private Pin CreatePin(MapView mapView, int pinNumber, string label, string address, double latitude, double longitude, IconTypes iconType)
		{
			var assembly = typeof(App).GetTypeInfo().Assembly;
			Position position = new Mapsui.UI.Forms.Position(latitude, longitude);
			Pin pin = new Pin(mapView)
			{
				Label = label,
				Address = address,
				Position = position,
				IsVisible = true,
				Type = PinType.Icon,
				Scale = 0.3f,
				Transparency = 0f
			};

			pin.Icon = assembly.GetManifestResourceStream(iconType.ToAssemblyFile()).ToBytes();
			pin.Feature.Geometry = SphericalMercator.FromLonLat(position.Longitude, position.Latitude);
			pin.Feature["name"] = label;

			//hack to ensure that _mapView.CreateCallout doesn't get called because Pin._callout is not null
			FieldInfo privateCalloutProp = pin.GetType().GetField("_callout", BindingFlags.NonPublic | BindingFlags.Instance);
			if (privateCalloutProp != null)
			{
				FieldInfo mapControlProp = mapView.GetType().GetField("_mapControl", BindingFlags.NonPublic | BindingFlags.Instance);
				if (mapControlProp != null)
				{
					Callout callout = new Callout((MapControl)mapControlProp.GetValue(mapView))
					{
						RectRadius = 5,
						ArrowHeight = 20,
						ArrowWidth = 20,
						ArrowAlignment = ArrowAlignment.Top,
						ArrowPosition = 0.5f,
						Title = label,
						Subtitle = address,
						Type = !string.IsNullOrEmpty(address) ? CalloutType.Detail : CalloutType.Single
					};
					if (!string.IsNullOrEmpty(address))
					{
[...]

PR #631, opened 2019-03-11
CHG: Fix candidate for #606

mmichelbb:
Accessing Pin.Callout leads into endless loop with latest Xamarin.Forms (Replaced while loop with thread-safe sync lock)

Issue occurred with latest Xamarin.Forms when using a .net Standard library for shared app code.


pauldendulk:
Thanks for the PR but it looks like there is a simpler solution now.

PR #646, opened 2019-03-31
Fix info event on forms version and fix fling velocity in forms

radderz:
The info event never got called in the forms map control. This calls it if it exists, to keep the same logic as the native versions.

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -142,14 +142,14 @@ namespace Mapsui.UI.Forms
                 var releasedTouch = _touches[e.Id];
                 _touches.Remove(e.Id);
 
-                double velocityX;
-                double velocityY;
-
-                (velocityX, velocityY) = _velocityTracker.CalcVelocity(e.Id, ticks);
-
                 // Is this a fling or swipe?
                 if (_touches.Count == 0)
                 {
+                    double velocityX;
+                    double velocityY;
+
+                    (velocityX, velocityY) = _velocityTracker.CalcVelocity(e.Id, ticks);
+
                     if (Math.Abs(velocityX) > 200 || Math.Abs(velocityY) > 200)
                     {
                         // This was the last finger on screen, so this is a fling
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -579,10 +579,13 @@ namespace Mapsui.UI.Forms
             if (args.Handled)
                 return true;
 
-            var eventReturn = InvokeInfo(screenPosition, screenPosition, 1);
+            var infoToInvoke = InvokeInfo(screenPosition, screenPosition, 1);
 
-            if (eventReturn != null)
-                return eventReturn.Handled;
+            if (infoToInvoke != null && !infoToInvoke.Handled)
+            {
+                OnInfo(infoToInvoke);
+                return infoToInvoke.Handled;
+            }
 
             return false;
         }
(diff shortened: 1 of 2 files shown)


radderz:
@charlenni - try out the fling velocity tracker with an animation. Use the below example.

`        public void OnFling(object sender, Mapsui.UI.SwipedEventArgs e)
        {
            e.Handled = true;
            var velocityX = -e.VelocityX;// / _scale;// / 5;
            var velocityY = -e.VelocityY;// / _scale;// / 5;

            var magnitudeOfV = Math.Sqrt((velocityX * velocityX) + (velocityY * velocityY));

            var animateMillis = magnitudeOfV / 10;
            
            if (magnitudeOfV < 200)
                return;

            e.Handled = true;

            this.AbortAnimation("FlingAnimation");

            Debug.WriteLine($"Animating Fling - X:{velocityX} - Y:{velocityY} - { animateMillis }ms");

            var i = 0;

            var animation = new Animation((v) =>
            {
                i++;
                try
                {
                    lock (_lockObject)
                    {
                        if (_mapControl == null)
                            return;
                        
                        var percentageV = v / 1000d;
                        var timeAmount = 16 / 1000d; // 16 millis

                        var xMovement = velocityX / _scale * percentageV * timeAmount;
                        var yMovement = velocityY / _scale * percentageV * timeAmount;

                        if (xMovement.IsNanOrInfOrZero())
                            xMovement = 0;
                        if (yMovement.IsNanOrInfOrZero())
                            yMovement = 0;

                        if (xMovement == 0 && yMovement == 0)
                            return;

                        var previous = _mapControl.Viewport.ScreenToWorld(0, 0);
                        var current = _mapControl.Viewport.ScreenToWorld(xMovement, yMovement);

                        var xDiff = current.X - previous.X;
                        var yDiff = current.Y - previous.Y;

                        var newX = _mapControl.Viewport.Center.X + xDiff;
                        var newY = _mapControl.Viewport.Center.Y + yDiff;

                        Debug.WriteLine($"Animating Fling - {i} - {xMovement}, {yMovement} - {xDiff}, {yDiff} - Current Center = {_mapControl.Viewport.Center.X}, {_mapControl.Viewport.Center.Y}");

                        _mapControl.Navigator.CenterOn(new Mapsui.Geometries.Point(newX, newY));
                    }
                }
                catch (Exception err)
                {

                }
            }, 1000d, 0, Easing.SinOut);

            animation.Commit(this, "FlingAnimation", 16, (uint)animateMillis, finished: ((d, b) => Debug.WriteLine($"Fling Finished - {i} [{d},{b}]")));
        }`

radderz:
@charlenni - this needs a little more work, on more sensitive screens sometimes the fling doesn't work. On ios it works nicely. Might need to average the last few samples up rather than just using last two.

pauldendulk:
I guess the missing info event on the XF MapControl went unnoticed so far because people are using the MapView which does have a info event. One possible issue comes to mind (without looking at the code), does the MapView still fire one info event after this fix, or will it fire both the MapView and MapControl info event?

charlenni:
@radderz I read about this. The problem could be, that on some devices, the last few events are all at the same place.

radderz:
> I guess the missing info event on the XF MapControl went unnoticed so far because people are using the MapView which does have a info event. One possible issue comes to mind (without looking at the code), does the MapView still fire one info event after this fix, or will it fire both the MapView and MapControl info event?

It should still only be called once, before it wouldn't have ever been called as the underlying mapcontrol never called info in the first place, the HandlerInfo couldn't have been called. In MapView there is the handletap which also called handlerinfo, but this should only get here if info doesn't mark the event as handled.

PR #660, opened 2019-05-16
Changed setter of Navigator in MapControl

charlenni:
Changed setter of Navigator in MapControl from private to public, so that it is possible to use own implementations of INavigator.

### DIFF
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -95,8 +95,12 @@ namespace Mapsui.UI.Wpf
         public INavigator Navigator
         {
             get => _navigator;
-            private set
+            set
             {
+                if (_navigator != null)
+                {
+                    _navigator.Navigated -= Navigated;
+                }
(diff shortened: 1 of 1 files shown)


radderz:
@charlenni not specifically, having it not changed over part way through the program running would just reduce any chance of unexpected behaviour unless you think you'd need to change the navigator part way through?

charlenni:
The problem is, that the default Navigator is created in the setter of Map in MapControl. So it is nearly impossible to set it without changing the API of MapControl. So I thought, that it would be the easiest to make the setter public. Property Map creates a default Navigator and you could change it later. Not the best solution, but the best for 2.0. So animated transitions could be made, but need no changes (except this repo) in 2.0.

Btw. the animated transitions are allready made (see [here](https://github.com/charlenni/Mapsui/blob/Animation/Mapsui/AnimatedNavigator.cs)) ;) I assume, that the most are suggestion from you.

radderz:
@charlenni ok, will merge it in

PR #671, opened 2019-06-25
Pull Request [190625]

zerocraft:
1.Fix Event Mapsui.UI.Forms.MapControl.OnLongTapped
2.Fix Mapsui.UI.Forms.MapView Clean Drawables On 0 Size
3.Add Event Mapsui.UI.Forms.MapView.PinLongTap (use PinClicked EventArgs)


radderz:
@zerocraft are you able to give a small blurb on what this is trying to achieve for each of the above points. Does this reference some bug somewhere?

pauldendulk:
@zerocraft Sorry for my very late reaction. The tap logic is too complicated at the moment. We should further simplify it ourselves first.

PR #675, opened 2019-07-09
Callout bug

jasells:
Fix for Issue #606 on UWP/WPF, at least.

### DIFF
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -495,7 +495,7 @@ namespace Mapsui.UI.Forms
             // since this seems to only be an issue on UWP (?), let's make our fix narrow. (Probably also an issue on WPF, though)
             if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
             {
-                // The root of this issue is that if the root of the cakk-stack reaching this method is a 
+                // The root of this issue is that if the bottom of the call-stack reaching this method is a 
                 // Touch event sourced from the MapControl, the internal SKGLView.Touch event is generated 
                 // on a Timer that is thread-based, and thus, not the UI thread (and Windows no like UI objects 
                 // created on worker threads).   
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -103,11 +103,79 @@ namespace Mapsui.UI.Forms
             {
                 if (PlatformRequiresUIcontext())
                 {
-                    Device.BeginInvokeOnMainThread(() => HandlerHover(s, e));
+                    Device.BeginInvokeOnMainThread(() => HandlerHovered(s, e));
                 }
                 else
                 {
-                    HandlerHover(s, e);
+                    HandlerHovered(s, e);
+                }
+            };
+
+            //new events added 
+            _mapControl.TouchStarted += (s, e) =>
+            {
+                if (PlatformRequiresUIcontext())
+                {
+                    Device.BeginInvokeOnMainThread(() => HandlerTouchStarted(s, e));
+                }
+                else
+                {
+                    HandlerTouchStarted(s, e);
+                }
+            };
+            _mapControl.TouchEnded += (s, e) =>
+            {
+                if (PlatformRequiresUIcontext())
+                {
+                    Device.BeginInvokeOnMainThread(() => HandlerTouchEnded(s, e));
+                }
+                else
+                {
+                    HandlerTouchEnded(s, e);
+                }
+            };
+            _mapControl.TouchMove += (s, e) =>
+            {
+                if (PlatformRequiresUIcontext())
+                {
+                    Device.BeginInvokeOnMainThread(() => HandlerTouchMove(s, e));
+                }
+                else
+                {
+                    HandlerTouchMove(s, e);
+                }
+            };
+            _mapControl.Swipe += (s, e) =>
+            {
+                if (PlatformRequiresUIcontext())
+                {
+                    Device.BeginInvokeOnMainThread(() => HandlerSwipe(s, e));
+                }
+                else
+                {
+                    HandlerSwipe(s, e);
+                }
+            };
+            _mapControl.Fling += (s, e) =>
+            {
+                if (PlatformRequiresUIcontext())
+                {
+                    Device.BeginInvokeOnMainThread(() => HandlerFling(s, e));
+                }
+                else
+                {
+                    HandlerFling(s, e);
+                }
+            };
+            _mapControl.Zoomed += (s, e) =>
+            {
+                if (PlatformRequiresUIcontext())
+                {
+                    Device.BeginInvokeOnMainThread(() => HandlerZoomed(s, e));
+                }
+                else
+                {
+                    HandlerZoomed(s, e);
                 }
             };
 
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -497,53 +497,10 @@ namespace Mapsui.UI.Forms
         /// <param name="position">Position of callout</param>
         public Callout CreateCallout(Position position)
         {
-            // this ref is only needed internally to this method, so let's scope it that way.
-            Callout _callout = null;
-
-            // since this seems to only be an issue on UWP (?), let's make our fix narrow. (Probably also an issue on WPF, though)
-            if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
-            {
-                //// wrap our mutex in a using block to clean it up asap.
-                //using (var manualReset = new System.Threading.ManualResetEventSlim(false))
-                //{
-                //    Device.BeginInvokeOnMainThread(() =>
-                //    {
-                        // the Callout ctor is doing all kinds of UI init, so this is the root of the cross-thread exception on UWP.
-                        _callout = CalloutFactory(position);
-
-                //        // signal the worker thread that we're done
-                //        manualReset.Set();
-                //    });
-
-                //    // since original impl had no timeout, we'll leave it that way here too...
-                //    // probably should fix this to have a timeout though?
-                //    manualReset.Wait(); // block the worker thread (better than spinning!)
-                //}
-                // This whole call stack would be better with await/async paradigm, but that requires larger refactor, as the caller is a property get method.
-            }
-            else // just create the callout object on the calling thread.
-            {
-                _callout = CalloutFactory(position);
-            }
-            //// My interpretation (PDD): This while keeps looping until the asynchronous call
-            //// above has created a callout.
-            //// An alternative might be to avoid CreateCallout from a non-ui thread by throwing
-            //// early.
-            //while (_callout == null) ;
-
-            //var result = _callout;
-            //_callout = null;
-
-            return _callout;
+            return new Callout(_mapControl) { Anchor = position }; ;
         }
       
-        /// <summary>
-        /// a method of convience to create a callout object completely and reduce copy-paste
-        /// </summary>
-        /// <param name="position"></param>
-        /// <returns></returns>
-        private Callout CalloutFactory(Position position) { return new Callout(_mapControl) { Anchor = position}; }
-       
+
         /// <summary>
         /// Shows given callout
         /// </summary>
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -66,10 +66,51 @@ namespace Mapsui.UI.Forms
             _mapControl.ViewportInitialized += HandlerViewportInitialized;
             _mapControl.Info += HandlerInfo;
             _mapControl.PropertyChanged += HandlerMapControlPropertyChanged;
-            _mapControl.SingleTap += HandlerTap;
-            _mapControl.DoubleTap += HandlerTap;
-            _mapControl.LongTap += HandlerLongTap;
-            _mapControl.Hovered += HandlerHover;
+
+            // Taps are being generated on worker thread in UWP (probably WPF too)
+            // since we don't know this is an issue on other platforms, apply a narrow fix
+            // can remove or make more explicit the platform check later...
+            EventHandler<TappedEventArgs> handler = (s, e) =>
+            {
+                if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
+                {
+                    Device.BeginInvokeOnMainThread(() => HandlerTap(s,e));
+                }
+                else
+                {
+                    HandlerTap(s, e);
+                }
+            };
+
+            _mapControl.SingleTap += handler;
+            _mapControl.DoubleTap += handler;
+
+            // do same check on these events...
+
+            _mapControl.LongTap += (s, e) =>
+            {
+                if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
+                {
+                    Device.BeginInvokeOnMainThread(() => HandlerLongTap(s, e));
+                }
+                else
+                {
+                    HandlerLongTap(s, e);
+                }
+            };
+
+            _mapControl.Hovered += (s, e) =>
+            {
+                if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
+                {
+                    Device.BeginInvokeOnMainThread(() => HandlerHover(s, e));
+                }
+                else
+                {
+                    HandlerHover(s, e);
+                }
+            };
+
             _mapControl.TouchMove += (s, e) =>
             {
                 Device.BeginInvokeOnMainThread(() => MyLocationFollow = false);
@@ -452,22 +493,22 @@ namespace Mapsui.UI.Forms
             // since this seems to only be an issue on UWP (?), let's make our fix narrow. (Probably also an issue on WPF, though)
             if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
             {
-                // wrap our mutex in a using block to clean it up asap.
-                using (var manualReset = new System.Threading.ManualResetEventSlim(false))
-                {
-                    Device.BeginInvokeOnMainThread(() =>
-                    {
+                //// wrap our mutex in a using block to clean it up asap.
+                //using (var manualReset = new System.Threading.ManualResetEventSlim(false))
+                //{
+                //    Device.BeginInvokeOnMainThread(() =>
+                //    {
                         // the Callout ctor is doing all kinds of UI init, so this is the root of the cross-thread exception on UWP.
                         _callout = CalloutFactory(position);
 
-                        // signal the worker thread that we're done
-                        manualReset.Set();
-                    });
+                //        // signal the worker thread that we're done
+                //        manualReset.Set();
+                //    });
 
-                    // since original impl had no timeout, we'll leave it that way here too...
-                    // probably should fix this to have a timeout though?
-                    manualReset.Wait(); // block the worker thread (better than spinning!)
-                }
+                //    // since original impl had no timeout, we'll leave it that way here too...
+                //    // probably should fix this to have a timeout though?
+                //    manualReset.Wait(); // block the worker thread (better than spinning!)
+                //}
                 // This whole call stack would be better with await/async paradigm, but that requires larger refactor, as the caller is a property get method.
             }
             else // just create the callout object on the calling thread.
@@ -485,44 +526,7 @@ namespace Mapsui.UI.Forms
 
             return _callout;
         }
-
-        public System.Threading.Tasks.Task<Callout> CreateCalloutAsync(Position position)
-        {
-            // a construct to properly bubble out either the result or an exception from this async task.
-            // The result *may* be new-ed on a different thread (the UI thread).
-            var completion = new System.Threading.Tasks.TaskCompletionSource<Callout>();
-
-            // since this seems to only be an issue on UWP (?), let's make our fix narrow. (Probably also an issue on WPF, though)
-            if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
-            {
-                // The root of this issue is that if the bottom of the call-stack reaching this method is a 
-                // Touch event sourced from the MapControl, the internal SKGLView.Touch event is generated 
-                // on a Timer that is thread-based, and thus, not the UI thread (and Windows no like UI objects 
-                // created on worker threads).   
-
-                Device.BeginInvokeOnMainThread(() =>
-                {
-                    try
-                    {
-                        //signal the calling thread that UI thread is done
-                        completion.SetResult(CalloutFactory(position));
-                    }
-                    catch (Exception e)
-                    {
-                        //signal the calling thread that UI thread had an error
-                        completion.SetException(e);
-                    }
-                });
-            }
-            else
-            {
-                // complete immediately, we're not on Windows
-                completion.SetResult(CalloutFactory(position));
-            }
-
-            return completion.Task;
-        }
-
+      
         /// <summary>
         /// a method of convience to create a callout object completely and reduce copy-paste
         /// </summary>
@@ -892,14 +896,14 @@ namespace Mapsui.UI.Forms
                 {
                     var args = new MapClickedEventArgs(_mapControl.Viewport.ScreenToWorld(e.ScreenPosition).ToForms(), e.NumOfTaps);
 
-                    MapClicked?.Invoke(this, args);
 
-                    if (args.Handled)
-                    {
-                        e.Handled = true;
-                        return;
-                    }
+                        MapClicked?.Invoke(this, args);
 
+                        if (args.Handled)
+                        {
+                            e.Handled = true;
+                            return;
+                        }
                     // Event isn't handled up to now.
                     // Than look, what we could do.
 
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -72,9 +72,9 @@ namespace Mapsui.UI.Forms
             // can remove or make more explicit the platform check later...
             EventHandler<TappedEventArgs> handler = (s, e) =>
             {
-                if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
+                if (PlatformRequiresUIcontext())
                 {
-                    Device.BeginInvokeOnMainThread(() => HandlerTap(s,e));
+                    Device.BeginInvokeOnMainThread(() => HandlerTap(s, e));
                 }
                 else
                 {
@@ -89,7 +89,7 @@ namespace Mapsui.UI.Forms
 
             _mapControl.LongTap += (s, e) =>
             {
-                if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
+                if (PlatformRequiresUIcontext())
                 {
                     Device.BeginInvokeOnMainThread(() => HandlerLongTap(s, e));
                 }
@@ -101,7 +101,7 @@ namespace Mapsui.UI.Forms
 
             _mapControl.Hovered += (s, e) =>
             {
-                if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
+                if (PlatformRequiresUIcontext())
                 {
                     Device.BeginInvokeOnMainThread(() => HandlerHover(s, e));
                 }
@@ -194,6 +194,16 @@ namespace Mapsui.UI.Forms
 
             _mapDrawableLayer.DataSource = new ObservableCollectionProvider<Drawable>(_drawable);
             _mapDrawableLayer.Style = null;  // We don't want a global style for this layer
+        }
+
+        /// <summary>
+        /// helper method to centralize checking the platform for need to switch context to UI thread.
+        /// </summary>
+        /// <returns></returns>
+        private static bool PlatformRequiresUIcontext()
+        {
+            return Device.RuntimePlatform.Equals(Device.UWP) 
+                || Device.RuntimePlatform.Equals(Device.WPF);
         }
 
         #region Events
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -439,8 +439,6 @@ namespace Mapsui.UI.Forms
 
         #region Callouts
 
-        private Callout _callout;
-
         /// <summary>
         /// Creates a callout at the given position
         /// </summary>
@@ -448,21 +446,44 @@ namespace Mapsui.UI.Forms
         /// <param name="position">Position of callout</param>
         public Callout CreateCallout(Position position)
         {
-            _callout = new Callout(_mapControl)
+            // this ref is only needed internally to this method, so let's scope it that way.
+            Callout _callout = null;
+
+            // since this seems to only be an issue on UWP (?), let's make our fix narrow. (Probably also an issue on WPF, though)
+            if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
             {
-                Anchor = position
-            };
+                // wrap our mutex in a using block to clean it up asap.
+                using (var manualReset = new System.Threading.ManualResetEventSlim(false))
+                {
+                    Device.BeginInvokeOnMainThread(() =>
+                    {
+                        // the Callout ctor is doing all kinds of UI init, so this is the root of the cross-thread exception on UWP.
+                        _callout = new Callout(_mapControl) { Anchor = position };
 
-            // My interpretation (PDD): This while keeps looping until the asynchronous call
-            // above has created a callout.
-            // An alternative might be to avoid CreateCallout from a non-ui thread by throwing
-            // early.
-            while (_callout == null) ;
+                        // signal the worker thread that we're done
+                        manualReset.Set();
+                    });
 
-            var result = _callout;
-            _callout = null;
+                    // since original impl had no timeout, we'll leave it that way here too...
+                    // probably should fix this to have a timeout though?
+                    manualReset.Wait(); // block the worker thread (better than spinning!)
+                }
+                // This whole call stack would be better with await/async paradigm, but that requires larger refactor, as the caller is a property get method.
+            }
+            else // just create the callout object on the calling thread.
+            {
+                _callout = new Callout(_mapControl) { Anchor = position };
+            }
+            //// My interpretation (PDD): This while keeps looping until the asynchronous call
+            //// above has created a callout.
+            //// An alternative might be to avoid CreateCallout from a non-ui thread by throwing
+            //// early.
+            //while (_callout == null) ;
 
-            return result;
+            //var result = _callout;
+            //_callout = null;
+
+            return _callout;
         }
 
         /// <summary>
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -458,7 +458,7 @@ namespace Mapsui.UI.Forms
                     Device.BeginInvokeOnMainThread(() =>
                     {
                         // the Callout ctor is doing all kinds of UI init, so this is the root of the cross-thread exception on UWP.
-                        _callout = new Callout(_mapControl) { Anchor = position };
+                        _callout = CalloutFactory(position);
 
                         // signal the worker thread that we're done
                         manualReset.Set();
@@ -472,7 +472,7 @@ namespace Mapsui.UI.Forms
             }
             else // just create the callout object on the calling thread.
             {
-                _callout = new Callout(_mapControl) { Anchor = position };
+                _callout = CalloutFactory(position);
             }
             //// My interpretation (PDD): This while keeps looping until the asynchronous call
             //// above has created a callout.
@@ -486,6 +486,50 @@ namespace Mapsui.UI.Forms
             return _callout;
         }
 
+        public System.Threading.Tasks.Task<Callout> CreateCalloutAsync(Position position)
+        {
+            // a construct to properly bubble out either the result or an exception from this async task.
+            // The result *may* be new-ed on a different thread (the UI thread).
+            var completion = new System.Threading.Tasks.TaskCompletionSource<Callout>();
+
+            // since this seems to only be an issue on UWP (?), let's make our fix narrow. (Probably also an issue on WPF, though)
+            if (Device.RuntimePlatform.Equals(Device.UWP) || Device.RuntimePlatform.Equals(Device.WPF))
+            {
+                // The root of this issue is that if the root of the cakk-stack reaching this method is a 
+                // Touch event sourced from the MapControl, the internal SKGLView.Touch event is generated 
+                // on a Timer that is thread-based, and thus, not the UI thread (and Windows no like UI objects 
+                // created on worker threads).   
+
+                Device.BeginInvokeOnMainThread(() =>
+                {
+                    try
+                    {
+                        //signal the calling thread that UI thread is done
+                        completion.SetResult(CalloutFactory(position));
+                    }
+                    catch (Exception e)
+                    {
+                        //signal the calling thread that UI thread had an error
+                        completion.SetException(e);
+                    }
+                });
+            }
+            else
+            {
+                // complete immediately, we're not on Windows
+                completion.SetResult(CalloutFactory(position));
+            }
+
+            return completion.Task;
+        }
+
+        /// <summary>
+        /// a method of convience to create a callout object completely and reduce copy-paste
+        /// </summary>
+        /// <param name="position"></param>
+        /// <returns></returns>
+        private Callout CalloutFactory(Position position) { return new Callout(_mapControl) { Anchor = position}; }
+       
         /// <summary>
         /// Shows given callout
         /// </summary>
--- a/Samples/Mapsui.Samples.Common/Maps/BingSample.cs
+++ b/Samples/Mapsui.Samples.Common/Maps/BingSample.cs
@@ -19,7 +19,7 @@ namespace Mapsui.Samples.Common.Maps
         {
             var map = new Map();
             var apiKey = "Enter your api key here"; // Contact Microsoft about how to use this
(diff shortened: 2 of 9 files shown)


charlenni:
Thank you for your help.

As I can see (did only a fast view), you only changed one of the samples for Forms. Could you please change the second too?

jasells:
@charlenni, It has been 6 weeks since I pushed this, so could you point me toward the other sample?

The _actual_ fix is in MapView and Pin, which are both common.    The edits I made to the MainPageLarge of the one sample were just to make it easier to use, and center on the host system's location.

Compare to March commit for MapView, and Feb commit for Pin to see the root issues that were fixed: 

1. Callout created on non-UI thread (Pin.cs) is no-good on Windows because Callout is a UI element/control.
2. root of above is that the tap/click events are being raised on a background thread (MapView.cs), which is unexpected on Windows (or Android, for that matter).  UI events are generally on the UI/main thread.

Mapsui.UI.Forms/MapView.cs
```diff
@@ -827,14 +863,14 @@ private void HandlerTap(object sender, TappedEventArgs e)
...
                     var args = new MapClickedEventArgs(_mapControl.Viewport.ScreenToWorld(e.ScreenPosition).ToForms(), e.NumOfTaps);
 
-                    MapClicked?.Invoke(this, args);
 
-                    if (args.Handled)
-                    {
-                        e.Handled = true;
-                        return;
-                    }
+                        MapClicked?.Invoke(this, args);
```
charlenni: What hast changed here and in the next lines?

Samples/Mapsui.Samples.Common/Maps/BingSample.cs
```diff
@@ -19,7 +19,7 @@ public static Map CreateMap()
         {
             var map = new Map();
             var apiKey = "Enter your api key here"; // Contact Microsoft about how to use this
-            map.Layers.Add(new TileLayer(KnownTileSources.Create(KnownTileSource.BingAerial, apiKey), 
+            map.Layers.Add(new TileLayer(KnownTileSources.Create(KnownTileSource.BingHybrid, apiKey),
```
charlenni: Is this needed for this PR?

charlenni:
@jasells It is very difficult for me to review a PR, which change more than one part at a time. It is the first time I ever review something. 

You changed the Callout handling, changed event handling and different parts of half of the sample (consits of MainPageLarge and MainPage/MapPage for different sizes).

It would be nice, if you could make three PR, each for one of the parts, and merge the current master before.

jasells:
it was current 6wks ago.  ;-)

I realized after I changed the demo app code, that it probably did more harm than good for the PR, BUT it does make repro-ing the issue and fix easier....   which is why I did it.

Samples/Mapsui.Samples.Common/Maps/BingSample.cs
jasells: Not related to the Callout issue.

Mapsui.UI.Forms/MapView.cs
jasells: All UI events are marshalled from the background thread generating them, to the main/UI thread (for WPF or UWP *only*  since there was no indication of a problem on other platforms).  This will avoid the thread-lock or illegal access from non-ui thread exception potential in the Demo app (or any other application code using Callouts).

jasells: Honestly, looking into Xamarin.Forms v 4.0+, they provide the *ideal* solution by adding a synchronous Xamarin.Forms.Device.InvokeOnMainThread() method in addition to the async BeginInvokeOnMainThread() that is only made available pre-XForms v 4.0.

charlenni:
>     1. Callout created on non-UI thread (Pin.cs) is no-good on Windows because Callout is a UI element/control.

Did you checked your code on Android and iOS too?

>     2. root of above is that the tap/click events are being raised on a background thread (MapView.cs), which is unexpected on Windows (or Android, for that matter).  UI events are generally on the UI/main thread.

Did you checked all the other events too? There are much more events than tap/click like fling and longtap.

radderz:
@jasells - yea the repo got a bit behind as paul is busy. I have completed a few pull requests in the past couple of days so we should be able to get these changes in if you split them up to make the pull request easier to test/understand.

Is there a specific reason you think these events need to be marshalled on the main thread? As your handler can handle the marshalling if a UI change was going to be made off handling of the event. Or is this internally breaking in the MapView itself?

jasells:
@radderz 

"Is there a specific reason you think these events need to be marshalled on the main thread?"

Yes.  

1. From experience, UI events are raised on UI thread in Windows (I *think* also in Android, I'd have to check).  The fact that SkiaSharp is raising them on a background thread is not expected, and leads to a Cross thread exception when creating the default Callout in Pin (on Windows). 
2. See below :

From [Microsoft](https://github.com/Microsoft/Windows-task-snippets/blob/master/tasks/UI-thread-access-from-background-thread.md):
"Code running on a background thread cannot manipulate UI elements directly without raising a cross-thread exception. Instead, you must run UI code using a dispatcher"


I wish I had a better link to explain it, but I have run these circles in the past several times on different projects on Windows.


Xamarin.Forms < v 4 only gives you the option for Xamarin.Forms.Device.BeginInvokeOnMainThread(), which internally calls the Dispatcher (asynchronously) on WPF/UWP.  

You can see my other reply [above ](https://github.com/Mapsui/Mapsui/pull/675#discussion_r314007280)for some more info, but the BLUF is that with Xamarin.Forms v 3.x, this is the best sln I can come up with that isn't ridiculously complicated (invlolving catpuring the UI thread ID during init, and storing it to compare in the Pin.Callout get method to make sure the Callout is created on the Main thread while also avoiding a deadlock.

The changes to Pin and MapView are not seperable, and the changes 
*only* affect UWP/WPF so that UI events behave as expected.  I intentionally wrote the code so that no other platforms will see a change from current.

`private static bool PlatformRequiresUIcontext() 
        { 
            return Device.RuntimePlatform.Equals(Device.UWP)  
                || Device.RuntimePlatform.Equals(Device.WPF); 
        }`

So, the UI events *only* get routed to the Main/UI thread for Windows, all other platforms behave per status quo.

charlenni:
I'm with Mark: why call this events in MainThread by default. Sometimes you don't want to do anything in the main thread. But if you want, you can do this. If I understand you correct, than you say, that all events are called from the main thread in XF. If this is correct, than we should change all events in the XF MapControl/MapView, not only a few. But I couldn't find any informations on this up to now. Do you have some link (source code of XF?).

The real information is, that BeginInvokeOnMainThread() is an asynchron call. Than it is clear, why creating the callout crashes. With this in mind, your lazy creation should work fine, because the callout is first accessed, when the Pin is created.

jasells:
> The real information is, that BeginInvokeOnMainThread() is an asynchron call. Than it is clear, why creating the callout crashes. 

Almost...  BeginInvokeOnMainThread() pushes an Action onto the Main thread's message (work to-do) queque, so any Action passed to BeginInvokeOnMainThread()  _will_ be executed by the UI/main thread.... at some point in the _future_.  Using BeginInvokeOnMainThread() will prevent the cross-thread exception and resulting crash, **BUT** you'd need some sync mech to know when it is done, and finish the property getter, and return the resulting Callout's ref.  And, to compound this,` async await` is not available in a ctor, nor in a Property getter/setter, and BeginInvokeOnMainThread() is async in the sense that it returns immediately, not blocking the caller until the main thread completes the Action, but _**_neither_** _  is it awatiable, so any sync attempt would involve an Action with a mutex of some sort to signal to the calling thread that the Main thread is complete.  This means you _***also***_  need to know the Main thread's ID so that you can *not* try to sync a call from the main thread to the Action passed to BeginInvokeOnMainThread(), since that will obviously result in a deadlock (main thread is waiting for the main thread to complete the Action....)

> With this in mind, your lazy creation should work fine, because the callout is first accessed, when the Pin is created

You are correct, if limited to *only* the ctor of Callout.  That was my hope was well, and why I put the lazy initializer in, **BUT**  Callout's ctor creates and manipulates grid(s), which are UI elements, and is the precise source of the *first* cross-thread exception in the demo, as written.  

See [here](https://github.com/jasells/Mapsui/blob/267daf98d13c178d6232f50324a40a79e4c5e2e4/Mapsui.UI.Forms/Objects/Callout.cs#L96):             `_grid = new Grid()`

So, the bottom line is that a Callout, as currently implemented, can _never_ be "newed" on a worker thread on Windows.  _Full Stop_.

_**But**_, the ctor is not the only member of Callout that can produce this exception, since it internally uses a Grid, and any manipulation of the internal Grid can throw the exception, including several properties that typically get set immediately after a Callout is created that ultimately call [Callout.UpdateMargin()](https://github.com/jasells/Mapsui/blob/267daf98d13c178d6232f50324a40a79e4c5e2e4/Mapsui.UI.Forms/Objects/Callout.cs#L588):

  `_grid.Margin = margin;`

This is not surprising behavior, for a UI element/control, on Windows.  (It is idiomatically _allowed_ on Windows that UI elements are not guaranteed to be thread-safe).  Though, I would argue that a well designed control would use the dispatcher synchronously in the ctor, or any other method internally, as needed (or Xamarin.Forms.Device.Invoke() in the case of a Xamarin control).  

Again, upgrading to Xforms v4+ would be the cleanest, as then you could use the synchronous InvokeOnMainThread().... either in Pin.Callout getter, or (preferably) internally to Callout anywhere Grid is manipulated.  But, if Windows is the only platform with this issue, then you'd probably still want to check the platform at runtime and only switch thread context for Windows, not other platforms.  Android especially is almost inverse in the fact that you want to do _as much as possible_ on worker threads, and avoid the UI thread, except when absolutely necessary.  But, that doesn't consider the dependency/support issue for a package already released.

_However_, the UI events being fired on a background thread _is_  atypical behavior for a Windows app, and as such, any Windows apps _using_ the MapView would be extremely prone to the same error, and make it difficult to track down (especially if it is an x-platform app, other platforms would work fine, it seems).   Popping them over to the UI thread internally, for Windows context _only_,  makes the MapView more idiomatically correct on Windows, while changing nothing for other platforms.  The Windows idiom is to use `async await Task` in UI event handlers (button clicks handlers, tap event handles, etc) to intentionally farm non-ui work to worker threads _because_ it is an implicit expectation that the thread raising the event _is_ the main thread.  In fact, this is one of the top use cases for which the `async await Task` API was created.

 I concede, it may be a different issue, but it is very tightly coupled to the exception in the demo, and why it was thought to be resolved previously, but wasn't, at least on Windows.


And... yes, it is complicated. And why I structured the proposed solution to try to contain the complication internally to MapsUI and shield any app-devs potententially using MapsUI from this Windows-specifc mess.  =)

jasells:
I'll try merging in current master today.

jasells:
FYI, the Lazy<Callout> discussion got me thinking.  I checked the mem profile and found that the Lazy<Callout>  is ~ 10x smaller than the Callout instance, so it does help memory usage in the case of a large # of Pins where you're not likely to need the callout for all of them.

I got to thinking about the simpler code for just creating a default Callout in the Pin ctor.  That would avoid most of the x-thread issue in trade for more memory used.

charlenni:
I'm not happy with the Callout implementation. It is implementad as View. The background is a Skia bitmap and on top of this is a grid and some other Views. This whole thing has now to be moved around on the screen when you move the map. While doing this, you could see, that the Callout moves a little bit slower than the map. 

I made it, because I thought, a callout belongs to a map. Users want to have one. But need they such a flexible one? The only advantage is, that you could add any View to the Callout, even an own designed. 

But the question is, if we need such a horrible thing? Wouldn't it be better to add a native Mapsui callout, that is drawn direct to the map in the same way as other features? Or only a simple one with header, text and an image? Or as SVG feature? That all would be much faster an would use less memory than the used implementation.

Up to now I didn't discussed this with anybody, so that it would be helpful to think about it, before we try to improve the existing code.

jasells:
Improving the implementation of the Callout is a different discussion.  Options you mention sound like ideas worth pursuing. I like the SVG idea myself for speed.

However, this PR is to address a bug in the as-is implementation, and would improve things regardless of future Callout implementation over-haul/refactoring.

charlenni:
It's worth to look at this, because if the 2.0 beta goes to 2.0 stable, than you couldn't change such a big thing.

jasells:
Still contend that is separate from this.   This primarily an issue of MapView behaving as expected on Windows with respect to UI events on UI thread.  Would still be applicable, regardless of Callout impl.

charlenni:
Ok, no discussion.

Regarding this PR: this PR changes 3 different things (Callout problem, event on MainThread, change of sample), which isn‘t always not a good thing. So I would suggest to close this PR and create 3 PRs, each for an individual problem. This is much clearer and could be understand much better.

For the MainThread problem: if you change the behavior of one or two events, than you should change the behavior of all events. I can‘t understand, why this should only be for some events. Or is there a reason?

jasells:
It is changed for all UI events (events triggered by user interaction via touch or mouse).

I'm not sure what u mean by the 'Callout problem'.....

The only 'problem' addresses with this PR wrt Callout is that the Demo tries to call `new Callout` on worker thread(s) on Windows.  Do you mean the use of `Lazy<Callout>` v. the original 'manual' lazy construction!

jasells:
Ok, [here's](https://github.com/jasells/Mapsui/tree/callout-bug-UWP-repro) a Repro branch showing *only* modified demo app to show:

1. Adding Pin via MapView.Clicked event in existing Demo app throws x-thread exception on UWP 
2. Adding Pin via Button.Clicked event works.

The mods to the demo app are to show the 2 call stacks and that the MapView.MapClicked event is *not* behaving as would be expected on Windows.

In order to completely avoid any user interaction with the MapView events, the demo gets the current location of the host system, and uses this location to geo-ref the Pin that is added when the "ctreatePinBtn" is clicked by the user.  

Note: I tried to run the WPF demo, and the demo from master doesn't even load.  I get exception very early in the initialization.  Not sure what the issue is at this time.  I suggest this issue be handled separately.

pauldendulk:
@jasells I am very sorry for my late reaction. I did not have any time for Mapsui in the seconds half of 2019. I appreciate the work you did but currently callout refactoring is going on your changes will be too much out of sink. I will have to close the PR due to absence from my side.

jasells:
@pauldendulk Xamarin Forms 4+ provides an easy fix with a synchronous `Device.RunOnMainThread`. Even with refactor, this issue will likely remain on UWP if the tap event source remains `SKGLView` (`MapControl`).  These events are raised on worker thread, which is a problem on windows since the event handlers will likely interact with UI controls/elements.

charlenni:
@jasells I searched and searched, but couldn‘t find any report, where someone talks about events and on which thread they are running. Do you have any information with link about this?

pauldendulk:
@jasells Mapsui's MapControl has a cross platform RunOnUIThread private method. Is there a reason we need the Xamarin Forms 4 version?

A while back I looked into the loop issue. The way is see it is that this code is complicated because it it is a workaround for a Mapsui limitation. The MapView adds layers on top of Mapsui's layers and keep it under its own control. But it has to use some tricks to get this done. The implementation could be a lot simpler if you could hook into Mapsui's layer rendering. So we need some changes at a lower level.

charlenni:
@pauldendulk That’s not correct. The problem with the layers was, that there isn’t the possibility to have layers, that stay always on top. We don’t have a order for layers. If someone than create their own layers, than this layers are on top of the pin and drawable layers. I solved this by always remove and read the layers, when something changes in Layers. Than work and there shouldn’t be a problem. 

The loop problem you mentioned is because of the callouts. They contain objects from Forms, which must be created in the UI thread. I did this, because I thought it would be good, if you could put anything you want into a callout. This way, you could add layouts, grids, lists and so on in a callout. But it seems, that there are race conditions. And when creating the callout in the constructor of the pin (what I do), than we get problems.

While looking throw this discussions, I have to say, that the long comment from @jasells at August 15. describes the problem very good.

What we can do, if you want to make a stable 2.0, we could remove all callout functionality. It would be nice to have a callout, because users expect this from a map library, but it is better to have a working one 😉. And it seems, that no one uses callouts, because there isn’t any discussion about this feature.

pauldendulk:
@charlenni Not being able to keep the MapView's layers on top is the Mapsui limitation I mention. Adding and removing the layers in the MapView is the workaround. 

The consequences is complicated code where an action somewhere else can cause a loop if it is on the wrong thread 🙄. The focus should not be on figuring out how we can fix things in this logic but on how we can avoid this kind of logic. So, we need to think about an option to order layers, or what I indicated, a way to render MapView's layers without adding them to Map.Layers (this way they will be completely hidden from the user).

How far removed are we from replacing the current callout with the skia callouts that you are developing in your branch?

jasells:
> @charlenni Not being able to keep the MapView's layers on top is the Mapsui limitation I mention. Adding and removing the layers in the MapView is the workaround.
> The consequences is complicated code where an action somewhere else can cause a loop if it is on the wrong thread 🙄. The focus should not be on figuring out how we can fix things in this logic but on how we can avoid this kind of logic. So, we need to think about an option to order layers, or what I indicated, a way to render MapView's layers without adding them to Map.Layers (this way they will be completely hidden from the user).
> How far removed are we from replacing the current callout with the skia callouts that you are developing in your branch?

skia-based call outs will help, to an extent.  However, calling this lib Mapsui.Forms is misleading, if your are required to use only skia UI elements...

It is not *just* creation that must be on main thread, it is also *any* update to the object.  I pushed examples that show the same issue outside of the Callouts soecifically, I believe above.

Any Xamarin.Forms UI element will show the same issue, if it is modified via a callback from the MapControl tap/touch events.

> Mapsui's MapControl has a cross platform RunOnUIThread private method. Is there a reason we need the Xamarin Forms 4 version?

Unsure, it depends on what the `MapControl` thinks is the UI thread...  I'd assume it is the thread already firing events, but i'd have to look.  If it is the UWP UI thread, then yes, that should work.

jasells:
> @jasells I searched and searched, but couldn‘t find any report, where someone talks about events and on which thread they are running. Do you have any information with link about this?

Here's a [link ](https://social.msdn.microsoft.com/Forums/windowsapps/en-US/d425e995-6822-4059-898f-0b5ff9586dfe/uwpcquotthe-application-called-an-interface-that-was-marshalled-for-a-different-thread-quot
)to a post about the same problem: updating UI from background thread.


[ObservableCollection ](https://stackoverflow.com/questions/57522322/correct-way-to-access-ui-from-background-thread-in-uwp)also behaves this way on windows

jasells:
> @jasells Mapsui's MapControl has a cross platform RunOnUIThread private method. Is there a reason we need the Xamarin Forms 4 version?
> A while back I looked into the loop issue. The way is see it is that this code is complicated because it it is a workaround for a Mapsui limitation. The MapView adds layers on top of Mapsui's layers and keep it under its own control. But it has to use some tricks to get this done. The implementation could be a lot simpler if you could hook into Mapsui's layer rendering. So we need some changes at a lower level.

`MapControl.RunOnUIThread` just calls [`Device.BeginInvokeOnMainThread(action)`](https://github.com/jasells/Mapsui/blob/383f7d03d89b81194195b1f3a57d47d90cc0ee12/Mapsui.UI.Forms/MapControl.cs#L641).  

This does avoid the cross threading exception, BUT it is async, with no way to know when it completes.  This is the root of the problem.  lol

Xamarin.Forms has a *synchronous* `Device.InvokeOnMainThread`  whech blcoks the caller until the call completes on the *other* thread.  This sould be a one line fix in the current code.

charlenni:
@jasells I know, that UI elemts have to be updated from the UI thread. But I don't know, on which thread I'm when a Pin is created. If I'm on not the UI thread, than I try to create the Callout in the UI thread and wait, until it is created. That's working. But when this is called on the UI thread, than I start creating the Callout on the UI thread by BeginInvokeOnMainThread and than wait for the creation. But this wouldn't happen, because the creation takes place, after my function, that waits for the creation, ends. So we have a lock situation. If I now don't use any part for the Callout, that has to be created on a UI thread, it should work perfect, but I loose the ability to use Xamarin.Forms parts for the Callout. That's ok for me. I draw them now direct to the Skia canvas. Advantage is, that they always have the same distance to their anchor point. Disadvantage is, that click detecting is much more difficult than with a Xamarin.Forms view.

So there is only one decision, we have to make: should all events fired on the UI thread or should the user do this? We have some Xamarin.Forms elements in the MapView. This is the SkiaSharp view and the buttons on the side for zoom, center and northing. Only changes to the buttons had to take place in the UI thread. I should look into this, if this is done right. And we have to decide, if Xamarin.Forms events are called on the UI thread or any other thread, so that the user had to switch to the UI thread, if she/he wants to manipulate something on the UI. The link you send says only, that changes of the UI had to be done on UI thread. But it don't said, that events in Xamarin.Forms always have to be fired on UI thread. That was the question: is there a statement from Xamarin, that all events have to be fired on UI thread. Could the user rely on this. I found this [link](https://docs.microsoft.com/de-de/xamarin/essentials/main-thread), where you could read in the second paragraph of background, that the user can't rely on this.

If you look into the code, I only propagate the events from SkiaSharp. And SkiaSharp propagate the events from the underlying os (in case of UWP see [here](https://github.com/mono/SkiaSharp/blob/master/source/SkiaSharp.Views.Forms/SkiaSharp.Views.Forms.UWP/SKTouchHandler.cs)). So all events should be called by their underlying event system.

sailman342:
We all know that we need to update the User Interface on the UI thread. This problem also arises with other packages when dealing with asynchronous events. I think Mapsui should not bother about that and the suser should manage it.

The way I do it is to call an init procedure (Initialize() ) of a ThreadHelper at the page contructor level (or appearing) and th check later when I update the UI.
```

namespace RoadBook.Threading
{
    public static class ThreadHelper
    {
        public static int MainThreadId { get; private set; }

        public static void Initialize(int mainThreadId)
        {
            MainThreadId = mainThreadId;
        }

        public static bool IsOnMainThread => Environment.CurrentManagedThreadId == MainThreadId;

        public static void RunOnMainThread(Action function)
        {
            ThreadLog("RunOnMainThread");

            if(ThreadHelper.IsOnMainThread)
            {
                function.Invoke();
            }
            else
            {
                Device.BeginInvokeOnMainThread(function);
            }
        }
    }
}
```

I then use the ThreadHelper to update the Ui from the UI thread


```
        // user decides to cancel his click, just go back to no info pin
        private void OnDeleteInfoPinRequest(RBEditPin pin)
        {
            ThreadLog("EditPage OnDeleteInfoPinRequest 1");

            ThreadHelper.RunOnMainThread(() =>
           {
               ThreadLog("EditPage OnDeleteInfoPinRequest 2", true);

               EditMapView.RemoveNativePinFromTheMap(infoPin);
               infoPin = null;
           });

        }
```

Note I also write a small TrhredLog procedure that logs the current thread and if required raises an exception if not on the UI thread (susage in  code above).

Implementing that in Mapsui would mean somehow Calling a Mapsui initialisation while being in the UI thread or to pass the UI thread to Mapsui …

Both solution doesn't prevent the user to care about UI thread in other asynchronous events, so in my humblre opinion it is OK like that ...

jasells:
> We all know that we need to update the User Interface on the UI thread. This problem also arises with other packages when dealing with asynchronous events. I think Mapsui should not bother about that and the suser should manage it.
> The way I do it is to call an init procedure (Initialize() ) of a ThreadHelper at the page contructor level (or appearing) and then check later when I update the UI.
> 
> namespace RoadBook.Threading
> {
>     public static class ThreadHelper
>     {
>         public static int MainThreadId { get; private set; }
> 
>         public static void Initialize(int mainThreadId)
>         {
>             MainThreadId = mainThreadId;
>         }
> 
>         public static bool IsOnMainThread => Environment.CurrentManagedThreadId == MainThreadId;
> 
>         public static void RunOnMainThread(Action function)
>         {
>             ThreadLog("RunOnMainThread");
> 
>             if(ThreadHelper.IsOnMainThread)
>             {
>                 function.Invoke();
>             }
>             else
>             {
>                 Device.BeginInvokeOnMainThread(function);
>             }
>         }
>     }
> }
> 
> I then use the ThreadHelper to update the Ui from the UI thread
>         // user decides to cancel his click, just go back to no info pin
>         private void OnDeleteInfoPinRequest(RBEditPin pin)
>         {
>             ThreadLog("EditPage OnDeleteInfoPinRequest 1");
> 
>             ThreadHelper.RunOnMainThread(() =>
>            {
>                ThreadLog("EditPage OnDeleteInfoPinRequest 2", true);
> 
>                EditMapView.RemoveNativePinFromTheMap(infoPin);
>                infoPin = null;
>            });
> 
>         }
> 
> Note I also wrote a small ThreadLog procedure that logs the current thread and if required raises an exception if not on the UI thread (ussage in  code above).
> Implementing that in Mapsui would mean somehow Calling a Mapsui initialisation while being in the UI thread or to pass the UI thread to Mapsui …
> Both solution doesn't prevent the user to care about UI thread in other asynchronous events, so in my humblre opinion it is OK like that ...

The issue here is that MapsuiView is a Xamarin.Forms View, and it is not an unreasonable expectation that it behaves like other Xamarin Views.  This the only time I've had to deal with this in Xamarin.

Adding this helper, with modification, could allow syncing properly (add a mutex so the calling thread knows when async action is complete), but Xamarin.TouchEffect provides this functionality, and is native to Xamarin, and cleaner.

charlenni:
@jasells I use only events from SkiaSharp.Views.Forms. So SkiaSharp call them in the wrong context. So we had to mention this for SkiaSharp and change this behavior there. Or is this wrong?

sailman342:
@jasells  If you use a simple Timer to get GPS position or other things you will be out of the UI thread anyway, this is not a special case ...

`ThreadLog : RunOnMainThread  Current Thread / UI Thread 23 / 3`

I agree with you for the mutex but I don't need in my app for now :-)

charlenni:
I now tested it with the Xamarin.Essential NuGet package and the MainThread.IsMainThread flag. The code than looks like

            if (MainThread.IsMainThread)
            {
                _callout = new Callout(_mapControl)
                {
                    Anchor = position
                };
            }
            else
            {
                // In Forms.UWP, the Callout needs to be created on the UI Thread
                Device.BeginInvokeOnMainThread(() =>
                {
                    _callout = new Callout(_mapControl)
                    {
                        Anchor = position
                    };
                });
            }

When I do it this way, than I see, that it crashes, when I call the BeginInvokeOnMainThread() function un UWP, because on UWP the event is called on main thread and so the callout creation is started with BeginInvokeOnMainThread and than is awaitet in the main thread. That's why the app stops. So all is ok. The only problem was, that I already was on the main thread.

And, if I test with this function, than on UWP the touch functions are called always on the main thread. No problem to call any UI updates.

But now I would say, that the new approach is the better way to create callouts.

sailman342:
Crashing … strange …


[Micosoft doc … ](https://docs.microsoft.com/en-us/xamarin/essentials/main-thread?context=xamarin/xamarin-forms)

```
You might wonder if you should check if code is running on a secondary thread before calling BeginInvokeOnMainThread, for example, like this:
C#

Copy

if (MainThread.IsMainThread)
{
    MyMainThreadCode();
}
else
{
    MainThread.BeginInvokeOnMainThread(MyMainThreadCode);
}
You might suspect that this check might improve performance if the block of code is already running on the main thread.
However, this check is not necessary. The platform implementations of BeginInvokeOnMainThread themselves check if the call is made on the main thread. 
There is very little performance penalty if you call BeginInvokeOnMainThread when it's not really necessary.
```

charlenni:
Lets look into the code for BeginInvokeOnMainThread: [UAP](https://github.com/xamarin/Xamarin.Forms/blob/85bd80e5c39b8de3992bd154ddfdf783de1e37ed/Xamarin.Forms.Platform.UAP/WindowsBasePlatformServices.cs#L37), [Android](https://github.com/xamarin/Xamarin.Forms/blob/85bd80e5c39b8de3992bd154ddfdf783de1e37ed/Xamarin.Forms.Platform.Android/Forms.cs#L622), [iOS](https://github.com/xamarin/Xamarin.Forms/blob/85bd80e5c39b8de3992bd154ddfdf783de1e37ed/Xamarin.Forms.Platform.iOS/Forms.cs#L205), [Tizen](https://github.com/xamarin/Xamarin.Forms/blob/85bd80e5c39b8de3992bd154ddfdf783de1e37ed/Xamarin.Forms.Platform.Tizen/TizenPlatformServices.cs#L109) and [GTK](https://github.com/xamarin/Xamarin.Forms/blob/85bd80e5c39b8de3992bd154ddfdf783de1e37ed/Xamarin.Forms.Platform.GTK/GtkPlatformServices.cs#L21).

For me it looks like the most systems don't check, if they are on the main thread, but start the action (put it in a queue) and go back. So there could be problems like the above.

@jasells Did you see any Mapsui.Forms event on UWP, that aren't called on the main thread?

sailman342:
```
ThreadLog : EditPage OnEditMapClicked 1  Current Thread / UI Thread 26 / 3
ThreadLog : RunOnMainThread  Current Thread / UI Thread 26 / 3
```
```

ThreadLog : EditPage OnEditMapClicked 1  Current Thread / UI Thread 7 / 3
ThreadLog : RunOnMainThread  Current Thread / UI Thread 7 / 3
ThreadLog : EditPage OnEditMapClicked 2  Current Thread / UI Thread 3 / 3
```

From time to time MapClicked is on another thread, but it is not systematic. Being on another thread is not always synonym to crash ...

charlenni:
@sailman342 Did you use OnEditMapClicked as your normal MapView.MapClick event? I checked for main thread in the function, that call MapClicked and never get a IsMainThread = false. So it would be nice, if you could check this.

sailman342:
Here is the code :

```
    public class RBMapView : MapView

    {
        public RBMapView() : base()

            this.MapClicked += OnMapClicked;
            this.PinClicked += OnPinClicked;


        private void OnMapClicked(object sender, MapClickedEventArgs eArg)
        {
            ThreadLog("RBMapView OnMapClicked");

            RBMapClickedEventArgs eventArgs = new RBMapClickedEventArgs(
                eArg.Point.Latitude,
                eArg.Point.Longitude
                );

            onMapClickHdlr((RBMapView)sender, eventArgs);
        }
```

And the threadlog you see in the OnMapClicked handler is below

`ThreadLog : RBMapView OnMapClicked  Current Thread / UI Thread 7 / 1`

This one is done on Forms.WPF because I am fighting with WPF for the time being :-(

charlenni:
Oh, I would say, that Xamarin.Forms support for WPF isn't perfect. So I would say, that we couldn't rely on this results.

sailman342:
This one is on UWP (like the ones I showed yesterday)

`ThreadLog : RBMapView OnPinClicked  Current Thread / UI Thread 26 / 3`

charlenni:
I checked on UWP and don't get ever a call of OnPinClicked, that isn't on main thread.

sailman342:
I am Swiss Citizen, possibly slow but not dreaming :-) :-)

Windows 10 family, Visual studio 2019, mapsui from the master 4 days ago …

What do you think I should do next !

What are the procédures you use to test UI Thread ?

I am just willing to help, these thing are OK for me as they are :-)

image

sailman342:
@jasells 

[Microsoft](https://docs.microsoft.com/en-us/xamarin/essentials/main-thread)

`Applications sometimes need to use events that call the event handler on a secondary thread of execution. (The Xamarin.Essentials classes Accelerometer, Compass, Gyroscope, Magnetometer, and OrientationSensor all might return information on a secondary thread when used with faster speeds.) If the event handler needs to access user-interface elements, it must run that code on the main thread. The MainThread class allows the application to run this code on the main thread.`

charlenni:
I don't want to say, that you aren't correct, but I couldn't see it. So, perhaps my changes to the callout creation changed something or we do different tests.

I added the lines as descibed [above](https://github.com/Mapsui/Mapsui/pull/675#issuecomment-582958990) and added Xamarin.Essential for the check of the main thread. Than I use a line like

            System.Diagnostics.Debug.WriteLine("PinClick.IsMainThread " + Xamarin.Essentials.MainThread.IsMainThread.ToString());

in MainPageLarge.xaml.cs in function OnPinClicked. The result is, that I always get a true.

If you have time, please try to add your thread logic at [PolylineSample](https://github.com/Mapsui/Mapsui/blob/master/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/PolylineSample.cs). Than start the polyline sample of Forms and click on the map around. You will see a red polyline, extending with each click. On which thread this click is called? On the same thread should OnPinClicked be called. Both are called from the touch event without changing the thread.

sailman342:
I was just joking I didn't felt offensed but I am a newbie with Xamarin ...

charlenni:
And thank you for all your efforts.

sailman342:
`ThreadLog :  Current Thread 14 (on UWP Ui Thread use to be 3 `

```
        public bool OnClick(object sender, EventArgs args)
        {
            System.Diagnostics.Debug.WriteLine($"ThreadLog :  Current Thread {Environment.CurrentManagedThreadId} (on UWP Ui Thread use to be 3 " );
```

charlenni:
I copied your line into my code, which I derived yesterday from master. The result:

grafik

As you see, each time I start the app, I get 3 as ID.

Perhaps another version of SkiaSharp? I use 1.68.2-preview.17. You have the same?

charlenni:
Changed the branch and used the master now. Added your line of code. Get always 3 as Id.

pauldendulk:
> However, calling this lib Mapsui.Forms is misleading, if your are required to use only skia UI elements...

How the callout supplied by Mapsui is implemented does not matter. The features on the map are also rendered with skia, but it is still a Xamarin.Forms component because it can be used in Xamarin.Forms. Users do not even need to know how it is implemented. Then again, we could add features that make it possible to let users add their own callout logic, using Xamarin.Forms elements.

sailman342:
@charlenni 

Added your code to mine 

```
ThreadLog : OnClick Current Thread 13 (on UWP Ui Thread use to be 3 ) 
OnClick.IsMainThread False
ThreadLog : OnClick Current Thread 15 (on UWP Ui Thread use to be 3 ) 
OnClick.IsMainThread False
ThreadLog : OnClick Current Thread 9 (on UWP Ui Thread use to be 3 ) 
OnClick.IsMainThread False
```

```
        public bool OnClick(object sender, EventArgs args)
        {
            System.Diagnostics.Debug.WriteLine($"ThreadLog : OnClick Current Thread {Environment.CurrentManagedThreadId} (on UWP Ui Thread use to be 3 ) " );
            System.Diagnostics.Debug.WriteLine("OnClick.IsMainThread " + Xamarin.Essentials.MainThread.IsMainThread.ToString());
```

sailman342:
image

charlenni:
I don't know what to do. I started the sample app in x86 and x64 on a Win 10 system with VS 2019 16.4.3. In both version I go to the "Add polyline Sample" and start clicking. I get always a 3 as Id. Nothing else.

sailman342:
What about @jasells who started the thing ?

I downloaded a fresh version of Mapsui Master, flushed my Nuget stuff and will restart building mapsui this evening, I will keep you posted :-)

jasells:
> @jasells  If you use a simple Timer to get GPS position or other things you will be out of the UI thread anyway, this is not a special case ...
> ThreadLog : RunOnMainThread  Current Thread / UI Thread 23 / 3
> I agree with you for the mutex but I don't need in my app for now :-)

GPS, other things are nit UI events (touch, tap, click, etc)

jasells:
> Lets look into the code for BeginInvokeOnMainThread: UAP, Android, iOS, Tizen and GTK.
> For me it looks like the most systems don't check, if they are on the main thread, but start the action (put it in a queue) and go back. So there could be problems like the above.
> @jasells Did you see any Mapsui.Forms event on UWP, that aren't called on the main thread?



> @jasells
> Microsoft
> Applications sometimes need to use events that call the event handler on a secondary thread of execution. (The Xamarin.Essentials classes Accelerometer, Compass, Gyroscope, Magnetometer, and OrientationSensor all might return information on a secondary thread when used with faster speeds.) If the event handler needs to access user-interface elements, it must run that code on the main thread. The MainThread class allows the application to run this code on the main thread.

This us what I've been saying..,.

jasells:
> However, calling this lib Mapsui.Forms is misleading, if your are required to use only skia UI elements...
> 
> How the callout supplied by Mapsui is implemented does not matter. The features on the map are also rendered with skia, but it is still a Xamarin.Forms component because it can be used in Xamarin.Forms. Users do not even need to know how it is implemented. Then again, we could add features that make it possible to let users add their own callout logic, using Xamarin.Forms elements.

Correct, and the problem currently is that any pin implemented as Xamairn.Forms.View or similar, will have same issue as Calllout.  Changing Callout impl to skia avoids the issue for Callout, but the general issue remains, waiting to spring on app debs using Mapsui.Forms.

jasells:
> What about @jasells who started the thing ?
> I downloaded a fresh version of Mapsui Master, flushed my Nuget stuff and will restart building mapsui this evening, I will keep you posted :-)

Its been 6  months....  We forked  4 months ago, but I'll  look at the repro branch I pushed then, see if needs syncing with current master

sailman342:
```
        public bool OnClick(object sender, EventArgs args)
        {
            System.Diagnostics.Debug.WriteLine($"Pin Sample : OnClick Current Thread {Environment.CurrentManagedThreadId}");
            System.Diagnostics.Debug.WriteLine("Pin Sample : OnClick.IsMainThread " + Xamarin.Essentials.MainThread.IsMainThread.ToString());
```

```
Pin Sample : OnClick Current Thread 3
Pin Sample : OnClick.IsMainThread True
```

```
        public bool OnClick(object sender, EventArgs args)
        {
            System.Diagnostics.Debug.WriteLine($"Polygon Sample : OnClick Current Thread {Environment.CurrentManagedThreadId}");
            System.Diagnostics.Debug.WriteLine("Polygon Sample : OnClick.IsMainThread " + Xamarin.Essentials.MainThread.IsMainThread.ToString());
```
I don't know if it is normal but I had to make multiple regenerations to get down to 23 warnings from 185 first try (I enabled automatic download of missing nugets)

We are always on main (UI) thread but pin exemple crahes ... polygon is Ok

And ployline as well

```
Polyline Sample : OnClick Current Thread 3
Polyline Sample : OnClick.IsMainThread True
Polyline Sample : OnClick Current Thread 3
Polyline Sample : OnClick.IsMainThread True
Polyline Sample : OnClick Current Thread 3
Polyline Sample : OnClick.IsMainThread True
Polyline Sample : OnClick Current Thread 3
Polyline Sample : OnClick.IsMainThread True
```
Are these warning meaningfull to you ?

image

charlenni:
Pin sample crashes are ok, that's the callout bug. But now you don't see any call on other threads than the UI thread? Is that correct?

sailman342:
Yes, we are inline now …

Do you have an idea about the skia warnings ?

Had a quick look on my project and the master rebuild, I was on Forms 4.4 while the master is on 4.3 !

I will make further investigations but it can be interesting to remember in case you would move to version 4.4 of forms (@jasells  can you kindly advise us on that ?)

Kindest and thanks again

charlenni:
@jasells Ok, the callout bug should be solved with the new type of callouts.

But there is always the problem with that, that you say, each event should be called on main thread. All I found is, that each change to the UI should be called on the UI thread. The event, from which such changes are fired, should call the event handler from the UI thread, that I couldn't find.

But I think, I could do this. It is a little bit of more effort. And I should use the MainThread functions from the Xamarin.Essential package. But that package contains no WPF version. So we have no Mapsui.Forms.WPF vrsion.

The other way would be to move this to the users and say, that they have to be carefull, because events mustn't be called on main thread.

jasells:
> @jasells Ok, the callout bug should be solved with the new type of callouts.

I just pulled master, and it gets stuck in `MapView.CreateCallout()`.

This appears to be because the tap events *are* now firing on the main thread from `MapControl`, which is a `SKGLView` (as expected), but `MapView.CreateCallout()` is trying to sync with creating the callout out on the main thread, but the sync is a crude spin wait, so the main thread is spinning waiting for itself to complete the `Device.BeginInvokeOnMainThread`....

This is due to the fact the SkiaSharp.Forms.Views has been updated from 1.6 to 1.68.2-preview.17.  The UI event on Main thread issue has been resolved there (the home lib of `SKGLView` which is the ultimate source of all UI events sources my the Map in MapsUI.Forms).

The *current* master branch *requires* SkiaSharp.Forms.Views v 1.68.2-preview.17, which in turn requires Xamarin.Forms v >= 4.2, which *removes* one of the self-imposed conditions on this original PR: **DO NOT CHANGE** the Xamarin.Forms version requirement.

SO....

Unless you guys want to fix this bug on the older release of MapsUI.Forms (1.?) Whatever it was when this thread started, it would probably be best to start a new thread about the *current* issue on UWP, and not mix up different bugs from different versions of Mapsui.UI.Forms.


The issue in *current* master (2.0.0-beta.25) is that `MapView.CreateCallout()` no longer needs to call `Device.BeginInvokeOnMainThread` and it should *never* try to sync without first *knowing* it is needed (the calling thread != UI thread, as sailman342 has shown with his examples).   

It appears to me that `MapView.CreateCallout()`  isn't even needed, since the UI events are firing on the expected thread!

The fix can be made in `Pin.Callout.get` and remove `MapView.CreateCallout()` 

See this [branch](https://github.com/jasells/Mapsui/tree/new-Callout-Create-UWP-Fix).

JAS

jasells:
> @jasells Ok, the callout bug should be solved with the new type of callouts.
> But there is always the problem with that, that you say, each event should be called on main thread. All I found is, that each change to the UI should be called on the UI thread. The event, from which such changes are fired, should call the event handler from the UI thread, that I couldn't find.
> But I think, I could do this. It is a little bit of more effort. And I should use the MainThread functions from the Xamarin.Essential package. But that package contains no WPF version. So we have no Mapsui.Forms.WPF vrsion.
> The other way would be to move this to the users and say, that they have to be carefull, because events mustn't be called on main thread.

**ON WINDOWS**  UI updates *must* be made by UI thread.  *Because* of this, *Windows HAS ALWAYS FIRED **UI EVENTS** ON MAIN THREAD.*   Full. Stop.

Not **ALL** events (like network data, GPS updates, etc. etc. etc).  

There are 2 categories of events on Windows:

1) UI events (clicks, taps, mouse move, keypress, etc)
2) everything else

jasells:
IF any decides to try to fix this issue in the older version of Maspui.UI.Forms, or needs it for their app: 

[Here is the BLUFF](https://github.com/Mapsui/Mapsui/issues/606#issuecomment-509472728)  that does not require upgrading Xamarin.Forms or SkiaSharp.

jasells:
> Yes, we are inline now …
> Do you have an idea about the skia warnings ?
> Had a quick look on my project and the master rebuild, I was on Forms 4.4 while the master is on 4.3 !
> I will make further investigations but it can be interesting to remember in case you would move to version 4.4 of forms (@jasells  can you kindly advise us on that ?)
> Kindest and thanks again

I pulled an built the master no problem...

PR #701, opened 2019-09-27
Added case insensitive dictionary for feature lookup

garykindel:


pauldendulk:
I decided not to change this. I was in doubt, though, because for the majority of cases this would make it easier for users. But there is a small chance that it will break somebodies code. Having two entries with the same characters and different casing does not make much sense but it does happen. For instance if the features are added by end users. Or if old databases are used that are not well maintained.

Issue #711, opened 2019-10-10
[idea] Create Polylines from Encoded string


galadril:
It would be cool to have the option to create a full polyline from an encoded string like described here:

https://developers.google.com/maps/documentation/utilities/polylinealgorithm
Polyline encoding is a lossy compression algorithm that allows you to store a series of coordinates as a single string. Point coordinates are encoded using signed values. If you only have a few static points, you may also wish to use the interactive polyline encoding utility.

The next util method/class could be used:
https://github.com/galadril/Xamarin.Forms.GoogleMaps/edit/master/Xamarin.Forms.GoogleMaps/Xamarin.Forms.GoogleMaps/Helpers/PolylineConverter.cs

PR #713, opened 2019-10-11
Create Polylines from Encoded string #711

galadril:
I've added the posibility to create a Polyline from encoded string


galadril:
Hmm i did the comment with another account without email address via VS2019... so i cant sign the CLA as Mark Heinis... what to do here?

galadril:
cool

PR #755, opened 2020-01-03
Sealed classes cannot be parent of custom extended implementation

SebastianRuzanowski:
Hello, 

Thanks for the Mapsui, we use it in our project :)
We have encountered on a problem whereas we cannot extend implementation of Pins, Circles, Polygons etc. as it is marked sealed. Is there any meaning behind this? 

Thanks,
Sebastian

### DIFF
--- a/Mapsui.UI.Forms/Objects/Circle.cs
+++ b/Mapsui.UI.Forms/Objects/Circle.cs
@@ -9,7 +9,7 @@ using System;
 
 namespace Mapsui.UI.Forms
 {
-    public sealed class Circle : Drawable
+    public class Circle : Drawable
     {
         public static readonly BindableProperty CenterProperty = BindableProperty.Create(nameof(Center), typeof(Position), typeof(Circle), default(Position));
         public static readonly BindableProperty RadiusProperty = BindableProperty.Create(nameof(Radius), typeof(Distance), typeof(Circle), Distance.FromMeters(1));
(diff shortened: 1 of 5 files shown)


charlenni:
What do you want to add? Why not making your own drawables?

SebastianRuzanowski:
> 
> 
> What do you want to add? Why not making your own drawables?

We just want to extend models of pins, point, polygons etc. for our own usings containing some custom properties, instead of making adapters of "Pin" and embracing logic. 
Let me know, what you think, maybe I am missing something.

charlenni:
Ok, know I understand, what you want to do. No problem and only a small PR.

The problem is, that the certificate of the Mapsui.Samples.Uwp sample is to old. @pauldendulk has to renew this one.

SebastianRuzanowski:
That was my first idea, that implementation would have been closed for extensions since sealed.
Thanks @charlenni for quick response! Much appreciated.

pauldendulk:
@SebastianRuzanowski @madik7 @krzysCogito @PawelTomkow I see three approvals on this PR. It is good to have more eyes on the code. Are you all working in the same company? Did you talk about this internally? If so, it is not bad to have some of those discussions here in this thread in public.

SebastianRuzanowski:
Yes, apparently we work in the same company and it became a blocker for us.
Therefore we have decided to bring this to your attention. I think we will
be using MapsUI as we will use maps functionality extensively. In case of
any problems or workarounds, we shall get back to you with any ideas around
this.

Best regards,
Sebastian

pon., 6 sty 2020 o 06:09 Paul den Dulk <notifications@github.com>
napisał(a):

SebastianRuzanowski:
@pauldendulk when can we expect newest nuget release?

pauldendulk:
@SebastianRuzanowski Just released 2.0.0-beta.31.

PR #761, opened 2020-01-08
Add Newtonsoft to Mapsui.Samples.Forms.Wpf

charlenni:


pauldendulk:
Why was this needed?

charlenni:
Good question. Newtonsoft is always in the last project installed. In this case in the samples. But up to now, no one checked the Forms.WPF sample. I checked it now, because the UWP samples lacks of the certificate ;).

charlenni:
Btw. both UWP samples (normal and Forms).

charlenni:
Oh, you did this allready. Thank you for your work.

pauldendulk:
But what is failing without the package?

charlenni:
Oh, you only updated the normal samples for UWP, not the Forms version.

pauldendulk:
> Oh, you only updated the normal samples for UWP, not the Forms version.

You are referring to the .pfx update?

charlenni:
Yes. And failing was Mapsui.Samples.Forms.Wpf.

charlenni:
To reproduce the problem with Newtonsoft start the Mapsui.Samples.Forms.Wpf sample and select <Geometries><1 Points>. Than you get an error.

pauldendulk:
> To reproduce the problem with Newtonsoft start the Mapsui.Samples.Forms.Wpf sample and select <1 Points>. Than you get an error.

ah yes, I see it fail, but why, the samples project has a reference to newtonsoft and it compiles, that should be enough.

charlenni:
I assume, that the DLL isn't copied to the correct folder. So it compiles, but when running, it fails.

pauldendulk:
> I assume, that the DLL isn't copied to the correct folder. So it compiles, but when running, it fails.

Perhaps it it seems to me this caused by an error, either in our csproj or in Xamarin.Forms. Since I don't know what causes it I will simple merge.

PR #762, opened 2020-01-08
Added wheel zoom in/out to MapControl for Forms in the UWP version

charlenni:
Copied and renamed the certificate from Mapsui.Sample.UWP to Mapsui.Samples.Forms.UWP.


pauldendulk:
@charlenni This is now possible because of the new skiasharp version and the only thing you needed to do is handle the event. Is that correct?

charlenni:
Yes, it is. It was already in version 1.68.1, but I lost the focus 😉

PR #763, opened 2020-01-08
Create Callout on main thread

ejkani:
This PR is solves a thread issue in UWP when creating a PIN.

Creating a PIN in UWP throws a thread exception. In UWP we need the creation of UI elements to be done on the UI thread. 

The exception occurs when creating `Label` as a part of the Callout
image

If we ignore the font settings of the label, everything works fine (except the missing font settings)
image

Instead of ignoring Font settings on the labels, we can run the Callout creation on the UI Thread. I believe this is only necessary in UWP. This PR uses this solution.
image

### DIFF
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -518,10 +518,13 @@ namespace Mapsui.UI.Forms
         /// <param name="position">Position of callout</param>
         public Callout CreateCallout(Position position)
         {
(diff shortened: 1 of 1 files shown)


charlenni:
@pauldendulk You are right, callouts should be drawn in the renderer. We this before, but didn‘t get to an end.

To do this, there should be a discussion, how such callouts look. Should there only a line of text? Or two? Perhaps an image? On any side? Or should the callouts drawn by a svg image?

PR #788, opened 2020-01-28
Rotation button

charlenni:
Added rotation of northing button.

Todo this, the images of the buttons have to be changed and SvgButton needed the option to rotate the image on the button.


pauldendulk:
@charlenni I had to resolve a conflict here and am not sure I did the right thing.

charlenni:
I couldn’t see this changes, but assume that is at the opacity of the buttons layout. For me it looks ok.

Thank you for your work.

PR #790, opened 2020-01-28
Bug, with not updating MapView, removed

charlenni:
Bug, with not updating MapView, if a property of Pin or Drawable changes, removed.

Now always the map is repainted, if something changes in Pins or Drawables.

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -44,6 +44,7 @@ namespace Mapsui.UI.Forms
         /// </summary>
         private const int touchSlop = 8;
 
+        private bool _initialized = false;
         private float _skiaScale;
         private double _innerRotation;
         private Dictionary<long, TouchEvent> _touches = new Dictionary<long, TouchEvent>();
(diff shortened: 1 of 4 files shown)


charlenni:
Added another change for another update problem of MapView. If you move away from MapView and come back, than the map isn't redrawn and even, if you move the map, the map stays blank.

With this change, it is possible to redraw MapView by moving around the map or by calling Refresh() if the page appears on the screen.

Mapsui.UI.Forms/MapControl.cs
```diff
@@ -259,7 +262,7 @@ private Geometries.Point GetScreenPosition(SKPoint point)
 
         public void RefreshGraphics()
         {
-            if (GRContext == null)
+            if (!_initialized && GRContext == null)
```
pauldendulk: Do you have an indication that the RefreshGraphics method is sometimes called before _initialized is true? Initialized is called in the constructor. It registers to the events. But if this could cause a call to the event before the constructor is finished this could impact many classes.

charlenni:
I tested it without the _initialized flag and it worked. But I'm not sure, if this is working on all devices. Or perhaps SkiaSharp has changed since the time, I included this query.

I would try, if there are any problems with this and if yes, look into this again.

pauldendulk:
@charlenni So, you did not actually see it fail but added the _initialize flag just to be sure, is that correct?

charlenni:
Sorry, yes 😩

PR #796, opened 2020-01-30
Added two more image formats for symbols:

charlenni:
1. SKPicture: SKPictures are in Skia recorded draw commands, which are than drawn onto the canvas.
2. SKDrawable: SKDrawable are objects, that draw themselfs onto a canvas. They have a OnDraw function, which gets a canvas for this. This is called, each time new.
Added two samples for this new functionalities.

The Drawable sample is a funny thing. The symbol is created new, each time the map gets a draw command. With that, it looks like a Qix opponent ;-) And you could see, how often the map is drawn.

### DIFF
--- a/Mapsui.Rendering.Skia/BitmapHelper.cs
+++ b/Mapsui.Rendering.Skia/BitmapHelper.cs
@@ -32,10 +32,17 @@ namespace Mapsui.Rendering.Skia
                 var image = SKImage.FromEncodedData(SKData.CreateCopy(stream.ToBytes()));
                 return new BitmapInfo {Bitmap = image};
             }
-
-            if (bitmapStream is Sprite sprite)
+            else if (bitmapStream is Sprite sprite)
             {
                 return new BitmapInfo {Sprite = sprite};
+            }
+            else if (bitmapStream is SKPicture picture)
+            {
+                return new BitmapInfo { Picture = picture };
+            }
+            if (bitmapStream is SKDrawable drawable)
+            {
+                return new BitmapInfo { Drawable = drawable };
             }
 
             return null;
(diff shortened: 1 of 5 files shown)


charlenni:
Btw. it won't work for the Xaml renderer. There is no implementation.

pauldendulk:
This is now the list of bitmap types.
- Bitmap
- Svg
- Sprite
- Picture
- Drawable
The first three are actual data types. You can store them on disk. The two new types are more abstract and can at this point only be understood if you know how they relate to skia data types. Perhaps there is a better solution. Maybe a Custom type, with a callback where the user can add

You add both a SKDrawable and SKPicture. Why both?

Could you explain a bit more about why you add this? Probably related to the callback, but how will it be used to draw the callback? I was thinking about a callout implementation more similar to how labels are drawn right now.

charlenni:
Yes, you are right. Both couldn't be stored on disk. The difference between SKPicture and SKDrawable is, that SKPicture is static. You record Skia drawing commands and could redraw this everytime you want. SKDrawable is dynamic. It has its own drawing command, that is called each time, when the SKDrawable should be drawn.

I started with the implementation of the callout. I thought, I need the SKPicture for this, but could handle it without any of this new formats. I implement it as a new style, called CalloutStyle. This style draws the bubble and hold all informations about the placement of the bubble relative to the feature. The content is a bitmap. Perhaps here comes the SKPicture in. If you want to make a callout with two lines of text, this had to be drawn adhoc. So it would be nice to draw the content in a SKPicture and set this as content of the callout. So you don't have to convert the content to a bitmap format, which could be loose some informations.

The SKDrawable format: I'm still think about vector maps. If you use the styling of Mapbox, than styling could change by zoom factor and other things. For example, street width could be zoom dependent and so on. For this, it would be helpfull, if the content is drawn, when it is displayed on the screen. So it would be possible to create tiles from vector data, which stores the needed data in the tile in a object derived from SKDrawable and then draw the data, when it should displayed on screen. But that is only possible, when Mapsui can handle SKDrawables.

This said, forget SKDrawable, because no one except me will use it ;-) But SKPicture will be usefull for adhoc created content.

pauldendulk:
Okay, that sounds good. So then I think we should close this PR and wait for your next one. Okay?

charlenni:
No. This PR is complete independent from all other PRs. Lets don't mix them up. The next big would be for callout, which don't need this one.

pauldendulk:
Well, I don't want to accept this PR in this form because it introduces two bitmap types which are too closely coupled with the SkiaSharp implementation. More in general reducing the api surface is important to keep this project maintainable.

charlenni:
With this in mind, you don't have any vector format, only bitmap formats except SVG. Correct?

So which is the correct way to extend the formats of bitmaps, that Mapsui could handle? For this, Mapsui should provide a system for a implementing custom renderers, so that different formats are possible. Such a bitmap format should have a Draw function, which provides a SKCanvas (or anything else for XAML), so that the object could draw itself to this canvas. With such a system, all formats are possible.

pauldendulk:
Side note: We are discussing complex things. The implementation is complex, and the ideas of how to set it up are complex and also not yet clear. This is hard but I try to do my best. There is bound to be some confusion but this should not discourage us.

> With this in mind, you don't have any vector format, only bitmap formats except SVG. Correct?
> 
> So which is the correct way to extend the formats of bitmaps, that Mapsui could handle? For this, Mapsui should provide a system for a implementing custom renderers, so that different formats are possible. 

I don't have a clear picture yet. The rough idea is that it should be similar to the custom widget renderer. So you can register a custom type, and the renderer tries to look for an implementation of the renderer for that type (this is how it works for widgets). So yes, a user could add any type, but that type will not be part of the Mapsui project. It will work as long as the user also provides its own renderer for that type. 

> Such a bitmap format should have a Draw function, which provides a SKCanvas (or anything else for XAML), so that the object could draw itself to this canvas. With such a system, all formats are possible.

Yes, similar to the widget renderer. 

```csharp
    public interface ISkiaWidgetRenderer : IWidgetRenderer
    {
        void Draw(SKCanvas canvas, IReadOnlyViewport viewport, IWidget widget, float layerOpacity);
    }
```

called like this:

```csharp
        public static void Render(object target, IReadOnlyViewport viewport, IEnumerable<IWidget> widgets,
            IDictionary<Type, IWidgetRenderer> renders, float layerOpacity)
        {
            var canvas = (SKCanvas) target;

            foreach (var widget in widgets)
            {
                ((ISkiaWidgetRenderer)renders[widget.GetType()]).Draw(canvas, viewport, widget, layerOpacity);
            }
        }
```
The render method get a generic target of type object. It is cast to an SKCanvas. So Mapsui core has no dependency on SkiaSharp. 

I am not sure if we should add a custom bitmap renderer, or perhaps a geometry renderer. and how to go about styles? Should there be custom styles as well? 

There is more to say but I have to bring the children to gymnastics.

charlenni:
> I am not sure if we should add a custom bitmap renderer, or perhaps a geometry renderer. and how to go about styles? Should there be custom styles as well?

Sounds like a big project. Will be a complete redesign of Mapsui. I‘m not sure, if it is worth the effort.

PR #807, opened 2020-02-06
Added MinVisible and MaxVisible to Pins and Drawables

charlenni:
Added MinVisible and MaxVisible to Drawables and Pins.
 Added extension functions for converting Mapsui resolutions to zoom levels and vice versa.


charlenni:
This are only convinient functions. You could have done this by direct change the values in the Feature, which is always exposed. All this is done to mimik the original MapView of Xamarin.Forms. If someone don't want to use it, than ther is the MapControl, which could be used in Xamarin.Forms, but without the convinient functions and in the same style as all other MapControls are used.

pauldendulk:
ah ok, I still need to checkout the regular MapView.

PR #808, opened 2020-02-07
New callout create uwp fix

jasells:
this avoids the dead-lock condition when creating a pin's callout on UWP.

### DIFF
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -511,33 +511,6 @@ namespace Mapsui.UI.Forms
 
         private Callout _callout;
 
-        /// <summary>
-        /// Creates a callout at the given position
-        /// </summary>
-        /// <returns>The callout</returns>
-        /// <param name="position">Position of callout</param>
-        public Callout CreateCallout(Position position)
-        {
-            // In Forms.UWP, the Callout needs to be created on the UI Thread
-            Device.BeginInvokeOnMainThread(() =>
-            {
-                _callout = new Callout(_mapControl)
-                {
-                    Anchor = position
-                };
-            });
-
-            // My interpretation (PDD): This while keeps looping until the asynchronous call
-            // above has created a callout.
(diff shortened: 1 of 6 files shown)


jasells:
This **_DOES_** address #606, but relies on version SkiaSharp.Forms.Views v 1.68.2-preview.17 changing the source thread of the tap events from `SKGLView` being the Main/UI thread on UWP/WPF.

This change makes the `SKGLView` behave idiomatically correct for Windows platforms.

charlenni:
Thank you for your support.

But meanwhile I did a complete redesign of callouts. You can find the code [here](https://github.com/charlenni/Mapsui/tree/NativeCallout). The new implementation has advantages, but also disadvantages, but it doesn't rely on Xamarin.Forms. So there are no problems with any events ;-)

Thank you very much for your help to understand where the problems came from.

jasells:
> Thank you for your support.
> But meanwhile I did a complete redesign of callouts. You can find the code here. The new implementation has advantages, but also disadvantages, but it doesn't rely on Xamarin.Forms. So there are no problems with any events ;-)
> Thank you very much for your help to understand where the problems came from.

I'll take a look!

But master still uses the Forms-based Callouts...

jasells:
As soon as this can get through, I have some related changes to make code maintenance easier on callouts, regardless of impl.

jasells:
> This DOES address #606, but relies on version SkiaSharp.Forms.Views v 1.68.2-preview.17 changing the source thread of the tap events from SKGLView to be the Main/UI thread on UWP/WPF.
> This change makes the SKGLView behave idiomatically correct for Windows platforms, in regard to the thread raising UI events like tap, touch, mouse move, etc, IS the UI thread.

@pauldendulk  To be clear, #606 exists in latest master *because* SkiSharp.Forms.Views  v 1.68.2-preview.17 fires tap events on Main/UI thread, but `MapView.CreateCallout()` puts the calling thread into a spin wait until the main thread creates the callout, leading to endless spin wait.  

No sync is needed now that Tap events are fired on main thread on UWP.

PR #816, opened 2020-02-14
Feature/feature info based on pixel changes

charlenni:
Added some more code to this new feature


pauldendulk:
@charlenni oh, you were merging to my branch, not master. Had not noticed. That's a smart thing to do. I will resolve the conflicts cause by my later push. 

You return the entire MapInfo object. Perhaps that makes it simpler. You do not call it yet. It will not be hard to replace it, right? 

We have to figure something out for WPF. Earlier I was thinking about using skia as well. This will add a skia dependency to WPF. Perhaps not that bad. Also an option is to call the old MapInfoHelper inside the IRenderInfo.GetMapInfo implementation.

charlenni:
> @charlenni oh, you were merging to my branch, not master. Had not noticed. That's a smart thing to do. I will resolve the conflicts cause by my later push.

Yes. No need for a PR to master up to now ;-)

> You return the entire MapInfo object. Perhaps that makes it simpler. You do not call it yet. It will not be hard to replace it, right?

What to replace? I return a MapInfo entry for each feature, which is hit. With that, you could create features, that are invisible for hit testing. Perhaps someone could use this. With this, GetMapInfo always return a list of MapInfos. We could make two function, the old one GetMapInfo, which returns the top most feature, and GetAllMapInfos, which return a list of all MapInfos.

> We have to figure something out for WPF. Earlier I was thinking about using skia as well. This will add a skia dependency to WPF. Perhaps not that bad. Also an option is to call the old MapInfoHelper inside the IRenderInfo.GetMapInfo implementation.

The later one is the correct one. Move MapInfoHelper to the WPF renderer.

charlenni:
Sorry, @pauldendulk, I created another PR for the speed changes. I think, the integration of your past changes didn't worked correct for me, so there are some extra changes in the PR.

You could increase speed dramatically (from 160 ms to under 10 ms) if you create a clipping rectangle around the point. So Skia only draws the parts, that are inside this rect.

PR #817, opened 2020-02-14
Some changes for speed

charlenni:


charlenni:
Perhaps it is the best to only incooperate the two first and the last commit to your code and drop the PR?

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -343,6 +343,9 @@ private bool WidgetTouched(IWidget widget, Point screenPosition)
         /// <inheritdoc />
         public MapInfo GetMapInfo(Point screenPosition, int margin = 0)
         {
+            var stop = new System.Diagnostics.Stopwatch();
```
pauldendulk: We should start using BenchMarkDotNet.

charlenni:
I know, but this was a fast approach to see, how fast this function is. And it was useful for this.

charlenni:
And it should be removed 😉

pauldendulk:
Thanks again. I added the performance optimization from the second commit manually.

I will merge the branch to master next.

I don't want to add timers in the code. I have worked in projects with all kinds of performance measurements in all kinds of places. This makes the code harder to read and harder to refactor. Instead performance could measure from external tests, we can use the unit test framework for this, and add timers there. Also I would like to start using BenchMarkDotNet.

PR #821, opened 2020-02-17
Added a function to replace bitmap data in BitmapRegistry

charlenni:
Added some comment

### DIFF
--- a/Mapsui/Styles/BitmapRegistry.cs
+++ b/Mapsui/Styles/BitmapRegistry.cs
@@ -3,6 +3,9 @@ using System.Collections.Generic;
 
 namespace Mapsui.Styles
 {
+    /// <summary>
+    /// Class for managing all bitmaps, which are registered for Mapsui drawing
+    /// </summary>
     public class BitmapRegistry
     {
         private static BitmapRegistry _instance;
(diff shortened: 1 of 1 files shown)


Mapsui/Styles/BitmapRegistry.cs
```diff
@@ -3,42 +3,89 @@
...
+        /// <summary>
+        /// Check bitmap data for correctness
+        /// </summary>
+        /// <param name="bitmapData">Bitmap data to check</param>
+        private void CheckBitmapData(object bitmapData)
+        {
+            if (bitmapData == null)
+                throw new ArgumentException("The bitmap data that is registered is null. Was the image loaded correctly?");
+
+            if (bitmapData is Sprite sprite)
```
pauldendulk: Good to do these checks to catch errors early. Unfortunate we have to cast in the blind. Perhaps we should start to work with a bitmap type at some point.

Mapsui/Styles/BitmapRegistry.cs
```diff
@@ -3,42 +3,89 @@
...
             return _register[id];
         }
+
+        /// <summary>
+        /// Set new bitmap data for a already registered bitmap
+        /// </summary>
+        /// <param name="id">Id of existing bitmap data</param>
+        /// <param name="bitmapData">New bitmap data to replace</param>
+        /// <returns>True, if replacing worked correct</returns>
+        public bool Set(int id, object bitmapData)
```
pauldendulk: Was thinking about 'Replace' as a name. Would be clear. 'Set' is also nice though. 'Update' would be another alternative. Let's keep 'Set'.

Mapsui/Styles/BitmapRegistry.cs
charlenni: Yes, but than we move the blind cast to this bitmap type. We could create a registry function separat for each type like RegisterBitmap, RegisterSVG, RegisterSprite and so on.

charlenni:
> Okay. Good to check for null and Sprite.Atlas early but I don't think this should be done inside a cache that has so far no knowledge of what is being cached. Good enough for now.

The null check was already part of this function when I started with Mapsui. I added the sprite check when I implemented sprites.

Mapsui/Styles/BitmapRegistry.cs
charlenni: I thought about all three names, but we have already a Get function, so I took Set as the counterpart.

charlenni:
Ok, this wouldn't work, because the Skia renderer has its own cache mechanism and so, the replaced bitmap in BitmapRegistry would be draw, but the old in the Skia SymbolCache. :-(

PR #822, opened 2020-02-18
Refactoring of callout

charlenni:
New version of callout with Skia rendering

### DIFF
--- a/Mapsui.Rendering.Skia/ExtensionMethods/ColorExtensions.cs
+++ b/Mapsui.Rendering.Skia/ExtensionMethods/ColorExtensions.cs
@@ -10,5 +10,10 @@ namespace Mapsui.Rendering.Skia
             if (color == null) return new SKColor(128, 128, 128, 0);
             return new SKColor((byte)color.R, (byte)color.G, (byte)color.B, (byte)(color.A * layerOpacity));
         }
+
+        public static Styles.Color ToMapsui(this SKColor color)
+        {
+            return new Styles.Color(color.Red, color.Green, color.Blue, color.Alpha);
+        }
     }
 }
--- a/Mapsui.UI.Forms/EventArgs/CalloutClickedEventArgs.cs
+++ b/Mapsui.UI.Forms/EventArgs/CalloutClickedEventArgs.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using Xamarin.Forms;
+
+namespace Mapsui.UI.Forms
+{
+    public sealed class CalloutClickedEventArgs : EventArgs
+    {
+        /// <summary>
+        /// Callout that is clicked
+        /// </summary>
+        public Callout Callout { get; }
+
+        /// <summary>
+        /// Point of click in EPSG:4326 coordinates
+        /// </summary>
+        public Position Point { get; }
+
+        /// <summary>
+        /// Point of click in screen coordinates
+        /// </summary>
+        public Point ScreenPoint { get; }
+
+        /// <summary>
+        /// Number of taps
+        /// </summary>
+        public int NumOfTaps { get; }
+
+        /// <summary>
+        /// Flag, if this event was handled
+        /// </summary>
+        /// <value><c>true</c> if handled; otherwise, <c>false</c>.</value>
+        public bool Handled { get; set; } = false;
+
+        internal CalloutClickedEventArgs(Callout callout, Position point, Point screenPoint, int numOfTaps)
+        {
+            Callout = callout;
+            Point = point;
+            ScreenPoint = screenPoint;
+            NumOfTaps = numOfTaps;
+        }
+    }
+}
--- a/Mapsui.UI.Forms/Extensions/TextAlignmentExtensions.cs
+++ b/Mapsui.UI.Forms/Extensions/TextAlignmentExtensions.cs
@@ -1,0 +1,33 @@
+﻿namespace Mapsui.UI.Forms.Extensions
+{
+    public static class TextAlignmentExtensions
+    {
+        /// <summary>
+        /// Convert Xamarin.Forms.TextAlignment to Mapsui/RichTextKit.Styles.Color
+        /// </summary>
+        /// <param name="textAlignment">TextAlignment in Xamarin.Forms format</param>
+        /// <returns>TextAlignment in Mapsui/RichTextKit format</returns>
+        public static Topten.RichTextKit.TextAlignment ToMapsui(this Xamarin.Forms.TextAlignment textAlignment)
+        {
+            Topten.RichTextKit.TextAlignment result;
+
+            switch (textAlignment)
+            {
+                case Xamarin.Forms.TextAlignment.Start:
+                    result = Topten.RichTextKit.TextAlignment.Left;
+                    break;
+                case Xamarin.Forms.TextAlignment.Center:
+                    result = Topten.RichTextKit.TextAlignment.Center;
+                    break;
+                case Xamarin.Forms.TextAlignment.End:
+                    result = Topten.RichTextKit.TextAlignment.Right;
+                    break;
+                default:
+                    result = Topten.RichTextKit.TextAlignment.Auto;
+                    break;
+            }
+
+            return result;
+        }
+    }
+}
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -23,8 +23,10 @@ namespace Mapsui.UI.Forms
         internal MapControl _mapControl;
 
         private readonly MyLocationLayer _mapMyLocationLayer;
+        private const string CalloutLayerName = "Callouts";
         private const string PinLayerName = "Pins";
         private const string DrawableLayerName = "Drawables";
+        private readonly MemoryLayer _mapCalloutLayer;
         private readonly MemoryLayer _mapPinLayer;
         private readonly MemoryLayer _mapDrawableLayer;
         private readonly StackLayout _mapButtons;
@@ -37,9 +39,9 @@ namespace Mapsui.UI.Forms
         private readonly SKPicture _pictMyLocationNoCenter;
         private readonly SKPicture _pictMyLocationCenter;
 
-        readonly ObservableCollection<Pin> _pins = new ObservableCollection<Pin>();
-        readonly ObservableCollection<Drawable> _drawable = new ObservableCollection<Drawable>();
-        readonly ObservableCollection<Callout> _callouts = new ObservableCollection<Callout>();
+        readonly ObservableRangeCollection<Pin> _pins = new ObservableRangeCollection<Pin>();
+        readonly ObservableRangeCollection<Drawable> _drawable = new ObservableRangeCollection<Drawable>();
+        readonly ObservableRangeCollection<Callout> _callouts = new ObservableRangeCollection<Callout>();
 
         /// <summary>
         /// Initializes a new instance of the <see cref="T:Mapsui.UI.Forms.MapView"/> class.
@@ -52,7 +54,8 @@ namespace Mapsui.UI.Forms
             IsClippedToBounds = true;
 
             _mapControl = new MapControl { UseDoubleTap = false };
-            _mapMyLocationLayer = new MyLocationLayer(this) { Enabled = MyLocationEnabled };
+            _mapMyLocationLayer = new MyLocationLayer(this) { Enabled = true };
+            _mapCalloutLayer = new MemoryLayer() { Name = CalloutLayerName, IsMapInfoLayer = true };
             _mapPinLayer = new MemoryLayer() { Name = PinLayerName, IsMapInfoLayer = true };
             _mapDrawableLayer = new MemoryLayer() { Name = DrawableLayerName, IsMapInfoLayer = true };
 
@@ -155,6 +158,9 @@ namespace Mapsui.UI.Forms
             _pins.CollectionChanged += HandlerPinsOnCollectionChanged;
             _drawable.CollectionChanged += HandlerDrawablesOnCollectionChanged;
 
+            _mapCalloutLayer.DataSource = new ObservableCollectionProvider<Callout>(_callouts);
+            _mapCalloutLayer.Style = null;  // We don't want a global style for this layer
+
             _mapPinLayer.DataSource = new ObservableCollectionProvider<Pin>(_pins);
             _mapPinLayer.Style = null;  // We don't want a global style for this layer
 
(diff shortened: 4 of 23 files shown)


charlenni:
Sorry, it seems, that this a bit more than a one line PR ;-)

Mapsui.Rendering.Skia/ExtensionMethods/ColorExtensions.cs
```diff
@@ -10,5 +10,10 @@ public static SKColor ToSkia(this Color color, float layerOpacity)
             if (color == null) return new SKColor(128, 128, 128, 0);
             return new SKColor((byte)color.R, (byte)color.G, (byte)color.B, (byte)(color.A * layerOpacity));
         }
+
```
pauldendulk: This method belongs here, but the existing method does not :). Nowadays I have as convention that the name of the class related to what it extends (the class after the this). So a SkiaExtension should extent a skia class. Color is not a skia class. Also, I do now use one file for each class. So the new method should be in SKColorExtensions and the old one in ColorExtensions. Also, als a folder name I now use 'Extensions' and not 'ExtensionMethods'. So what was here before was all wrong! ;)

Mapsui.UI.Forms/EventArgs/CalloutClickedEventArgs.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
```
pauldendulk: There is a lot of overlap with the new MapInfo here. Perhaps there is a way we could merge the two. In this case there is a Callout field, this is style. With the new MapInfo we get a Style and a Feature.

Mapsui.UI.Forms/Extensions/TextAlignmentExtensions.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Mapsui.UI.Forms.Extensions
```
pauldendulk: Ah, this class is fully in line with the Extension conventions I described above.

Mapsui.UI.Forms/Extensions/TextAlignmentExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public static class TextAlignmentExtensions
+    {
+        /// <summary>
+        /// Convert Xamarin.Forms.TextAlignment to Mapsui/RichTextKit.Styles.Color
+        /// </summary>
+        /// <param name="textAlignment">TextAlignment in Xamarin.Forms format</param>
+        /// <returns>TextAlignment in Mapsui/RichTextKit format</returns>
+        public static Topten.RichTextKit.TextAlignment ToMapsui(this Xamarin.Forms.TextAlignment textAlignment)
```
pauldendulk: A common convention is to never use the namespace unless there is a need to resolve some conflict (which might be the case here).

Mapsui.UI.Forms/MapView.cs
```diff
@@ -915,7 +904,11 @@ private void HandlerTap(object sender, TappedEventArgs e)
...
 
                 // A feature is clicked
-                HandlerInfo(sender, new MapInfoEventArgs { MapInfo = mapInfo, Handled = e.Handled, NumTaps = e.NumOfTaps });
+                var mapInfoEventArgs = new MapInfoEventArgs { MapInfo = mapInfo, Handled = e.Handled, NumTaps = e.NumOfTaps };
+
+                HandlerInfo(sender, mapInfoEventArgs);
+
+                e.Handled = mapInfoEventArgs.Handled;
             }
         }
```
pauldendulk: MapView is big

Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
```diff
@@ -27,6 +27,7 @@
   <ItemGroup>
     <PackageReference Include="SkiaSharp" Version="1.68.2-preview.17" />
     <PackageReference Include="SkiaSharp.Views.Forms" Version="1.68.2-preview.17" />
+    <PackageReference Include="Topten.RichTextKit" Version="0.1.118" />
```
pauldendulk: Perhaps you do not need to add it here when it is in Mapsui.Rendering.Skia

Mapsui.UI.Forms/Objects/Callout.cs
```diff
@@ -344,394 +446,190 @@ public string Subtitle
...
-
-        private void MapControlSizeChanged(object sender, EventArgs e)
+        internal void Update()
         {
-            UpdateScreenPosition();
+            UpdateContent();
+            UpdateCalloutStyle();
         }
 
         /// <summary>
```
pauldendulk: Callout.cs is also big. What is the responsibility of this file now that the callout is rendered at a lower level? Suppose I would like to use callouts in WPF would I also need a 700 lines callout.cs file there?

Mapsui.UI.Forms/Objects/Pin.cs
```diff
@@ -16,7 +17,8 @@ public class Pin : BindableObject, IFeatureProvider
     {
         private int _bitmapId = -1;
         private byte[] _bitmapData;
-        private readonly MapView _mapView;
+        private bool _wasCalloutVisible;
+        private MapView _mapView;
```
pauldendulk: This is common in the current implementation of the forms part but something I should mention. In general it is a good thing to avoid the number of pointers from one part of the code to another part. It is easier keep track of things of you pass the minimal information needed to a method or class. The MapView is a big class with a lot of methods, by passing it to the Pin, which also has multiple instantiations, can make it hard to keep track of things. I do not have a specific solution here, and there can be cases where this is the best solution, but it is something to keep in mind when making decisions.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
```diff
@@ -55,6 +55,9 @@
     <Reference Include="FormsViewGroup, Version=2.0.0.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\..\..\packages\Xamarin.Forms.4.3.0.991211\lib\MonoAndroid90\FormsViewGroup.dll</HintPath>
     </Reference>
+    <Reference Include="HarfBuzzSharp, Version=1.0.0.0, Culture=neutral, PublicKeyToken=0738eb9f132ed756, processorArchitecture=MSIL">
```
pauldendulk: ah, HarfBuzzSharp. Something I run into now and then but have not worked with it yet. Not sure how this related to RichTextKit.

pauldendulk: oh, RichTextKit has a dependency on HarfBuzzSharp.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.WPF/Mapsui.Samples.Forms.WPF.csproj
```diff
@@ -235,11 +241,13 @@
...
   <Import Project="..\..\..\packages\SQLitePCLRaw.lib.e_sqlite3.osx.1.1.13\build\net35\SQLitePCLRaw.lib.e_sqlite3.osx.targets" Condition="Exists('..\..\..\packages\SQLitePCLRaw.lib.e_sqlite3.osx.1.1.11\build\net35\SQLitePCLRaw.lib.e_sqlite3.osx.targets')" />
   <Import Project="..\..\..\packages\Xamarin.Forms.4.3.0.991211\build\Xamarin.Forms.targets" Condition="Exists('..\..\..\packages\Xamarin.Forms.4.3.0.991211\build\Xamarin.Forms.targets')" />
   <Import Project="..\..\..\packages\SkiaSharp.1.68.2-preview.17\build\net45\SkiaSharp.targets" Condition="Exists('..\..\..\packages\SkiaSharp.1.68.2-preview.17\build\net45\SkiaSharp.targets')" />
   <Import Project="..\..\..\packages\SkiaSharp.Views.Forms.1.68.2-preview.17\build\net45\SkiaSharp.Views.Forms.targets" Condition="Exists('..\..\..\packages\SkiaSharp.Views.Forms.1.68.2-preview.17\build\net45\SkiaSharp.Views.Forms.targets')" />
+  <Import Project="..\..\..\packages\HarfBuzzSharp.2.6.1\build\net45\HarfBuzzSharp.targets" Condition="Exists('..\..\..\packages\HarfBuzzSharp.2.6.1\build\net45\HarfBuzzSharp.targets')" />
```
pauldendulk: Is there something special with HarfBuzzSharp that it needs this condition? There probably is, I just don't know what.

pauldendulk:
> Sorry, it seems, that this a bit more than a one line PR ;-)

Today I pushed a commit with 103 files changed. :) I think it is my record. All related to a single change of interface though.

Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
pauldendulk: I had to add them to Mapsui.Samples.Wpf. I do not know why this is needed for some nugets and not for others.

Mapsui.UI.Forms/EventArgs/CalloutClickedEventArgs.cs
pauldendulk: In the WPF sample the callout style is indeed returned (as expected). Perhaps this can be used instead. This class probably came about earlier, but if you can delete it after some rework that would be worth the effort. Deleting is the best thing you can do in software development.
image

Samples/Mapsui.Samples.Common/Maps/CalloutSample.cs
```diff
@@ -0,0 +1,137 @@
...
+                var bitmapId = -1;
+                // Get text size
+                SKRect bounds;
+                using (SKPaint paint = new SKPaint())
+                {
+                    paint.Color = new SKColor((byte)Random.Next(0, 256), (byte)Random.Next(0, 256), (byte)Random.Next(0, 256));
+                    paint.Typeface = SKTypeface.FromFamilyName(null, SKFontStyleWeight.Bold, SKFontStyleWidth.Normal, SKFontStyleSlant.Upright);
+                    paint.TextSize = 20;
+
+                    using (SKPath textPath = paint.GetTextPath(c.Name, 0, 0))
```
pauldendulk: I made some changes to this file. I created separate functions for creating the callout image and callout style. I like functions like that that get some input they merely read and return a new create object as output. They don't change the state. In functional programming they call this a [pure function](https://en.wikipedia.org/wiki/Pure_function). State causes much of the complexity in software. Everywhere we create state we should see this as a problem (although often necessary). 

A particular problematic state full thing in Mapsui is the BitmapRegistry. State full and static so it can be accessed from all places. So far I works okay, and I have not thought of a better alternative. But it is good to be aware that this is a bad thing.

charlenni:
Coming home a little bit late, I saw, that you looked into the PR. Thank you very much for this.

I try to answer your questions and comments tomorrow, if I find some time.

Mapsui.UI.Forms/Extensions/TextAlignmentExtensions.cs
charlenni: This is the case. Both called TextAllignment and I wanted to make clear, which belongs to where.

Mapsui.UI.Forms/MapView.cs
charlenni: Yes, very big. Sorry for this. I don’t know, if I could reduce the size. Have to look into this.

Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
charlenni: I forgot to remove this in the UI.Forms part. Is an old entry.

Mapsui.UI.Forms/Objects/Callout.cs
charlenni: This is a bendable wrapper around CalloutStyle. If you use CalloutStyle direct in WPF, you don’t need this.

Mapsui.UI.Forms/Objects/Pin.cs
charlenni: I need it, if something changes in Pin, that I could redraw the MapControl.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
charlenni: As Skia too, RichTextKit uses HarfBuzz to convert characters to path and positions.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.WPF/Mapsui.Samples.Forms.WPF.csproj
charlenni: I don’t see any special thing here. Looks for HarfBuzz the same as for the others.

Samples/Mapsui.Samples.Common/Maps/CalloutSample.cs
charlenni: I know, that the samples are not very good style, but I use them for testing only.

pauldendulk: Yes, it is not important, but I thought it was good to mention because this strategy can also be used in other places.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.WPF/Mapsui.Samples.Forms.WPF.csproj
pauldendulk: You are right.

Mapsui.UI.Forms/Objects/Callout.cs
pauldendulk: If it is not needed in WPF it makes you wonder why so much code is needed in Forms. In general less code is better. A large file like this is a maintenance burden. I was asking about it because I want to now what it does, what functionality does it add? If I know this it becomes possible to think about alternatives. 'bendable wrapper' does not tell me that much. My guess is that it makes it more convenient to work with it. Perhaps there are other ways to serve that purpose. I think we should not go further into this specific question. It could become a lengthy discussion. But I still like to place remarks like this to explain how I usually think about code. I am constantly looking for ways to get less of it. I struggle hard to organize the code so that I can easily see the bigger picture when I go though it. 'Single responsibility' is an important concept. Also I strive for 'self documenting' code as much as possible.

Mapsui.UI.Forms/Objects/Pin.cs
pauldendulk: Sure, you will not add it for no reason, but there may be other ways to achieve the same. Now pointers go in both directions, from Pin to MapView and MapView has a list of Pins. This is something we should try to avoid. It is hard to keep track of how events flow in such a system. The lengthy discussion about thread exception and the loop we had in de layer update mechanism are examples of problems we should not try to fix but try to avoid. 

This is why I don't want PropertyNotifications in the Callout. This allows an outside system to mingle with the rendering logic. We use the invalidate pattern instead. This means there is only one point from which the renderer is called, in the render loop. From all sites it is possible to invalidate the render state, but this is nothing more than setting a flag. It is up to the renederer to deal with the invalid state. A typical problem that arises when you do not use that pattern is that many redrawn are called an a queue of messages stacks up which slows the system down.

Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
pauldendulk: Maybe it is necessary as in WPF. Not sure why. Perhaps for packages using the 'bait and switch' setup.

PR #836, opened 2020-02-23
Refactoring callout

charlenni:
Draw callouts more directly to the canvas without using images.

I assume, that I'm not ready. But mention it a little bit earlier, so that changes could be made easier.

### DIFF
--- a/Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
+++ b/Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
@@ -24,8 +24,34 @@ namespace Mapsui.Rendering.Skia
 
                 RenderCallout(calloutStyle);
             }
-            // Reuse ImageStyleRenderer because the only thing we need to do is to draw an image
-            ImageStyleRenderer.Draw(canvas, calloutStyle, destination, symbolCache, opacity, (float)viewport.Rotation);
+
+            // Now we have the complete callout rendered, so we could draw it
+            if (calloutStyle.BitmapId < 0)
+                return;
+
+            var picture = (SKPicture)BitmapRegistry.Instance.Get(calloutStyle.BitmapId);
+
+            // Calc offset (relative or absolute)
+            var symbolOffsetX = calloutStyle.SymbolOffset.IsRelative ? picture.CullRect.Width * (float)calloutStyle.SymbolOffset.X : (float)calloutStyle.SymbolOffset.X;
+            var symbolOffsetY = calloutStyle.SymbolOffset.IsRelative ? picture.CullRect.Height * (float)calloutStyle.SymbolOffset.Y : (float)calloutStyle.SymbolOffset.Y;
+
+            var rotation = (float)calloutStyle.SymbolRotation;
+            if (calloutStyle.RotateWithMap) rotation += (float)viewport.Rotation;
+
+            // Save state of the canvas, so we could move and rotate the canvas
+            canvas.Save();
+
+            // Move 0/0 to the Anchor point of Callout
+            canvas.Translate((float)destination.X - symbolOffsetX, (float)destination.Y - symbolOffsetY);
+            canvas.Scale((float)calloutStyle.SymbolScale, (float)calloutStyle.SymbolScale);
+
+            // 0/0 are assumed at center of image, but Picture has 0/0 at left top position
+            canvas.RotateDegrees(rotation);
+            canvas.Translate((float)calloutStyle.Offset.X, (float)calloutStyle.Offset.Y);
+
+            canvas.DrawPicture(picture);
+
+            canvas.Restore();
         }
 
         public static void RenderCallout(CalloutStyle callout)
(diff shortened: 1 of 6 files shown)


charlenni:
Now it should be ready to review and, if ok, approving.

Thank you for your work.

Samples/Mapsui.Samples.Common/Maps/CalloutSample.cs
```diff
@@ -81,19 +81,19 @@ private static IStyle CreateCalloutStyle(int bitmapId)
             {
                 case 0:
                     calloutStyle.ArrowAlignment = ArrowAlignment.Bottom;
-                    calloutStyle.Offset = new Geometries.Point(0, SymbolStyle.DefaultHeight * 0.5f);
+                    calloutStyle.Offset = new Offset(0, SymbolStyle.DefaultHeight * 0.5f);
```
pauldendulk: ah, reuse of offset. Good. It is almost the same as Point but as a developer it helpt me to see how an object was intended.

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
```diff
@@ -12,20 +12,46 @@ public class CalloutStyleRenderer : SymbolStyle
...
             float opacity, Point destination, CalloutStyle calloutStyle)
         {
-            if (calloutStyle.BitmapId < 0 || calloutStyle.Invalidated)
+            if (calloutStyle.BitmapId < 0 || calloutStyle.Invalidated || calloutStyle.TitleFont.Invalidated || calloutStyle.SubtitleFont.Invalidated)
             {
                 if (calloutStyle.Content < 0 && calloutStyle.Type == CalloutType.Custom)
                     return;
 
-                if (calloutStyle.Invalidated)
+                if (calloutStyle.Invalidated || calloutStyle.TitleFont.Invalidated || calloutStyle.SubtitleFont.Invalidated)
```
pauldendulk: Perhaps we could create property for the Invalidated status of all.

Mapsui/Styles/Font.cs
```diff
@@ -10,9 +15,62 @@ public Font(Font font)
...
-        public bool Bold { get; set; }
+        public string FontFamily
+        {
+            get => _fontFamily;
+            set
+            {
+                if (value != _fontFamily)
+                {
+                    _fontFamily = value;
+                    Invalidated = true;
```
pauldendulk: Clever solution. No notifications used, but it is still possible to know if the children are Invalidated, while the user does not have to set Invalidated explicitly. 

A disadvantage is all the code necessary. If we want to use this approach consistently we might have to do it in many other classes. 

To implement PropertyChanged without all the code I have used [Fody ProperyChanged](https://github.com/Fody/PropertyChanged) in other projects. I don't like to add another dependency, and I don't like the magic of it. Had trouble fixing some bugs. Perhaps we could also write some 'weaving' (that is apparently what it uses) ourselves to generate this code automatically for a class.

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
```diff
@@ -212,16 +246,27 @@ private static void DrawContent(CalloutStyle callout, SKCanvas canvas, BitmapInf
 
                 var offset = new SKPoint(offsetX, offsetY);
 
-                switch (bitmapInfo.Type)
+                if (callout.Type == CalloutType.Custom)
```
pauldendulk: Was this something I broke?

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
```diff
@@ -12,20 +12,46 @@ public class CalloutStyleRenderer : SymbolStyle
...
                 RenderCallout(calloutStyle);
             }
-            // Reuse ImageStyleRenderer because the only thing we need to do is to draw an image
-            ImageStyleRenderer.Draw(canvas, calloutStyle, destination, symbolCache, opacity, (float)viewport.Rotation);
+
+            // Now we have the complete callout rendered, so we could draw it
+            if (calloutStyle.BitmapId < 0)
+                return;
+
+            var picture = (SKPicture)BitmapRegistry.Instance.Get(calloutStyle.BitmapId);
```
pauldendulk: This is a big chunk of new code. What does it do? Why is it necessary in this new version?

pauldendulk:
@charlenni When running the sample callout sample in wpf I it has an exception when fetching the bitmap from the cache.

charlenni:
@pauldendulk So, I'm back.

Do you have the problems with Mapsui.Samples.WPF or Mapsui.Samples.Forms.WPF?

Samples/Mapsui.Samples.Common/Maps/CalloutSample.cs
charlenni: Yes, I thought, if we have it, we should use it. Code is much clearer.

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
charlenni: Than you have to in CalloutStyle, if the fonts changed. Could be made, should be done on the fly.

Mapsui/Styles/Font.cs
charlenni: Yes, it is some more code, but it is worth. I assume, that user often forget this.

In this case it isn't usefull to make changes. I see the advantages of OnPropertyChanged, but you said, it is to much effort for this, so we dropped it.

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
charlenni: No, all ok. But we must not decide between mor than two cases, so a if is enough.

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
charlenni: It is, because we don't draw into a canvas and than create an image of this canvas, but now draw direct to the canvas. This couldn't be done by the ImageStyleRenderer.

pauldendulk:
Forms wpf and uwp are okay.
Regular wpf and uwp are not okay.

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
pauldendulk: Yes, the only complication is that you could also be able to set the Invalidated status of the callout itself, separately. So, there are two things to keep track of: The Invalided status of the callout on it's own and the of itself and its children. How does this work? Two properties? Invalidated, DeepInvalidated (hmm, not great). Perhaps an _invalidated internal field to keep track of the callout itself, while the property getter checks the internal field and the children. This means the property can not be set from the outside so we need to cover all cases.

pauldendulk:
> Forms wpf and uwp are okay.
> Regular wpf and uwp are not okay.

@charlenni Is this something you could look into?

PR #837, opened 2020-02-23
Design study for new owner drawn styles

charlenni:
A first attampt for owner drawn styles. For discussion only :^)

PR #839, opened 2020-02-25
Added handling of owner drawn styles in Skia and WPF

charlenni:
Added a design study, how it would possible to add owner drawn styles in Mapsui.
Added a sample for this in "Special" - "Special Style".

Couldn't add it to WPF, because function `RenderFeature()` is static and couldn't access `StyleRenderers`, because it isn't static. Don't know, how to solve this without removing static from all the rendering functions. And the drawing function of sample isn't complete. Would be to time consuming for me to understand, how to do it. Sorry.

This replaces PR #837, because this has not needed parts. So it is easier to understand, when creating a new PR.

### DIFF
--- a/Mapsui.Rendering.Skia/MapRenderer.cs
+++ b/Mapsui.Rendering.Skia/MapRenderer.cs
@@ -6,6 +6,7 @@ using Mapsui.Geometries;
 using Mapsui.Layers;
 using Mapsui.Logging;
 using Mapsui.Providers;
+using Mapsui.Rendering.Skia.SkiaStyles;
 using Mapsui.Rendering.Skia.SkiaWidgets;
 using Mapsui.Styles;
 using Mapsui.UI;
@@ -28,6 +29,11 @@ namespace Mapsui.Rendering.Skia
         public ISymbolCache SymbolCache => _symbolCache;
 
         public IDictionary<Type, IWidgetRenderer> WidgetRenders { get; } = new Dictionary<Type, IWidgetRenderer>();
+
+        /// <summary>
+        /// Dictionary holding all special renderers for styles
+        /// </summary>
+        public IDictionary<Type, IStyleRenderer> StyleRenderers { get; } = new Dictionary<Type, IStyleRenderer>();
 
         static MapRenderer()
         {
@@ -135,6 +141,21 @@ namespace Mapsui.Rendering.Skia
 
         private void RenderFeature(SKCanvas canvas, IReadOnlyViewport viewport, ILayer layer, IStyle style, IFeature feature, float layerOpacity)
         {
+            // Check, if we have a special renderer for this style
+            if (StyleRenderers.ContainsKey(style.GetType()))
+            {
+                // Save canvas
+                canvas.Save();
+                // We have a special renderer, so try, if it could draw this
+                var result = ((ISkiaStyleRenderer)StyleRenderers[style.GetType()]).Draw(canvas, viewport, layer, feature, style, _symbolCache);
+                // Restore old canvas
+                canvas.Restore();
+                // Was it drawn?
+                if (result)
+                    // Yes, special style renderer drawn correct
+                    return;
+            }
+            // No special style renderer handled this up to now, than try standard renderers
             if (feature.Geometry is Point)
                 PointRenderer.Draw(canvas, viewport, style, feature, feature.Geometry, _symbolCache, layerOpacity * style.Opacity);
             else if (feature.Geometry is MultiPoint)
@@ -192,6 +213,7 @@ namespace Mapsui.Rendering.Skia
                     var color = pixmap.GetPixelColor(intX, intY);
 
                     VisibleFeatureIterator.IterateLayers(viewport, layers, (v, layer, style, feature, opacity) => {
+                        surface.Canvas.Save();
                         // 1) Clear the entire bitmap
                         surface.Canvas.Clear(SKColors.Transparent);
                         // 2) Render the feature to the clean canvas
@@ -200,6 +222,7 @@ namespace Mapsui.Rendering.Skia
                         if (color != pixmap.GetPixelColor(intX, intY))
                             // 4) Add feature and style to result
                             list.Add(new MapInfoRecord(feature, style, layer));
+                        surface.Canvas.Restore();
                     });
                 }
 
--- a/Mapsui.Rendering.Skia/MapRenderer.cs
+++ b/Mapsui.Rendering.Skia/MapRenderer.cs
@@ -45,6 +45,8 @@ namespace Mapsui.Rendering.Skia
             WidgetRenders[typeof(Hyperlink)] = new HyperlinkWidgetRenderer();
             WidgetRenders[typeof(ScaleBarWidget)] = new ScaleBarWidgetRenderer();
             WidgetRenders[typeof(ZoomInOutWidget)] = new ZoomInOutWidgetRenderer();
+
+            StyleRenderers.Add(typeof(VectorStyle), new VectorStyleRenderer(this));
         }
 
         public void Render(object target, IReadOnlyViewport viewport, IEnumerable<ILayer> layers,
--- a/Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
+++ b/Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
@@ -1,0 +1,148 @@
+﻿using Mapsui.Geometries;
+using Mapsui.Layers;
+using Mapsui.Providers;
+using Mapsui.Styles;
+using SkiaSharp;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Mapsui.Rendering.Skia.SkiaStyles
+{
+    /// <summary>
+    /// Demo VectorStyleRenderer
+    /// </summary>
+    public class VectorStyleRenderer : ISkiaStyleRenderer
+    {
+        private MapRenderer mapRenderer;
+        private static bool flag = false;
+
+        public VectorStyleRenderer(MapRenderer renderer)
+        {
+            mapRenderer = renderer;
+        }
+
+        public bool Draw(SKCanvas canvas, IReadOnlyViewport viewport, ILayer layer, IFeature feature, IStyle style, ISymbolCache symbolCache)
+        {
+            if (feature.Geometry is Point destination)
+            {
+                // Draws a dounat at point
+                DrawPoint(canvas, viewport.WorldToScreen(destination), (VectorStyle)style, (float)layer.Opacity);
+                return true;
+            }
+            else if (feature.Geometry is Raster raster)
+            {
+                // Draws a pink frame and cross over raster image
+                DrawRaster(canvas, viewport, layer, feature, style);
+                return true;
+            }
+
+            return false;
+        }
+
+        public static void DrawPoint(SKCanvas canvas, Point destination, VectorStyle vectorStyle, float opacity)
+        {
+            var width = (float)SymbolStyle.DefaultWidth;
+            var halfWidth = width / 2;
+
+            var linePaint = CreateLinePaint(vectorStyle.Outline, opacity);
+            var fillPaint = CreateLinePaint(vectorStyle.Outline, opacity);
+
+            linePaint.StrokeWidth = halfWidth / 2;
+            fillPaint.StrokeWidth = halfWidth / 4;
+            fillPaint.Color = SKColors.Pink;
+
+            canvas.Translate((float)destination.X, (float)destination.Y);
+
+            if (linePaint != null && linePaint.Color.Alpha != 0) canvas.DrawCircle(0, 0, halfWidth, linePaint);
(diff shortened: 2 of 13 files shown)


charlenni:
Added one more sample. Here you could see, how built in VectorStyle renderer could be extended/replaced.

One part draws a pink frame and cross on some raster tiles. The other part replaces the VectorStyle for drawing Points. You could see this in the Label Sample.

All other geometries are drawn normal with the built in renderers.

charlenni:
@pauldendulk Did you looked into this PR? It is only a small change. Most of the code is about the sample. And never mind, that the tests not work. It is, because I changed the normal drawing for VectorStyle, which break the tests.

charlenni:
Btw. shouldn't it be `WidgetRenderers` instead of `WidgetRenders`?

charlenni:
Or if you don't want to do it: how can I create an own NuGet package to use this PR in an own project?

pauldendulk:
> Btw. shouldn't it be `WidgetRenderers` instead of `WidgetRenders`?

Yes, that is an error.

pauldendulk:
> @pauldendulk Did you looked into this PR? It is only a small change. Most of the code is about the sample. And never mind, that the tests not work. It is, because I changed the normal drawing for VectorStyle, which break the tests.

I could not make time for it today but will do so as soon as I can.

Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
```diff
@@ -0,0 +1,148 @@
...
+            {
+                var priorMatrix = canvas.TotalMatrix;
+                var matrix = CreateRotationMatrix(viewport, boundingBox, priorMatrix);
+
+                canvas.SetMatrix(matrix);
+
+                destination = new BoundingBox(0.0, 0.0, boundingBox.Width, boundingBox.Height).ToSkia();
+            }
+
+            if (flag)
```
pauldendulk: What is the meaning of the flag? Could it have a more meaningful name?

charlenni: Oh, never mind. This class should be dropped, because it is only a silly example, what could be done. Don't include this class in your review or any PR. It replaces normal VectorStyle drawing and leads to the Tests problem.

The flag itself decides, if the cross above the raster image is drawn or not.

Mapsui.Rendering.Xaml/MapRenderer.cs
```diff
@@ -204,6 +210,17 @@ private static Canvas RenderLayerStatic(IReadOnlyViewport viewport, ILayer layer
 
         private static void RenderFeature(IReadOnlyViewport viewport, Canvas canvas, IFeature feature, IStyle style, SymbolCache symbolCache, bool rasterizing)
         {
+            //// Check, if we have a special renderer for this style
```
pauldendulk: This is how a custom renderer for xaml would look like? I would not mind at all to leave it out. Xaml will be removed at some point.

Samples/Mapsui.Samples.Wpf/XamlSpecialStyleRenderer.cs
```diff
@@ -0,0 +1,50 @@
...
+using Mapsui.Rendering.Xaml.XamlStyles;
+using Mapsui.Styles;
+using System;
+using System.Windows.Controls;
+using XamlMedia = System.Windows.Media;
+using XamlShapes = System.Windows.Shapes;
+using XamlPoint = System.Windows.Point;
+using XamlColors = System.Windows.Media.Colors;
+
+public class XamlSpecialStyleRenderer : IXamlStyleRenderer
```
pauldendulk: XamlSpecialStyleRenderer? How does this work? Is this just an example? Or is everything in xaml rendered by one style? Perhaps not support custom renderers for xaml at all and leave everything as is as much as possible.

pauldendulk:
@charlenni What is the status of this PR? Is it just an example of the general setup? (which is fine btw, just need to know). 

In general I like it. It is similar to widgets, which is something I know and we know actually works. 

What I liked most about the idea of 'a callback on the style' is that it would keep things together for the user. For this (and for widget renderers) you need to do a few separate things, create a style, create a style renderer and register the renderer to the style. Perhaps this could be simplified a bit, but if not, then let's do it like this.

I would say for xaml we do not support custom rendering at all.

Mapsui.Rendering.Xaml/MapRenderer.cs
charlenni: Yes, this is the way it should work: Look if we have a StyleRenderer for this style and if yes, call Draw function with Canvas.

Samples/Mapsui.Samples.Wpf/XamlSpecialStyleRenderer.cs
charlenni: I had to add this here, because the common samples could used by Skia and Xaml renderers. The SpecialStyleRenderer for Skia is in the SpecialStyleSample (line Sample.Commons has on all platforms Skia, but not on all Xaml). It is only to show, how the XamlStyleRenderer should look like.

pauldendulk:
@charlenni What's important about this feature is that it gives much freedom to users. Often when a new feature is needed some PR is needed, which might takes time. Users can only start using the feature after the PR has been merged and released. Now many thing can be added without changes to the core.

I can also image users start developing their own style/stylerenderer pair and when they make progress later submit this as a new feature.

charlenni:
> @charlenni What is the status of this PR? Is it just an example of the general setup? (which is fine btw, just need to know).

From my side it is ready. The only disadvantage is, that we have an check mor in the renderer, but if there are no StyleRenderers, nothing changes to the behavior up to now.

> In general I like it. It is similar to widgets, which is something I know and we know actually works.

I thought, it would be the best.

> What I liked most about the idea of 'a callback on the style' is that it would keep things together for the user. For this (and for widget renderers) you need to do a few separate things, create a style, create a style renderer and register the renderer to the style. Perhaps this could be simplified a bit, but if not, then let's do it like this.

You could create only one object, that is Style and StyleRenderer, but this would only work, if you know your type of renderer and you have only one type of renderer. Most user have this, so it would work very well.

> I would say for xaml we do not support custom rendering at all.

Oh, it shouldn't be much work to get it running. Is the same as with WidgetRenderers, so it sould be easy to do.

> @charlenni What's important about this feature is that it gives much freedom to users. Often when a new feature is needed some PR is needed, which might takes time. Users can only start using the feature after the PR has been merged and released. Now many thing can be added without changes to the core.

That was my intention. Always I want to have something new, you have work to do :-D. With this, I could draw all my things by my own.

> I can also image users start developing their own style/stylerenderer pair and when they make progress later submit this as a new feature.

Yes, this could be happen. It would be nice to have more styles than now.

Samples/Mapsui.Samples.Wpf/XamlSpecialStyleRenderer.cs
pauldendulk: oh, just for the sample. Right. I am now doubting whether we should or should not add this for xaml. What I was now thinking of is that with the setup it might be easier and clearer to accept differences between te platforms. Now people often find out afterwards that something is not implemented in xaml. But with this setup there is a list of StyleRenderers which make very clear what is supported. Even for us as developers this is clearer.

charlenni: I always try to create all things for both. I do this, until you say, that the WPF renderer isn't supported anymore. Makes life easier.

pauldendulk:
> From my side it is ready. The only disadvantage is, that we have an check more in the renderer, but if there are no StyleRenderers, nothing changes to the behavior up to now.

Can you explain a bit more about 'check more in the renderer'?

The unit tests fail. Did you look at them? Do you know what is going on? 

This is a big change and I want to first play with it myself (I will use the branch in your repository). I won't have time to merge it today (off the rest of the day). But I hope to wrap this up in the beginnen of the next week. Is that okay?

charlenni:
> Can you explain a bit more about 'check more in the renderer'?

If is the first part, where I check, if we have a StyleRenderer in StyleRenderers for this Style. It is new and is called for each Feature that is rendered.

> The unit tests fail. Did you look at them? Do you know what is going on?

The unit test fails because I added a VectorStyleRenderer, that changes the normal VectorStyle behavior, which is tested in one of the tests. Remove the second commit (you don't loose any importent thing) and try again. It will work without this VectorStyleRenderer sample. I only want to show, that it is possible with this approach to change even the build in styles.

> This is a big change and I want to first play with it myself (I will use the branch in your repository). I won't have time to merge it today (off the rest of the day). But I hope to wrap this up in the beginnen of the next week. Is that okay?

It is ok. Don't hurry. Take your time. It is a small change, but have big impact :-)

pauldendulk:
Okay, thanks for that info, I can work with that. Have a nice weekend! 😎

charlenni:
Same to you.

Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
```diff
@@ -0,0 +1,148 @@
...
+using SkiaSharp;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Mapsui.Rendering.Skia.SkiaStyles
+{
+    /// <summary>
+    /// Demo VectorStyleRenderer
+    /// </summary>
+    public class VectorStyleRenderer : ISkiaStyleRenderer
```
pauldendulk: VectorStyleRenderer is now in the Mapsui.Rendering.Skia, while the existing renderers are also still in there. I can image this was intended to show how it could work if we ported every renderer to the new stylerenderer plugin structure. Is this correct?

I would like to release a version with this plugin functionality soon, so there will be no time to port all existing renderer (and I am not sure yet if we should). What would need to be done for that is to remove the registration of the VectorStyleRenderer and move the VectorStyleRenderer to a samples project. And rename it. .... or perhaps delete it since there is already a SkiaSpecialStyleRenderer that could function as sample. 

There will probably be more improvements but with just a few changes I could release a version tomorrow.

charlenni:
It was intended to show, how actual renderers could be replaced.

As I said above, remove the second commit and the VectorStyleRenderer is history.

Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
pauldendulk: > It was intended to show, how actual renderers could be replaced.
> 
> As I said above, remove the second commit and the VectorStyleRenderer is history.

ah, right. Thanks.

charlenni:
@pauldendulk You now added the demo VectorStyleRenderer. With this, the tests are brocken, aren’t they? This shouldn’t happen.

It could stay as demo, but shouldn’t be called because of the above.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -39,6 +45,8 @@ public MapRenderer()
             WidgetRenders[typeof(Hyperlink)] = new HyperlinkWidgetRenderer();
             WidgetRenders[typeof(ScaleBarWidget)] = new ScaleBarWidgetRenderer();
             WidgetRenders[typeof(ZoomInOutWidget)] = new ZoomInOutWidgetRenderer();
+
+            StyleRenderers.Add(typeof(VectorStyle), new VectorStyleRenderer(this));
```
charlenni: This line should be commented. Than the tests are working again.

charlenni: Perhaps this could go into a sample. Should have made this earlier. Sorry for that 😬

pauldendulk:
> @pauldendulk You now added the demo VectorStyleRenderer. With this, the tests are brocken, aren’t they? This shouldn’t happen.
> 
> It could stay as demo, but shouldn’t be called because of the above.

I removed the registration of the VectorStyleRenderer to fixed the test (did not do this through an MR, sorry, otherwise you would have known). I have to think about what to do with the VectorStyleRenderer. We need a sample for the custom rendering, but there is already something for this, right? And for replacing the default renderers it is perhaps enough for now that we understand how it could work.

pauldendulk:
Also I want to think about how to structure the renderers more in general. So far the primary entry was through geometry type, and then style, but now the style becomes more important. So, should style become the primary entry now? And should the style renderer detect the geometry type? This is something you would like to do only once. Or should we perhaps split everything up in smaller parts. Where a style renderer takes just one geometry type.

pauldendulk:
Just looked at the current code. Perhaps the current solution is the best. Now you have to detect the Geometry type within the style renderer. The disadvantage is that different renderers all need to do this type checking. On the other hand most renderers will just check one type, most often it will be Point, and will ignore the rest. So this is still simple. Also this solution, where we do not bother detecting the geometry type, leaves a lot of freedom for the implementer. Users could even add their own IFeature which has new kind of Geometries (unknown to Mapsui), or perhaps their IFeature implementation could have new properties that affect rendering.

charlenni:
> leaves a lot of freedom for the implementer

I assume, that this is the main reason to do it this way. So let it do this way and look, if it works ok in real life.

PR #840, opened 2020-02-25
Added a SKPaint to all DrawPicture calls to improve drawing quality

charlenni:


pauldendulk:
The BitmapHelper has a default SKPaint which has ```FilterQuality = SKFilterQuality.Low```. This might seem as a performance optimization as you would expect Medium to be the default, but it is not, SKFilterQuality.None is the default. The quality of Low is actually a lot better than defaut None. Medium and High are noticeably better as well, but significantly slower. 

I played around with antialiased back in the day but, and a lot of other parameters that I did not fully understand (like dither) and ended up using filter quality only. Perhaps you could check the results you get using SKFilterQuality.Low.

Furthermore, I am of the opinion that we need a performance benchmark. 😄

charlenni:
@pauldendulk Here it is a problem of copying ready drawn vector graphics to the map. SVG images are created as SKPicture and than drawn with canvas.DrawPicture(). SKPicture Stores no bitmap but drawing commands, that than drawn in DrawPicture(). And it seems, that in OpenGL you have to use a SKPaint with IsAntialiase=true in the DeawPicture() function. Would be no problem, if the Skia scale is 1, but on Android it is 3 and there start the problem.

To be clear: it is only a problem with OpenGL and SVG images as it seems. Has nothing to do with filter quality.

> Furthermore, I am of the opinion that we need a performance benchmark. 😄

Did you mention this before? 😉

pauldendulk:
ah, you are right, nothing to do with bitmaps. I was confused because the first change is in BitmapHelper. Perhaps this could be improved as well. Let me take a look.

charlenni:
No, I don't think so. It is only a problem, when SKPicture is drawn. I checked the other places already. There shouldn't be any problem.

pauldendulk:
> No, I don't think so. It is only a problem, when SKPicture is drawn. I checked the other places already. There shouldn't be any problem.

I understand there is no technical problem but I was confused and I have tried to improve the code in the PR to your PR to avoid this confusion.

PR #844, opened 2020-02-28
Did a null check that could crash the MapView when checking on isAround

pauldendulk:
I clicked around and got an exception. Simple null check fix.


charlenni:
Thank you.

PR #848, opened 2020-03-04
Added a conversion from bitmap to texture backed bitmap

charlenni:
Added a conversion from bitmap to texture backed bitmap for Forms and Android for test.

DO NOT MERGE THIS PR!


pauldendulk:
Closing for now because we have a better solution in the other PR. 

In general I would like to aim for the best performance possible. So eventually we will need to cache skia objects that can be renderer fast, without further conversions.

charlenni:
Yes, this would worth a look. In the moment, we create always new CPU backed SKImages, that have to be transfered to GPU memory. If we could avoid this, perhaps we get more performance. And higher performence isn't always good ;-)

PR #853, opened 2020-03-06
Use Viewport copy constructor in paint methods

pauldendulk:


pauldendulk:
@charlenni Merging PR with copy constructor.

charlenni:
Yes, now it looks much better. The problem was, that the copy function didn't worked with a IReadOnlyViewport. That was also the reason, why I said, that Viewport is confusing.

PR #864, opened 2020-03-08
Tile dispatcher

charlenni:
Added the ability to load tiles with formats other than byte[] with the TileFetchDispatcher and TileLayer.


pauldendulk:
An initial reaction.

I see the tile fetcher as just returning raw data. In this case the format chosen is a byte[]. But a byte array could be anything. In this case it is an image but it does not have to be. I choose for byte array (long ago) because that is a format that is more clear to me than a stream. A stream could have different implementations and may or may not need to be disposed. But these days most data comes as streams over http, and serializing a stream to a specific format is common, so perhaps it should be a stream instead. I would prefer this over returning something of type object. 

Still, now we do a conversion to bitmap, but for vector tiles (that is your purpose, right?) something else needs to be done. Thinking about a solution.

pauldendulk:
In the 'vectortiles' branch there is a GenericTileLayer<T> and a TileLayer which is derived from that and uses a TileParser as argument. Not sure I like that. TileParser is still quite abstract. You would expect the T to be of the type of tile that it returns. In the case of the derived TileLayer that is a Raster (an IGeometry), but that is a type that is used nowhere in the interface. You just put a Raster in a Feature, a list of Features is returned and the geometry is cast to Raster later on in the renderer.

TileParser just turns the byte array into a Raster. Since the raster is nothing more than the raw data there is no real conversion here, but this is the place where such conversion would be appropriate. 

What about just injecting a method for parsing to geometry? No generics needed.

charlenni:
> Still, now we do a conversion to bitmap, but for vector tiles (that is your purpose, right?) something else needs to be done. Thinking about a solution.

Perhaps it is possible to return a interface like `ITile`. Than Raster implements ITile.

> What about just injecting a method for parsing to geometry? No generics needed.

Isn't this, what I'm doing? I use `FetchToFeature` to generate a feature from the result. But at two places, `byte[]` is hard coded, so I had to replace this.

I don't want to change big things. You know, V2 is comming. So I try to get, what I want, with so small changes as possible. Ok, callout wasn't a good example for this ;-)

Do you have an enterprise version with profiler? Sometimes the map movement is a little bit slow. Would be nice to see some profiles, for what the time is wasted.

pauldendulk:
> > What about just injecting a method for parsing to geometry? No generics needed.
> 
> Isn't this, what I'm doing? I use `FetchToFeature` to generate a feature from the result. But at two places, `byte[]` is hard coded, so I had to replace this.

Yes, this is what you are doing given the current architecture. But in the currently architecture you need to inject the converter into the TileLayer and then into the TileDispatcher. I am now thinking of away in which the TileDispatcher could be more stupid. Maybe it should not contain a conversion to Geometry. 

> Do you have an enterprise version with profiler? Sometimes the map movement is a little bit slow. Would be nice to see some profiles, for what the time is wasted.

I could take a look at that. What kind of situations do you see this? Which samples? What do you do?

pauldendulk:
Also now there are two methods injected, perhaps there could be one.

pauldendulk:
Also a more meta remark about these discussions around MRs. I wrote most of the original code in here, but it was a long time ago and it takes some time to get the full picture back. Usually I try to start a conversation before I have the full picture. The conversation itself helps me to get it back. So my initial remarks might be off the mark. I am sorry about this but I prefer to start the conversation with partial knowledge instead of studying it more because the conversation helps me to get my mind going on a topic.

charlenni:
First of all, I would like to say, that one reason, why I like Mapsui and BruTile is their flexibility. With only a few changes, you could do fantastic new things.

This PR is one of this things. With only a few changed lines, the same code code be used for raster and vector tiles. Ok, if you have both in mind, you would have made it here and there a little bit better, but it works.

I assume, that you not very deep in the world of Mapbox GL vector tiles, so I would give you a short information about them. First of all, I use Mapbox GL format for feature data and styling. There are other formats out there (one of the mostly know is Mapsforge), which I always have in mind. Mapbox GL uses a so called style file in JSON format. This contains general informations about the file, all sources, that are used and the styling for this sources. Sources could be raster and vector tiles. The styling is divided in in layers, where the layers are drawn in that order, that they have in the style file. Each style layer refers to a source and a layer in the source data (if it is a vector source). This styles are dynamically and could change with zoom and other context dependent parts.

When loading a tile, you get a number of style layers (normally around 100) with features, that belong to this layer. Some features belong to more than one style layer. This could be e.g. roads, which drawn twice (bigger and smaller to get framed roads). 

So how to bring this to Mapsui? Loading the style file is made with Newtonsoft.Json. Reading the features is made by BruTile, which could read MBTile files and online MVT data. At the end, I have style layers with their belonging features.

Now I have to bring this informations into Mapsui and on the screen. There are different ways to do this. First I could create one layer for each style layer and add a provider, which returns the right features. Second I could return a tile, containing all style layers and features, that are belong to this. Here are again two ways: return a ready drawn tile as drawable or return a list of features with belonging styles. A third would be to create a raster tile from the data.

In the moment I create a drawable tile with prerendered pathes, that are drawn each time with the correct style. The path is created when loading the feature data. It is one path per style layer. So I loose the feature information for MapInfo. But it works fast and relayable. For this I need this PR.

And I have another problem with the TileLayer. It works good, when a tile isn't in the tile source. It makes a fallback and looks, if there is a tile with a lower zoom level and uses this to draw the map. Works good for raster tiles, because they overwrite underlaying lower resolution tiles. But with vector tiles, this doesn't work. They draw only lines and have free parts between them. If you than draw two tiles, you see the underlying tile between the tile on top. For this, TileLayer should have a clipping mechanism. Only that tile part, that is missing, should be drawn. For example, if you draw a map in zoom level 7 and you only have one tile out of four for this and you get the tile in zoom level 6, you should draw only this parts of tile in zoom level 6, that you don't have in zoom level 7. Up to now I don't know how to do this. Perhaps you have an idea for this.

pauldendulk:
> And I have another problem with the TileLayer. It works good, when a tile isn't in the tile source. It makes a fallback and looks, if there is a tile with a lower zoom level and uses this to draw the map. Works good for raster tiles, because they overwrite underlaying lower resolution tiles. But with vector tiles, this doesn't work. They draw only lines and have free parts between them. If you than draw two tiles, you see the underlying tile between the tile on top. For this, TileLayer should have a clipping mechanism. Only that tile part, that is missing, should be drawn. For example, if you draw a map in zoom level 7 and you only have one tile out of four for this and you get the tile in zoom level 6, you should draw only this parts of tile in zoom level 6, that you don't have in zoom level 7. Up to now I don't know how to do this. Perhaps you have an idea for this.

A quick response to this part. There is an IRenderGetStrategy which has two implementations in Mapsui, the RenderGetStrategy which is used as default and the MinimalRenderGetStrategy. You can pass in the MinimalRenderGetStrategy in the TileLayer constructor and it will not draw the higher level tiles. There will be holes when a tile is not fetched yet, but you will not have tiles drawn on top of each other. 

About clipping. Not looked at the current code yet, but I know I have implemented this at some point, perhaps in another renderer. I remember because I was very surprised the clipping was a lot slower than actually drawing the full image. So, for performance no clipping was the best option. But without clipping there is already a problem in the current implementation if you have semi-transparent tile layers. It looks really ugly if you have tiles with just roads and a transparent background.

pauldendulk:
This is the virtual earth layer with semi transparency while not all tiles are loaded yet. The brighter parts is where two semi transparent tile levels are drawn on top of each other. Here it is not so problematic but some tile services are slower and then it is better to use the MinimalRenderGetStrategy. The MinimalRenderGetStrategy could also be used to improve performance, the default RenderGetStrategy has a fancy recursive search which could take more time (but we never measured).

image

pauldendulk:
This is the virtual earth layer with semi transparency while not all tiles are loaded yet. The brighter parts is where two semi transparent tile levels are drawn on top of each other. Here it is not so problematic but some tile services are slower and then it is better to use the MinimalRenderGetStrategy. The MinimalRenderGetStrategy could also be used to improve performance, the default RenderGetStrategy has a fancy recursive search which could take more time (but we never measured).

image

pauldendulk:
If you would like to keep the recursive strategy it should be possible to set a clip rectangle when drawing part of a 'bigger' tile to fill the area of a smaller tile that is missing. The clip rectangle should of course be the size of that missing smaller tiles.

charlenni:
Good to know. Than I test the MinimalRenderGetStrategy with my own overdraw version, which could create each tile from a tile with lower zoom level. No problem with vector tiles, which are always look good :-)

charlenni:
Regarding the profiler: I have a little speed problem. If I test on Android, than it renders the map after a `Navigator.ZoomIn()` some times. Is fast enough, but if the zoom is very high, this is getting slow. Up to zoom level 14 (it is the zoom level, where the mbtiles file has entries), the render times are between 10 and 80 ms. If I go to zoom level 15, then the times are around 200 ms, but saw also 800 ms. On zoom level 26 we are around 1700 ms. That is something you could see. And you see on the screen four lines. On level 14 you see complete Monaco with hundreds of lines. So I'm not sure, what's going on. It is always the same amount of paths (all is rendered from zoom level 14 tiles). Perhaps the lines are bigger and so more pixels have to been drawn.

So, sometimes a profiler would be helpful :-)

charlenni:
Btw. you didn't say something about this PR. Do you think, it is possible to do this? Perhaps without many changes, because of V2? It is a small change, that should not have any impact on others.

pauldendulk:
> Btw. you didn't say something about this PR. Do you think, it is possible to do this? Perhaps without many changes, because of V2? It is a small change, that should not have any impact on others.

I am looking into a way to do this. But I would prefer one method that is injected. To allow this the TileLayer and TileDispatcher should be organized a little different. Maybe the TileDispatcher should be more like the Delayer, which is a class that does not know what the actions it has to execute are about.

pauldendulk:
Is there a way I could reproduce the performance problem?

charlenni:
> I am looking into a way to do this. But I would prefer one method that is injected. To allow this the TileLayer and TileDispatcher should be organized a little different. Maybe the TileDispatcher should be more like the Delayer, which is a class that does not know what the actions it has to execute are about.

But this is a bigger thing. We could add some problems before release V2. So it would be better to do this small change and do the rest later.

> Is there a way I could reproduce the performance problem?

You could use my branch “[VectorTileTest](https://github.com/charlenni/Mapsui/tree/VectorTilesTest)” and the project “[VectorTiles](https://github.com/charlenni/VectorTiles)” to test it. There must be the VectorTiles project in the same directory as the Mapsui folder.

In the meantime I found the profiler function of VS 2019. And I tested to create an extra path for each vector tile feature. If I do this, than it is going extrem slow. So creating one path per layer is the fastest way.

I would say, that the problem is, that the clipping isn’t correct. Perhaps the complete tile is drawn in extra high resolution, but only a small part is visible. Will try to get this.

charlenni:
I tested a little bit more, but it is strange.

If I zoom into the map, the number of `OnPaintSurface()` increases. Higher zoom level are followed by more calls to `OnPaintSurface()`. On zoom level 27 the function is called 26 times. There is only one tile visible and there is no mouse movement. Even, if the map isn't visible on the screen, because another window is hiding the app, than `OnPaintSurface()`is called. Very strange.

pauldendulk:
Do you use the MinimalRenderGetStrategy?

charlenni:
No, I use the normal one.

charlenni:
Ok, one thing I found is, that when exit the map (move over a button and so on), a `RefreshData()` is called. This call updates all layers, which leads to a call `OnPaintSurface()` for each layer. Perhaps we should set a flag, so that only after the refresh of the layer a invalidate is called. But only `TileLayer()` invalidates the canvas. And if I only have one tile on screen (zoom level 19), the invalidate is called 30 times. I checked this with a normal tile source like in the samples and found, that there invalidate is called only two times when you on such zoom levels, where only one tile fills the complete screen. I don't understand this. Why calls my TileLayer 30 times invalidate, where as the normal tile layer with a OSM provider calls only twice.

But now, I assume, that I found the problem. It is, because vector tiles work up to zoom level 30 (and higher), I try to add resolutions up to zoom level 30.  But than the tile fetcher trys to fetch many tiles, which it doesn't find. But each look up leads to a refresh of the map. 

So the question is, how to work with a tile source, that has zoom levels 0 to 14, but could be zoomed in up to 30. If the zoom is e.g. 24, the tile source should return the zoom level 14 tile (and look only for this, because tile sources max zoom level is 14).

pauldendulk:
The TileSchema specifies the levels (Resolutions). The max zoom level is 14 then the resolution should not have levels 'deeper' than that. If it has levels specified where it has no tiles the schema is considered incorrect. In the past an exception was thrown when a tile was not found but later we got a bit less strict.

Below is an example of TileSource created with a schema with min level 0 en max level 14:

```csharp
private static HttpTileSource CreateTileSource()
{
    return new HttpTileSource(new GlobalSphericalMercator(0, 14),
        "https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png",
        new[] { "a", "b", "c" }, name: "OpenStreetMap",
        attribution: OpenStreetMapAttribution, userAgent:"OpenStreetMap in Mapsui");
}
```

pauldendulk:
oh, there is still an Exception, perhaps only the MBTiles returns no exception. Take a look at the TileFetchDispatcher.FetchOnThread, where the exception is caught, and TileFetchDispatcher.FetchCompleted. The exception is passed on as an Error field to the MapControl, the MapControl does no refresh when the Error field is set. Perhaps this is different in your implementation.

pauldendulk:
Btw, there is also the IFetchStrategy, which also has a minimal version, the MinimalFetchStrategy. This might be confusing. There is a strategy for fetching tiles from the cache (IRenderGetStrategy) and a strategy to fetch tiles from the web service, or the mbtiles file (IFetchStrategy). The IFetchStrategy might amplify your problem because it always fetches the tiles at the higher levels. This is to have some tiles to fall back on for the IRenderGetStrategy

charlenni:
I set my IFetchStrategy to 30, so I always get a tile, even if I have only zoom level 0 in the mbtiles file.

Ok, the problem is this: I have a mbtiles file with up to zoom level 14. Now I want to draw the map in, say, zoom level 20. In this zoom level I need, say, 8 tiles. All this 8 tiles have the same source, the tile in zoom level 14, the max of the mbtiles file. So I want to stop to check for this zoom level 20 tiles and return a zoom level 14 tile with larger extent. This is than drawn on the canvas. This would be the best way.

Another way could be to do the same in tile creation. I have this, so it would be possible to create all 8 tiles for zoom level 20. But this isn't so good. All this tiles waste space in the cache and it needs time to create them.

The question is: which is the best way to do this? Could TileLayer/FetchTileDispatcher be used for this?

charlenni:
Ok, one thing I found is, that when exit the map (move over a button and so on), a `RefreshData()` is called. This call updates all layers, which leads to a call `OnPaintSurface()` for each layer. Perhaps we should set a flag, so that only after the refresh of the layer a invalidate is called. But only `TileLayer()` invalidates the canvas.

But now, I assume, that I found the problem. It is, because vector tiles work up to zoom level 30 (and higher), I try to add resolutions up to zoom level 30.  But than the tile fetcher trys to fetch many tiles, which it doesn't find. But each look up leads to a refresh of the map.

Now I changed it in this way, that my TileSource only delivers tiles up to the max zoom level (in this case 14). Beyond this level, it returns null. That work great. But than the ViewportLimiter stops zooming in. Ok, so I set ZoomMode to ZoomMode.Unlimited which should help. This doesn't work, because ZoomHelper doesn't respect this entry. It returns always the last resolution entry from Map.Resolutions, which in turn returns the resolutions of the layers. So what to do? ZoomMode isn't in interface IViewportLimiter and so I don't know, if Map.Limiter has such a property, because it is of type IViewportLimiter.

But if this is all solved, it is lightnig fast :D

So the question is, how to work with a tile source, that has zoom levels 0 to 14, but could be zoomed in up to 30. If the zoom is e.g. 24, the tile source should return the zoom level 14 tile (and look only for this, because tile sources max zoom level is 14).

pauldendulk:
@charlenni Sorry, I could not find time for this this week. The skiasharp testing took more time than planned and starting up homeschooling takes more time.

charlenni:
Oh, and I thought, that this change find its way into the next Nuget. So it is very complicated to test the vector tile sample. Should have someone, which could confirm my performance problems.

charlenni:
@pauldendulk Short reminder. And happy easter!

pauldendulk:
@charlenni Sorry, got lost in the other PRs. Will give this priority next week. Happy easter to you too!

charlenni:
@pauldendulk A short reminder. 2.0.0-rc.3 is out, but this PR isn’t part of it.

pauldendulk:
> You could use my branch “[VectorTileTest](https://github.com/charlenni/Mapsui/tree/VectorTilesTest)” and the project “[VectorTiles](https://github.com/charlenni/VectorTiles)” to test it. There must be the VectorTiles project in the same directory as the Mapsui folder.

If I checkout that branch some projects can not be loaded:

image

pauldendulk:
@charlenni What I would like to avoid is to use ```object``` in the interface. We now have a ```byte[]```, Perhaps a ```Stream``` would be better, since the source is usually a stream coming from an http request, database or file system. However, the current byte[] can function as the equivalent, it is basically a ```Stream``` in another format. That stream can subsequently be converted to any other type in the ```TileFetchDispatcher```. It can easily be converted to ```Stream``` with ```new MemoryStream(myByteArray)```. The downside of this is performance, It is converted twice, once inside BruTile, once inside Mapsui. This only happens when fetching the tile, not in the render loop, so the impact is not that big. 

My idea was to show how this worked in your branch, but I can into the load error.

pauldendulk:
@charlenni Ik started thinking about a better solution in the tilelayer-rework branch but radical changes come too mind. Wat about creating a specific VectorTileLayer at this point, which just copies a lot of code from the TileLayer. Merging more of that code should be done in the beta of 3 which could start soon.

charlenni:
A normal bitmap TileSource delivers a ready rendered image as byte[] or stream. My TileSource returns no byte[] or stream, but an object. In this case a Drawable, something that could draw itself in a canvas when this is invalidated. So I couldn't change this to a stream nor a byte[].

So the only way, if you don't want to use objects, you have to create an interface, that has one function, which could create a feature of the returned object.

Mapsui is completely designed to use bitmap data for the map. The TileProvider/TileSource concept is shorten to this. If you use vector data, than you have more parts: a provider for vector data (could be a mbtiles file or a http source), a style file, a styler, which combines vector data and style data and a renderer, which draws this all to canvas.

Regarding the error: you need the Mapsui branch and the complete [VectorTile](https://github.com/charlenni/VectorTiles) project, which should be in the same folder than Mapsui.

    projects/Mapsui
    projects/VectorTiles

Did you download the other project too? Had to do this, because of the missing PR here. So I need a special Mapsui version to test my project. Couldn't use the normal Mapsui NuGet package.

charlenni:
> @charlenni Ik started thinking about a better solution in the tilelayer-rework branch but radical changes come too mind. Wat about creating a specific VectorTileLayer at this point, which just copies a lot of code from the TileLayer. Merging more of that code should be done in the beta of 3 which could start soon.

Yes, could be made. But the only changes are contained in this PR.

Do you like this approach more? Should I create my own VectorTileLayer and copy the complete code?

charlenni:
I thought about it. A TileDispatcher should fetch tiles, but shouldn't know anything about the type of the tile. It gets tile indexes and a tile source and fetches this tiles. For that, it must not know anything about the structure of this tiles. Tiles could be bitmaps, vector data, drawables or even sounds.

In this case the TileDispatcher converts tile data into bitmap features. This is the problem. And the problem is even earlier: `TileSource.GetTile()` returns also a `byte[]`. Both problems are solved by this PR. With this PR TileDispatcher need not to know anything abut the tile data. It gets a function for retriving tiles from TileSource and convert them to a feature.

If you use streams instead of byte[], than you make nothing better. TileDispatcher is not variable.

pauldendulk:
> Regarding the error: you need the Mapsui branch and the complete [VectorTile](https://github.com/charlenni/VectorTiles) project, which should be in the same folder than Mapsui.
> 
> ```
> projects/Mapsui
> projects/VectorTiles
> ```

Perhaps I don't understand but it seems to me that the VectorTiles project is not inside your git folder. The Mapsui project folder has .git folder. Everything you want to add to the git repository should be in that folder or a subfolder, but you VectorTiles project seems to be outside of it. In the solution (which is in the root) you point to a csproj that with "..\".


```
Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "VectorTiles", "..\VectorTiles\VectorTiles\VectorTiles.csproj", "{4272E906-8F41-41C7-8B10-9F392F3B061A}"
```

pauldendulk:
> Do you like this approach more? Should I create my own VectorTileLayer and copy the complete code?

No, i dislike it, but close to a release a workaround like this might be the best solution. We should strive for reuse of simple basic components. Our components should be designed for this. You remark about the fetcher not knowing about the type of tile fits into such a design. But the exact way to do this needs some time.

pauldendulk:
The use of a Drawable changes things. I have to form a picture of how this works.

This is a Xamarin.Forms [Drawable](https://docs.microsoft.com/en-us/dotnet/api/android.graphics.drawables.drawable?view=xamarin-android-sdk-9)?

Converting to platform specific types is an important performance improvement. Something like that should be done. Why a Drawable and not a SkiaSharp object? It only works in Forms right? 

Do you use MapInfo, or how does that work? Is it just the click event of Forms? Does the Drawable itself contain the info?

charlenni:
In this case a drawable is a SKDrawable from SkiaSharp. It is a object, that has a Draw function, with which the object could draw itself to a SKCanvas. With this, I could draw the vector tile with right settings for e.g. given zoom factor. So the streets could be drawn with a stroke width which is specially calculated for the given float zoom factor. To be clear: a VectorTile is an object, that draws itself each time a OnPaintSurface event is called.

VectorTiles is an independent project and has nothing to do with Mapsui except, that it uses Mapsui as a library. You told me, that this is the way you prefer for extensions. But I could only test this, if I have the right Mapsui (see this PR). Without this special Mapsui version, VectorTiles wouldn’t work. That’s the reason, why I like to have this PR. Than it would be possible to develop VectorTiles with the normal NuGet package of Mapsui. If not, everyone has to create his/her own Mapsui version, which is difficult. 

So to use ist now, you have to download the mentioned VectorTiles project, place it beside Mapsui and use the VectorTilesTest branch. There I added the projects (universal VectorTiles, VectorTiles.MapboxGL and a Forms example) to test. VectorTiles is independent from Forms and could be used for all other platforms of Mapsui. I use normally Forms, because it’s fast and easy to build and test it on Android, iOS and UWP, if you want to see, how fast it is. VectorTiles uses a user defined renderer and works only with the SkiaSharp renderer, because (see above) it draws itself each time to the canvas.

charlenni:
For SKDrawable see the [description](http://skia-doc.commondatastorage.googleapis.com/doxygen/doxygen/html/classSkDrawable.html). I added this to SkiaSharp especially for for this VectorTiles implementation.

pauldendulk:
The drawable approach sounds good. For v3 of Mapsui I want all the transformations to skia objects to be done in a background thread so the rendering itself can be as fast as skiasharp can be. Perhaps something similar can be used. Mapsui will be 2.0.0 will be out this weekend or soon after, so work in v3 can start soon.

Do you think it would be an option to transform all Mapsui features (so, points, lines, polygons and tiles) into a SKDrawable?

My idea has always been that various developers maintain their own libraries where each can reuse the others. For this to work effectively there have to be clear boundaries between these projects, clear responsibilities. For instance I would love it if there was a solid shapefile reading library that I could use. So yes, i think this is a good idea. Work needs to be done on defining clear interfaces and responsibilities. For the MapView it turned out there was too much entanglement, that's why that did not work out in it's initial form. 

About this PR. I do not like the object in the interface. Perhaps there is not better approach, but that is hard for me to judge at this point. If you can help me get to get the run your use case on my machine I could help to find a solution.

You say that without this PR it is not possible to use Mapsui as a library. That is a serious problem I would like to help fix, but right now I do not understand why this is the case. A workaround would be to make a copy of the tilelayer, right? Perhaps simplify that copy to the bare minimum you need for you use case. Is this not correct?

charlenni:
I’m out, but will answer this evening.

charlenni:
> The drawable approach sounds good. For v3 of Mapsui I want all the transformations to skia objects to be done in a background thread so the rendering itself can be as fast as skiasharp can be. Perhaps something similar to this can be used. Mapsui 2.0.0 will be out this weekend or soon after, so work in v3 can start soon.

Sounds good.

One point where this can help is by converting polylines and polygons once to a SKPath and reuse this to draw it again and again.

> Do you think it would be an option to transform all Mapsui features (so, points, lines, polygons and tiles) into a SKDrawable?

Should be no problem. A `SKDrawable` is a little bit like a `SKPicture` (saves all draw commands and replys them when used with `SKCanvas.DrawPicture()`), but draws itselft instead of predrawn commands. So you could decide what to draw in the moment the canvas is updated. The new custom renderers are like SKDrawable. Thinking about it, I would say, they are the same. So expanding the custom renderer concept would be enough. With this, I would say, that all features and their renderers should be moved to this concept.

> My idea has always been that various developers maintain their own libraries where each can reuse the others. For this to work effectively there have to be clear boundaries between these projects, clear responsibilities. For instance I would love it if there was a solid shapefile reading library that I could use. So yes, i think this is a good idea. Work needs to be done on defining clear interfaces and responsibilities. For the MapView it turned out there was too much entanglement, that's why that did not work out in it's initial form.

No, I would say, it worked. But there are many things, that have to be changed for this concept. There was the change to have a common MapControl part for all MapControls. That wasn't a thing for Forms, but for all implementations.

I assume, that there must be more changes to go in this direction. But Mapsui is a very good starting point, because it is very easy to implement new things.

> About this PR. I do not like the object in the interface. Perhaps there is not better approach, but that is hard for me to judge at this point. If you can help me get to get the run your use case on my machine I could help to find a solution.

I will try.

If your Mapsui folder is at `C:\Projects\Mapsui` than download the VectorTiles repository and place it at `C:\Projects`, so that you have the following two folders

    C:\Projects\Mapsui
    C:\Projects\VectorTiles

Now open your VS and select your Mapsui project. Change the Branch to `VectorTilesTest`. Now you should see some additionally projects like `VectorTiles`, `VectorTiles.MapboxGL` and some samples like `Mapsui.Samples.Forms.Android`, `Mapsui.Samples.Forms.iOS` and `Mapsui.Samples.Forms.UWP`. Be carefull. They are not in the Mapsui sample folder. Start the `Mapsui.Samples.Forms.UWP` sample to test it.

> You say that without this PR it is not possible to use Mapsui as a library. That is a serious problem I would like to help fix, but right now I do not understand why this is the case. A workaround would be to make a copy of the tilelayer, right? Perhaps simplify that copy to the bare minimum you need for you use case. Is this not correct?

You have not only to copy the TileLayer, but also for the TileFetchDispatcher and its objects.

The TileLayer is so flexible, that you could use it for bitmap and vector tiles with only such a small change. To be correct, the TileLayer is capable of handling both, but the TileFetchDispatcher, which is created by the TileLayer, isn't. With this PR you could fetch anything, that is tile oriented. And because you don't know, what you fetch with TileFetchDispatcher, there should be an `object` instead of a `byte[]`. Not all tiles are `byte[]`. That's all. And this is difficult to do with an interface.

pauldendulk:
Thanks for you update. I got your sample to work and started work on some changes, but there are too many changes to add them to this release. 

Instead I will just accept your PR, so you can continue your work on vectortiles with more ease. However, I will start on a different setup for TileLayer in 3.0.0-alpha.1 soon after the v2 release. So, more changes on your side will be needed there. 

See here more work on the new TileLayer:
https://github.com/Mapsui/Mapsui/blob/feature/tilelayer-rework/Mapsui/Layers/TileLayer.cs#L152

The TileFetchDispacher is now free of the byte[] type, similar to what you proposed. The TileLayer only used the byte[] in the conversion to Feature, in the two methods at the bottom. I think will already work for your problem. I am not sure if this will stay like this.

pauldendulk:
@charlenni So, my plan is the release the current master branch as 2.0.0.

charlenni:
This all sounds good.

Worked the samples for you? Did you tried it? Fast enough up to now? Do you think it is a good way to go?

pauldendulk:
Yes, they worked. The performance was okay but not spectacular yet. My profiler showed a lot of processing in the fetcher, which makes sense because the conversion is now there. 

I disabled fetching on majorChange and did mostly panning but performance of panning was still not very good and the profiler still showed high use of fetcher logic, which I don't really understand. 

Performance should be one of the main things to focus on in v3.

pauldendulk:
When I disable the timer it seems to be significantly faster.

```csharp
        protected override void OnDraw(SKCanvas canvas)
        {
            canvas.Save();

            //var watch = new System.Diagnostics.Stopwatch();
            //watch.Start();

            foreach (var bucket in Buckets)
            {
                if (bucket != null)
                    bucket.OnDraw(canvas, Context);
            }

            //watch.Stop();
            //System.Diagnostics.Debug.WriteLine($"Draw VectorTile ({TileInfo.Index.Level}/{TileInfo.Index.Col}/{TileInfo.Index.Row}): {watch.ElapsedMilliseconds}");

            canvas.Restore();
        }
```

This is how I disabled fetching on majorChange
```csharp
        public override void RefreshData(BoundingBox extent, double resolution, bool majorChange)
        {
            if (!majorChange) return;

            if (Enabled && extent.GetArea() > 0 && _tileFetchDispatcher != null && MaxVisible >= resolution && MinVisible <= resolution)
            {
                _tileFetchDispatcher.SetViewport(extent, resolution);
                _fetchMachine.Start();
            }
        }
```

PR #873, opened 2020-03-18
Make ZoomHelper fit for overzoom

charlenni:
Changed ZoomHelper, so that it respects the ZoomMode of IViewportLimiter


pauldendulk:
In what kind of situation did this cause problems?

pauldendulk:
pff, complicated. From the Navigator eventually the RestrictedViewport is called. But the problem is that the ZoomHelper restricts before the limiter is involved. The original problem is that the ZoomHelper meddles with things the Limiter is responsible for. The solution you propose solves this problem where it occurs, but a better solution would be if the problem would stay contained within the limiter. Not sure how to deal with this yet.

Perhaps the LimitedViewport was not a good idea to begin with. Or perhaps the ZoomHelper should be called inside the limiter implementation. Perhaps the limiter should be a layer through which the all zoom functions should be called. Thoughts in progress.

charlenni:
It isn’t possible to zoom beyond the last entry in the resolutions array. With this, it isn’t possible to zoom vector tiles more than the last resolution, even when it would be possible.

pauldendulk:
There is an IViewportLimiter interface. There are various implementations, three right now, but users could add more. It would be nice of the limiting logic could be contained within such an interface implementation. In this case there is an issue related to one such implementation, ViewportLimiter, and to solve this it fields are added to the interface and logic wrt limiting is done outside of limiter. 

An alternatieve: What if never limit in the ZoomHelper altogether? For the calls from Navigator this is no problem, SetResolution is called on the limiter. And then there is a call from from the WPF mousewheel and this ends up calling the Navigator.

pauldendulk:
I will merge and then remove the ZoomMode from the interface.

PR #874, opened 2020-03-20
Feature/use pixeldensity in rasterizinglayer

pauldendulk:


pauldendulk:
@tonyfonager This is something I started two weeks ago. Should have been a quick thing but I could not spend much time on it. The rasterizinglayer is now pixelperfect.

After struggling with passing in the PixelDensity I now think I would be best if the MapRenderer is in charge of rasterization. Perhaps just a boolean on the layer. It could even be determined dynamic based on how long it takes to draw a layer. 

@charlenni This also affects the snapshot functionality that you added. After this fix the quality should now be the same as what you see on screen.

pauldendulk:
Below is a image comparison of the rasterized and direct-draw image. The image below shows the differences in purple highlighting. There is only some in the top bar, none in the image itself.

image

@tonyfonager btw, you mentioned some wiggling in the position of the features wrt the background. They may have been caused (or at least enhanced) by a rounding error that was fixed with this.

tonyfonager:
Its SOOO cool and great news. I will test it first thing tomorrow. Thanks alot Paul.

tonyfonager:
If I have to test this, I have to leave the "research" branch with the SkiaSharp fix, correct ?

pauldendulk:
Yes, it is only master. There is no need to hurry though. Testing skiasharp is more important now.

tonyfonager:
I have had NO crashes so far - just sat with the iOS version on a device for 30 minutes, zooming, panning and absolutely no issues 👍

PR #887, opened 2020-04-09
Navigator updates

charlenni:
Added a _updating flag to Forms.MapControl, so that it is possible to prevent refreshing of graphics. This is than used to abonden refreshing when removing and adding special layers.
Removed some more unneeded refreshings of graphics.


charlenni:
I found, that the workaround for the layer order problematic (remove and then add all Forms layers) is problematic for performance, because the map is redrawn each time, a layer is removed or added. With 4 special layers, the map is redrawn 8 times for each layer you add to a map. This is very expensive. So I thought of a solution to not do this. The fastest way was to add a flag, which prevents the redraw of the map in special cases.

If you have a better idea to do this, then I like to know it. Perhaps we should have a public flag in the shared code of MapControl, which prevents a redraw of the map.

charlenni:
You found a part of the PR by yourself in #891 ([a0b397f](https://github.com/Mapsui/Mapsui/pull/887/commits/a0b397faa55c5b76444a33522a7d74c42ef38241)), but another part has the same problem: [76b0cce](https://github.com/Mapsui/Mapsui/pull/887/commits/76b0cce639deae3fbbbcbb1906e2e0c41f71be4e). Should be added to source code.

pauldendulk:
> You found a part of the PR by yourself in #891 ([a0b397f](https://github.com/Mapsui/Mapsui/pull/887/commits/a0b397faa55c5b76444a33522a7d74c42ef38241)), but another part has the same problem: [76b0cce](https://github.com/Mapsui/Mapsui/pull/887/commits/76b0cce639deae3fbbbcbb1906e2e0c41f71be4e). Should be added to source code.

ah, right. i will remove those refreshes as well.

pauldendulk:
@charlenni Removed https://github.com/Mapsui/Mapsui/pull/895

pauldendulk:
@charlenni As a better solution I propose to change the layercollection to this: https://github.com/Mapsui/Mapsui/issues/598#issuecomment-471181466. It will be 2.1 one though. Or maybe we even 3.0.

Btw, 3.0 could be released real quick if we keep to bug count low at all times. This long beta period for 2.0 is mismanagement from my part, and part caused by dependencies.

charlenni:
@pauldendulk Could work, but for this, users have to know how the adding of layers in MapView should work. If they don't know about the existing layers, they add their own layers on top of all others.

pauldendulk:
@charlenni I think that will not be a big problem. As you can see the group argument has a default value of 0 and can be left out of the call for the most common cases. In our samples we will usually use the overloads without the group parameter. For example:
```csharp
Map.Layers.AddToTop(myLayer);
```
Only when you want to work with overlays you add the group parameter.
```csharp
Map.Layers.AddToTop(myGpsLocationWhichShouldAlwaysBeOnTop, 10);
```
So this gps layer will be in group 10 and will stay in group 10.

Only the Clear() method without a group parameter might cause problems. I already mentioned that in the comment to the interface, and now I think it should not remove all layers. If the paremeter is left out it should only remove group 0. 

Perhaps there should be a separate ClearAllGroups, with a warning in the doc summary. Perhaps there should be no such method at all (and users should just know which group to clear if they want to remove everything. If we do this we might want to add an: IEnumerable<int> GetAllGroups().

pauldendulk:
@charlenni I made an update to the interface following my previous comment. Clear should default to group 0 and I think adding a ClearAllGroups is not very confusing and makes it easy if they need to be cleared.

PR #890, opened 2020-04-12
Do not Limit zoom in ZoomHelper

pauldendulk:
@charlenni I now removed the limiting in ZoomHelper. The only place there is there is limiting is in the ViewportLimiter. It is of course better to have it in one place. There was already a problem with the current solution, the ViewportLimiter allowed one level more than the ZoomHelper. I now removed this extra level in the ViewportLimiter to keep the behavior for mouse wheel the same but I am not sure I changed it for some other use case. Either way, the current logic is straight forward and if users want an extra level this is possible through another ViewportLimiter implementation.

### DIFF
--- a/Mapsui/AnimatedNavigator.cs
+++ b/Mapsui/AnimatedNavigator.cs
@@ -185,7 +185,7 @@ namespace Mapsui
         /// </summary>
         public void ZoomIn()
         {
-            var resolution = ZoomHelper.ZoomIn(_map.Resolutions, _viewport.Resolution, _map.Limiter.ZoomMode == UI.ZoomMode.Unlimited);
+            var resolution = ZoomHelper.ZoomIn(_map.Resolutions, _viewport.Resolution);
 
             ZoomTo(resolution);
         }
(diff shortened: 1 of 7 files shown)


charlenni:
So, if I understand it correct, I have to create a new ViewportLimiter to get, what I want (set a range of zoom levels that could be displayed independent from the zoom levels the TileSource has. Is this correct?

pauldendulk:
You can set ViewportLimited.ZoomMode = ZoomMode.Unrestricted to allow zooming outside of the zoom levels in the layers. I think this is rarely what users want.

You can set ViewportLimited.ZoomMode = ZoomMode.KeepWithinResolutions and set ViewportLimiter.ZoomLimits to extend (or restrict) the range outside of the list of Resolutions. This might be sufficient for your use case.

Further, you could implement your own IViewportLimiter for alternative use cases. And there are currently already two other IViewportLimiter implementations that might be useful. 

It is not possible to override the full list of Resolutions.  It would make sense to add this possibility. It is easy to understand. The ZoomHelper logic would have to change to allow this.

charlenni:
Thank you for your help. I will try the second or the first version.

PR #891, opened 2020-04-12
Remove some needless callbacks

pauldendulk:


charlenni:
This was part of #887. Do you than add the other changes of #887 by yourself?

pauldendulk:
@charlenni Not sure how to go about this. I would prefer not to add more state with _updating field. Let's continue the discussion in #887.

Issue #884, opened 2020-04-02
iOS broke between beta37 and beta38


spaddlewit:
I get a blank, frozen map. Works fine on beta37 and below. I thought the Skia dependency might be related (beta37 used preview17, beta38 uses preview39), but I rolled back to beta37 with preview39 and it works just fine.

spaddlewit:
beta37 works fine with Skia preview45, as well. Something going on in Mapsui itself.

pauldendulk:
Thanks for the details. I did not test anything on iOS lately. Will check.

Do you use iOS Forms? or plain iOS?

@tonyfonager You do not have a problem with iOS right? What could be the difference?

pauldendulk:
@spaddlewit Did you also try 2.0.0-rc.1?

tonyfonager:
@pauldendulk I am running on 2.0.0-rc.1 right now, and it works without any issues on both iOS and Android

spaddlewit:
Yes, I'm using Xamarin.Forms.
My base layer is also an mbtiles layer. What's yours, @tonyfonager ?

spaddlewit:
Yes, I tried 2.0.0-rc.1. They all act the same post-beta37.

pauldendulk:
wow, thanks for the fast responses.

tonyfonager:
I currently have disabled my mbtiles layer, and use the built in OSM map provider while testing.

`Map.Layers.Add(OpenStreetMap.CreateTileLayer());`

pauldendulk:
There could be issues with the mbtiles packages. Did some updates and they are very confusing.

tonyfonager:
@pauldendulk I could try testing with mbtiles file again, but I am running on a modified version of BruTiles, so it might not give the expected result ...

spaddlewit:
I'll also note that UWP, other than my previous PopAsync bug, works fine on 2.0.0-rc1. So it's something that changed iOS-side in beta38...

spaddlewit:
Alright, some more results from testing.. in beta38..
I loaded an online tile service instead of mbtiles-
If I navigate to another screen, then navigate back, I noticed that the map flashed for a quick moment in a greyish-blue color.. if I hold down my finger and pan around, it starts drawing the map, but it looks to be using the farthest-zoomed-out tile zoom level, while being zoomed in closer than what that tile level would normally be displayed at.

Sebastian1989101:
[...]

PR #897, opened 2020-04-20
Removed storing of PixelDensity

charlenni:
With this, PixelDensity is calculated each time new. There was a problem with storing this value in the Forms implementation (IsVisible = true after adding MapControl/MapView). All other MapControls except WPF don't have to calculate this value. For WPF it could be good to save this calculated value.

For the problem see #884.

### DIFF
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -63,16 +63,9 @@ namespace Mapsui.UI.Wpf
             }
         }
 
-        private float? _pixelDensity;
-
         public float PixelDensity
         {
-            get
-            {
-                if (_pixelDensity == null || _pixelDensity <= 0)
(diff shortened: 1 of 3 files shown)


pauldendulk:
Would prefer to store it, but if this simple change fixes that annoying issue that would be very good news.

pauldendulk:
Is the IsVisible still necessary if you do not store the PixelDensity? Of would it just look weird (changing the density at startup?). 

What if the PixelDensity is set at the event where you set IsVisible to true?

charlenni:
I tried to set it in the OnPropertyChanged event,  it there the CanvasSize wasn’t correct. I’m not sure, when it is set correct. All places I tried it was 2 and lead to a PixelDensity of 0.0067.

Perhaps it could be stored and checked, if it is below 1. If this is the case, calculate it. Would this be a solution?

Issue #898, opened 2020-04-23
After zooming or panning, clicking doesnt work


johnhs91:
Hello

I got a map, in little view, when you double click it, it will open in full screen.
Problem is, if I pan or zoom. (Via 2 fingers) something is triggered in Mapsui, so I can't double click anymore.
Everytime I double click afterwards, it still triggers pan or zoom.

I've debugged, and it go into SKTouchAction.Moved, if I just use 1 finger and double click.
Worst part of this is, when I go into OnTouchMove, the method that is triggered in touch. It actually belive I used 2 fingers on the map. 

I've tried to make a workaround where I disable pan and zoom on the little view. But it doesn't help, same bug occurred.

I'm using Mapsui 2.0 beta 35

Br. John Sørensen

charlenni:
You didn’t mention, which Mapsui control do you use. Android, iOS, UWP or Forms?

johnhs91:
I'm using forms sorry.
The issue is on both android and iOS.

charlenni:
I tried to reproduce this here on iOS, but couldn’t find a problem. Panning and zooming with two fingers and than double click on the map works perfect. Tested it with the “Add Pin Sample” for Forms.

Do you use two different MapViews for this? Could it be, that you didn’t set the double click flag to true?

johnhs91:
I use the same view for both.
I've been debugging in control to check, if it was my code. But it's all in Mapsui i find the issue.

OnTouch in MapControl (Forms) trigger only moved, if I moved with 2 fingers once. And when it call OnTouchMove it always find 2 fingers. So it make it pan or zoom instead of click somehow.

To be honest I'm not sure if iOS and android is the same problem. I've only debugged your code on android. So I don't know how there can be a difference here.
If you try to do same with android, does it work too then?

johnhs91:
I can easily double click the map, if I don't zoom or pan before that.
And I got no code that manipulate with mapsui code. So I have no clue why this happens.
Can it be the skiasharp version? or is it fixed ?

charlenni:
Ok, I now checked it on Android with the "Add Pin Sample". No problem. Moving/zooming with two fingers around and than make a double click on the map leads to a OnDoubleTapped event for the MapControl.

Could you test it with the normal Forms sample? Could you provide some code to test it here?

johnhs91:
Wierd part from my side is, that it only happen when the map is small.
[...]

PR #902, opened 2020-04-30
[Forms] MapControl clears all touches when gesture is cancelled

charlenni:
When gesture in MapControl for Forms is cancelled, all touches are cleared. See #898.


charlenni:
Ah, @radderz, online again. That’s nice. I hope, you are well, especially with all this fires down under.

You said sone time ago, you have a good fling implementation for Mapsui. Is this right? Would it be possible to share this with us?

Issue #908, opened 2020-05-05
In WPF at mousewheel zoom-out the Map sticks to the mousepointer


pauldendulk:
What happens is that if you zoom out with the mouse wheel and at the same time mouse the mouse outside of the edge of the visible map the map will move along with the mouse as if it is glued to it.

pauldendulk:
### The problem

This was hard to debug. See the code below. There are several steps to it. What this code did not anticipate is that the _viewport here is the LimitedViewport and it will limit the already at the first SetCenter which messes up the logic. 

```csharp
        public void ZoomTo(double resolution, Point centerOfZoom)
        {
            // 1) Temporarily center on the center of zoom
            _viewport.SetCenter(_viewport.ScreenToWorld(centerOfZoom));

            // 2) Then zoom 
            _viewport.SetResolution(resolution);

            // 3) Then move the temporary center of the map back to the mouse position
            _viewport.SetCenter(_viewport.ScreenToWorld(
                _viewport.Width - centerOfZoom.X,
                _viewport.Height - centerOfZoom.Y));

            Navigated?.Invoke(this, EventArgs.Empty);
        }
```

### The Fix
To solve this I added this logic to the Viewport itself so that from here only _viewport.SetResolution(resolution, centerOfZoom) needs to be called.

### What could be improved
To solve this some more logic needed to be added to the Viewport which I would rather keep simple. Perhaps the layering of Viewport and LimitedViewport is not necessary anymore now that the Navigator is added. Perhaps the LimitingViewport should be deleted and the ViewportLimiter should be called from the Navigator.

pauldendulk:
Fixed with #909

PR #909, opened 2020-05-05
Fix the map sticks to mouse problem #908

pauldendulk:


charlenni:
Congratulation! The first found bug in the new version. 2.0.1 is comming :D

pauldendulk:
@charlenni Not sure if this is meant as a criticism, but I made no illusions about this happening. This is the way things always go, and I don't think we should have shipped earlier. There is a distinction in bugs you can ship and bugs you can't ship. You can't ship crashes that happen at random moments that have no workaround and I think the iOS startup bug is not shippable https://github.com/Mapsui/Mapsui/issues/884.

btw, this bug was present for about a year and was never reported. Also, it was already the second bug. @tonyfonager found the first, it is more serious. https://github.com/Mapsui/Mapsui/pull/907.

charlenni:
@pauldendulk No, no, it was a funny comment. I know, that this is the way things always go. If you say, now we can ship it, than you find the next bug. It is the same when you write a paper: if you leave the copy shop, you find again a mistake :-)

PR #923, opened 2020-05-15
Replace SQLite path for MBTiles files with full path

charlenni:
Replace SQLite path for MBTiles files with full path, so that doesn't throw an exception on UWP


charlenni:
Tested it on UWP and Android.

PR #929, opened 2020-05-24
Feature/cumulative resolutions on wpf mouse wheel

pauldendulk:
When using the new Navigator animations in WPF this failed to cumulate the mouse wheel event. With cumulative mouse wheel event I mean that if a mouse wheel event is fired and during the animation a new mouse wheel event comes in, then the target resolution is not just the next resolution but the one after that. 

Without this mechanism a second mouse wheel event would start a new animation to the same target which would only delay the time it would take to reach that target.

### DIFF
--- a/Mapsui.UI.Wpf/MapControl.cs
+++ b/Mapsui.UI.Wpf/MapControl.cs
@@ -5,7 +5,6 @@ using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Input;
 using System.Windows.Media;
-using System.Windows.Media.Animation;
 using System.Windows.Shapes;
 using Mapsui.Layers;
 using Mapsui.Providers;
@@ -29,20 +28,15 @@ namespace Mapsui.UI.Wpf
 
     public partial class MapControl : Grid, IMapControl
     {
-        // ReSharper disable once UnusedMember.Local // This registration triggers the call to OnResolutionChanged
-        private static readonly DependencyProperty ResolutionProperty =
-            DependencyProperty.Register(
-                "Resolution", typeof(double), typeof(MapControl),
-                new PropertyMetadata(OnResolutionChanged));
-
         private readonly Rectangle _selectRectangle = CreateSelectRectangle();
-        private readonly DoubleAnimation _zoomAnimation = new DoubleAnimation();
         private Geometries.Point _currentMousePosition;
         private Geometries.Point _downMousePosition;
         private bool _mouseDown;
         private Geometries.Point _previousMousePosition;
         private RenderMode _renderMode;
         private double _toResolution = double.NaN;
+        private int _mouseWheelTickCount;
+        private const int _mouseWheelAnimationDuration = 1000;
         private bool _hasBeenManipulated;
         private double _innerRotation;
         private readonly FlingTracker _flingTracker = new FlingTracker();
@@ -106,7 +100,7 @@ namespace Mapsui.UI.Wpf
                 Visibility = Visibility.Collapsed
             };
         }
-        
+
         public Canvas WpfCanvas { get; } = CreateWpfRenderCanvas();
 
         private SKElement SkiaCanvas { get; } = CreateSkiaRenderElement();
@@ -153,12 +147,6 @@ namespace Mapsui.UI.Wpf
             };
         }
 
-        [Obsolete("Use Viewport.ViewportChanged", true)]
-        // ReSharper disable once UnusedMember.Global
-#pragma warning disable 67
-        public event EventHandler<ViewChangedEventArgs> ViewChanged;
-#pragma warning restore 67
-
         public event EventHandler<FeatureInfoEventArgs> FeatureInfo; // todo: Remove and add sample for alternative
 
         public void RefreshGraphics()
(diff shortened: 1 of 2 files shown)


pauldendulk:
@charlenni @radderz Could you take a look?

pauldendulk:
I also added Easing.QuarticOut because this was what was used before. When comparing the new animation with the previous one I still see a difference in smoothness but it is hard to pinpoint what makes the difference. Setting the Animation.Interval to 4 did not improve this. Possible causes that I can think of:
- The new animation is heavier which makes everything run slower (not very likely)
- Somewhere some extra processing is done because we set both Resolution and Center in these animations (did not check this).
- The stop and start of the Animation timer causes some jumps in smoothness. It should be possible to keep the timer running and only change the time it ends.

pauldendulk:
The 'cumulative resolutions' are now implemented by checking if the TickCount is more than 1000 ms ago. An alternative would be to have some callback to indicate if the timer is completed. This was not available but actually I think it is better this way. Callback complicate things.

pauldendulk:
There is one other difference with the old implementation and this is that during the animation the current mouse position was used to center. The new implementation only uses the mouse center at the start of the animation. Now sure if there is an good fix for this. One possibility, if the previous mouse wheel event is less than 1000 ms ago a mouse move event would restart the animation to a new target.

btw, the restart during animation used here only works (meaning, feels okay to the user) because an ease *out* animation is used, this starts fast. An ease that start slow would make it look weird.

pauldendulk:
btw, the cumulative logic is not easy to understand. Perhaps the code should be a bit more self documenting.

pauldendulk:
If i try to take into account the current mouse wheel position (see the code below) the animation becomes clearly less smooth. So my guess is that starting and stopping the animation has a big impact. I would think it does not have to be. @charlenni indicated in another thread that the values are calculated at the moment they are needed (which I think is good) so setting a new target for the timer should have nearly no impact. Just set a new value to animate to and change the moment the timer should stop. 

```csharp
private void MapControlMouseMove(object sender, MouseEventArgs e)
{
   // Removed code
    if (_mouseDown)
    {
        // Removed code
    }
    else
    {
        if (DuringMouseWheelAnimation())
        {
            Navigator.ZoomTo(_toResolution, _currentMousePosition, _mouseWheelAnimationDuration, Easing.QuarticOut);
        }
    }
}
```

pauldendulk:
Now looking at the Animation code. Currently we work with a Duration field. Instead we could work with EndTime. A new animation would calculate a new EndTime from currentTime + duration. In the HandleTimerElapse Stop is called and the current time has passed the EndTime.

charlenni:
Perhaps you need a flag to check, if an animation is running? I don’t like the 1000 ms approach. It is hardcoded value, which brings problems.

Regarding the center point: I‘m not sure, what other libraries do? Both sound good for me. To zoom around the point, where the mouse cursor is, or to zoom and bring the mouse cursor to the center.

charlenni:
Perhaps the animation should go into the paint surface function: check, if there is a running animation and, if yes, then update to the new values. This values calculated by the time past since the start of the animation. Then redraw the surface. And the timer do a invalidate surface each time it ticks. Perhaps this is smoother.

radderz:
Stopping and starting could have a problem where the first tick is always 16ms away from now. I.e. we dont do the first animation frame straight away. So if you stop and start multiple animations you'll get a jittering during those stop/starts. You could have been 15ms since last frame, and then you stop and start, which would add 16ms extra till the next change. We could make the first frame be done straight away somehow or not have a timer that stops, which is always running somehow. Is our timer on the main thread? As if it isnt a ui thread timer maybe this comes at a cost when redrawing?

pauldendulk:
> Perhaps you need a flag to check, if an animation is running? I don’t like the 1000 ms approach. It is hardcoded value, which brings problems.

The 1000 has always been hard coded. What are good alternatives? We could add a setter for this specific value. But if we do we could also make the Easing settable. These are very specific things to set and for the user not very clear what they do unless they inspect the code. A better way may be move the whole animation more to the outside. Perhaps some kind of Animation object should be settable on some MouseWheelAnimation field.

pauldendulk:
> Stopping and starting could have a problem where the first tick is always 16ms away from now. I.e. we dont do the first animation frame straight away. So if you stop and start multiple animations you'll get a jittering during those stop/starts. You could have been 15ms since last frame, and then you stop and start, which would add 16ms extra till the next change. We could make the first frame be done straight away somehow or not have a timer that stops, which is always running somehow. Is our timer on the main thread? As if it isnt a ui thread timer maybe this comes at a cost when redrawing?

A timer that does not stop seems a good option. 

About the UI-thread/background thread, I had not thought about this. What would be the performance impact? A timer on the main thread consumes power of the ui thread? ... I can also image a background thread could be bad for performance because you have to dispatch it on the main thread.

Another option would be to work with an refresh mode. I mentioned this earlier. No timer at all. At the end of the renderloop we check if an animation is running and if so call Invalidate. The skia renderloop will automatically start a new render event but this will be capped by the max fps of skia.

pauldendulk:
> Regarding the center point: I‘m not sure, what other libraries do? Both sound good for me. To zoom around the point, where the mouse cursor is, or to zoom and bring the mouse cursor to the center.

This is not the issue I was thinking about. I was only talking about the 'zoom around the point' mode. But there are two ways it could function:
- Take the the mouse position at the  moment the mouse wheel is used.
- Take into account the mouse move during the animation. This is how it previously worked and this is what I liked as a user. While the map is zooming in you can guide it to the exact location you are interested in.

charlenni:
> Another option would be to work with an refresh mode. I mentioned this earlier. No timer at all. At the end of the renderloop we check if an animation is running and if so call Invalidate. The skia renderloop will automatically start a new render event but this will be capped by the max fps of Skia

Yes, and it does nothing else but updating the screen, because it runs again and again in the main thread.

PR #930, opened 2020-05-24
Changed system of updating animation

charlenni:
Changed system of updating animation: Timer ticks invalidate canvas and repaint of canvas updates navigator animations before redrawing the canvas. So each time the canvas should be repaint, the animation is updated.

### DIFF
--- a/Mapsui.UI.Android/MapControl.cs
+++ b/Mapsui.UI.Android/MapControl.cs
@@ -93,6 +93,9 @@ namespace Mapsui.UI.Android
 
             e.Surface.Canvas.Scale(PixelDensity, PixelDensity);
 
+            if (Mapsui.Utilities.Animation.NeedsUpdate)
+                Mapsui.Utilities.Animation.UpdateAnimations();
+
             Renderer.Render(e.Surface.Canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
         }
 
@@ -148,6 +151,9 @@ namespace Mapsui.UI.Android
         private void CanvasOnPaintSurfaceGL(object sender, SKPaintGLSurfaceEventArgs args)
         {
             args.Surface.Canvas.Scale(PixelDensity, PixelDensity);
+
+            if (Mapsui.Utilities.Animation.NeedsUpdate)
+                Mapsui.Utilities.Animation.UpdateAnimations();
 
             Renderer.Render(args.Surface.Canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
         }
--- a/Mapsui.UI.Android/MapControl.cs
+++ b/Mapsui.UI.Android/MapControl.cs
@@ -9,6 +9,7 @@ using Android.Util;
 using Android.Views;
 using Mapsui.Geometries.Utilities;
 using Mapsui.Logging;
+using Mapsui.Utilities;
 using SkiaSharp.Views.Android;
 using Math = System.Math;
 using Point = Mapsui.Geometries.Point;
@@ -74,6 +75,8 @@ namespace Mapsui.UI.Android
 
             Map = new Map();
             Touch += MapView_Touch;
+
+            Mapsui.Utilities.Animation.AnimationTimer = new AnimationTimer(this);
 
             var listener = new MapControlGestureListener();
 
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -250,6 +250,9 @@ namespace Mapsui.UI.Forms
         {
             if (PixelDensity <= 0) return;
 
+            // Update any running animation
+            ((Navigator)_navigator).UpdateRunningAnimation();
+
             skPaintSurfaceEventArgs.Surface.Canvas.Scale(PixelDensity, PixelDensity);
 
             var canvas = skPaintSurfaceEventArgs.Surface.Canvas;
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using Mapsui.Geometries.Utilities;
 using Xamarin.Forms;
 using System.Threading.Tasks;
+using Mapsui.Utilities;
 
 namespace Mapsui.UI.Forms
 {
@@ -87,6 +88,8 @@ namespace Mapsui.UI.Forms
         {
             Map = new Map();
             BackgroundColor = Color.White;
+
+            Mapsui.Utilities.Animation.AnimationTimer = new AnimationTimer(this);
 
             EnableTouchEvents = true;
 
@@ -251,7 +254,8 @@ namespace Mapsui.UI.Forms
             if (PixelDensity <= 0) return;
 
             // Update any running animation
-            ((Navigator)_navigator).UpdateRunningAnimation();
+            if (Mapsui.Utilities.Animation.NeedsUpdate)
+                Mapsui.Utilities.Animation.UpdateAnimations();
 
             skPaintSurfaceEventArgs.Surface.Canvas.Scale(PixelDensity, PixelDensity);
 
--- a/Mapsui/Navigator.cs
+++ b/Mapsui/Navigator.cs
@@ -31,6 +31,7 @@ namespace Mapsui
         }
 
         public EventHandler Navigated { get; set; }
+        public EventHandler<AnimationEventArgs> AnimationTick { get; set; }
 
         public Navigator(Map map, IViewport viewport)
         {
@@ -121,6 +122,7 @@ namespace Mapsui
 
                 _animation = new Animation(duration);
                 _animation.Entries.AddRange(animations);
+                _animation.Ticked += (s, e) => { AnimationTick?.Invoke(s, e); };
                 _animation.Start();
             }
         }
@@ -174,6 +176,7 @@ namespace Mapsui
 
                 _animation = new Animation(duration);
                 _animation.Entries.AddRange(animations);
+                _animation.Ticked += (s, e) => { AnimationTick?.Invoke(s, e); };
                 _animation.Start();
             }
         }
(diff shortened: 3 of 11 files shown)


Mapsui.UI.Forms/MapControl.cs
```diff
@@ -250,6 +250,9 @@ void OnPaintSurface(object sender, SKPaintGLSurfaceEventArgs skPaintSurfaceEvent
         {
             if (PixelDensity <= 0) return;
 
+            // Update any running animation
+            ((Navigator)_navigator).UpdateRunningAnimation();
```
pauldendulk, line 254: ah right, this is closer to what I had in mind when talking about the animated points. Now the calculation is actually done before the render. The calculation in HandleTimerElapse was done right before the Invoke, but and Invoke does not have to be followed immediately by an render update, so that will not work well in some cases. Can the calculation in HandleTimerElapse now be removed? ... if we add this to all render loops.

Mapsui/Utilities/Animation.cs
```diff
@@ -94,6 +94,11 @@ public void Stop(bool gotoEnd = true)
         /// <param name="e">Timer tick arguments</param>
         private void HandleTimerElapse(object sender, ElapsedEventArgs e)
         {
+            Ticked?.Invoke(sender, new AnimationEventArgs((_stopwatch.ElapsedTicks - _stopwatchStart) / _durationTicks));
```
pauldendulk, line 97: oh the code is removed from HandleTimerElapse already.

Mapsui/Navigator.cs
```diff
@@ -487,6 +494,7 @@ public void FlingeWith(double velocityX, double velocityY, long maxDuration)
 
             _animation = new Animation((long)animateMillis);
             _animation.Entries.AddRange(animations);
+            _animation.Ticked += (s, e) => { AnimationTick?.Invoke(s, e); };
```
pauldendulk, line 497: Lot's of events handlers to add. I would hope it is possible to keep the Animation and just alter it instead of re-instantiating.

charlenni: Yes, you are right. That‘s all, because I don’t know anything about Navigator or MapControl in Animation. So I have to do this with events.

Perhaps it would be possible to move the timer from Animation to Navigator. So it will be one event less. But then, Animation wouldn‘t work independent.

Any ideas, how to solve this?

charlenni:
The animation is ok, but how to invalidate the canvas from the animation? That is the problem. So how could we inform the MapControl, that it should start a timer, that regularly invalidates the canvas? Or should we inform Animation about the MapControl?

radderz:
We can do the calculation just before, but I think we will also need to know how far through the animation this is and how long since the previous tick for things like fling. As fling works by looking at the velocity, and the duration since last tick and adds that distance on then the velocity gets reduced by the easing. Maybe you guys can think of a better approach to that process but if know the number of ms since last timer tick then it'd enable this change without affecting fling.

The good thing about this approach is that we can invalidate the canvas at the beginning of the animation function, and we can just update the timer end time and let the timer continue running. We could probably just leave the timer running indefinitely (low cost) while the map is visible.

charlenni:
Why not create a IAnimationTimer with only two functions: Start() and Stop(). This is additive to Animation as

    public static IAnimationTimer AnimationTimer;

This timer is created platform dependent in the MapControl and is added in the constructor of MapControl to Animation.

The Start() and Stop() functions need a counter, so that it is possible to call it from more than one animation. Start increase a counter, stop decrease the counter. If in stop counter gets zero, the underplaying timer is stopped. If start is called with counter is zero, then the timer is started.

With this approach, we can drop all events. And we can use any counter we want.

charlenni:
> We can do the calculation just before, but I think we will also need to know how far through the animation this is and how long since the previous tick for things like fling. As fling works by looking at the velocity, and the duration since last tick and adds that distance on then the velocity gets reduced by the easing. Maybe you guys can think of a better approach to that process but if know the number of ms since last timer tick then it'd enable this change without affecting fling.

Nothing changed in the animation itself. So you still get a value between 0 and 1 calculated by the ticks since starting the animation. Now you don’t get them when the timer ticks, but before the canvas is redrawn. The timer does only one thing: send a command, so That the canvas redraws.

> The good thing about this approach is that we can invalidate the canvas at the beginning of the animation function, and we can just update the timer end time and let the timer continue running. We could probably just leave the timer running indefinitely (low cost) while the map is visible.

No, that isn’t a good idea. The timer invalidates the canvas each 16 ms, so we redraw the map each 16 ms.

I have to do some rework on this. I want to use an interface for start and stop the timer as suggested above. And I want to have a list of all animations (`public static List<Animation> Animations`) so that it is possible to have any type of animation independent from Navigator (e.g. for a pulsing my location icon). Then we will check in the render function, if there are any animations in the list and, if yes, call all their ticks. The first animation starts the timer, the last animation stops the timer. We should use locks for adding and removing animations to the list and when using it in the render function.

Do you think, it would work this way?

pauldendulk:
> > The good thing about this approach is that we can invalidate the canvas at the beginning of the animation function, and we can just update the timer end time and let the timer continue running. We could probably just leave the timer running indefinitely (low cost) while the map is visible.
> 
> No, that isn’t a good idea. The timer invalidates the canvas each 16 ms, so we redraw the map each 16 ms.

I agree with charlenni. There should be no redraw. From a performance perspective this is no problem but it will drain the battery and heat up the processor. I worked on projects where this was unacceptable for the customer.

Perhaps radderz had something else in mind, let the timer run but only update if something changed. We would not some kind of flag for that.

pauldendulk:
> I have to do some rework on this. I want to use an interface for start and stop the timer as suggested above. And I want to have a list of all animations (`public static List<Animation> Animations`) so that it is possible to have any type of animation independent from Navigator (e.g. for a pulsing my location icon). Then we will check in the render function, if there are any animations in the list and, if yes, call all their ticks. The first animation starts the timer, the last animation stops the timer. We should use locks for adding and removing animations to the list and when using it in the render function.
> 
> Do you think, it would work this way?

I would really like to avoid locks. My idea was that the timer would stop if all animations have halted, but during the animations we should just set a few fields if the animation changes.
- New target
- New origin (the current state based on the animation).
- new endtime.

btw, I am also thinking about separating the animation from navigator. It is should become something mirroring Viewport (a bit like the LimitingViewport).

charlenni:
During the animation we haven’t any locks. We have locks at start and stop of animation To update the timer count and one when copying the list of animations in redraw to be sure, that the list of animations don’t change in this moment.

charlenni:
The animation has nothing to do with Navigator not with Viewport. To be clear, with this animation you could do, what you want, even move buttons on the screen around.

radderz:
> I agree with charlenni. There should be no redraw. From a performance perspective this is no problem but it will drain the battery and heat up the processor. I worked on projects where this was unacceptable for the customer.
> 
> Perhaps radderz had something else in mind, let the timer run but only update if something changed. We would not some kind of flag for that.

The timer could keep running, but we would only invalidate the canvas when there is an animation remaining..

charlenni:
Ok, made a redesign of the animation. Works now without events. Uses a timer, which implements IAnimationTimer. This timer gets a IMapControl when created and uses this to call a canvas refresh. With this it is possible to use animations for other things than Navigator. See for this the new sample.

One problem I have. if you use the new sample with Forms, Forms is completely frozen. It seems, that for Forms the duration of 16 ms is to low. Or anything else happens, that I don't understand.

Mapsui.UI.Android/MapControl.cs
```diff
@@ -93,6 +93,9 @@ private void CanvasOnPaintSurface(object sender, SKPaintSurfaceEventArgs e)
 
             e.Surface.Canvas.Scale(PixelDensity, PixelDensity);
 
+            if (Mapsui.Utilities.Animation.NeedsUpdate)
```
pauldendulk, line 96: Random thoughts without an overview: I like a separate Animation object with which the MapControl interacts. I like the update here in the loop. As I mentioned somewhere else I think an option for refresh would be to call Invalidate after Render is called. Something like this:

var needsUpdate = Mapsui.Utilities.Animation.NeedsUpdate;
if (needsUpdate) Mapsui.Utilities.Animation.UpdateAnimations();
Renderer.Render(e.Surface.Canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
if (needsUpdate) Invalidate(); // will trigger another render update.

pauldendulk: I guess the Animation has static state. This is something I try to avoid. Not sure how hard that would be in this case.

pauldendulk:
> Works now without events.

nice

> One problem I have. if you use the new sample with Forms, Forms is completely frozen. It seems, that for Forms the duration of 16 ms is to low.

I hope I can look at this today.

Mapsui.UI.Android/MapControl.cs
charlenni: If you do it this way, then you have a endless loop until all animations are stopped. You update the animations, render the canvas and call a update for the canvas again. This would lead to an updated animation, which leads again to a repaint and that to an updated animation...

Animation has two static things: the AnimationTimer, so it could be call Start() and Stop(), and an array with all Running animations, so that MapControl update all animations. If you have a better idea, how to do this, I‘m open.

pauldendulk: > If you do it this way, then you have a endless loop until all animations are stopped.

Isn't this exactly what we want? This is what the timer is doing, calling refresh updates until the animation ends. 

What you may fear is that this would cause a stack overflow or perhaps that updates are called too often, but this is not how such a render loop works (or should work). If you call Invalidate this does not immediately trigger a the OnPaint method, it merely sets a flag to indicate that a repaint is necessary. It is up to the renderloop to schedule this repaint event.  The timing of the next OnPaint depends on how the max fps is set. 

With the timer we we need to figure out how to match the timer events to the OnPaint events. This problem does not exist if we directly work with the OnPaint events.

charlenni: I thought, that Skia tries to be as fast as possible. But it could perhaps work. You could try it.

PR #937, opened 2020-06-03
Feature/mouse move during mouse wheel animation

pauldendulk:
This PR is for research only!


pauldendulk:
I do not have the intention to merge this pull request. 

It is an attempt to work with a fixed set of animations that are added in the Navigator constructor. These animations can be enabled and disabled but will not be removed. This allows us to keep the timer running at all time. When all animations are disabled no map refresh should be called but the timer could keep running. 

Another thing I tried to do at the same time, is work only with animations for the 3 main variables:
- Center
- Resolution
- Rotation
All animation should go through these animations. For this the fling animation should be changed, but this is possible. What is not possible with this setup is the two step animation of FlyTo. Perhaps this could by solved by making some changes to the animation, but I don't want to go that way right now.

I learned that the ZoomTo resolution and centerOfZoom was already problematic in the implementation now in master. I think that has to be rewritten. The current 3 step logic should be replaced with separate Center and Resolution updates. This may be one of the first things I will do next.

Also I will look for another way we could do animation while keeping the timer running at all times. Perhaps by adding all current animation entries at the start. Or finding a solution where we can add new items without start and stap. Perhaps using a queue to send new animations.

Mapsui/Navigator.cs
```diff
@@ -639,10 +613,5 @@ private void FlingTick(AnimationEntry entry, double value)
 
             Navigated?.Invoke(this, EventArgs.Empty);
         }
-
```
charlenni, line 642: That’s not correct. At the end should be a call to ViewportChanged.

PR #940, opened 2020-06-05
Replace tree step centerOfZoom with direct set of center and resolution

pauldendulk:


charlenni:
If I understand it correct, then you want to zoom and set the center of the map to the point, where the user clicked on the map. Both should move together. So why do you not make two AnimationEntrys, one for the zoom and one for the move?

pauldendulk:
The previous ResolutionTick method also updated the center. This had grown like that because the 'center of zoom' logic worked with a three step calculation, that was just because that was understandable for me back then when I wrote it. When navigation was added it also was implemented with the three step logic, as a consequence center had to be updated in the ResolutionTick. This is in itself somewhat confusing and caused problems for further refactorings.

charlenni:
No. The three step zoom was, because the point, where to zoom, wasn't the center of the map. If you clicked in the left top corner, the map zoomed and the point you clicked stands in the left top corner. So the point to zoom has to be brought to the center, resolution is changed, and the map is brought back.

Now you make it completely different: the point you clicked moves, while resolution changes, to the center of the map.

pauldendulk:
> If I understand it correct, then you want to zoom and set the center of the map to the point, where the user clicked on the map. Both should move together. So why do you not make two AnimationEntrys, one for the zoom and one for the move?

The current code is functionality equivalent to what we had. It is one step is the rewrite I started in the other PR. What I want is a system that performs well. My plans are now to let the timer run, don't add and remove animations, don't lock, and we can do a single update for all changing viewport fields.

pauldendulk:
> No. The three step zoom was, because the point, where to zoom, wasn't the center of the map.

I wrote that logic so I think I know what it was intended for. The current code is functionally equivalent.

PR #947, opened 2020-06-12
Pass on ChangeType in RefreshData calls

pauldendulk:


pauldendulk:
The ChangeType is passed on to the RefreshData call. The specific layers determine what should happen. 
- The TileLayer starts fetching with any change
- The ImageLayer and Layer ignore all Continuous changes. They also have their own mechanism to fetch delay. These mechanisms could be made smarter if they use the ChangeType parameter.

PR #971, opened 2020-08-02
Migrate packages.config to PackageReference

inforithmics:
Migrated all packages.config to PackageReference
Migrated Mapsui.Samples.Forms.Shared to .Net Standard 2.0 Project


inforithmics:
1. The reason i made this pull request is that currently in the project there is a mixture of packages.config und projectreferences
2. The reason I Migrated Mapsui.Samples.Forms.Shared  to .Net Standard 2.0 Project is that
    - Mapsui.samples.Forms.WPF wouldn't compile with the shared project (Very strange errors).
    - Best Practice with Xamarin.Forms is now a .Standard 2.0 project instead of a shared project (All Default templates use now a 
       .NET a Standard Project instead of a Shared project).
3. The Consolidate checkin I made because in Mapsui.Samples.iOS.csproj old versions of Brutile.mbtiles and skiaSharp.views where used.

The migration was done with right click on packages.config migrate to ProjectReferences.

inforithmics:
When this gets in I plan to make a Pull Request for migrating
from SkiaSharp.Svg to Svg.Skia (Because skiasharp.svg is still based on skia 1.60 and development stopped) and Svg.Skia is the successor.

pauldendulk:
Ah, loos good!

The svg migration is also very welcome. The situation with the older skiasharp dependency has to be resolved.

Only for the MapControl a shared project should be used. I saw there is another shared project that could be netstandard (Do not have the project open right now).

I see app.configs with bindingredirects are added. Are they really necessary?

inforithmics:
I'll check if the binding redirects are needed probably needed only for the WPF Project (on .NET Framework).

dedmen:
> I plan to make a Pull Request for migrating
from SkiaSharp.Svg to Svg.Skia SkiaSharp.Svg is still based on skia 1.60 and development stopped) and Svg.Skia is the successor.

Yes please :heart:, I just tripped over that https://github.com/mono/SkiaSharp.Extended/issues/91

inforithmics:
So all binding redirects are removed because before it worked too without them.
So in my opinion this task is done.

As a side note I tried locally tried to migrate to svg.skia and the migration went pretty smooth (All SVG symbols worked)
I have to make a side by side comparison.
But this will be a different pull request (I don't want to do to much in one pull request)

pauldendulk:
> As a side note I tried locally to migrate to svg.skia and the migration went pretty smooth (All SVG symbols worked)
> I have to make a side by side comparison to see if everything looks the same (dimensions).
> But this will be a different pull request (I don't want to do to much in one pull request)

A yes, please keep the PR small like this. 

btw, merging can go fast but sometimes can take much longer. This time because of holidays but there can be all sorts of other reasons. That is just how it is.

pauldendulk:
@inforithmics btw, I you work on svg I would appreciate it if you add a unit test for this. Mapsui does not have a lot of tests at the moment, so I am asking more from you than from myself, but I intend to get the test coverage up. Perhaps you can use the rendering tests described here:
https://mapsui.com/documentation/rendering-tests.html
They are more like regression tests than proper unit tests but they helped me avoiding a lot of bugs. Rendering is hard to unit test, the setup of these tests help with that. Basically it compares images but some tolerance is allowed.

mattleibow:
Hi folks! Thanks for using SkiaSharp.SVG when it first came out. I know it has been a rocky run, especially with me not actually maintaining it.
However, due to SVG being huge and I am not able to implement all the features - as well as a very good replacement library now existing, I am officially deprecating SkiaSharp.SVG.

As @dedmen states https://github.com/wieslawsoltes/Svg.Skia is a much better, more complete implementation.

Thanks for your support all this time, Mapsui is a really great library and to see it using SkiaSharp has always been awesome for me.

Issue #991, opened 2020-10-06
BitmapHelper does not load SVG files


ChaseMarsh:
I've run into the issue several times that the svg assets I receive (often exported from tools) contain text before the "<svg" string that is used by BitmapHelper.LoadBitmap to identify svg files. I'm requesting that this identification be a little smarter. At least disregard text up until it finds "<svg" element. I'm sure we don't want to read the whole file, so maybe a reasonable byte limit and/or skip "<?" and "<!" lines? Or back to a way to identify the type as Svg?

My current use case is that I have over 4000 svgs files (yes, seriously) in a folder that are retrieved dynamically. These files came from another source as a tool export. They contain this at the top (plus starting '<'):

?xml version="1.0" encoding="utf-8"?>
!-- Generator: Adobe Illustrator 16.0.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
<svg

PR #995, opened 2020-10-11
BitmapHelper does not load SVG files Fix

inforithmics:
Fixes https://github.com/Mapsui/Mapsui/issues/991


Mapsui/Utilities/StreamHelper.cs
```diff
@@ -0,0 +1,87 @@
+﻿using System.IO;
+using System.Text;
+
+
+namespace Mapsui.Utilities
+{
+    public static class StreamHelper
```
pauldendulk, line 7: Nowadays (not sure if I have done this in Mapsui so far) I prefer to add such extension methods in an StreamExtensions class in a Extensions folder. I use extensions more and more btw.

pauldendulk (approved):
Great work. Unit tests for the StreamHelper as well

PR #1005, opened 2020-11-01
Cleanup: fix some warnings etc

janusw:
This PR only does some minor cleanup:
* It fixes a couple of code warnings shown by VS (mostly SkiaSharp deprecation warnings).
* It removes two unnecessary calls to `SKMatrix.MakeIdentity`.
* It removes an auto-generated file `Resource.Designer.cs` from the git repo.


pauldendulk:
> It removes an auto-generated file Resource.Designer.cs from the git repo.

At some point (long ago) I tried to remove those but got some problem because of that. But since I can't remember which it is good to remove them, if the problem persists we could learn what they are.

pauldendulk (approved):
Thanks a lot. Cleanup like this is often not of interest to contributors. They are focused on getting their functionality in. Continuous improvements like these are essential to keep the project maintainable.

PR #981, opened 2020-08-17
Svg.skia

inforithmics:
Initial Svg.Skia Implementation
- [x] Updated Skia to 2.80.1 (needed by svg.skia)
- [x] Better image caching. Fixed Svg Symbol Sample
- [X] Add Unit Tests (Image Compare)

PR #1018, opened 2020-11-07
Add logging to svg sample

inforithmics:
Addresses the comment in this pull request (add logging)

https://github.com/Mapsui/Mapsui/pull/981


pauldendulk (approved):
Fabulous. Logging could be improved in other places as well. There should be an easy way to view logs as well. Perhaps there should be an option to log to the map canvas, writing text in skiasharp. Perhaps that kind of logging should be on by default in Debug mode, and it should be possible to turn it on in Release mode.

PR #1029, opened 2020-12-04
Upgrade the project file to the new format

IOL0ol1:
Upgrade the project file to the new format like Mapsui.csproj


IOL0ol1:
Something went wrong after  ffb6011

inforithmics:
If you want to make a pull request that is targeting a lower version than 4.8 I wouldn't go below 4.7.2. 
Because only with this Version all .Net Standard Bugs are fixed. 
Using 4.6.2 introduces a lot of nuget dependencies. Because .Net Standard isn't integrated into the Framework 4.6.2.
Only in 4.7.2 The whole .Net Standard 2.0 is working properly.

IOL0ol1:
> If you want to make a pull request that is targeting a lower version than 4.8 I wouldn't go below 4.7.2.
> Because only with this Version all .Net Standard Bugs are fixed.
> Using 4.6.2 introduces a lot of nuget dependencies. Because .Net Standard isn't integrated into the Framework 4.6.2.
> Only in 4.7.2 The whole .Net Standard 2.0 is working properly.
> 
> This is only my personal opinion.

I cloned a branch to test the nuget package, only to add net462 support to the .csproj file and package it independently.
I packaged each project individually, and their dependencies were handled automatically by nuget.
At the moment, it's working well.
I started a new issue about packaging, and dependency processing. #1034

inforithmics:
My pull request that simplified the projects got accepted, so maybe you can make another pull request with your changes? Especially the partioning into several packages. I think this is a good idea.

PR #1030, opened 2020-12-10
Csharp8

inforithmics:
Set the CSharp Language Version only in Directory.Build.props

This Pull Request Changes the CSharp Language Version to C# 8.0 so that nullable annotations could be made.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,8 @@
+<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!-- Common Project Properties -->
+  <PropertyGroup>
+    <LangVersion>8.0</LangVersion>
+  </PropertyGroup>
+
+</Project>
--- a/Mapsui.Desktop/Mapsui.Desktop.csproj
+++ b/Mapsui.Desktop/Mapsui.Desktop.csproj
@@ -4,7 +4,6 @@
     <TargetFramework>net48</TargetFramework>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
-    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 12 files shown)


pauldendulk:
@inforithmics Thanks for the PR. How does this work? I see you removed the 'latests' node. Is 'latest' not enough for C# 8? Or was the node only added for part of the csproj files?

I just googled Directory.Build.props. https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#directorybuildprops-and-directorybuildtargets Did not know about it. Thanks for teaching me something new. It seems you can just put the file next to your sln and do not need to include it anywhere. Nice.

pauldendulk (approved):
I like the Directory.Build.props. Could be used for other purposes as well. 

I am going to merge this but still interested in the question I asked in the comments (How does this work? I see you removed the 'latests' node. Is 'latest' not enough for C# 8? Or was the node only added for part of the csproj files?)

inforithmics:
Yes, latest would have been sufficient. But with Directory.Build.props You can set it to latest centrally.
You can even have other Directory.Build.props in Lower folders. It is Filesystem hierarchical.

The reason why the language version has to be specified is, because the default langauge version is dependent on the Framework Version

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version
For example .NET Framework 4.8 has a default of 7.3 Language Version.

What you can do is set all the settings of PropertyGroup and ItemGroup

for example

<PropertyGroup>
      <Deterministic>true</Deterministic> 
      <AnalysisLevel>5</AnalysisLevel> 
<ItemGroup>
    <PackageReference Include="Microsoft.Net.Compilers" Version="3.8.0">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
    </PackageReference>
  </ItemGroup>

pauldendulk:
>The reason why the language version has to be specified is, because the default language version is dependent on the Framework Version

Thanks, this answers my question.

inforithmics:
Only to be clear

`<LangVersion>latest</LangVersion>` sets it to C# 9 for all Frameworks independent of the Default Language Level.

pauldendulk:
> <LangVersion>latest</LangVersion> sets it to C# 9 for all Frameworks independent of the Default Language Level.

oh, that was indeed not clear to me. 

Perhaps it's best to choose a specific language version so it does not depend on what a developer has installed. That said, perhaps we could already go for C# 9. I like records. What do you think?

inforithmics:
In my opinion it's a good idea, because in c# 9 there are many nullable fixes too.

for example the generic declaration List<T?> means in c#8 that the Generic Type T needs to be a class and in c#9 it can be a class or a nullable valuetype

PR #1037, opened 2020-12-20
Add netcoreapp3.1 to Wpf

Liklainy:
* Added `netcoreapp3.1` target to Wpf projects
* Changed wpf projects to `Sdk.WindowsDesktop` with `UseWPF`
* Updated `Svg2Xaml` to `0.3.0.7` which supports `netcoreapp3.1` target
* Added separate nuspec for `Mapsui.Wpf`

Wpf package with `netcoreapp3.1` needs `SkiaSharp.Views.WPF` which I was not able to add to `Mapsui` package, because of this I added separate nuspec for `Mapsui.Wpf` as was done with `Mapsui.Forms`
But I didn't remove wpf stuff from `Mapsui` package, so if someone has wpf app which targets `net48` he can still simply use `Mapsui` package.

After updating `Svg2Xaml` package there are two tests failing
* `RenderBitmapAtlas` - which has a tiger svg image
  It failes because in new version `ViewBox` support was enabled, and this svg has this `viewBox` attribute which was previously ignored. As result the tiger in generated images is slightly moved.

0.3.0.5             |  0.3.0.7           |  diff
:---------------:|:---------------:|:------------------:
bitmap_atlas_generated_5 | bitmap_atlas_generated_7 | bitmap_atlas_diff

* `RenderPointWithSvgSymbols` - for this test I'm still not sure why the behaviour was changed

0.3.0.5             |  0.3.0.7           |  diff
:---------------:|:---------------:|:------------------:
points_with_svgsymbolstyle_generated_5 | points_with_svgsymbolstyle_generated_7 | points_with_svgsymbolstyle_diff


pauldendulk:
Sorry, I postponed this MR because it was hesitant. And now I caused all kinds of merge conflicts.

I saw some #if's and Conditions and it made me wonder if this was a good idea.  I have worked with conditional compiling in the past and found it very hard to maintain. Now scrolling to the change it did not seem that bad, just 3 ifs.

If you could update for the current master I would be much obliged. 😌

pauldendulk:
There are some things I need to think about though. Like the Wpf nuspec.

pauldendulk:
Thanks for the analysis of the test fails! This happens from time to time. These renderers just of some small changes in them on many updates. What I do is to check the difference, exactly like you did, and if that looks acceptable I copy the newly generated image to the reference image to fix the test.

Liklainy:
Thank you for attention to this PR! 😄
I rebased changes to actual master and replaced images for two tests.
I would like to remove `#if`'s introduced by me, to do that can I remove caching though `HttpContext`?

pauldendulk:
@Liklainy I removed the HttpContext in Shapefile. It made only sense when running it on a server. Please update your branch for this change.

pauldendulk:
> I rebased changes to actual master and replaced images for two tests.

There are still many conflicts, did you push?

pauldendulk:
In general I would prefer to stick to a single nuspec. It is possible to have different versions of a dll in their own folder. See image below of the Mapsui nuget.

Also I think the desktop nuget should be ignored. We are in the process of moving the providers in the Desktop project to the Mapsui project. I do not intend to publish a next version of that project. 

image

Liklainy:
Decided to split this PR
First will be updating `Svg2Xaml` lib #1099

PR #1046, opened 2021-01-12
Callout.IsClosableByClick: make setter public

### LINKED ISSUE #1045: Unable to avoid callouts from being closed on click
I'm observing since a while that the callouts of my pins are automatically being closed when they are clicked, and I'd like to change that (I have another event that is triggered on callout click) ...

I recently found `Mapsui.UI.Forms.Callout.IsClosableByClick`, which seems to be `true` by default. Unfortunately I cannot just set it to `false`, since it has a private setter :/

So the question is: Why is this read-only? Is there another way to avoid callouts from being closed on click? What am I missing?

janusw:
This fixes #1045. The setter of `Callout.IsClosableByClick` was private, so one could never change this property (which is true by default) and as a result the callout was always closed by clicking on it.

This PR makes it possible to have callouts that stay visible after clicking (and demonstrates this in the PinSample test).

### DIFF
--- a/Mapsui.UI.Forms/Objects/Callout.cs
+++ b/Mapsui.UI.Forms/Objects/Callout.cs
@@ -370,7 +370,7 @@ namespace Mapsui.UI.Forms
         public bool IsClosableByClick
         {
             get { return (bool)GetValue(IsClosableByClickProperty); }
-            private set { SetValue(IsClosableByClickProperty, value); }
+            set { SetValue(IsClosableByClickProperty, value); }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


charlenni:
Could be avoided by handling CalloutClicked event and setting handled to true. See Callout.cs at line 553.

janusw:
> Could be avoided by handling CalloutClicked event and setting handled to true. See Callout.cs at line 553.

Indeed that seem to work (and thus fixes #1045).

Nevertheless, what's the purpose of the property `Callout.IsClosableByClick`, if it can't be set? IMHO it still makes sense to make the setter public. (Otherwise the property should be removed, since it cannot be used and is just unnecessary code bloat.)

pauldendulk (approved):
I am not very knowledgeable on the forms parts. You argument sounds reasonable, so I will just merge it. 

btw. Removing might also be a good idea. For projects to succeed in the long run deleting is essential.

Issue #1043, opened 2021-01-06
Mapsui Xamarin.Forms 5 compatability


BurkusCat:
```NU1608: Detected package version outside of dependency constraint: Mapsui.Forms 2.0.3 requires Xamarin.Forms (>= 4.3.0 && < 5.0.0) but version Xamarin.Forms 5.0.0.1874 was resolved.```

The NuGet package (https://www.nuget.org/packages/Mapsui.Forms/) lists Xamarin.Forms (>= 4.3.0 && < 5.0.0) as a dependency. A new version of Xamarin.Forms has just been released 5.0.0.1874 so this shows a warning when installing packages.

With Xamarin.Forms 5 installed, Mapsui appears to work fine for me.

sailman342:
Yes, same for me.  No problem on UWP and Android, but no way to build on Forms WPF :-(

Furthermore, I rebuild Mapsui from sources with Xamarin Forms 5.and I had 4 lines to change in Mapsui to get it building and working fine ! This is beacause new Xamarin has a Brush class and we have to mention we use the MApsui one :-).

I believe the dependency can be safely relaxed !

:-)

Sebastian1989101:
@pauldendulk +1 for this. Can't see any issues so far with XF 5.0 and Mapsui. Because this is a service release before the MAUI integration into XF, it should be safe to increase the package requirement.

pauldendulk:
This has been implemented with https://github.com/Mapsui/Mapsui/pull/1050

PR #1050, opened 2021-01-23
Update Xamarin.Forms to version 5.0

janusw:
This should fix issue #1043. I have tested it Android & iOS. Both seem to be working fine.


pauldendulk (approved):
Guess that's okay.

PR #1062, opened 2021-02-08
update BruTile to version 3.1.2

janusw:
This PR updates the three BruTile packages ...

* BruTile
* BruTile.Desktop
* BruTile.MbTiles

... from version 3.0.0 to 3.1.2.

The update is done via VS for Mac. The changes include a couple of linebreak changes (automatically done by VS) and some line removals in the ProjectReferences of `Mapsui.Rendering.Skia.Tests`. I don't really understand the latter, but since this is also an automatic change by VS, I assume it is harmless.

To be on the safe side, I tried to run those unit test, and all 11 fail on the branch, but in fact the also fail on master already (so it's definitely not due to my changes).

### DIFF
--- a/Mapsui.Desktop/Mapsui.Desktop.csproj
+++ b/Mapsui.Desktop/Mapsui.Desktop.csproj
@@ -11,7 +11,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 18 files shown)


pauldendulk (approved):
Thanks!

PR #1067, opened 2021-02-12
MapView: adjust button layout on iOS

### LINKED ISSUE #1059: MapView: buttons can overlap with status bar on some iPhone models
When using Mapsui.UI.Forms.MapView to display a map on iOS, it can happen that the zoom buttons overlap with the status bar on certain iPhone models.

This happens in particular when the MapView is basically 'full-screen' and without any other UI elements above. Since in iOS the status bar (with the clock and icons for battery, WiFi, etc) is transparent by default, the map is shown beneath it. The map buttons (for zooming etc) are rather close to the top of the map and they can even overlap with the status bar icons for certain screen sizes (e.g. with the iPhone 12 or iPhone XR).

I can imagine two different ways to fix this:

1. Adjust the default placement of the buttons (moving them slightly downward). Possibly only apply this adjustment on iOS, since the problem does not occur on Android.
2. Provide some way to adjust the button placement via the Mapsui API, so that an app can adjust it to its needs.

The first one is certainly simpler to implement, and perfectly sufficient for my purpose.

janusw:
* fixes #1059 for me
* move buttons a bit downward, so that they do not
  overlap with the status bar on some iPhone models

### DIFF
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -145,7 +145,12 @@ namespace Mapsui.UI.Forms
             _mapButtons.Children.Add(_mapSpacingButton2);
             _mapButtons.Children.Add(_mapNorthingButton);
 
-            AbsoluteLayout.SetLayoutBounds(_mapButtons, new Rectangle(0.95, 0.03, 40, 176));
+            if (Device.RuntimePlatform == Device.iOS)
+                // the buttons are placed a bit lower on iOS to avoid overlap with the transparent status bar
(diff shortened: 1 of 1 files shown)


janusw:
Forgot to mention that this PR only modifies the layout on iOS and leaves other platforms untouched (it's pretty obvious from the code, though).

@charlenni Do you have an opinion on this? Is this ok with you or would you prefer a different solution?

charlenni:
For me, it is ok. It could be made better, if you expose this values as properties, but I assume, one could move the buttons around by the described child method or disable them and make new ones.

pauldendulk (approved):
Not a great solution, but no better alternative is known and it solves a real existing practical problem.

Issue #1072, opened 2021-02-13
WFS support


janusw:
It would be nice to have full WFS support in Mapsui. There is already a WfsProvider and a WfsSample, but both probably need to be checked and updated. See the discussion started in #1034.

janusw:
FYI, a couple of openly accessible WFS service are available from the German BKG, as listed [here](https://gdz.bkg.bund.de/index.php/default/webdienste.html), e.g.:

- https://sgx.geodatenzentrum.de/wfs_clc5_2018?request=GetCapabilities&service=wfs
- https://sgx.geodatenzentrum.de/wfs_dlm250?request=GetCapabilities&service=wfs
- https://sgx.geodatenzentrum.de/wfs_gnde?request=GetCapabilities&service=wfs
- https://sgx.geodatenzentrum.de/wfs_kfz250?request=GetCapabilities&service=wfs
- https://sgx.geodatenzentrum.de/wfs_vg1000?request=GetCapabilities&service=wfs
- etc

Those can be useful for testing, I hope. Maybe others have more example of open WFS services.

PR #1074, opened 2021-02-13
Multiple WFS improvements

JBou:
This PR tries to make the WFSProvider work again. Some changes are based on [SharpMap](https://github.com/SharpMap/SharpMap)
There is a working demo of my changes in the separate branch: [JBou/suedtirol_map](https://github.com/JBou/Mapsui/tree/suedtirol_map)

#### The following changes have been made:

- [x] Fixed WFS 1.0.0 request when using a filter with POST request
There was a missing Opening tag
- [x] Fixed get request wrongly adding PostData in WFSProvider
- [x] Support multiple labels for WFS
Added FeatureTypeInfo.Elements to store all the feature's elements
based on https://github.com/SharpMap/SharpMap/commit/5fd5d7e7aaa641bc291ebe7b5a56d62970c0fdb4
- [x] Allow BoundingBox being null on WFS request
This allows to search for features on the whole map.
Using Double.MaxValue as Boundingbox returned a Stackoverflowexception from the server.
- [x] Made the WFS request more readable and properly url-encode the parameters
- [x] Fixed "srsName" parameter to properly support data reprojection for WFS
See [this comment](https://github.com/Mapsui/Mapsui/pull/1074#issuecomment-778791052)
- [x] Fix Axis Order for WFS request > 1.0.0 [see SharpMap commit](https://github.com/SharpMap/SharpMap/commit/1b6eac0f61506d3bb782ded76ac74666f3ee9e0b#diff-9a06870322de76edb927937e8d239ff978d6c97624891644f30558c8dc9881a7)
(see [this comment](https://github.com/Mapsui/Mapsui/pull/1074#issuecomment-778623545) for detailed informations).
more information on the axis order can be found  in the [geoserver docs](https://docs.geoserver.org/stable/en/user/services/wfs/axis_order.html#wfs-basics-axis).
Before this PR, WFS > 1.0.0 always returned inverted points. Maps with normal axis order work now as expected in 1.1.0 too.
- [x] We could create a new Method in `WFS_ 1_1_0_TextResouces` and `WFS_ 1_1_0_TextResouces` named e.g. `createFilter`, where we create the <Filter> and use it in the GET (as query parameter) and POST request (as body). This way we have the exact same filter string and avoid future bugs if we fix something at one place and forget about the other place.
- [x] change code to use commas instead of multiple propertyName parameters? see [this comment](https://github.com/Mapsui/Mapsui/pull/1074#issuecomment-778756652).
- [x] Merge [these changes](https://github.com/SharpMap/SharpMap/commit/72d4dba61c1b3e767757c5bc1288a0c90e2220d2)
- [x] Merge [these changes](https://github.com/SharpMap/SharpMap/commit/9f23dd01db62436ade82d7bb4f065fb30a2c3e05#diff-9a06870322de76edb927937e8d239ff978d6c97624891644f30558c8dc9881a7)
- [x] Add code documentation from [here](https://github.com/SharpMap/SharpMap/commit/7975a94d5192c3b28eac5caed2ff08fe28990bdf#diff-9a06870322de76edb927937e8d239ff978d6c97624891644f30558c8dc9881a7)

Other:
- [ ] Use BruTiles CrsAxisOrderRegistry instead of our own?
- [x] Add Samples for WFS.

#### Postponed / Ideas
- [ ] Test a map with inverted axis order ([example wfs service](http://services.sandre.eaufrance.fr/geo/zon_FXX?REQUEST=GetCapabilities&service=WFS&VERSION=1.0.0) - `EPSG:4326`) with both WFS versions. It should invert the points of the response properly, but I think it doesn't currently invert the points when sending an `ExecuteIntersectionQuery`.
- [ ] Transform coordinates from other CRS in ExecuteIntersectionQuery and transform them back when receiving the response to support different CRS's? It's an idea, I don't know how you want to handle this?
- [ ] add a sample using a Transformation to test Transforming a WFS Layer (`ESPG:25832`) on a Map with different CRS (`EPSG:3857`)
- [ ] Implement WFS version 2.0.0
---

This is part of #1072


JBou:
I added the possibility to use multiple labels, because i need to get all the metadata for a specific feature and haven't found another solution. You can then access them using `feature["key"]`. To get all field keys use `wfsProvider.FeatureTypeInfo.Elements`.

JBou:
Currently I had a look at the Axis Order for the CRS 25832. A lot of information on this topic can be found [here](https://docs.geoserver.org/stable/en/user/services/wfs/axis_order.html).

As of now, if we use WFS 1.0.0 ([example link](http://geoservices.buergernetz.bz.it/geoserver/wfs?SERVICE=WFS&Version=1.0.0&REQUEST=GetFeature&TYPENAME=p_bz-cadastre_public%3aParticelle_poligoni_validate_pubb&SRS%20=25832&FILTER=%3CFilter+xmlns%3d%22http%3a%2f%2fwww.opengis.net%2fogc%22+xmlns%3agml%3d%22http%3a%2f%2fwww.opengis.net%2fgml%22+xmlns%3ap_bz-cadastre_public%3d%22http%3a%2f%2fgeoserver.d41.prov.bz%2fp_bz-cadastre_public%22%3E%3CBBOX%3E%3CPropertyName%3Ep_bz-cadastre_public%3aPPOL_SHAPE%3C%2fPropertyName%3E%3Cgml%3aBox+srsName%3d%2225832%22%3E%3Cgml%3acoordinates%3E745088.38142615778%2c5177750.4882910615+745088.38142615778%2c5177750.4882910615%3C%2fgml%3acoordinates%3E%3C%2fgml%3aBox%3E%3C%2fBBOX%3E%3C%2fFilter%3E)) geoserver returns a response in gml2, which contains a `<gml:coordinates>` xml tag, and uses the first number as x and the second as y. In every other case, for example when using WFS 1.1.0 ([example link](http://geoservices.buergernetz.bz.it/geoserver/wfs?SERVICE=WFS&Version=1.1.0&REQUEST=GetFeature&TYPENAME=p_bz-cadastre_public%3aParticelle_poligoni_validate_pubb&SRS%20=25832&FILTER=%3CFilter+xmlns%3d%22http%3a%2f%2fwww.opengis.net%2fogc%22+xmlns%3agml%3d%22http%3a%2f%2fwww.opengis.net%2fgml%22+xmlns%3ap_bz-cadastre_public%3d%22http%3a%2f%2fgeoserver.d41.prov.bz%2fp_bz-cadastre_public%22%3E%3CBBOX%3E%3CPropertyName%3Ep_bz-cadastre_public%3aPPOL_SHAPE%3C%2fPropertyName%3E%3Cgml%3aBox+srsName%3d%2225832%22%3E%3Cgml%3acoordinates%3E745088.38142615778%2c5177750.4882910615+745088.38142615778%2c5177750.4882910615%3C%2fgml%3acoordinates%3E%3C%2fgml%3aBox%3E%3C%2fBBOX%3E%3C%2fFilter%3E)) the geoserver returns gml3 with a `<gml:posList>` and uses the first number as y and the second as x. this logic can be seen here:https://github.com/Mapsui/Mapsui/blob/af2bf64d3f45c0a3a7b91d2d58cc2a4fff3d13d3/Mapsui.Desktop/Wfs/Utilities/GeometryFactories.cs#L113


I tried to implement the use of AxisOrderRegistry like in SharpMap (based on [this commit](https://github.com/SharpMap/SharpMap/commit/1b6eac0f61506d3bb782ded76ac74666f3ee9e0b#diff-9a06870322de76edb927937e8d239ff978d6c97624891644f30558c8dc9881a7)), but it's not working correctly. You can see these changes [here](https://github.com/JBou/Mapsui/commit/2882e434d14a2dc6fb99a639d264ece0b4a8c575)(it has been [taken from BruTile](https://github.com/BruTile/BruTile/blob/e741767d78774d09dd974935160fb9339b807b7c/BruTile/Wmts/CrsAxisOrderRegistry.cs), maybe we shouldn't copy the class and fix this right in BruTile itself and reuse the same class). The tests to create the BitField are located [here](https://github.com/BruTile/BruTile/blob/e741767d78774d09dd974935160fb9339b807b7c/Tests/BruTile.Desktop.Tests/Crs/CrsAxisOrderRegistryTest.cs).

The AxisOrderRegistry returns inversed order for CRS 25832, but in my opinion thats not correct (as you can see in the example links above, the order remains the same). I started a geoserver instance, when to the page listed on [this documentation](https://docs.geoserver.org/stable/en/user/services/wfs/axis_order.html#srslist-axis-order) and searched for 25832. 

<details>
  <summary>The internal definition is this</summary>

```
PROJCS["ETRS89 / UTM zone 32N", 
  GEOGCS["ETRS89", 
    DATUM["European Terrestrial Reference System 1989", 
      SPHEROID["GRS 1980", 6378137.0, 298.257222101, AUTHORITY["EPSG","7019"]], 
      TOWGS84[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0], 
      AUTHORITY["EPSG","6258"]], 
    PRIMEM["Greenwich", 0.0, AUTHORITY["EPSG","8901"]], 
    UNIT["degree", 0.017453292519943295], 
    AXIS["Geodetic longitude", EAST], 
    AXIS["Geodetic latitude", NORTH], 
    AUTHORITY["EPSG","4258"]], 
  PROJECTION["Transverse_Mercator", AUTHORITY["EPSG","9807"]], 
  PARAMETER["central_meridian", 9.0], 
  PARAMETER["latitude_of_origin", 0.0], 
  PARAMETER["scale_factor", 0.9996], 
  PARAMETER["false_easting", 500000.0], 
  PARAMETER["false_northing", 0.0], 
  UNIT["m", 1.0], 
  AXIS["Easting", EAST], 
  AXIS["Northing", NORTH], 
  AUTHORITY["EPSG","25832"]]
```
</details>
<details>
  <summary>and the EPSG definition this</summary>

```
PROJCS["ETRS89 / UTM zone 32N", 
  GEOGCS["ETRS89", 
    DATUM["European Terrestrial Reference System 1989", 
      SPHEROID["GRS 1980", 6378137.0, 298.257222101, AUTHORITY["EPSG","7019"]], 
      TOWGS84[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0], 
      AUTHORITY["EPSG","6258"]], 
    PRIMEM["Greenwich", 0.0, AUTHORITY["EPSG","8901"]], 
    UNIT["degree", 0.017453292519943295], 
    AXIS["Geodetic latitude", NORTH], 
    AXIS["Geodetic longitude", EAST], 
    AUTHORITY["EPSG","4258"]], 
  PROJECTION["Transverse_Mercator", AUTHORITY["EPSG","9807"]], 
  PARAMETER["central_meridian", 9.0], 
  PARAMETER["latitude_of_origin", 0.0], 
  PARAMETER["scale_factor", 0.9996], 
  PARAMETER["false_easting", 500000.0], 
  PARAMETER["false_northing", 0.0], 
  UNIT["m", 1.0], 
  AXIS["Easting", EAST], 
  AXIS["Northing", NORTH], 
  AUTHORITY["EPSG","25832"]]
```
</details>

We see that the Axis of the BASE CRS are inverted, but the axis of the CRS 25832 itself aren't. In my opinion, this is the source of the problem. We shouldn't check


Last but not least, when creating the ExecuteIntersectionQuery, we have to invert the axis ordere there too, if needed.


If you have a look on another CRS, for example 4326, i have an example request for WFS 1.0.0 [here](http://services.sandre.eaufrance.fr/geo/zon_FXX?REQUEST=GetFeature&service=WFS&VERSION=1.0.0&typename=sa:SegClassContinuiteEco_Liste2&count=1&) and an example request for WFS 1.1.0 [here](http://services.sandre.eaufrance.fr/geo/zon_FXX?REQUEST=GetFeature&service=WFS&VERSION=1.1.0&typename=sa:SegClassContinuiteEco_Liste2&count=1&). We see, that the axis are inverted as expected. You can compare the geoserver definitions below. This CRS doesn't have a BASE CRS:

<details>
  <summary>The internal definition is this</summary>

```
GEOGCS["WGS 84", 
  DATUM["World Geodetic System 1984", 
    SPHEROID["WGS 84", 6378137.0, 298.257223563, AUTHORITY["EPSG","7030"]], 
    AUTHORITY["EPSG","6326"]], 
  PRIMEM["Greenwich", 0.0, AUTHORITY["EPSG","8901"]], 
  UNIT["degree", 0.017453292519943295], 
  AXIS["Geodetic longitude", EAST], 
  AXIS["Geodetic latitude", NORTH], 
  AUTHORITY["EPSG","4326"]]
```

</details>
<details>
  <summary>and the EPSG definition this</summary>

```
GEOGCS["WGS 84", 
  DATUM["World Geodetic System 1984", 
    SPHEROID["WGS 84", 6378137.0, 298.257223563, AUTHORITY["EPSG","7030"]], 
    AUTHORITY["EPSG","6326"]], 
  PRIMEM["Greenwich", 0.0, AUTHORITY["EPSG","8901"]], 
  UNIT["degree", 0.017453292519943295], 
  AXIS["Geodetic latitude", NORTH], 
  AXIS["Geodetic longitude", EAST], 
  AUTHORITY["EPSG","4326"]]
```

</details>

The last thing to think about, is how to choose when we need to inverse the axis order. First we have to fix the AxisOrderRegistry to support CRS like 25832 (as written above). Then we need to decide if the difference is between WFS 1.0.0 and WFS 1.1.0 or between the different representation as listed in [this table](https://docs.geoserver.org/stable/en/user/services/wfs/axis_order.html#axis-ordering).

The current implementation is wrong for sure. It inverses the axis order based on `coordinates` (gml2) or `poslist` (gml3), as written above.
It can't be caused by the gml version returned, because we can append `&outputformat=gml2` and `&outputformat=gml3` to the request, and see that it doesn't make a difference:
Examples: 
normal order: [WFS 1.0.0 & gml2](http://services.sandre.eaufrance.fr/geo/zon_FXX?REQUEST=GetFeature&service=WFS&VERSION=1.0.0&typename=sa:SegClassContinuiteEco_Liste2&count=1&outputformat=gml2) - [WFS 1.0.0 & gml3](http://services.sandre.eaufrance.fr/geo/zon_FXX?REQUEST=GetFeature&service=WFS&VERSION=1.0.0&typename=sa:SegClassContinuiteEco_Liste2&count=1&outputformat=gml3)
inversed order: [WFS 1.1.0 & gml2](http://services.sandre.eaufrance.fr/geo/zon_FXX?REQUEST=GetFeature&service=WFS&VERSION=1.1.0&typename=sa:SegClassContinuiteEco_Liste2&count=1&outputformat=gml2) - [WFS 1.1.0 & gml3](http://services.sandre.eaufrance.fr/geo/zon_FXX?REQUEST=GetFeature&service=WFS&VERSION=1.1.0&typename=sa:SegClassContinuiteEco_Liste2&count=1&outputformat=gml3)

In every request the `srsName="EPSG:4326"` is the same, so i think that the WFS version decides if we need to invert the axis or not. My idea: Always use normal order (x,y) for WFS 1.0.0, and pass the CRS to AxisOrderRegistry to find out the axis order when using 1.1.0 or above.

Some information on the different WFS version can be found [here](https://docs.geoserver.org/stable/en/user/services/wfs/basics.html).

I hope you can understand it a bit, it's a lot of information...

JBou:
Before this PR, WFS > 1.0.0 always returned inverted points. Maps with normal axis order work now as expected in 1.1.0 too. Just pushed 2 commits to fix this.

TODO: Test a map with inverted axis order. It should invert the points of the response properly, but I think it doesn't currently invert the points when sending an `ExecuteIntersectionQuery`.

---
What do you think about this changes as of now?

janusw:
Thanks for all the work so far. I haven't looked at your modifications of the WfsProvider yet, but I tried to run your Südtirol sample. Unfortunately that did not work, because I'm working on a Mac and cannot run WPF-specific code.

I saw that you started the sample in Mapsui.Samples.Common, but then moved it to Mapsui.Samples.Wpf. Could you possibly set it up as a cross-platform sample, like the other Mapsui samples in Mapsui.Samples.Common? Cross-platform support is really one of the big strengths of Mapsui, and it would be quite useful to have a working WFS sample along with all your improvements.

pauldendulk:
Guys! ❤️❤️❤️ This is awesome!

pauldendulk:
Some reactions before a close my laptop for today.

- This is awesome. You fixed some serious issues.

- It's important to have a working sample as part of this. The SüdtirolMap sample looks great. Is it possible to add this to the main Mapsui repo? Or is this not a public WFS? Could one of the WFSes mentioned  by @janusw be added? Perhaps WFS should have it's own sample category.

- Axis order is known to be complicated. This does not need to be fully solved right now.

- "Transform coordinates from other CRS in ExecuteIntersectionQuery". So, a WFS supports a number of CRSes, and the client could request just one of those. But what you are talking about is the situation where you need to request a CRS that is not supported?

JBou:
> I saw that you started the sample in Mapsui.Samples.Common, but then moved it to Mapsui.Samples.Wpf. Could you possibly set it up as a cross-platform sample, like the other Mapsui samples in Mapsui.Samples.Common? Cross-platform support is really one of the big strengths of Mapsui, and it would be quite useful to have a working WFS sample along with all your improvements.

The sample has all the functionality i need for my custom project. I had to move it to the WPF namespace, because i needed a way to get the `MouseLeftButtonUp` (Map clicked) and `MouseMove` event. Could you say me how to implement this with `IMapControl`?

In addition, the `WmsProvider` used in the sample is part of `Mapsui.Desktop`. Can i reference this project in `Mapsui.Samples.Common` or can the `WmsProvider` be moved to the `Mapsui` namespace?

At the end I show a `MessageBox` with all the Labels in it. Is there another possibility in .Net Standard 2.0?

JBou:
> It's important to have a working sample as part of this. The SüdtirolMap sample looks great. Is it possible to add this to the main Mapsui repo? Or is this not a public WFS? Could one of the WFSes mentioned  by @janusw be added? Perhaps WFS should have it's own sample category.

The WFS is a public service by the government of South Tyrol, so we should be able to use it. Once i resolve the issues in the previous comment, I can clean it up and document it better and add it to this PR. I can already move it to a separate category. Afterwards we can find possibilities to use the WFSes from @janusw.

> Axis order is known to be complicated. This does not need to be fully solved right now.

It's already a lot better than before. CRSes with normal axis order are working now. We could use [`BruTile.Wmts.CrsAxisOrderRegistry`](https://github.com/BruTile/BruTile/blob/master/BruTile/Wmts/CrsAxisOrderRegistry.cs#L10) instead of or own copy, because we depend on it. Makes maintaining changes easier, it's almost the same logic. What do you prefer?

> "Transform coordinates from other CRS in ExecuteIntersectionQuery". So, a WFS supports a number of CRSes, and the client could request just one of those. But what you are talking about is the situation where you need to request a CRS that is not supported?

Yes, exactly. Add transformations for CRSes not supported by the service. We could use a `EPSG:25832` -only WFS service in a `EPSG:4326` map.

The WFS service used in the sample only supports only `EPSG:25832` for example. We would have to transform the coordinates locally, somethings [like this](https://github.com/SharpMap/SharpMap/blob/develop/SharpMap/Data/Providers/WFSClient.cs#L1104-L1118). I don't know if the SharpMap implementation is working correctly, though. It's a bit complicated and needs more tests, but it is an idea for the future.

The samples provided by @janusw have multiple Supported CRSes, they should be parsed and listed in the `featureTypeInfo`. In this cases, we have to append `&srsName=CRS` to the WFS request, in the other cases we need to transform it in Mapsui itself.

JBou:
I just saw that there is already added a parameter called `SRS` on GET Requests, and a parameter `srsName` on POST requests. I think the `SRS` parameter used in the GET request is wrong, it should also be `srsName`.

More information on this topic can be found at the end of [this chapter](https://docs.geoserver.org/latest/en/user/services/wfs/reference.html#getfeature). As far as I understand it, the `srsName` specifies the CRS of the **returned** data, and to request a bbox with a different CRS, we need to add it at the end of the bbox parameter: `bbox=a1,b1,a2,b2,CRS`

---

In the same chapter it also says:
> For more than one property from a single feature, use a comma-separated list of values for propertyName

Maybe I should change my code to use commas instead of multiple `propertyName` parameters here:
https://github.com/Mapsui/Mapsui/pull/1074/commits/b2890c8cefe3151d6c13c975755e530dbf8ce7b4#diff-e843360fa5dd27a9966ee9f0a6f9177730c422fce3ab1f970071007e378be820R127-R131

janusw:
> The sample has all the functionality i need for my custom project.

That's great, but maybe it is a bit overblown for the Mapsui samples. Note that Mapsui does not have *any* WFS sample so far, so one could really start with something simple ...

I'd suggest to leave out all handling of mouse clicks and message boxes. Just have static map (without any WMS, maybe just based on OSM) and  put some WFS layer on top of that.


> In addition, the `WmsProvider` used in the sample is part of `Mapsui.Desktop`. Can i reference this project in `Mapsui.Samples.Common` or can the `WmsProvider` be moved to the `Mapsui` namespace?

WMS is really a separate topic that should be tackled after this PR is merged.

And yes, I hope we can also move WmsProvider to the Mapsui core project, but I'm not sure if there are any larger obstacles there. Let's not do too much at once here ...

JBou:
The latest commit makes use of the [geoserver reprojection feature](https://docs.geoserver.org/latest/en/user/services/wfs/vendor.html#reprojection). Now you can specify a different CRS than the `DefaultSRS` and the server will transform the coordinates of the requested boundingbox and response geometries.
I'm now able to use my `EPSG:25832` WFS service with a `EPSG:3857` map. Just set the `CRS` of the WfsProvider to `EPSG:3857`, the WFS service will do the rest.

 As stated in the article linked before, with geoserver it even works on version 1.0.0. It even works if the `GetCapabilities` request doesn't list the CRS, or the CRS is not listed as `OtherCRS`.
I tested the reprojection feature on both WFS versions and GET and POST requests. I can confirm it works for all.

I think this will solve most of the transformation issues we talked about earlier.

@janusw Okay, I will try to create some basic WFS samples in the coming week.

pauldendulk:
> In addition, the WmsProvider used in the sample is part of Mapsui.Desktop. Can i reference this project in Mapsui.Samples.Common or can the WmsProvider be moved to the Mapsui namespace?

The WmsProvider should move to Mapsui proper but let's do that in a later PR. A simple sample without a background layer and mouse manipulation is enough as a first step. Let's improve it later on.

pauldendulk:
> At the end I show a MessageBox with all the Labels in it. Is there another possibility in .Net Standard 2.0?

There is no way to show the MessageBox in .NET Standard itself but most of the other code can be in .NET Standard. Tools to get info need a small platform dependent part (Although we could add Widget for this as well). There is something like this in the 'Map Info' sample. In that case the request is for the features that are in memory within Mapsui. There should be an alternative to get feature info from the data source on the server. In the past something like this was used for WMS. Not sure what the current state is. It probably will not work in its current form but let's try to get this working in a generic platform independent way.

pauldendulk:
> We could use BruTile.Wmts.CrsAxisOrderRegistry instead of or own copy

That is better but perhaps there is a way to avoid it altogether. The whole axis order thing is very unfortunate. The way the OGC designed the spec forces us to have knowledge of specific projections. In later version things were improved. I vaguely remember there is a way to specify the axis order in the request.

pauldendulk:
> The WFS service used in the sample only supports only EPSG:25832 for example. We would have to transform the coordinates locally, somethings like this. I don't know if the SharpMap implementation is working correctly, though. It's a bit complicated and needs more tests, but it is an idea for the future.

This is means to do local coordinate projections. This is a very big topic within geospatial. Mapsui has a simple interface for projection (ITransformation) but no implementation for projections other than for SphericalMercator. If we support that we will use https://github.com/NetTopologySuite/ProjNet4GeoAPI. Also because we intend to use NTS itself as well at some point.

For now lets postpone this. Projection by the service is a good alternative for most scenarios.

pauldendulk:
## The problem of the source extent

In Mapsui it is possible to specify the extent of the Map but in many situations the better solution is to let the data source provide the extent, and let the Map zoom to that extent on startup.

In the current implementation of WFS there is a problem in this logic, because the initial request for features as no extent yet, therefor no feature are returned, therefore the extent of the source stays empty. 

How it should work is that the capabilities specifies the extent, and it actually is available. Unfortunately this extent is only specified in GPS coordinates (WGS84 or EPGS:4326). So here again the spec forces us to include a projection library. Perhaps geoserver has a solution for this as well.

pauldendulk:
@JBou I forgot what the status was on this MR. I know I would like to have a working sample. I think you had one in your own repository, but what was the problem with that?

JBou:
Sorry for the long break, i have not much time left atm. 
I rearranged the TODO list, there are only some small changes to do to finish this. I moved the more complicated stuff down to the "Postponed" list, we should try to get the basic WFS implementation to work.

As for the sample, I had a look some weeks ago. My sample uses a basemap and wasn't useful for a generic WFS sample. We should create one to showcase the basic usage of WFS layers. But the sample i tried to create was lagging because there were a lot of geometries generated. I would have to try out a wfs layer with not that many geometries. I try to find some time the next week to fix the remaining tasks

pauldendulk:
If there is a sample that just show some WFS data in the map that would be enough.

JBou:
I just added some more commits and a simple WFS sample. 
In the sample I had to add ViewportLimiterKeepWithin and a BaseTileMap so that the map centers on the extend. I didn't know about any other WFS service and which other features to show, so I just used a known one to me. You have to click once on the map currently (caused by #1075) and wait for 5-10 seconds (the WFS requests are not that fast...).
I hope this helps, otherwise @pauldendulk could you create a sample yourself, you have more knowledge on MapsUi and how to improve performance?

pauldendulk:
> I just added some more commits and a simple WFS sample.
> In the sample I had to add ViewportLimiterKeepWithin and a BaseTileMap so that the map centers on the extend. I didn't know about any other WFS service and which other features to show, so I just used a known one to me. You have to click once on the map currently (caused by #1075) and wait for 5-10 seconds (the WFS requests are not that fast...).
> I hope this helps, otherwise @pauldendulk could you create a sample yourself, you have more knowledge on MapsUi and how to improve performance?

Excellent, I can go with this.

JBou:
Can this be merged/released so we have a working version with WFS support before switching to NTS?

For me, this PR is now okay. Could you review it? :)

janusw:
> Can this be merged/released so we have a working version with WFS support before switching to NTS?

That would be awesome. Thanks again for your work here! 👍 

> For me, this PR is now okay. Could you review it? :)


Some comments from my side:
* I was not able to run `Mapsui.Samples.Forms.Android` from your branch ( `
Android.Content.Res.Resources+NotFoundException: Resource ID #0x7f0b0081` ). Does anyone else see this or is it some problem on my side?
* `Mapsui.Samples.Forms.Mac` works well, and I am able to run your sample. It's pretty heavy, though (performance on my MacBook is not great). Maybe we'll need some performance improvements, but that can be done later.
* For now, its great that it works at all. I see an aerial map of Südtirol, which is covered by solid red segments. I hope that's the expected result
* There are a few errors in the Debugger console. Is this something one should worry about?
```
Mapsui.Samples.Forms.Mac.exe Error: 0 : An exception occured during a HTTP request to https://geoservices.buergernetz.bz.it/geoserver/wfs?SERVICE=WFS&Version=1.1.0&REQUEST=GetFeature&TYPENAME=p_bz-cadastre%3aCOMUNI_CATASTALI&srsName=EPSG%3a25832&FILTER=%3cFilter+xmlns%3d%22http%3a%2f%2fwww.opengis.net%2fogc%22%3e%3cBBOX%3e%3cPropertyName%3ep_bz-cadastre%3aCCAT_SHAPE%3c%2fPropertyName%3e%3cgml%3aEnvelope+srsName%3d%22http%3a%2f%2fwww.opengis.net%2fgml%2fsrs%2fepsg.xml%2325832%22+xmlns%3agml%3d%22http%3a%2f%2fwww.opengis.net%2fgml%22%3e%3cgml%3alowerCorner%3e413644.42867281759+4950195.1738821855%3c%2fgml%3alowerCorner%3e%3cgml%3aupperCorner%3e686355.57132718235+5149804.8261178145%3c%2fgml%3aupperCorner%3e%3c%2fgml%3aEnvelope%3e%3c%2fBBOX%3e%3c%2fFilter%3e&NAMESPACE=xmlns(p_bz-cadastre=http%3a%2f%2fgeoserver.d41.prov.bz%2fp_bz-cadastre): Cannot re-call start of asynchronous method while a previous call is still in progress.
Mapsui.Samples.Forms.Mac.exe Error: 0 : An exception occured while parsing a polygon geometry: The request was aborted: The request was canceled.
```
* It looks like most of the map is covered by red shapes, but there a some small holes in the southern parts. Could this be related to the errors above? Or are the holes expected?
* Finally, if this new WFS sample works well, we should remove the old one from the sample app (since it does not work at all).

janusw:
>     * I was not able to run `Mapsui.Samples.Forms.Android` from your branch ( ` Android.Content.Res.Resources+NotFoundException: Resource ID #0x7f0b0081` ). Does anyone else see this or is it some problem on my side?

I see this also on the master branch, so either it's a local problem with my setup or a general problem with Mapui. In any case it's not specific to your branch, so nevermind.

JBou:
> For now, its great that it works at all. I see an aerial map of Südtirol, which is covered by solid red segments. I hope that's the expected result

Yes, it should show the municipalities. Each solid red segment is a minicipality. The aerial map is just here to make it work, without it the limits are extend is worng and the map is blank.

> There are a few errors in the Debugger console. Is this something one should worry about?

The errors should be debugged later on, one by one. Maybe zooming or panning too fast (without waiting for the response) throws these errors. 

> Finally, if this new WFS sample works well, we should remove the old one from the sample app (since it does not work at all).

I have to say that the old sample works only for the states and POI layer, which returns the states of USA as red shapes (just like my sample). The other layers (SelectedStatesAndHouseholds, New, Roads, Landmarks, labels) are not working. YOu need to have a local [geoserver ](http://geoserver.org/download/)running, though. Even the old sample throws errors (
 I can remove it If you want to stick with the new sample for now.

If you have a WMS layer and just want to use WFS to get features on a specific point, it works great now. If you want to show the whole map using WFS as layer, performance is not good and it throws some errors. This needs to be fixed in a later PR. But even using other software (I tried it with QGis), the responses are much slower than with other Providers. But they may be using local caches to speed this up. SharpMap also implemented some kind of cache, but I don't know how it works.

JBou:
I first though the holes are expected, just missing data. But now I had a look in QGIS, and there it shows like that:
So there is a problem with rendering features that are not just a single shape.

2021-04-27 18_43_52-_Unbenanntes Projekt — QGIS

pauldendulk:
@janusw Thanks for the review. 

It's okay to fix the exceptions later but important to keep track of them.

Yes, the old sample can be removed. This can be part of the PR. Installing geoserver locally is too much to ask anyway.

JBou:
Alright, I removed the old sample. I just added a GetMapInfo to the new sample, so when you click on a shape it shows you a property of the feature. In addition, I added a LabelLayer which shows the name of the municipality.

It seems that having a WFS layer and a LabelLayer for the same provider generates duplicate requests, we should check this later on and if so, we should improve performance.
In addition, when zooming we should maybe add a little delay to the request, sometimes it send a wfs request multiple times while zooming, because the mouse wheel has a little delay while zooming.

pauldendulk:
> In addition, when zooming we should maybe add a little delay to the request, sometimes it send a wfs request multiple times while zooming, because the mouse wheel has a little delay while zooming.

There is a Delayer class that is used in the Layer class which could be used for fine tuning behavior like this. It has properties like MillisecondsToWait and StartWithDelay.

pauldendulk:
> It seems that having a WFS layer and a LabelLayer for the same provider generates duplicate requests, we should check this later on and if so, we should improve performance.

Perhaps this can be resolved with a LabelStyle on the same layer.

pauldendulk (approved):
Important improvements!

PR #1077, opened 2021-02-14
Update Svg.Skia

janusw:
... from version 0.4.1 to 0.5.0, update the nuspec dependency and update SkiaSharp to the latest version (in a place where it was missed apparently).

### DIFF
--- a/Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj
+++ b/Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SkiaSharp" Version="2.80.1" />
+    <PackageReference Include="SkiaSharp" Version="2.80.2" />
     <Folder Include="Properties\" />
   </ItemGroup>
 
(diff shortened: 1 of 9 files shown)


pauldendulk (approved):
Sometime packages updates are missed because on mac not all projects are opened. Does that play a role here?

janusw:
> Sometime packages updates are missed because on mac not all projects are opened. Does that play a role here?

I don't think so. The strange thing is that `Mapsui.Rendering.Skia.csproj` had two different references to SkiaSharp, only of of which was not up to date:

```
  <ItemGroup>
    <PackageReference Include="SkiaSharp" Version="2.80.1" />
    <Folder Include="Properties\" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="SkiaSharp" Version="2.80.2" />
    <PackageReference Include="Svg.Skia" Version="0.4.1" />
    <PackageReference Include="Topten.RichTextKit" Version="0.3.134" />
  </ItemGroup>
```

I don't quite understand this. Maybe the first one can even be removed completely?

pauldendulk:
@janusw Yes, I think this is just some form of corruption. Let's remove it. If there is a reason for it we will find out.

janusw:
> @janusw Yes, I think this is just some form of corruption. Let's remove it. If there is a reason for it we will find out.

Alright, I opened #1078 for it.

PR #1080, opened 2021-02-16
Set CSharp Language version to 9.0

inforithmics:
Here is a pull request to set the C# Language Version to 9.0 as discussed in

https://github.com/Mapsui/Mapsui/pull/1030

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <!--Specify Compiler for Mac Builds-->
-    <PackageReference Include="Microsoft.Net.Compilers" Version="3.8.0">
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="3.8.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -5,4 +5,12 @@
     <LangVersion>9.0</LangVersion>
   </PropertyGroup>
 
+  <ItemGroup>
+    <!--Specify Compiler for Mac Builds-->
+    <PackageReference Include="Microsoft.Net.Compilers" Version="3.8.0">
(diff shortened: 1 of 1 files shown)


inforithmics:
To use Csharp on Mac I had to specify the .NET Compiler (3.8), so on Windows on Mac the same Compiler is used which should solve problems of different Compilers.

pauldendulk:
@inforithmics What happens if you don't specify the compiler on Mac? 

Specifying a compiler as nuget package is new to me. I would assume this usually depends on what is installed. Referring to a nuget package makes it less dependent on a developers machine, so I guess it's a good thing, but I am unfamiliar with how that works.

pauldendulk:
I did little else in bitrise than import the project. I tried to update it later, got confused and then imported it again, which went smooth. We could do that again. 

Exactly what problem does it solve? Perhaps your situation is that you are building on windows and noticed a fail of the bitrise build and fixed it with this nuget package reference. Perhaps the nuget package reference is better, not sure.

inforithmics:
When I don't specifiy the Compiler following error occurs

"/Users/vagrant/git/Mapsui.iOS.sln" (Build target) (1) ->
"/Users/vagrant/git/Mapsui/Mapsui.csproj" (default target) (2) ->
"/Users/vagrant/git/Mapsui.Geometries/Mapsui.Geometries.csproj" (default target) (3:2) ->
(CoreCompile target) -> 
  CSC : error CS1617: Invalid option '9.0' for /langversion. Use '/langversion:?' to list supported values. [/Users/vagrant/git/Mapsui.Geometries/Mapsui.Geometries.csproj]
    0 Warning(s)
    1 Error(s)

Which means an older Csharp compiler is used on MacOs specifiying the Compiler explicitly solves this.

inforithmics:
I saw that this nuget package that I used is deprecated, changed it to use the correct nuget package. This is a 'temporary' workaround until MacOs gets a newer Csharp Compiler which usually occours sometime later. Then this nuget package can be removed.

pauldendulk (approved):
ok, well thought through.

PR #1082, opened 2021-02-21
README: update supported platforms

janusw:
* mention all platforms that are supported via Xamarin.Forms
* in particular MacOS was missing here


pauldendulk (approved):
Great

Issue #1084, opened 2021-02-26
SVG Pins: performance issues


janusw:
I suspect that the memory handling for SVG Pins is not quite optimal, and might have some issues.

Let me illustrate this with the following scenario:
* I have a map based on Mapsui.UI.Forms.MapView with a potentially larger number of pins.
* There are a number of different types of pins, where each type has its own image (e.g. given by an svg or png file). Let's say we have something like 10-20 types, and each image file has a few kB.
* The map will contain many pins of each type (say: hundreds or thousands).

Since there are many pins, I'd like to avoid duplicating the memory allocated for their images, for reasons of performance and memory footprint. I can see some potential issues in this direction (I've been looking at the `Pin` class and `Pin.CreateFeature` in particular):

1. For svg pins, the svg data is stored as a string (`Pin.Svg`). In `Pin.CreateFeature` this is then converted (copied) into a byte array via `Encoding.UTF8.GetBytes`. This means that for every Pin object I create, I make a separate copy of my svg data, right? For small svgs this might not be a problem, for for more complex ones that have several kB in size, it adds up. And it seems to be avoidable by making Pin.Svg a byte array (just like Pin.Icon) instead of a string. Correct?

2. IIUC the `BitmapRegistry` class could be used to avoid duplication of image data, but it seems to me that for every new pin a new entry in the BitmapRegistry is generated (since `Pin.CreateFeature` alwas seems to call `BitmapRegistry.Instance.Register`). Is that right, or have I simply not looked deep enough into the code? For png pins it might not be a problem: We might have multiple registry entries, but ultimately they all point to the same data in memory. But for svg (since we make a copy along the way, see first point above), we have multiple entries and they all point to their own private copy of the data, right?

@pauldendulk @charlenni 
I'd be grateful for any kinds of comments on this. Am I missing something here?

charlenni:
No, you are correct. `BitmapRegistry` doesn't check, if a "Bitmap" is already registered. And the Forms implementation is very dump in this case too. I didn't implement a check, if the SVG string is already registered. As always: MapView is a convenience wrapper around MapControl for Forms. If you want to do more than the basics of Mapsui, then you should use MapControl. There you could check, if your SVG is registered, and if yes, then use the ID again.

The conversion to a byte array was done, because `BitmapRegistry` once upon a time only accepted this format.

janusw:
[...]

PR #1085, opened 2021-02-26
change Pin.Svg from string to byte[]

janusw:
* this avoids a conversion in Pin.CreateFeature
* it is more consistent with Pin.Icon (which is also a byte array)
* and it makes the usage easier
  (avoiding the need for a StreamReader)
* see issue #1084

### DIFF
--- a/Mapsui.UI.Forms/Objects/Pin.cs
+++ b/Mapsui.UI.Forms/Objects/Pin.cs
@@ -1,7 +1,6 @@
 ﻿using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
-using System.Text;
 using Mapsui.Providers;
 using Mapsui.Rendering.Skia;
 using Mapsui.Styles;
@@ -25,7 +24,7 @@ namespace Mapsui.UI.Forms
         public static readonly BindableProperty LabelProperty = BindableProperty.Create(nameof(Label), typeof(string), typeof(Pin), default(string));
         public static readonly BindableProperty AddressProperty = BindableProperty.Create(nameof(Address), typeof(string), typeof(Pin), default(string));
         public static readonly BindableProperty IconProperty = BindableProperty.Create(nameof(Icon), typeof(byte[]), typeof(Pin), default(byte[]));
-        public static readonly BindableProperty SvgProperty = BindableProperty.Create(nameof(Svg), typeof(string), typeof(Pin), default(string));
+        public static readonly BindableProperty SvgProperty = BindableProperty.Create(nameof(Svg), typeof(byte[]), typeof(Pin), default(byte[]));
         public static readonly BindableProperty ScaleProperty = BindableProperty.Create(nameof(Scale), typeof(float), typeof(Pin), 1.0f);
         public static readonly BindableProperty RotationProperty = BindableProperty.Create(nameof(Rotation), typeof(float), typeof(Pin), 0f);
         public static readonly BindableProperty RotateWithMapProperty = BindableProperty.Create(nameof(RotateWithMap), typeof(bool), typeof(Pin), false);
@@ -148,9 +147,9 @@ namespace Mapsui.UI.Forms
         /// <summary>
         /// String holding the Svg image informations
         /// </summary>
-        public string Svg
+        public byte[] Svg
         {
-            get { return (string)GetValue(SvgProperty); }
+            get { return (byte[])GetValue(SvgProperty); }
             set { SetValue(SvgProperty, value); }
         }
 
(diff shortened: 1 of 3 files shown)


janusw:
> I would not expect, that SVG, which is a readable format, would be stored as byte[].

Why not? Technically speaking, it makes everything easier and avoids extra conversion code, as you can see from the diff.

> You could create an SVG with StringBuilder.

I guess you could, but is that a use case that you have seen in any real-world code? Of course it's still possible to do, but will require a conversion string -> byte[].

I think the far more widespread case is the one where you read the svg data from a file, and it's the only one covered by the samples. IMHO the important case should be the simplest to use, not the rare one.

charlenni:
As I described earlier, MapView is a convenience wrapper around the basic MapControl. And for this, the objects that are used are convenience objects, which should be usable very easy. If you now make SVG a byte array, all have to convert their strings into byte arrays by their own. If I use the library, I wouldn't expect this. For me is an SVG in a plain text format. I wouldn't expect a byte array for this.

If you want to use this library in a more advanced manner, than you could create your own layers and us them with special techniques. You could bypass many of the slow conversions.

Ask yourself: what type would I expect for SVGs? And if byte array is the format you would think about first, then we should change it. If you would expect a string, then we should leave it, even if it is slower.

janusw:
> As I described earlier, MapView is a convenience wrapper around the basic MapControl. And for this, the objects that are used are convenience objects, which should be usable very easy.

Right. They should be usable and they should be performant.


> If you now make SVG a byte array, all have to convert their strings into byte arrays by their own.

Changing the API is not nice, but progress can not be avoided completely. Mapsui v3 is in alpha state. API breakage should be possible, and this will not be the only case.

Also the conversion is really simple (and more importantly it improves performance).

> If I use the library, I wouldn't expect this. For me is an SVG in a plain text format. I wouldn't expect a byte array for this.

Actually this PR is not about your expectations and not about mine. Expectations are subjective.

This PR is about performance. The code for svg pins currently contains an unnecessary conversion from string to byte[]. It may not be the biggest performance killer in the svg game, but it's completely unnecessary and avoidable if one uses byte[] in the first place.


> If you want to use this library in a more advanced manner, than you could create your own layers and us them with special techniques. You could bypass many of the slow conversions.

a) I don't want to do that.
b) I don't think there are many "slow conversions". And if there were, they should be removed.

> Ask yourself: what type would I expect for SVGs? And if byte array is the format you would think about first, then we should change it. If you would expect a string, then we should leave it, even if it is slower.

That's a strange way to think about things.

I'm asking myself: Which option is the most natural one? That is not necessarily the same as my (possibly flawed) first expectation, and I may need to think twice about it. The most natural option is the one which requires the least conversion code, corresponds to the typical usage scenario and is most performant.

Actually another possible choice for Pin.Svg might be SkSvg. But if you still want to continue supporting the Xaml renderer, that's probably not a good one. For a Skia-only implementation it would be perfect.

Also, one more thing: If I open a PR which touches code that you wrote, it doesn't mean: "I hate your code. It sucks." If that was the case, I wouldn't even invest any time in it, but just move on to another library. Instead it means: "Your code is great. Let's make it even better." You should take it as a kudo, not as an insult.

charlenni:
I know, that you want to help, but I'm sorry to say, that I see this case different then you.

I would expect, that SVG is a string. That was the reason, why I implemented this way. Further I assume, that the additional conversion from string to byte array isn't your performance problem. Or did you use a profiler to check this?

I try to comment at #1084, so we have only one place to discuss. I hope, this is ok.

charlenni:
As you can see [here](https://github.com/wieslawsoltes/Svg.Skia/blob/master/src/Svg.Skia/SKSvg.cs), other developers think too, that SVG is a string.

janusw:
> I know, that you want to help, but I'm sorry to say, that I see this case different then you.

Alright, I guess then we just agree to disagree.


> I would expect, that SVG is a string. That was the reason, why I implemented this way. Further I assume, that the additional conversion from string to byte array isn't your performance problem. Or did you use a profiler to check this?

No, I didn't. And yes, this is probably not my primary performance problem.

It's just a small and close-to-obvious modification that changes everything for the better AFAICS (simpler code, better performance, etc).

I wouldn't even care about it if there was a proper cache/registry, which would recognize that it knows a particular svg already and would not have to store it again, thus performing the unnecessary conversion for all the thousands of pins that I might throw at it (even if they all have the same svg).

If I add 1000 pins, I can add a cache in my user-side code which prevents me from loading the same svg a thousand times, but I cannot prevent the string -> byte[] conversion from happening a thousand times, and that is the problem here.

@pauldendulk Do you have an opinion on all of this?

tmijieux:
maybe instead of changing the type of the Svg property we could add and SvgBytes property and both properties would refer to the same backing field (getters optional?)
```csharp
public byte[] SvgBytes { set => IntoDrawReadyRepresentation(value); }
public string Svg {  set => IntoDrawReadyRepresentation(value); }
```

charlenni:
So, I now created a PR #1098 which doesn't change SVG to a byte array, but treats them as string. I think, this is the logically correct way.  

Additionally the strings and rendered pins are stored in a cache to save memory.

This doesn't change anything with your performance problems, because it makes only Pins with PinType.Pin and the same color faster.

pauldendulk:
I'll close this since I merged charlennis changes. Perhaps some more changes follow.

PR #1086, opened 2021-03-02
Improve Forms.Pin

charlenni:
This is an example for @janusw how a possible cache for Forms.Pin could look like. Also, it demonstrates, how SVGs could be implemented without any byte array or stream. Should work on both render platforms (Xaml and Skia) where new features work only on Skia renderer. But this isn't a problem, because only Forms implementation uses it.

Regarding the speed I had to say, that I don't know what's going on. The SKPicture is much slower than the SKImage. If I use the built in profiler, I could see with the many pin sample the following diagram:

grafik

There you could see, that only a small part of the CPU time is used to render the screen. So what does the program meanwhile? Has anyone a better profiler to take a deep dive in?

@pauldendulk Sorry for the additional commits. Only the last is valid.


janusw:
Thanks for looking into this, @charlenni. I'll try to have a look at it soon.


> @pauldendulk Sorry for the additional commits. Only the last is valid.

Maybe you should rebase your branch to Mapsui:master? With an interactive rebase (`git rebase -i`) you can easily get rid of the extra commits.

charlenni:
Thank you. I will try it if I return to my computer.

tmijieux:
```
SKPicture Class
Recorded drawing operations made to a SKCanvas to be played back at a later time.
```
It basically means the underlying SKia library will perform mathematical computation each time to compute what pixel to draw what color(if it does not find a way to optimize that somehow).
If it is done once into a buffer (basically rasterizing into a SKImage) and if that image is shared among all the pins that use the same svg, the chance it will be faster that way is great because the buffer will be loaded into memory cache, and drawing the pin will just be a blit into the framebuffer(memory copy between two memory location that fits together in the cache can be very fast) but theorically situations where the reverse could be true are conceivable  if, the buffer is large(big image dimensions, framebuffer not fitting in cache) and the draw command is very simple (for example filling a large rectangle with a single, opaque, color) then may the SKPicture could be faster in that case because that would require less memory pressure.

> There you could see, that only a small part of the CPU time is used to render the screen. So what does the program meanwhile? Has anyone a better profiler to take a deep dive in?

If by some mean hardware acceleration is used you're not gonna see much on the processor, probably lots of time is just lost transfering data and/or kernels between cpu and gpu.
If that is the case the same optimization than before probably applies.

charlenni:
Thank you @tmijieux. That is, what I expected. Perhaps I hoped, that there something to optimize, but there seems no room for this. 

So scaling the canvas, drawing the SVG and than use this image for later would be the best.

But the other improvements are worth to go into the code.

tmijieux:
I dont know about the implementation details and internals of SkiaSharp but according my general knowledge about performance and hardware acceleration this is my best guess right now and this is why i am most interested in finding a way to improve the quality when using the "rasterizing method". Definitive answer will be found by trying different implementation and benchmarking them

charlenni:
One point is the correct resizing before conversion, so we have no quality loss. I think we do this. Another point is to use high quality when drawing the image. That we don‘t do. As you could see in the code, we use low. This is the case, because, if I remember it correct, the high quality image copying, which was done for map tiles, was much slower than the low quality process.

@tmijieux Do you have quality problems with your SVG Pins?

tmijieux:
> One point is the correct resizing before conversion, so we have no quality loss. I think we do this.



In [this code](https://github.com/Mapsui/Mapsui/blob/master/Mapsui.UI.Forms/Objects/Pin.cs#L462-L489) the SKBitmap is created with the dimension of the svg, and the scaling is done in the renderer with svg.Scale() after the bitmap was rasterizing. Isn't that the contrary of what you are stating ?

> Another point is to use high quality when drawing the image. That we don‘t do. As you could see in the code, we use low. This is the case, because, if I remember it correct, the high quality image copying, which was done for map tiles, was much slower than the low quality process.

you mentionned that "Low quality" before but i dont understand what you are refering to, where is this parameter ?

> @tmijieux Do you have quality problems with your SVG Pins?

yes in my app, I tried to do the scaling before(scale > 1 and put SymbolScale = 1 into the Feature...) but the final resolution is bad, i can tell that the shapes borders are not very neat just by looking at it

```csharp
            var svg = new SKSvg();
            svg.Load(data);

            var info = new SKImageInfo(
                (int)(svg.Picture.CullRect.Width * scale),
                (int)(svg.Picture.CullRect.Height * scale)) {
                AlphaType = SKAlphaType.Premul
            };

            var bitmap = new SKBitmap(info);
            var canvas = new SKCanvas(bitmap);
            canvas.Clear();
            canvas.Scale((float)scale);

            // paint.ColorFilter = SKColorFilter.CreateBlendMode(
            //     Color.ToSKColor(),
            //     SKBlendMode.SrcIn
            // ); // use the source color

            using var paint = new SKPaint() { IsAntialias = true };
            canvas.DrawPicture(svg.Picture, paint);

            using var image = SKImage.FromBitmap(bitmap);
            using var bitmapData = image.Encode(SKEncodedImageFormat.Png, 100);
            var bitmapDataArray = bitmapData.ToArray();
            var stream = new MemoryStream(bitmapDataArray);
            var bitmapId = BitmapRegistry.Instance.Register(stream);
```

tmijieux:
ewwww found my culprit [here](https://github.com/Mapsui/Mapsui/blob/master/Mapsui.UI.Forms/MapControl.cs#L255-L260)

charlenni:
Regarding the quality: look at BitmapRenderer.cs in Mapsui.Rendering.Skia in line 11(?). There the quality is set for the used SKPaint. Try to change this to High.

tmijieux:
oh indeed i see how this could affect the quality when upscaling a bitmap, 
according to the doc from upstream [cpp skia](https://github.com/google/skia/blob/master/include/core/SkFilterQuality.h)
"Low" is not the lowest actually. Contrary to what is said in [skiasharp doc]( https://docs.microsoft.com/fr-fr/dotnet/api/skiasharp.skfilterquality?view=skiasharp-2.80.2) "None" is not unspecified:  it is the lowest qualify where no interpolation is done at all and it should be even faster.

Mapsui.UI.Forms/Objects/Pin.cs
```diff
@@ -449,43 +454,69 @@ private void CreateFeature()
...
-                        var canvas = new SKCanvas(bitmap);
-                        // Now draw Svg image to bitmap
-                        using (var paint = new SKPaint() { IsAntialias = true })
                         {
-                            // Replace color while drawing
-                            paint.ColorFilter = SKColorFilter.CreateBlendMode(Color.ToSKColor(), SKBlendMode.SrcIn); // use the source color
-                            canvas.Clear();
-                            canvas.DrawPicture(svg.Picture, paint);
+                            // Load the SVG document
+                            hash = Svg.GetHashCode();
```
tmijieux, line 460: Using `GetHashCode()` there is a chance that different svg will conflict and the user will end up having the wrong svg drawing , I understand that you want to avoid full string comparison but this code is not even supposed to run on the UI thread, this is the view "preparation" that is done once when you create all your pins which isnt going to happen every frame , so i would personnaly use the svg string as the key to the dictionnary directly instead for correctness that does not depend on hash conflict chance.

Mapsui.UI.Forms/Objects/Pin.cs
```diff
@@ -449,43 +454,69 @@ private void CreateFeature()
...
+                                }
+                            }
                         }
-                        // Now convert canvas to bitmap
-                        using (var image = SKImage.FromBitmap(bitmap))
-                        using (var data = image.Encode(SKEncodedImageFormat.Png, 100))
+                        break;
+                    case PinType.Pin:
+                        var skColor = Color.ToSKColor();
+                        var colorInHex = $"{skColor.Red.ToString("X2")}{skColor.Green.ToString("X2")}{skColor.Blue.ToString("X2")}";
```
tmijieux, line 479: maybe for the PinType.Pin we could use a second dictionnary with just the color hexstring as a key ?

Also xamarin forms Color already have a string Color.ToHex() method that does exactly that with the alpha component (format) #ARGB

Mapsui.UI.Forms/Objects/Pin.cs
charlenni: Your are right, not the best idea. But if you use the whole string, you need a big part of memory. Perhaps you have any other idea, what to use?

Mapsui.UI.Forms/Objects/Pin.cs
charlenni: Did you checked, if a ARGB value is usable at this place?

charlenni:
Than you have to create a SVG with the correct size and set the Scale property to 1. Shouldn’t be very complicated.

tmijieux:
> 
> 
> Than you have to create a SVG with the correct size and set the Scale property to 1. Shouldn’t be very complicated.

what i mean by that is: "never ever calling the Scale method before drawing bitmaps unless we explicitly want this", even this one (L256) should not be called for bitmaps to be sure the implementation will not try to apply any bitmaps scaling algorithm (because this is heavy for performances)
https://github.com/Mapsui/Mapsui/blob/fca6076ddb0dbdfee740bab883db97307fa2f276/Mapsui.UI.Forms/MapControl.cs#L255-L260

I also found this article that may hold some interesting informations (this is confirming what i want to do)
https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Optimizing_canvas

Mapsui.UI.Forms/Objects/Pin.cs
tmijieux: svg needs RGBA :sob:

charlenni: I tried this, but found, that #RGBA doesn‘t work. So I used #RGB. Have to investigate this.

A special array for pin would be a possibility. I assume, that I have to make this PR new.

charlenni:
I made a short [test](https://github.com/charlenni/VTMap) with 2000 pins on a map. Tried it on UWP and would say, it is fast enough. It draws the pins as SKPicture on the map. The SVG image is converted and colored before drawing.

charlenni:
Ok, on my old Android phone it runs not so smooth, but not so slow then with Mapsui. It seems, that Mapsui is drawing too much or updates too often. In the demo, I try to update the screen each 16 ms. If this isn't possible, because the last drawing isn't ready, then the update of the screen is dropped.

I assume, that there is a high potential for optimizations of Mapsui. So don't hesitate to make PRs ;-)

charlenni:
@tmijieux I updated the above mentioned demo project, so that each of the 2000 pins is scaled before it is drawn. If you try the demo, you would see, that there isn't a big difference between with and without scaling while drawing. So I assume, that this isn't our problem.

@janusw To different things:
1. Rebasing didn't work. As you have already expected, I'm no Git expert. If we are ready with the changes we want, I create them with a new PR
2. Your problem is not the conversion from string to byte[] and back. There could be made optimizations, but all of them make the drawing not much faster. Perhaps we could reduce the memory consumption a little smaller, but that's all.

So, to make Mapsui draw faster, there had to be made other improvements. Any idea?

janusw:
> @janusw To different things:
> 
>     1. Rebasing didn't work. As you have already expected, I'm no Git expert. If we are ready with the changes we want, I create them with a new PR

Here's my standard workflow for such a case:
1. Check out your branch.
2. `git rebase -i master`
3. Pick only the last commit, forget about the other one (which updates the animation sample and is totally unrelated).
4. You then get a conflict in `Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/AnimationSample.cs`. Fix it by getting rid of any changes in the AnimationSample.
5. `git add Samples` and `git rebase --continue`
6. You get a brand-new branch that is based on the current master and has only one commit on top of it, and consequently this PR shows only one commit as well.

tmijieux:
> I assume, that there is a high potential for optimizations of Mapsui. So don't hesitate to make PRs ;-)

I have ideas running in my head at the moment, i will probably do tests and maybe propose a PR if i have  time to do it and good results.


> @tmijieux I updated the above mentioned demo project, so that each of the 2000 pins is scaled before it is drawn. If you try the demo, you would see, that there isn't a big difference between with and without scaling while drawing. So I assume, that this isn't our problem.

ok, i will read the whole thing to be sure it corresponds to what i was thinking about, i'm pretty sure having to apply linear interpolation on each pixels each time we do a bitmap draw does account for something but maybe that something is very small overhead because hardware acceleration maybe handle this very well, or maybe they other thing that slows down even more and we cannot currently tell if there is a problem with that.

> But what I discovered is the following. That every operation causes more than 1 Refresh Graphics call.

I would have bet that the the InvalidateSurface() does not call the PaintSurface event synchronously but rather wait for the user code stack to be empty(giving control back to framework) Does someone know about this ?


(the way I think about this is, InvalidateSurface() set a boolean somewhere to say the view need drawing, but calling invalidatesurface multiple time would set the same boolean, and when the user code give control back to framework the PaintSurface is called once for all the InvalidateSurface() calls, this is basically the way it works in browsers if you modify the DOM in js, it basically waits for the js stack to be empty before redrawing and does not schedule a redraw immediately after first modification)

charlenni:
Closed for #1091

inforithmics:
> > I assume, that there is a high potential for optimizations of Mapsui. So don't hesitate to make PRs ;-)
> 
> I have ideas running in my head at the moment, i will probably do tests and maybe propose a PR if i have time to do it and good results.
> 
> > @tmijieux I updated the above mentioned demo project, so that each of the 2000 pins is scaled before it is drawn. If you try the demo, you would see, that there isn't a big difference between with and without scaling while drawing. So I assume, that this isn't our problem.
> 
> ok, i will read the whole thing to be sure it corresponds to what i was thinking about, i'm pretty sure having to apply linear interpolation on each pixels each time we do a bitmap draw does account for something but maybe that something is very small overhead because hardware acceleration maybe handle this very well, or maybe they other thing that slows down even more and we cannot currently tell if there is a problem with that.
> 
> > But what I discovered is the following. That every operation causes more than 1 Refresh Graphics call.
> 
> I would have bet that the the InvalidateSurface() does not call the PaintSurface event synchronously but rather wait for the user code stack to be empty(giving control back to framework) Does someone know about this ?
> 
> (the way I think about this is, InvalidateSurface() set a boolean somewhere to say the view need drawing, but calling invalidatesurface multiple time would set the same boolean, and when the user code give control back to framework the PaintSurface is called once for all the InvalidateSurface() calls, this is basically the way it works in browsers if you modify the DOM in js, it basically waits for the js stack to be empty before redrawing and does not schedule a redraw immediately after first modification)

Yes something along this way I thought could be implementended and will help Mapsui Performance a lot because, currently the logic is 

PropertyChanged -> Causes a redraw.
Another PropertyChanged -> Causes a second redraw, the redraws are scheduled on the UI Thred so the redraws have to wait for a the last redraw to finish and often render the same content.

tmijieux:
@inforithmics I just checked the SKGLView already does this mostly , it seems that the drawing is done in a background thread, and if you call InvalidateSurface a lots of time during the same user code bit( event handler for example) , it sometimes does 2 draws because on of the InvalidateSurface lands after the first draw started.(if the draws are long the problem may get worse)
(for 1,2 or 3 i always have 1 draw only, if i call InvalidateSurface 300 times (without any operation in between) i never get more than 2 draws) I suspect that if there is time between InvalidateSurface, or if the draw is long, maybe more draws will be scheduled)

PR #1088, opened 2021-03-03
Fix #1087

### LINKED ISSUE #1087: Navigator.NavigateTo(center, resolution, ..., ...) doesn't work if using a viewport limiter due to order of operation
https://github.com/Mapsui/Mapsui/blob/b4df2b9388cc94d61f54a09fea7f99967a392e54/Mapsui/Navigation/Navigator.cs#L88

If you trigger the center change before the resolution change and you are fully zoomed out, this can cause the center Y to end up as 0 but the resolution is updated in the case where a viewport limiter is in place.

The order should be to set the resolution first, then the center.

radderz:
#1087

This fixes the incorrect order of updating the viewport for when fully zoomed out in a limited viewport. If Center is set first then the Y value will end up as 0 as the zoom happened second.

### DIFF
--- a/Mapsui/Navigation/Navigator.cs
+++ b/Mapsui/Navigation/Navigator.cs
@@ -85,8 +85,8 @@ namespace Mapsui
 
             if (duration == 0)
             {
-                _viewport.SetCenter(center);
                 _viewport.SetResolution(resolution);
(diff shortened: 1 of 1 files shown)


pauldendulk (approved):
Good find

PR #1091, opened 2021-03-06
Added some improvements for Mapsui.Forms Pin

charlenni:
Forms: Added a cache for bitmap of Pin
Forms: Added new SKPicture renderer
Forms: Changed Pin with SVG image to vector instead of bitmap drawing

This time clean and, I hope so, without errors.

### DIFF
--- a/Mapsui.Rendering.Skia/BitmapHelper.cs
+++ b/Mapsui.Rendering.Skia/BitmapHelper.cs
@@ -1,21 +1,36 @@
-using System.IO;
-using System.Text;
 using Mapsui.Extensions;
 using Mapsui.Styles;
 using SkiaSharp;
 using Svg.Skia;
+using System.IO;
 
 
 namespace Mapsui.Rendering.Skia
 {
     public static class BitmapHelper
     {
-        public static BitmapInfo LoadBitmap(object bitmapStream)
+        public static BitmapInfo LoadBitmap(object bitmap)
         {
             // todo: Our BitmapRegistry stores not only bitmaps. Perhaps we should store a class in it
             // which has all information. So we should have a SymbolImageRegistry in which we store a
             // SymbolImage. Which holds the type, data and other parameters.
-            if (bitmapStream is Stream stream)
+            if (bitmap is string str)
+            {
+                if (str.ToLower().Contains("<svg"))
+                {
+                    var svg = new SKSvg();
+                    svg.FromSvg(str);
+
+                    return new BitmapInfo { Svg = svg };
+                }
+            }
+
+            if (bitmap is SKPicture picture)
+            {
+                return new BitmapInfo { Picture = picture };
+            }
+
+            if (bitmap is Stream stream)
             {
                 if (stream.IsSvg())
                 {
@@ -29,7 +44,7 @@ namespace Mapsui.Rendering.Skia
                 return new BitmapInfo {Bitmap = image};
             }
 
-            if (bitmapStream is Sprite sprite)
+            if (bitmap is Sprite sprite)
             {
                 return new BitmapInfo {Sprite = sprite};
             }
--- a/Mapsui.Rendering.Skia/BitmapInfo.cs
+++ b/Mapsui.Rendering.Skia/BitmapInfo.cs
@@ -8,7 +8,8 @@ namespace Mapsui.Rendering.Skia
     {
         Bitmap,
         Svg,
-        Sprite
+        Sprite,
+        Picture
     }
 
     public class BitmapInfo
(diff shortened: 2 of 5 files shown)


tmijieux:
>  Forms: Changed Pin with SVG image to vector instead of bitmap drawing

so how does the manypins samples that was using the bitmap compare to the previous implementation performancewise ?

charlenni:
@tmijieux It seems slower than using bitmaps.

@tmijieux and @inforithmics In the mentioned demo I used a timer, that each 16 ms (60 Hz) checks, if a flag is set or an animation is running. If one of this two is true, the surface is redrawn. If a redraw is started, no other redraw could be made. The timer ticks returns without any operation.

tmijieux:
```csharp

        void Draw(SKCanvas canvas)
        {
            if (_drawing)
                return;

            lock (_syncObject)
            {
                _drawing = true;

                canvas.Clear(SKColors.LightGray);

                foreach (var renderer in Map.Layers.Renderers)
                    renderer.Draw(canvas, Viewport);

                foreach (var renderer in Map.Overlays.Renderers)
                    renderer.Draw(canvas, Viewport);

                NeedsRedraw = false;

                _drawing = false;
            }
        }
```
this code theorically does not prevent two or more draw happening on the same data(even if it is unlikely)
Two threads could arrive on the same time on the lock line before _drawing is set to true
consider the following scenario , first time your enter `OnTimerCallback` you call InvalidateSurface, then , say a thread will be started to do the draw, but because all the core are currently busy, and for obscure scheduling reason, busy the thread will not start before the next tick of `OnTimerCallback`, another InvalidateSurface is called and now the two thread will arrive on the same time on the lock line (using 2 cores)

what you could do here is add a check on the NeedsRedraw just before the canvas.Clear() and return if it is false, but that would let the second thread blocking for the whole duration of the first thread draw()...
or you could be using a try/lock or atomic exchange to address this kind of race patterns...
https://stackoverflow.com/questions/8546/is-there-a-try-to-lock-skip-if-timed-out-operation-in-c
https://docs.microsoft.com/fr-fr/dotnet/api/system.threading.interlocked.exchange?view=net-5.0

charlenni:
So you will say, that this isn't a good idea? Any other idea?

tmijieux:
> 
> 
> So you will say, that this isn't a good idea? Any other idea?

no this definitely not a bad idea, just the way you ensure "only one draws happen a a time" is not 100% safe to enforce exactly that just like in the link a posted you could do something like
```csharp
        void Draw(SKCanvas canvas)
        {
            if (_drawing)
                return;

            if (Monitor.TryEnter(_syncObject))
            {
                try {
                    _drawing = true;

                    canvas.Clear(SKColors.LightGray);

                    foreach (var renderer in Map.Layers.Renderers)
                        renderer.Draw(canvas, Viewport);

                    foreach (var renderer in Map.Overlays.Renderers)
                        renderer.Draw(canvas, Viewport);

                    NeedsRedraw = false;


                }
                catch(Exception e)
                {
                    Console.WriteLine($"error during draw:{e}");
                }
                finally
                {
                     _drawing = false;
                    Monitor.Exit(someobject)
                }
            }
        }
```

charlenni:
Thank you. I try to change this.

Do you have problems with getting tiles? I always get a System.Net.Http.HttpRequestException when trying to get tiles.

tmijieux:
another problem that i spoted in your vtmap is that the NeedsRedraw must not be udpated during the Draw call(especially not at the end) because the draw call may be on another thread (not the ui thread). If we do that, we may skip some required draws(if a new draw is required while a draw is currently going, ... that new draw will be skipped), the most simple method to avoid race conditions is to serialize all operations with side effects on the UI on the same thread, this is the principle of the unique UI thread, you're never supposed to update the view outside of the UI thread
you should do if (redraw) RunOnUIThread(() => { NeedsRedraw = false; Invalidate(); })
and modify Invalidate so that it does not call RunOnUIThread a second time

tmijieux:
> 
> 
> Thank you. I try to change this.
> 
> Do you have problems with getting tiles? I always get a System.Net.Http.HttpRequestException when trying to get tiles.

on what samples are you getting this errors ?

charlenni:
On all Forms samples. It seems, that there are problems with BruTiles.

tmijieux:
i have no problem with the tiles

janusw:
> @tmijieux It seems slower than using bitmaps.

Yes, the ManyPinsSample is much slower on this branch (compared to master). This is certainly not a good idea (at least not the part that concerns PinType.Pin).

The changes for PinType.Svg may still be helpful, haven't tried that yet.

You're really doing too many things at once here. At the very least, the unrelated parts of this branch should be in separate commits, so that reach of them can be benchmarked on their own. The PinType.Pin changes definitely need to be removed.

charlenni:
You are right. Than I close this PR and make a new one, if I have spare time.

PR #1092, opened 2021-03-06
Added a render loop instead of always calling InvalidateSurface()

charlenni:
Added a render loop instead of always calling InvalidateSurface() each time anything changes.

The render loop is called each 16 ms and checks, if something has changed (needs redraw or an animation is running). If a redraw is already running, the frame is dropped.

This is a PR for discussion and isn't tested very heavy. I'm not sure, if this is much faster than the old system. Should be tested carefully.

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -1,3 +1,4 @@
+using Mapsui.Geometries.Utilities;
 using Mapsui.Rendering;
 using Mapsui.UI.Utils;
 using SkiaSharp;
@@ -5,11 +6,12 @@ using SkiaSharp.Views.Forms;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
 using System.Linq;
-using Mapsui.Geometries.Utilities;
-using Xamarin.Forms;
+using System.Threading;
 using System.Threading.Tasks;
-using Mapsui.Utilities;
+using Xamarin.Forms;
 
 namespace Mapsui.UI.Forms
 {
@@ -54,6 +56,16 @@ namespace Mapsui.UI.Forms
         private int _numOfTaps = 0;
         private readonly FlingTracker _flingTracker = new FlingTracker();
         private Geometries.Point _previousCenter;
+        private bool _sizeChanged = false;
+
+        // Timer for rendering loop
+        System.Threading.Timer _timer;
+        // Values for drawing loop
+        readonly Stopwatch _stopWatch = new Stopwatch();
+        double _fpsAverage = 0.0;
+        const double _fpsWanted = 60.0;
+        int _fpsCount = 0;
+        object _lockObj = new object();
 
         /// <summary>
         /// Saver for angle before last pinch movement
@@ -71,6 +83,11 @@ namespace Mapsui.UI.Forms
         {
             Initialize();
         }
+
+        /// <summary>
+        /// Flag for redrawing view
+        /// </summary>
+        public bool NeedsRedraw { get; set; } = true;
 
         public float ScreenWidth => (float)Width;
 
@@ -95,13 +112,16 @@ namespace Mapsui.UI.Forms
             Touch += OnTouch;
             SizeChanged += OnSizeChanged;
 
+            // Create timer for redrawing
+            _timer = new System.Threading.Timer(OnTimerCallback, null, TimeSpan.FromMilliseconds(1000.0 / _fpsWanted), TimeSpan.FromMilliseconds(1000.0 / _fpsWanted));
+
             _initialized = true;
         }
 
         private void OnSizeChanged(object sender, EventArgs e)
         {
             _touches.Clear();
-            SetViewportSize();
+            _sizeChanged = true;
         }
 
         private async void OnTouch(object sender, SKTouchEventArgs e)
@@ -249,15 +269,33 @@ namespace Mapsui.UI.Forms
 
         void OnPaintSurface(object sender, SKPaintGLSurfaceEventArgs skPaintSurfaceEventArgs)
         {
-            if (PixelDensity <= 0) return;
+            if (_sizeChanged)
+            {
+                SetViewportSize();
+                _sizeChanged = false;
+            }
 
-            Navigator.UpdateAnimations();
-
-            skPaintSurfaceEventArgs.Surface.Canvas.Scale(PixelDensity, PixelDensity);
-
+            bool lockTaken = false;
             var canvas = skPaintSurfaceEventArgs.Surface.Canvas;
 
-            _renderer.Render(canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
+            try
+            {
+                Monitor.TryEnter(_lockObj, ref lockTaken);
+
+                if (lockTaken)
+                {
+                    skPaintSurfaceEventArgs.Surface.Canvas.Scale(PixelDensity, PixelDensity);
+
+                    _renderer.Render(canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
+
+                    NeedsRedraw = false;
+                }
+            }
+            finally
+            {
+                if (lockTaken)
+                    Monitor.Exit(_lockObj);
+            }
         }
 
         private Geometries.Point GetScreenPosition(SKPoint point)
@@ -274,7 +312,7 @@ namespace Mapsui.UI.Forms
                 return;
             }
 
-            RunOnUIThread(InvalidateSurface);
+            NeedsRedraw = true;
         }
 
         /// <summary>
@@ -737,6 +775,43 @@ namespace Mapsui.UI.Forms
         {
             if (Width <= 0) return 0;
             return (float)(CanvasSize.Width / Width);
+        }
+
+        // See http://codetips.nl/skiagameloop.html
+        void OnTimerCallback(object state)
+        {
+            // Get the elapsed time from the stopwatch because the 1/fps timer interval is not accurate and can be off by 2 ms
+            var dt = _stopWatch.Elapsed.TotalSeconds;
+
+            // Restart the time measurement for the next time this method is called
+            _stopWatch.Restart();
+
+            // Workload in background
+            var redraw = NeedsRedraw || Navigator.UpdateAnimations();
+
+            // Calculate current fps
+            var fps = dt > 0 ? 1.0 / dt : 0;
+
+            // When the fps is to low, reduce the load by skipping the frame
+            if (fps < _fpsWanted / 2)
+                return;
+
+            // Calculate an averaged fps
+            _fpsAverage += fps;
+            _fpsCount++;
+
+            if (_fpsCount == 20)
+            {
+                fps = _fpsAverage / _fpsCount;
+                Debug.WriteLine($"FPS {fps.ToString("N3", CultureInfo.InvariantCulture)}");
+
+                _fpsCount = 0;
+                _fpsAverage = 0.0;
+            }
+
+            // Called if needed
+            if (redraw)
+                Device.BeginInvokeOnMainThread(() => InvalidateSurface());
         }
     }
 }
--- a/Mapsui/Navigation/AnimatedNavigator.cs
+++ b/Mapsui/Navigation/AnimatedNavigator.cs
@@ -79,7 +79,7 @@ namespace Mapsui
             throw new NotImplementedException();
         }
 
-        public void UpdateAnimations()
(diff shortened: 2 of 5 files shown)


janusw:
I'm not sure if I'm the right person to review your code, but I can ask some relevant questions:

- Which problem are you trying to solve here?
- Which samples should benefit from it the most?
- How to construct an extreme case which showcases the potential performance speedup of this change?
- Could there be cases where performance gets worse?

charlenni:
@janusw You are as good as any other person 😉

I don‘t know, if there are any performance improvements. I didn‘t thought about this. We talked about calling InvalidateSurface not so often. This would be one way to do this.

The greatest benefit would be to have a correct base for any type of animations, not only for navigation.

tmijieux:
https://github.com/Mapsui/Mapsui/issues/1084#issuecomment-798925603

I'm pretty sure xamarin forms implementation already does something like this (see my answer in the issue)
The first problem that needs to be assigned in order to improve performance is proper timing of the rendering. (i strongly believe that benchmarking code needs to be commited into the main mapsui repository, potentially with  conditions likes `#if MAPSUI_BENCHMARK ... #endif` if we having timing code in the final nuget is not desirable)
Then we need to benchmark and save the results of every single samples on various platforms and devices, and then start to comparing what improvements would bring something like this.

pauldendulk:
> The greatest benefit would be to have a correct base for any type of animations, not only for navigation.

I would suggest a solution where all animations follow the design of the navigation animations. So calculate an update of the animation state right before rendering. It will not be possible to calculate an animation path beforehand. So this is a constraint which in some cases will make things harder. But then again, changes to the animation path due to user interventions are a problem here anyway. An important advantage is that you do not have to synchronize an animation update frequency with the render loop frequency.

pauldendulk:
> i strongly believe that benchmarking code needs to be committed into the main mapsui repository

Yes, that is the best way to make real improvement.

inforithmics:
I have a Performance Problem with Mapsui

DisplayMap

As you can See Iterate Layers is Called more than 4000 Times

With this Patch It is only called 182 Times, that's an improvement of 2000%

DisplayMapRenderLoop

I tested it on other samples and it became faster too. But not this much. In this sample I have a lot of jpg for rendering. So If nobody is oposed to it I would like to approve this patch.

inforithmics:
Another advantage of this patch is, that the UI Thread gets some time to do other tasks than rendering. For example with this patch the longClick works more reliable.
Maybe I should point out that the Profiling was done on a 4K Screen on a 1K (1920x1080) Screen it is not so bad.

charlenni:
Yes, this is one thing, that this render loop could solve. One more is the problem of removing and adding the Forms 4 top most layers again and again. I assume, it is faster then 16 ms to remove and add them again. So there is only one redraw of the screen.

With this render loop, all drawing operations of the last 16 ms are collected and drawn together. This should be faster then to draw each change by its own.

@pauldendulk Regarding animations: with this loop, all types of animations are checked each 16 ms if they need an update. If yes, the screen is redrawn. If not, then it is checked, if the rest needs a redraw. So animations would work perfect.

If we not use a render loop, we have to collect updates for a certain time and then redraw the screen/map, to update it. This redraw should normally happen all 16 ms. Longer times lead to a jumping map, shorter times are not visible. If something is drawn more often then all 16 ms, then you waste to much CPU time, because no one could see the changes.

@tmijieux Perhaps there is something similar in Forms, but not in all other implementations of Mapsui. So we have to create our own.

inforithmics:
Removing and adding Layers improves this patch

https://github.com/Mapsui/Mapsui/pull/1185

inforithmics:
When no render thread is used the waiting could be done with Task.Delay(16). This would introduce asynchronous Processing into Mapsui.

janusw:
Sorry to chime in here so late. I also just tried the branch, and realized I should have done this much earlier. It does in fact improve the performance a lot, in particular for the very bad extreme-cases that I was seeing earlier.

> So If nobody is oposed to it I would like to approve this patch.

Not at all, I fully agree. This is a great improvement. Thanks, @charlenni!

janusw (approved):

janusw:
I think this pretty much fixes #1084 for me.

charlenni:
> I would suggest a solution where all animations follow the design of the navigation animations.

@pauldendulk I looked again into your code of the navigation animation. It isn't good regarding the performance. You call many times a Viewport.Changed event. If you fly from one point to another you change zoom and center at the same time. But you redraw the map once for zoom and once for center. Better to calculate all things at one time and then redraw the map. If you have one hundred animations (for pins and so) you redraw the map one hundred times. If you update all animations in one point and then redraw the map, it would be much faster. So a central render loop would be much better.

janusw:
@charlenni It seems this branch has conflicts with another recent commit (also from you, I think). Could you please fix those conflicts (e.g. by rebasing to current master)?

charlenni:
I could do this, but it takes some time.

charlenni:
This was only a demo for what could be made to speed up Mapsui. If you look into it, it improves only the Forms part. Perhaps it would be better to think of a solution, which works on all platforms. Should be possible. At the end Mapsui should run on all platforms that are supported by SkiaSharp.

janusw:
> I could do this, but it takes some time.

I can also do it, if you are ok with it. (I hope the conflict should not be too hard to fix.)

charlenni:
No problem.

The CPU/GPU selection part introduces two OnPaintSurface implementations instead of one as used here. So this changes should move into the common part of the CPU/GPU OnPaintSurface functions.

charlenni:
The next step, if there is a render loop, would be to change the animation objects. Then you could add and remove animations very easy. I have this already in the VTMap demo. So we could copy this. The AnimationTimer.cs can then be deleted. And a refactoring of Navigator.cs has to be done.

janusw:
> This was only a demo for what could be made to speed up Mapsui. If you look into it, it improves only the Forms part. 

Yes, I have noticed that.

> Perhaps it would be better to think of a solution, which works on all platforms. Should be possible.

I was hoping that your modifications to the MapControl in the Forms part could be applied analogously to the MapControls on other platforms as well. If that's the case, I could try to look into it (I don't know the codebase as well as you do, though).

In any case, I'll try to get the rebasing done first, in order to fix the conflicts.


> At the end Mapsui should run on all platforms that are supported by SkiaSharp.

Certainly. Although from my perspective the Forms part is by far the most important. Essentially it covers all platforms that are supported by Mapsui (Android, iOS, Mac, UWP), only WPF is missing AFAICS.

From that point of view, things like Mapsui.UI.Android, Mapsui.UI.iOS and Mapsui.UI.Uwp can be considered as "legacy code", since they could be completely replaced by Mapsui.UI.Forms. Right?

janusw:
> The next step, if there is a render loop, would be to change the animation objects.

Agreed, but I think this should be done in a separate PR as a follow up. This PR already gives great performance improvements for many uses cases that do not involve animations at all.

Cleaning up the animations implementations certainly makes a lot of sense, but it's a separate topic IMHO.

I'd really like to get the performance improvements into a (pre-)release very soon, and I'm absolutely sure a huge number of users will profit from this.

janusw:
Alright I rebased to current master and fixed the conflicts. However, I could not update the PR branch (since it's in your private fork), so I have opened a new PR with the rebased branch: #1187

pauldendulk:
> From that point of view, things like Mapsui.UI.Android, Mapsui.UI.iOS and Mapsui.UI.Uwp can be considered as "legacy code", since they could be completely replaced by Mapsui.UI.Forms. Right?

Perhaps, but it is my impression that WPF is still used a lot. And more importantly Xamarin.Forms will also be legacy at some point. MAUI's release is planned for November https://github-wiki-see.page/m/dotnet/maui/wiki/Roadmap. So, I think we should strife for mostly platform independent code.

charlenni:
@pauldendulk 👍

pauldendulk:
@inforithmics Thanks for the performance screenshot. The screenshot shows the number of times the VisibleFeatureIterator is called. In the second screenshot the renderer is is called every 16ms, which should be sufficient, yet is takes only a fraction of the iterations of the current implementation, so we can conclude that the current renderer is doing more iterations than needed. So, there is room for improvement. However, the number of render iterations are not directly related to performance. We could set the render loop to once every 32ms, that would half the number of VisibleFeatureIterator calls, but that is not a measure of performance. To make decisions here we should also look at how the user experience.

inforithmics:
I think here the Developer using Mapsui should be able to decide how many FPS he wants to support. So the FPS should be a property that can be set externally. So for static maps it could be low and for animations it could be higher. But I think this has to be decided by the Developer because he nows the Usecase the best. 
As an initial value 60 fps is okay for me.

inforithmics:
This patch https://github.com/Mapsui/Mapsui/pull/1185
reduces the render Iterations from 182 to 52 a nice improvement.

inforithmics:
This patch has already an adaption in Framerate if it can't reach the wanted fps it starts skipping frames. So during high load it avoids refreshing to often.

  // When the fps is to low, reduce the load by skipping the frame
            if (fps < _fpsWanted / 2)
                return;

So this would make 32ms Render Time when updating a lot.

janusw:
> > From that point of view, things like Mapsui.UI.Android, Mapsui.UI.iOS and Mapsui.UI.Uwp can be considered as "legacy code", since they could be completely replaced by Mapsui.UI.Forms. Right?
> 
> Perhaps, but it is my impression that WPF is still used a lot.

Yes, I was *not* talking about WPF. It's not covered by Xamarin.Forms :)

> And more importantly Xamarin.Forms will also be legacy at some point. MAUI's release is planned for November https://github-wiki-see.page/m/dotnet/maui/wiki/Roadmap.

Also true. We certainly need a Mapsui.UI.MAUI rather soon (but probably after 3.0).

> So, I think we should strife for mostly platform independent code.

Definitely!

janusw:
> I think here the Developer using Mapsui should be able to decide how many FPS he wants to support. So the FPS should be a property that can be set externally. So for static maps it could be low and for animations it could be higher. But I think this has to be decided by the Developer because he nows the Usecase the best.
> As an initial value 60 fps is okay for me.

Sounds reasonable. But we should move this discussion to #1187. This one is outdated and should be closed, I think.

charlenni:
> @inforithmics Thanks for the performance screenshot. The screenshot shows the number of times the VisibleFeatureIterator is called. In the second screenshot the renderer is is called every 16ms, which should be sufficient, yet is takes only a fraction of the iterations of the current implementation, so we can conclude that the current renderer is doing more iterations than needed. So, there is room for improvement. However, the number of render iterations are not directly related to performance. We could set the render loop to once every 32ms, that would half the number of VisibleFeatureIterator calls, but that is not a measure of performance. To make decisions here we should also look at the user experience.

The render loop has nothing to do with the performance of drawing the map. The only job this Loop does, is to collect screen update requests from Mapsui. You could do this by other methods, but I could not think of one, that needs so less effort and changes to the code. You get this by few lines of code. You could see this by the performance checks from @inforithmics: we go from 4000 redraws of the screen to 182 and have no loose of quality. It doesn't decide, which update is needed or not. It checks only all 16 ms, if the screen should updated or not. More often isn't needed, because the user don't see it. If the load is high, then drop a update and try it later again. This load could be a screen redraw or anything else.

pauldendulk:
@charlenni Yes, I understand that it is not a performance improvement. That is what I am trying to explain. The performance tests show valuable information, it is important that there are 20 x less redraws but that does not mean a 20 x performance improvement.

charlenni:
Did I said, that this is a performance improvement for drawing the map? I don't think so. I said, that the map is more reactive. And I assume, that this is the case.

BTW. On iOS it is a performance improvement, because iOS seems to save all requests for a redraw and then redraws the next 20 s the Same map again and again. It is impossible to use the map in this time.

And it is good to have such a loop for animations, so you don't have to create for each animation a new timer, that ask for a redraw of the map. It is than one timer, that updates all animations and then redraw the map if needed.

pauldendulk:
@charlenni My remark was related to the profiler screenshot and how it should be interpreted. I was not trying to say that anyone was wrong. I was saying that we should distinguish two kind of performance. This makes a difference in how we should look at the numbers. When doing regular performance it is just the the faster the better, that is simple. In this case it's about responsiveness of what could be called 'perceived performance'. The numbers don't tell everything. Less calls to VisibleFeatureIterator are not simply better. We need to look at the app and how it is behaves when used. We should thing of other statistics to make descisions. For instance a simple FPS counter (which I used to have at the start of this project) is currently missing.

Further I think responsiveness is more important than absolute performance. This was in fact one of the main goals when I first started Mapsui, and what I had in mind when I created this page many years ago. https://mapsui.com/documentation/async-fetching.html.

Issue #1093, opened 2021-03-08
Callouts missing on iOS & MacOS


janusw:
Recently I observe that callouts are not being drawn on iOS any more. I cannot pinpoint it to a specific Mapsui commit, but it seems to be related to the update of Visual Studio for Mac to version 8.9 that I did a few days ago (which also included updates of Mono and Xamarin.iOS etc).

To reproduce, build Mapsui.Samples.Forms.iOS and run it. Select the "Add Pin Sample" test, and tap on the map repeatedly. You will see pins appearing. On Android they still have callouts, but on iOS the callouts are missing now. This was still working recently.

(Also note that on iOS the Bing map does not show properly, but I think that's a separate issue. IIRC I have observed this already some time ago.)

Can anyone reproduce the problem with the missing callouts? Any hints are appreciated why they are gone. This is not necessarily a Mapsui bug, but might also be a bug in Xamarin.

charlenni:
I assume, that it is a problem with RichTextKit (see [here](https://github.com/toptensoftware/RichTextKit/issues/32)).

charlenni:
I assume, the Font object of Xamarin.Forms changed. You could experiment with this by setting TitelFont and SubtitleFont of callout.

janusw:
> I assume, that it is a problem with RichTextKit (see [here](https://github.com/toptensoftware/RichTextKit/issues/32)).

Thanks for the hint. How did you get the error mentioned there?

Also, are you actually able to reproduce the problem as I described it?

One thing that I just tried is to update the RichTextKit references in the Mapsui project from 0.3.134 to the current stable version 0.4.138. But that did not make any difference apparently.

janusw:
> (Also note that on iOS the Bing map does not show properly, but I think that's a separate issue. IIRC I have observed this already some time ago.)

I opened #1094 for this.

charlenni:
As always: if Mapsui is running smooth, but does have problems to render some part, set a breakpoint [here](https://github.com/Mapsui/Mapsui/blob/fca6076ddb0dbdfee740bab883db97307fa2f276/Mapsui.Rendering.Skia/MapRenderer.cs#L118). This line catches all errors that are thrown while rendering the map.

Xamarin.Core.Font.cs hasn't changed since 2019. RichTextKit has mostly changed the documentation since 0.3.134. So I would say, something happened in Xamarin.iOS.

And yes, an update doesn't solve the problem. It was the first thing I did after knowing what was going on.

[...]

PR #1095, opened 2021-03-09
Mapsui.Samples.Forms: write log messages to the debug console

janusw:
* helps to debug problems like #1093


pauldendulk (approved):
ah yes, this should be on all platforms.

PR #1097, opened 2021-03-11
Mapsui.Samples.Forms.iOS: allow http URLs

### LINKED ISSUE #1094: Bing map does not work on iOS
The Bing map does not show up on iOS. This can be seen in the BingSample or the PinSample. Both just show a black background. I see the following error in the console output:

```
Mapsui.Samples.Forms.iOS[2938:1486955]
App Transport Security has blocked a cleartext HTTP (http://) resource load since it is insecure.
Temporary exceptions can be configured via your app's Info.plist file.
```

So apparently the Bing map is accessed via a http URL, and there could be two ways to fix the problem:
* Use a https URL if possible.
* Add an exception via Info.plist.

janusw:
* thus making the Bing maps work
  (in BingSample and PinSample)
* this fixes #1094


pauldendulk (approved):
A fix is expected of the Bing KnownTileSource but i assume this could also help for other http layers.

PR #1098, opened 2021-03-12
Added a string based bitmap type for SVG

charlenni:
Added a string based bitmap type for SVG images.
Added a cache for Mapsui.Forms Pin type to save some memory, if always the same SVG or Pin color is used.

### DIFF
--- a/Mapsui.Rendering.Skia/BitmapHelper.cs
+++ b/Mapsui.Rendering.Skia/BitmapHelper.cs
@@ -10,13 +10,13 @@ namespace Mapsui.Rendering.Skia
 {
     public static class BitmapHelper
     {
-        public static BitmapInfo LoadBitmap(object bitmapData)
+        public static BitmapInfo LoadBitmap(object bitmapStream)
         {
             // todo: Our BitmapRegistry stores not only bitmaps. Perhaps we should store a class in it
             // which has all information. So we should have a SymbolImageRegistry in which we store a
             // SymbolImage. Which holds the type, data and other parameters.
 
-            if (bitmapData is string str)
+            if (bitmapStream is string str)
             {
                 if (str.ToLower().Contains("<svg"))
                 {
@@ -27,7 +27,7 @@ namespace Mapsui.Rendering.Skia
                 }
             }
 
-            if (bitmapData is Stream stream)
+            if (bitmapStream is Stream stream)
             {
                 if (stream.IsSvg())
                 {
@@ -41,7 +41,7 @@ namespace Mapsui.Rendering.Skia
                 return new BitmapInfo {Bitmap = image};
             }
 
-            if (bitmapData is Sprite sprite)
+            if (bitmapStream is Sprite sprite)
             {
                 return new BitmapInfo {Sprite = sprite};
             }
--- a/Mapsui.Rendering.Skia/BitmapHelper.cs
+++ b/Mapsui.Rendering.Skia/BitmapHelper.cs
@@ -10,13 +10,13 @@ namespace Mapsui.Rendering.Skia
 {
     public static class BitmapHelper
     {
-        public static BitmapInfo LoadBitmap(object bitmapStream)
+        public static BitmapInfo LoadBitmap(object bitmapData)
         {
             // todo: Our BitmapRegistry stores not only bitmaps. Perhaps we should store a class in it
             // which has all information. So we should have a SymbolImageRegistry in which we store a
             // SymbolImage. Which holds the type, data and other parameters.
 
-            if (bitmapStream is string str)
+            if (bitmapData is string str)
             {
                 if (str.ToLower().Contains("<svg"))
                 {
@@ -27,7 +27,7 @@ namespace Mapsui.Rendering.Skia
                 }
             }
 
-            if (bitmapStream is Stream stream)
+            if (bitmapData is Stream stream)
             {
                 if (stream.IsSvg())
                 {
@@ -41,7 +41,7 @@ namespace Mapsui.Rendering.Skia
                 return new BitmapInfo {Bitmap = image};
             }
 
-            if (bitmapStream is Sprite sprite)
+            if (bitmapData is Sprite sprite)
             {
                 return new BitmapInfo {Sprite = sprite};
             }
--- a/Mapsui.Rendering.Skia/BitmapHelper.cs
+++ b/Mapsui.Rendering.Skia/BitmapHelper.cs
@@ -15,6 +15,18 @@ namespace Mapsui.Rendering.Skia
             // todo: Our BitmapRegistry stores not only bitmaps. Perhaps we should store a class in it
             // which has all information. So we should have a SymbolImageRegistry in which we store a
             // SymbolImage. Which holds the type, data and other parameters.
+
+            if (bitmapStream is string str)
+            {
+                if (str.ToLower().Contains("<svg"))
+                {
+                    var svg = new SKSvg();
+                    svg.FromSvg(str);
+
+                    return new BitmapInfo { Svg = svg };
+                }
+            }
+
             if (bitmapStream is Stream stream)
             {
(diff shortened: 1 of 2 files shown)


janusw:
The title sounds a bit odd, but after quickly glancing over your changes, I'd say I like what you're doing 😄

Mapsui.UI.Forms/Objects/Pin.cs
```diff
@@ -440,28 +444,45 @@ private void CreateFeature()
                 {
                     // There is already a registered bitmap, so delete it
                     BitmapRegistry.Instance.Unregister(_bitmapId);
+                    if (!string.IsNullOrEmpty(_bitmapIdKey))
```
tmijieux, line 447: unregistering a bitmapId is a bug, if multiple pins are already using the Id, that will leave all the other pins id hanging..., if the possibility of unregistering bitmap from the bitmap registry is absolutely needed here we need some kind of reference counting

tmijieux (changes requested):

Mapsui.UI.Forms/Objects/Pin.cs
charlenni: You are right. We should leave the bitmap in memory.

Mapsui.Rendering.Skia/BitmapHelper.cs
```diff
@@ -15,6 +15,18 @@ public static BitmapInfo LoadBitmap(object bitmapStream)
             // todo: Our BitmapRegistry stores not only bitmaps. Perhaps we should store a class in it
             // which has all information. So we should have a SymbolImageRegistry in which we store a
             // SymbolImage. Which holds the type, data and other parameters.
+
+            if (bitmapStream is string str)
+            {
+                if (str.ToLower().Contains("<svg"))
```
janusw, line 21: As long as you only store svg data as string in the BitmapRegistry, this extra check is not really necessary, is it? From a performance perspective, it could be harmful ...

tmijieux: i'd like something like `TrimStart().Substr(4).ToLower() == "<svg"` but there is the xml header problem

janusw: My point is: Svgs are the only thing we store in BitmapRegistry as string right now. When they were stored as stream, we needed to check the content to distinguish them from other stream data (actual bitmaps). Now that should not be necessary any more, since the type of the stored object (string vs stream) is sufficient to make the distinction. Correct?

Mapsui.Rendering.Skia/BitmapHelper.cs
```diff
@@ -10,13 +10,13 @@ namespace Mapsui.Rendering.Skia
 {
     public static class BitmapHelper
     {
-        public static BitmapInfo LoadBitmap(object bitmapStream)
+        public static BitmapInfo LoadBitmap(object bitmapData)
```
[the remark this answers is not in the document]
janusw: And of course: `BitmapRegistry` -> `ImageRegistry`. I'm not saying all this renaming is absolutely necessary, but it might make things clearer and avoid confusion.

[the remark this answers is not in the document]
charlenni: All this renaming breaks the existing code only for some smaller confusion.

Mapsui.Rendering.Skia/BitmapHelper.cs
charlenni: To be correct: Svgs are the only thing we store **up to now** in BitmapRegistry as string. In the early times, there were only byte[] as type for bitmaps. Than other types arrive and additional checking was inserted. Now you say, that a string could only be a SVG image. But what, if someone diceide to add Base64 based images? They are represented as strings too.

Mapsui.Rendering.Skia/BitmapHelper.cs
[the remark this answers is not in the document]
janusw: Well, it doesn't "break" anything. The `BitmapHelper` class should never show up in any user code, right?

Having a messed-up naming scheme in your private toy project is your own business, but for an open source project like Mapsui (which currently counts 37 contributors) a certain amount of code hygiene is kind of important.

Note that there is already a todo note in the `BitmapHelper` class about this topic ("Our BitmapRegistry stores not only bitmaps ..."). This really needs to be cleaned up at some point instead of adding to the confusion.

Mapsui.Rendering.Skia/BitmapHelper.cs
janusw: Sure, once any other string-based data is added, the check is needed again. As long as this is not the case, it could be commented out. Is your Base64 example something that you expect to come soon, or is it more of a hypothetical possibility?

Mapsui.Rendering.Skia/BitmapHelper.cs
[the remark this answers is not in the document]
charlenni: That's not correct. BitmapHelper is only hide by Mapsui.Forms. If you work direct with Features, you need to register bitmaps.

But ok, a renaming isn't a real problem for changes.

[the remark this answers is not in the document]
janusw: Anyway, all this renaming does not need to be done here (can be handled in a follow-up issue).

Mapsui.Rendering.Skia/BitmapHelper.cs
charlenni: No, I wouldn’t implement this in the moment. But do we remember this next time?

Mapsui.Rendering.Skia/BitmapHelper.cs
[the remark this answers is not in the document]
charlenni: Ok, then I will revert this. But be sure, I will never do this again. I don’t create a PR for renaming a variable.

Btw. this is one of the reasons why the development of Mapsui is stagnating. If you shouldn’t change anything on the API and each PR should have only one or two lines, then there isn’t much more you can do beside removing bugs.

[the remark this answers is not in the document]
janusw: > Ok, then I will revert this.

Revert what? `bitmapData` is still better than `bitmapStream`, so the renaming you've done already can certainly stay.

> But be sure, I will never do this again. I don’t create a PR for renaming a variable.

I think I will do exactly this: Open a PR for renaming all the things I listed above. You might say "it's just a name", but if there are a lot of wrong or misleading names, people make wrong assumptions when reading (and extending) the code, and that will lead to bugs. Again: I don't expect you to do this, but I think it's worth to fix it.


> Btw. this is one of the reasons why the development of Mapsui is stagnating. If you shouldn’t change anything on the API and each PR should have only one or two lines, then there isn’t much more you can do beside removing bugs.

I really don't get your point here:

- You were the one who complained that the renaming would break the API, and now complain that one cannot make any progress without breaking the API (that would have been my part 😉). What do you want: Make progress or keep the status quo?
- Who said that each PR should only have one or two lines? You created a PR that changes about 50 lines, I proposed even more changes. Where's the problem?

Mapsui.Rendering.Skia/BitmapHelper.cs
janusw: > No, I wouldn’t implement this in the moment. But do we remember this next time?

I would just leave the commented-out check in the code (maybe with a comment), so that the next person who touches the `BitmapHelper` class has a chance to see that he might need to activate this check.

Anyway, I'm also fine with keeping the check, if you insist that it's necessary.

pauldendulk (approved):
I'll merge this and see if I can make some more changes. Perhaps the BitmapRegistry should have BitmapInfo objects as content. And since we are now still in alpha we can also afford some breaking changes by renames.

PR #1099, opened 2021-03-15
Update Svg2Xaml

Liklainy:
After updating `Svg2Xaml` there were two failing tests
* `RenderBitmapAtlas` - which has a tiger svg image
  It failes because in new version `ViewBox` support was enabled, and this svg has this `viewBox` attribute which was previously ignored. As result the tiger in generated images is slightly moved.

0.3.0.5             |  0.3.0.7           |  diff
:---------------:|:---------------:|:------------------:
bitmap_atlas_generated_5 | bitmap_atlas_generated_7 | bitmap_atlas_diff

* `RenderPointWithSvgSymbols`

0.3.0.5             |  0.3.0.7           |  diff
:---------------:|:---------------:|:------------------:
points_with_svgsymbolstyle_generated_5 | points_with_svgsymbolstyle_generated_7 | points_with_svgsymbolstyle_diff

Changed Origin images to pass tests


pauldendulk (approved):
Great! and thanks taking care of the tests as well. This is how it works. Updates of rendering libs often have a slight change in rendering. If the images seem the same and the diff can be explained you can just update the reference images.

Issue #1100, opened 2021-03-16
Circle ignoring StrokeColor and StrokeWidth?


eltomek:
The following code generates transparent circles but the `StrokeColor` and `StrokeWidth` properties seem to be ignored:
```
Circle circle = new Circle
{
    Center = waypointNode.Value.Position,
    Radius = Distance.FromKilometers(1),
    Quality = 60,
    FillColor = Xamarin.Forms.Color.Transparent,
    StrokeColor = Xamarin.Forms.Color.Yellow,
    StrokeWidth = 10
};
mapView.Drawables.Add(circle);
```
![image](https://user-images.githubusercontent.com/3283040/111368052-ae73b600-8695-11eb-92a9-ad07d998c94f.png)

No matter what `StrokeColor` I assign (even the `Xamarin.Forms.Color.Transparent`) and `StrokeWidth` value the circle is always drawn with 1px thick gray color. `FillColor` works ok.

charlenni:
This was a bug. Should be solved by PR #1101.

pauldendulk:
#1101 Was merged.

PR #1101, opened 2021-03-16
Removed a bug where Circles couldn't have different colors and widths

charlenni:
Removed a bug where Circles couldn't have different colors and widths for the outline. See issue #1100.

LineStrings use Line as Pen for the line, Polygons use Outline as Pen for the line. Why? Don't ask me. So Circles have to use Outline in Style instead of Line.

Checked it in Polygons, where it is right.


pauldendulk (approved):
Clear, thanks

PR #1113, opened 2021-04-10
Benchmarking

tmijieux:
this is a first draft proposition for benchmarking, i did at first for skia renderers and then added similar to xaml renderers but after thinking about it i think this will not measure the actual rendering time for the xaml renderers, but rather just the xaml elements construction time ... the rendering is probably done by the framework in the ui loop.

I added a simple UI to visualize the capabilities of these code on the Xamarin.Forms samples project (phone idiom only)

There is also some modifications to some projects/configuration, some were intentional, others were just side effects of me changing stuff on visual studio in order to build and debug, ofc most can be revert is they raise problems.


tmijieux:
i do not really understand why but although all tests passed on test runner i still have the RenderLabels (xaml) test that does not work on my local machine

charlenni:
First, thank you for doing this.

For me all tests work correct. No problem found.

- When running the samples on UWP, there isn't any menu for choosing benchmarking.
- Perhaps it is possible to set a fixed length for times in the list. Would be much more readable.
- There are more changes than we need for benchmarking. Makes the PR not very easy to read.
- I'm not sure, if the times are correct. Normally the drawing of hyperlinks need around 1/3 ms, but sometimes need 33 ms.

tmijieux:
>     * When running the samples on UWP, there isn't any menu for choosing benchmarking.
yeah as i said for now I only did for "phone idiom". uwp probably have "Desktop idiom" (unless you have a windows phone?)
I will add views on sample for other xf idioms and try to add for other framework( wpf, ios, android, regular uwp) as well.

>     * Perhaps it is possible to set a fixed length for times in the list. Would be much more readable.
ok what fixed length do you think is the best? i think we need to keep a large number of digits to have details for small timings.

> 
>     * I'm not sure, if the times are correct. Normally the drawing of hyperlinks need around 1/3 ms, but sometimes need 33 ms.

reasons why individual timings have a high variability could be of many:
- "cold start" code is not jitted, cache not populated, ...
- cache miss
- thread preempted by operating system / scheduled on another core
- managed thread interruped by runtime for garbage collection
- or eventually it could be a real performance problems that does not happens everytime for some reason

charlenni:
I think, 3 digits should be enough. The problem is, that sometimes only 2 digits are shown.

Another point is, that it would be nice to have a timestamp for when the rendering is started. Would be helpful to see, how often the renderer is called at which time.

tmijieux:
I added a navigation view for forms uwp, to allow navigation to the benchmark page, since the timings were often very small compared to a millisecond on my pc, i put 5 digits for now

charlenni:
Thank you again. Not it is usable on UWP. Not my favorite platform, but the fastest one for testing.

The times are no fixed numbers. More digits, but not always the same amount. And I'm missing a ticks entry to see, how much time is between each call of the drawing routine. Perhaps it is possible to add one.

I assume, that there are to many features counted. If you use the "Add pin sample" you see 36 features, but only 18 are drawn. That's because the 18 features drawn with layer style and again counted in the features loop, even if they have now feature style.

So the questions is: what should features count? Drawn features? Or checked features?

The drawing of tiles seams to be cached in some way. Each first rendering it takes some more time (around 10x) as the next rendering. Even if you move around it is fast. But if you go to the benchmarking path and come back, then it is slow again. Perhaps a GC problem.

pauldendulk:
Thanks for the PR. I am sorry I have not found time for it yet.

inforithmics:
I had a look at your pull Request and in my opinion it is great. That you Implemented it on Mobile is great too, because on Desktop you can use a Profiling tool. But on Mobile Devices there is no good Profiler I now of. And often the problems occour on Mobile Devices.

It would be great if you could update your pull request, with the latest changes of Mapsui.

pauldendulk:
The part  I didn't like is that it needs changes to the core logic in order to measure the performance. My understanding is that this is mainly to test performance during application use. Nevertheless that may be necessary.

The kind of performance test that I would like to have is one where only the speed with which datasources are rendered to the skia canvas are measured. Using BenchmarkDotNet they could be run like unit tests.

charlenni:
I thought many times about this, but didn't get, how to do this. You need a canvas for this and then call the rendering functions of a special part of the map. It isn't very easy to do.

pauldendulk:
@charlenni We have rendering unit tests where we write in memory geometries to a bitmap using skia. The tests are meant to check for unwanted changes in the rendering. Such tests could be used for performance I assume. Or do you see a problem there? It will not tests every part of the system, but for the renderer part it will have quite clear results.

inforithmics:
BenchmarkDotNet would be the Best solution, because than you could compare Performance Checkins. How much they improve Performance or other checkins if they regress it.
Because Profilers add a certain overhead that, skew the results.

charlenni:
Yes, you are right, but now there are some changes, that are very deeply. Think about the benchmarking of the render loop: is it faster to use a render loop or the old approach. I assume, that this is a little bit trickier. All examples that I see for benchmarking are about checking, if one or another function is faster, but not the whole app.

charlenni:
@pauldendulk, @inforithmics: Did you find any guid to use BenchmarkDotNet with Xamarin.Forms? All I found was about using it with console apps.

inforithmics:
I made a quick google search and found this.

https://github.com/jonathanpeppers/Xamarin.Forms.Benchmarks

I didn't look at it though,

pauldendulk:
> Yes, you are right, but now there are some changes, that are very deeply. Think about the benchmarking of the render loop: is it faster to use a render loop or the old approach. I assume, that this is a little bit trickier. All examples that I see for benchmarking are about checking, if one or another function is faster, but not the whole app.

The rendering performance tests I had in mind would not measure the behavior of the render loop. So, we need other ways to measure this. Performance tests of the app as a whole will still be useful. 

Also it is not always clear what is meant with performance. The most common way is the speed with which a some amount of work is done, like rendering a full map. But for the user there is also something I call *perceived performance* . The user wants an application to respond immediately to its actions, no lagging and no speed bumps. This could involve making things asynchronous, or lowering the number of render updates.

inforithmics:
I have the same opinion. The render/display loop needs to be fast and responsive. The data processing like loading/calculating needs to be done in background processes. The best way to do this is use async/Tasks.

tmijieux:
I saw that you added the timer + performance object so i think we dont need this pull request any more ?
Even if we wanted to , i think the base code a probably changed too much for a rebase of this code to be meaningful. 
Unless you think there is still interest in this merge request, I am going to close it.

pauldendulk:
> I saw that you added the timer + performance object so i think we dont need this pull request any more ? Even if we wanted to , i think the base code a probably changed too much for a rebase of this code to be meaningful. Unless you think there is still interest in this merge request, I am going to close it.

Okay I will close it. This is also due to the fact that I neglected this PR for so long, sorry for this.

Issue #1118, opened 2021-04-23
Design to support NTS


pauldendulk:
This issue is to discuss the way in which we should start integrating NTS.

pauldendulk:
# Changes to prepare for NTS

The PR [here](https://github.com/Mapsui/Mapsui/pull/1119) is a proposal on how to proceed with NTS. Some ideas behind this PR:
- It should, at least for a while, be possible to support both Mapsui.Geometries and NTS. Also there will be other types of things which will need specific renderers which are not geometrie, like Rasters, VectorTiles. So we need some abstraction that allows for those different types.
- To allow flexibility I have chosen for a solution where the Geometry was removed from the IFeature and an IGeometryFeature was introduced with a Geometry field. The core classes use the IFeature while specific layers use IGeometry. Later an INtsFeature could be introduced with it's own layers and renderers. 
- Where to make the cut? In classes where IGeometryFeature is used we can have type save Geometries, that's nice. But since the core classes use IFeature we need to do some casting somewhere. In some situations this can be prevented using generics, but not everywhere. The currently version compiles and seems to function correctly (after quickly clicking to through all the samples) but there may be better ways to make the cut. More thought will help.

## Some things that have changed
- Removed Geometry from IFeature
- Add IGeometryFeature with a Geometry
- The IProvider now has a generic argument ```public interface IProvider <out T> where T : IFeature```
- The ILayer has no generic argument. Not sure if it should. 

## More changes to come
- **MPoint and MRect for all core classes**: Currently the Mapsui.Geometries Point and BoundingBox are used in many many places. Often this does not relate to geometries at all, like in the Widgets. In other cases it is an extent of a geometry but not a geometry itself. Instead of replacing them with NTS Point and BoundingBox I propose to replace them with two new Mapsui specific classes. I was thinking of calling them **MPoint** and **MRect**. A first step would be to copy the current Point and BoundingBox and replace them. We will need to add extension method converters both ways (ToMPoint, ToMRect and ToGeometry).
- We will need to split the Mapsui project into two parts:
  - The Mapsui core which has no dependency on Mapsui.Geometry or NTS, and instead uses MPoint and MRect
  - The other (have no name for it yet) that depends on Mapsui.Geometries which will hold most of the current layers. 
  - And perhaps the TileLayer could go into it's own project.
- A new Mapsui.Nts project will have a dependency on Mapsui core only.
- Some classes that need to stay in Mapsui core (and will need to make use of MPoint and MRect).
[...]

PR #1119, opened 2021-04-23
WIP: Feature/add igeometryfeature

pauldendulk:
Big rewrite to prepare for NTS. See #1118 to discuss.


radderz (approved):
This works seems pretty straight forward, I have reviewed the code and don't see any issues for an upgrade and think removing the xaml renderer is a good idea since the skia renderer does everything it used to do anyway. I haven't had a chance to test the code itself and possibly won't for another week or two as I am struggling for time at the moment. 

Have you considered using an RTree (NTS provides a few variants, I have also built one I could contribute if you'd like to not use NTS in the library that contains the GetFeaturesInView function) for finding which features are in view rather than doing a bounding box intersect? This would slow down the initial addition of features, however it will make the searching much faster. Assuming there is sufficient features anyway. This could be turned off by default and turned on with a setting perhaps, although with few features, the cost of the RTree would be minimal so it may be worth just leaving it on.

janusw:
It's great to see some activity towards NTS. But could I maybe suggest to publish a stable release 3.0, before tackling all this major refactoring?

There are people who depend on a 'stable' version of Mapsui. The 2.0 release is pretty much a full year old by now, and the master branch has received many changes and fixes over the last year. I am currently using the 3.0.0-alpha tags as replacement for a a stable version in my projects, since 2.0.x does not have the necessary features and fixes. (And I assume there are others who do the same, or maybe some people are simply stuck with 2.0.x since they don't dare to try anything labeled 'alpha').

Considering this situation, I'd highly recommend to prepare a stable release now and approach the rocky road toward NTS after that (which could then result in a version 4.0?). A stable release once a year is not a bad idea, and it would certainly help many users to have some solid ground to step on, instead of treading through an ever-changing swamp of alpha versions ;)

@pauldendulk Do you have any thoughts on this?

pauldendulk:
@janusw Yes, that makes sense. There should be a 3.0 release without NTS or preparations for NTS. There will probably be some more fixes coming in after the release but we will have to update the NTS branch for this.

janusw:
> @janusw Yes, that makes sense.

Good to hear that you agree :)

> There should be a 3.0 release without NTS or preparations for NTS. There will probably be some more fixes coming in after the release but we will have to update the NTS branch for this.

Yeah, the 3.0 release does not need to happen  immediately, of course. But maybe one could either branch off a stable release-3.0 branch before applying the NTS-related changes, or alternatively make the NTS preparations on a devel/unstable/future branch? (Not sure where you'd like your master branch to be.)

pauldendulk:
> Yeah, the 3.0 release does not need to happen immediately, of course. But maybe one could either branch off a stable release-3.0 branch before applying the NTS-related changes, or alternatively make the NTS preparations on a devel/unstable/future branch? (Not sure where you'd like your master branch to be.)

Yes, this is how we did it in the past. Branch the release and develop in master. Do some patches from the release branch. Yet, I only want to do the really necessary work in the release branch because there is too little time to maintain branches that are far apart.

Issue #1121, opened 2021-04-29
Restoring nuget packages does not work on Mac any more


janusw:
With the latest commits on master, restoring nuget packages does not work any more in VS for Mac. I get the following errors:

```
/usr/local/share/dotnet/sdk/5.0.202/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(63,5): error NETSDK1100: Windows is required to build Windows desktop applications. [/Users/janus/github/Mapsui/Mapsui.Rendering.Xaml/Mapsui.Rendering.Xaml.csproj]
/usr/local/share/dotnet/sdk/5.0.202/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(63,5): error NETSDK1100: Windows is required to build Windows desktop applications. [/Users/janus/github/Mapsui/Mapsui.UI.Wpf/Mapsui.UI.Wpf.csproj]
/usr/local/share/dotnet/sdk/5.0.202/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(63,5): error NETSDK1100: Windows is required to build Windows desktop applications. [/Users/janus/github/Mapsui/Samples/Mapsui.Samples.Wpf.Editing/Mapsui.Samples.Wpf.Editing.csproj]
/usr/local/share/dotnet/sdk/5.0.202/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(63,5): error NETSDK1100: Windows is required to build Windows desktop applications. [/Users/janus/github/Mapsui/Samples/Mapsui.Samples.Wpf/Mapsui.Samples.Wpf.csproj]
```

I think this is happening since a99ae93a830b39f7abe0dad8f62e1a8feb2491b8, and it's preventing me from building Mapsui on MacOS.

See also https://github.com/dotnet/sdk/issues/3803 and https://github.com/dotnet/sdk/issues/3592.

Any suggestions on how to fix it?

pauldendulk:
Are you building Mapsui.sln or Mapsui.iOS.sln?

janusw:
> Are you building Mapsui.sln or Mapsui.iOS.sln?

I am building Mapsui.sln. That used to work well.

What exactly is Mapsui.iOS.sln?

janusw:
[...]

PR #1053, opened 2021-01-25
Mapsui.Samples.Forms.Mac: fix project configurations

janusw:
This PR should fix issue #1051. It particular it makes the build in Debug config work by setting the correct target architecture.

Furthermore, it enables the project for the platforms AnyCPU, x64 and x86, but disables it for ARM and iPhone.

PR #1123, opened 2021-05-02
Update Mapsui.iOS.sln

janusw:
Update Mapsui.iOS.sln to include all projects that can be built on MacOS, and rename it to Mapsui.Mac.sln (as proposed in #1121).

Also fix the configuration mappings (cf. #1053).


pauldendulk (approved):
Good improvement. These kind of contributions really help with the maintenance.

janusw:
@pauldendulk Bitrise seems to fail here. Can you tell what's wrong? (I can't access the logs.)

PR #1133, opened 2021-05-20
Update RichTextKit

janusw:
In the light of recent RichtTextKit-related issues (like #1093), it's probably not a bad idea to update this library to the latest version.

Note that the version change does not seem to have any effect on the above issue. Also there don't seem to be any incompatible changes.

### DIFF
--- a/Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj
+++ b/Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj
@@ -10,7 +10,7 @@
   <ItemGroup>
     <PackageReference Include="SkiaSharp" Version="2.80.2" />
     <PackageReference Include="Svg.Skia" Version="0.5.0" />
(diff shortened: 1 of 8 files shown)


m1dst:
I manually applied these updates and rebuilt the nuget which solved the issue I was having with the linker.  Would be great to get this change into an official pre-release nuget package.

m1dst (approved):

janusw:
> I manually applied these updates and rebuilt the nuget which solved the issue I was having with the linker.

Good to hear, thanks for the comment!


> Would be great to get this change into an official pre-release nuget package.

Absolutely. @pauldendulk Can we please merge this into master?

janusw:
I bumped the RichTextKit version once more, since another minor update was released in the meantime.

buu-huu (approved):
Makes sense

pauldendulk (approved):
Thanks for taking care of this.

PR #1146, opened 2021-06-15
Wfs Fixes: HttpClientUtil Close, Appending Query to Url

inforithmics:
Fix the Close of the HttpClient Util. So that it doesn't throw an exception


inforithmics:
Fixed the Appending of A query to an url that already is a query

pauldendulk (approved):
High quality MR, including two new unit test classes, great.

Issue #1135, opened 2021-05-22
Create cross-platform build script


janusw:
The current build script (`Scripts/buildpack.bat`) only works on Windows, but not on MacOS.

In order to be able to build from the command line and create nuget packages also on MacOS, one could either:
* add a Mac-specific build script or
* create a cross-platform build script (maybe based on [cake](https://cakebuild.net/))

Option 2 is clearly preferable IMHO, but option 1 is still better than the current state of things.

janusw:
One limitation of the build procedure on Mac is that it cannot build the Windows-specific parts (WPF and UWP), but everything else should actually be transferable.

I am starting to work on a Mac-specific build script in a first step. Once this is done, one could possibly meld the platform-specific scripts into a cake-based cross-platform script.

pauldendulk:
Yes, this needs improvements. Not sure which is the best way to proceed. I prefer a single script of course, but the main problem is not the script but the fact that we have components that can only be build on 1 of the platforms.

Not inclined to add Cake. Linux sh scripts are actually not that bad because they could also be built with WSL2 on Windows. I expect this to become a more common approach in general.

Eventually I would like to use more github actions and also release nugets through github actions.

pauldendulk:
I also consider dropping support for things that can not be build cross platform, or move them to a separate github repository.

janusw:
> I also consider dropping support for things that can not be build cross platform, or move them to a separate github repository.

Huh, that sounds problematic: UWP cannot be built on Mac, iOS cannot be build on Windows (without a Mac build server). I would not want to deprecate or remove any of those two.

Instead I think that for each platfrom one should only build the supported components.

For auto-building the nupkg one should stick to Windows (plus Mac build server), since that config basically supports all components.

pauldendulk:
Valid points, still I am thinking of a way to make it more maintainable. With your new PR we have two nuspecs. The nuspec is typically something we (meaning mostly me 🙄) forget to update. Now we have two of those. What if there is one nuspec for the shared functionality and separate nuspecs for Mac and UWP/WPF?

Also it is also possible to work without nuspecs, use the solution to decide what should be in the nuget.

janusw:
> Also it is also possible to work without nuspecs, use the solution to decide what should be in the nuget.

I was also thinking in that direction, but it might require some restructuring of the projects in the VS solution.

janusw:
[...]

PR #1156, opened 2021-07-18
add a build script for Mac

janusw:
* analogous to the Windows build script
* since the Windows-specific parts can not be built,
  we need a separate nuspec file as well
* a cleanup script is also included
* cf. #1135

### DIFF
--- a/NuSpec/Mapsui.Mac.nuspec
+++ b/NuSpec/Mapsui.Mac.nuspec
@@ -1,0 +1,86 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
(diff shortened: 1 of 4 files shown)


pauldendulk (approved):
Excellent! It would be nice to have a way to avoid the duplicate nuspec. Perhaps three separate nuspecs (mac/windows/shared). Or no nuspec (work with a specific sln for nuget release - we would still need 3 slns).

janusw:
> Excellent! It would be nice to have a way to avoid the duplicate nuspec. Perhaps three separate nuspecs (mac/windows/shared). Or no nuspec (work with a specific sln for nuget release - we would still need 3 slns).

The last option here sounds most attractive to me. We already have two slns, and I'm not sure if we'd need a third one.

PR #1170, opened 2021-08-07
fix #1144 Shapefile

### LINKED ISSUE #1144: Shapefile.GetObjectIDsInView throw Exception as database is closed
What is correct use of Shapefile .GetObjectIDsInView ?

Code extract:
```
    if (!_isOpen)
                throw (new ApplicationException("An attempt was made to read from a closed datasource"));
```
How am I supposed to open it ?

Thanks for your help

franky235:
Fix GetGeometriesInView and GetGeometryexternal


janusw (approved):
This looks good, I think. The diff seems to be slightly different from the previous PR (additional whitespace changes?), but basically it looks ok.

Issue #1034, opened 2020-12-14
How about separating packaging(nuget)?


IOL0ol1:
I don't think it's a normal practice to make an all-in-one nuget package.
Is there any special purpose in the current packaging method ?
I forked a branch and tried to separate the dll packaging, while adding support for the lower net framework(net462).
I released the [mapsui3 series of nuget packages](https://www.nuget.org/packages?q=mapsui3) for my  wpf(net47) project.
We should select the nuget package for the corresponding platform without having to install it all.

pauldendulk:
Each dll in a separate package was not the best practice when Mapsui started. Changing it will be a breaking change, but could be acceptable. We can remove all the nuspecs and will not have to update them (I frequently forget about that). So perhaps we should go for it. 

Anyone else objections? @janusw What do you think?

janusw:
@IOL0ol1 What would be the major advantage that you gain from breaking up Mapsui in several packages? What's your motivation for doing that?

IMHO your breakup into
* Mapsui
* Mapsui.Geometries
* Mapsui.Rendering.Xaml
* Mapsui.Rendering.Skia
* Mapsui.Desktop
* Mapsui.UI.Wpf

is a bit of an overkill, also it's not complete (you're missing things like Mapsui.UI.Android, Mapsui.UI.iOS, etc), right?

@pauldendulk  In my opinion Mapsui may need a bit of restructuring first, before breaking up the package. For example, many of the classes that are included in the Mapsui.Desktop VS project actually live in namespaces like `Mapsui.Providers`. That should be cleaned up first. Possibly Mapsui.Desktop can be removed completely? Looks pretty much like a collection of legacy code to me ...

janusw:
> @pauldendulk In my opinion Mapsui may need a bit of restructuring first, before breaking up the package. For example, many of the classes that are included in the Mapsui.Desktop VS project actually live in namespaces like `Mapsui.Providers`. That should be cleaned up first. Possibly Mapsui.Desktop can be removed completely? Looks pretty much like a collection of legacy code to me ...

An example of what I'm talking about here is the `WFSProvider` and related code. AFAICS its namespace suggests it should go into the Mapsui project: https://github.com/janusw/Mapsui/commits/wfs

[...]

PR #1172, opened 2021-08-11
Remove the project Mapsui.Geometries (move all code into Mapsui base project)

janusw:
This is sort of a follow-up to #1165, and a preparation for #1034.

The namespace Mapsui.Geometries is handled in a special manner right now. Most of it lives in the separate project Mapsui.Geometries (which is a dependency of the Mapsui base project), while individual classes also live in the Mapsui project (e.g. ReadOnlyPoint, or extension methods like BoundingBoxExtensions or GeometryExtensions).

This PR moves all the code from Mapsui.Geometries into the Mapsui base project, while keeping the namespace structure. This makes sense, since packaging-wise the Geometries stuff is also included in the Mapsui base package, and I see no important argument for keeping it separate. (The Mapsui base project already has numerous subfolders and sub-namespaces.)

This simplifies the project structure and makes it more consistent with the namespace and packaging structure.


janusw:
My major complaints about the current situation (before this PR) are actually:
* The `Mapsui.Geometries` namespace is torn apart: Most of it is included in the `Mapsui.Geometries` project, but some parts are in the `Mapsui` base project.
* What makes the `Geometries` namespace so special that it deserves a separate project (as opposed to all the other namespaces like `Layers`, `Projection`, `Providers`, etc)?

> Also relevant. The plan was to replace the geometries project with NTS, still part of the code will need to be moved the Mapsui core project.

I don't think the plan of moving to NTS is affected by this PR. If the geometries code is going to be replaced by NTS, this can be done, no matter if it is in a separate project or not.

In addition, this PR makes many thing simpler:
* The nuspecs become simpler, since Mapsui.Geometries.dll is gone and everthing is in Mapsui.dll.
* The project structure becomes simpler; all the project references to Mapsui.Geometries are gone.

In summary, I really see no reason why Mapsui.Geometries should be in a separate project.

pauldendulk:
@janusw In the discussion we had yesterday I said I had no objection because it would soon be replaced by NTS. But I was no looking at my NTS branch and I realize to plan was to have a gradual migration from Mapsui.Geometries to NTS. So this makes it better to keep it separated. I want to create projects which have a dependency on NTS and no dependency on Mapsui.Geometries, while some other projects have it the other way round. Would it be a problem for you if this was not merged?

charlenni:
What is the reason to have NTS and Mapsui.Geometries at the same time? If we want to go with NTS, then we should go only with NTSand not with an additionally other library.

There are good reasons for an own library (small and dedicated to Mapsui). But to support more then one?

janusw:
> Would it be a problem for you if this was not merged?

No, it's not a big deal for me. It's just 'cleanup', after all. I thought it makes sense to do it, but if you have good reasons to keep the Geometries stuff separate, then that's fine with me.

In particular if it creates conflicts with NTS stuff that you already started, then we should certainly defer all of this discussion until the move to NTS is done. Maybe the situation becomes clearer afterwards.

The only thing I really dislike conceptually is that the Mapsui.Geometries namespace is spread over the Mapsui and Mapsui.Geometries packages. That is kind of inconsistent and IMHO a clear sign that something is not quite right with the structure.

pauldendulk:
> What is the reason to have NTS and Mapsui.Geometries at the same time? If we want to go with NTS, then we should go only with NTSand not with an additionally other library.

Because it is a very big change that needs to be done in several steps. So the end goal (as I see it now) is to have Mapsui.Geometries removed entirely. Perhaps there should even be a v4 release which contains both so users have the option to migrate from Mapsui.Geometry to NTS while having a working app all the time. The next version should have Mapsui.Geometries removed.

> There are good reasons for an own library (small and dedicated to Mapsui). But to support more then one?

So in the end there will be only one.

pauldendulk:
> The only thing I really dislike conceptually is that the Mapsui.Geometries namespace is spread over the Mapsui and Mapsui.Geometries packages. That is kind of inconsistent and IMHO a clear sign that something is not quite right with the structure.

Yes and no. 

No. Mapsui is a mapping library so it is bound to have a big dependency on some form of geometry. NTS will also be used heavily. 

Yes. Actually in the design I had in mind the core Mapsui project would not have a dependency on NTS. Only the layers would have that. The core would use a few own geometric classes. [Here](https://github.com/Mapsui/Mapsui/issues/1118) I proposed an MPoint and MRect.

pauldendulk:
And indeed this PR would interfere with the work done in the branch.

PR #1175, opened 2021-08-14
Separate packaging (add Mapsui.Native package)

janusw:
... as proposed in https://github.com/Mapsui/Mapsui/issues/1034#issuecomment-894852662:

Add a separate Mapsui.Native package (as opposed to Mapsui.Forms), including:
* Mapsui.UI.Android
* Mapsui.UI.iOS
* Mapsui.UI.Uwp
* Mapsui.Ui.Wpf
* Mapsui.Rendering.Xaml


janusw:
Btw, apart from smaller packages sizes, another advantage of separate packages is that we get better usage figures (via the nuget.org download statistics). In particular this will give us a better idea of how many people use the native UI (as opposed to Mapsui.Forms).

janusw:
More advantages of this PR that have not been mentioned yet:
* The Mapsui base package can be fully built on Mac now (which is great)!
* In addition we get rid of Mapsui.Mac.nuspec, since Mapsui.nuspec can be used for building the Mapsui package on Mac.

pauldendulk (approved):
Good improvement. If only for the package size issue but I think also a general it is good to have this split off.

janusw:
Awesome. Thanks for merging! :D

PR #1176, opened 2021-08-14
WPF: Adding MouseMove handler control

franky235:
Removing temporarily MouseMove handler is convenient to draw something on the map with left clic, and probably also to reassigne left clic as requested here: https://stackoverflow.com/questions/52260870/change-the-panning-button-in-mapsui-control (seems it is not possible to remove events when you are not the owner)


pauldendulk:
The MouseMove handler is public. You could set it to null I assume. But if you do it is not possible to attach it again because MapControlMouseMove is private. Is that the problem you ran into?

If you want to draw while the map is fixed you could also use MapControl.Map.PanLock. Would that solve your problem? See the PanLockSample.

image

franky235:
Thanks Paul, you're right, PanLock does the job in my case, that's the right thing to do !
I close the PR that is not "clean code"

Issue #1151, opened 2021-06-27
Pin.Callout: anchor wrong when RotateWithMap is enabled


janusw:
When using a `Pin` with `Callout`, both of which rotate with the map (i.e. `Pin.RotateWithMap=true` and `Callout.RotateWithMap=true`), the callout anchor is placed incorrectly when the map is rotated off the default orientation.

### Reproduction

This bug can be observed in the `PinSample` if `Pin.RotateWithMap` is set to true. I have set up a branch where this is done: https://github.com/janusw/Mapsui/commits/callout_anchor_bug_repro

To reproduce the bug, build the sample app from this branch (I'm using Mapsui.Forms.Samples.Android), run it, then select "Add Pin Sample" from the category "Forms". Tap on the map to create a pin with callout. The anchor is placed on the top of the pin. Then rotate the map and see the anchor moving off the pin. It should stay fixed on the top of the pin, but it does not.

janusw:
Some screenshots to make it clearer. Default orientation:
![Screenshot_20210627-130138883](https://user-images.githubusercontent.com/484108/123542675-37959100-d74b-11eb-95cd-f3f52ff49f1f.jpg)

Rotated:
![Screenshot_20210627-130150476](https://user-images.githubusercontent.com/484108/123542677-3c5a4500-d74b-11eb-807d-426e66ad27f2.jpg)

PR #1179, opened 2021-08-16
Removed a bug where the callout has wrong anchor while rotating

charlenni:
Removed a bug where the callout didn't have the correct anchor if the map is rotated. See #1151.


janusw (approved):
Awesome, thanks a lot for this! I verified that it fully fixes the problem.

PR #1180, opened 2021-08-16
Changed the MapControl in Mapsui.UI.Forms to ContentView

charlenni:
Changed the MapControl in Mapsui.UI.Forms in that way, that you could use GPU (SKGLView) or CPU (SKCanvasView) backend. In normal cases, the GPU backend is used. If you want in special cases the CPU backend, you could set the flag `MapControl.UseGPU` before creating the MapControl object, e.g. in App.xaml.cs before `InitializeComponent()` with

            Mapsui.UI.Forms.MapControl.UseGPU = false;

That's it. It is helpful when there are problems with the GPU backend.

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -16,8 +16,10 @@ namespace Mapsui.UI.Forms
     /// <summary>
     /// Class, that uses the API of all other Mapsui MapControls
     /// </summary>
-    public partial class MapControl : SKGLView, IMapControl, IDisposable
+    public partial class MapControl :  ContentView, IMapControl, IDisposable
     {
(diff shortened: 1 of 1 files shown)


charlenni:
From time to time you get issues, where the map is black or Artefakts are shown. I assume, that this are problems with the OpenGL implementation/driver of the smartphone version. With this, it is possible to use the CPU backend as workaround.

But now I'm not sure, if this all I should to for the Forms implementation. Perhaps it would better, to insert the SKView into a Layout. With that, MapView could derive from MapControl and many of the MapView wrappers could be deleted. Then MapControl has the same functions then all other MapControls and MapView add the Pin layer and other special Forms easy-to-use things.

Any comments on this?

PR #1182, opened 2021-08-22
svg.skia from 0.4.1. to 0.5.0 in nuspec files

bertt:


janusw:
What's the idea here? The svg.skia update from 0.4.1 to 0.5 was done in #1077, and AFAIR the nuspec dependency was left to include 0.4.1 intentionally, because no API change was necessary for the update, so in principle Mapsui should still work with the old version. Or do you have any indication that it does not any more (by now)?

bertt:
When following the Mapsui for Xamarin.Forms getting started (https://mapsui.com/documentation/getting-started-xamarin-forms.html) using Mapsui.Forms 3.0.0-rc.1 , there is a runtime error (tested on UWP): 

System.IO.FileLoadException: 'Could not load file or assembly 'Svg.Skia, Version=0.5.0.0, Culture=neutral, PublicKeyToken=dafe96fe6c845a74'. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)'

With these changes this runtime error does not occur.

janusw:
> When following the Mapsui for Xamarin.Forms getting started (https://mapsui.com/documentation/getting-started-xamarin-forms.html) using Mapsui.Forms 3.0.0-rc.1 , there is a runtime error (tested on UWP):

Indeed I can reproduce this problem. Only happens when running on UWP though (not on Android).

So, yeah, sounds like a reasonable fix to me. Thanks!

janusw (approved):
Ok with me.

PR #1185, opened 2021-08-31
Added methods to make mass edit methods to LayerCollection

Sebastian1989101:
This methods can prevent lag issues when multiple layers are added/removed/modified on espacially iOS. Currently iOS makes a redraw for each add/remove which causes the app to lag heavily for multiple seconds.

### DIFF
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -92,13 +92,7 @@ namespace Mapsui.UI.Forms
             AddLayers();
 
             // Add some events to _mapControl.Map.Layers
-            _mapControl.Map.Layers.LayerAdded += HandlerLayerChanged;
-            _mapControl.Map.Layers.LayerMoved += HandlerLayerChanged;
-            _mapControl.Map.Layers.LayerRemoved += HandlerLayerChanged;
-
-            _mapControl.Map.Layers.MultipleLayersAdded += HandlerMultipleLayersChanged;
-            _mapControl.Map.Layers.MultipleLayersRemoved += HandlerMultipleLayersChanged;
-            _mapControl.Map.Layers.MultipleLayersModified += HandlerMultipleLayersModified;
+            _mapControl.Map.Layers.Changed += HandleLayersChanged;
 
             AbsoluteLayout.SetLayoutBounds(_mapControl, new Rectangle(0, 0, 1, 1));
             AbsoluteLayout.SetLayoutFlags(_mapControl, AbsoluteLayoutFlags.All);
@@ -700,37 +694,21 @@ namespace Mapsui.UI.Forms
         {
             ViewportInitialized?.Invoke(sender, e);
         }
-
-        private void HandlerLayerChanged(ILayer layer)
+        
+        private void HandleLayersChanged(object sender, LayerCollectionChangedEventArgs args)
         {
-            if (layer == MyLocationLayer || layer == _mapDrawableLayer || layer == _mapPinLayer || layer == _mapCalloutLayer)
-                return;
-
-            // Readd MapView layers, so that they always on top
-            ReAddLayers();
-        }
-
-        private void HandlerMultipleLayersChanged(IEnumerable<ILayer> layers)
-        {
-            var localLayers = layers.ToList();
-            if (localLayers.Contains(MyLocationLayer) || localLayers.Contains(_mapDrawableLayer) || localLayers.Contains(_mapPinLayer) || localLayers.Contains(_mapCalloutLayer))
-                return;
-
-            // Readd MapView layers, so that they always on top
-            ReAddLayers();
-        }
-
-        private void HandlerMultipleLayersModified(IEnumerable<ILayer> layersRemoved, IEnumerable<ILayer> layersAdded)
-        {
-            var localRemovedLayers = layersRemoved.ToList();
-            var localAddedLayers = layersAdded.ToList();
+            var localRemovedLayers = args.RemovedLayers?.ToList() ?? new List<ILayer>();
+            var localAddedLayers = args.AddedLayers?.ToList() ?? new List<ILayer>();
 
             if (localRemovedLayers.Contains(MyLocationLayer) || localRemovedLayers.Contains(_mapDrawableLayer) || localRemovedLayers.Contains(_mapPinLayer) || localRemovedLayers.Contains(_mapCalloutLayer) || 
                 localAddedLayers.Contains(MyLocationLayer) || localAddedLayers.Contains(_mapDrawableLayer) || localAddedLayers.Contains(_mapPinLayer) || localAddedLayers.Contains(_mapCalloutLayer))
                 return;
 
-            // Readd MapView layers, so that they always on top
-            ReAddLayers();
+            // Remove MapView layers
+            RemoveLayers();
+
+            // Readd them, so that they always on top
+            AddLayers();
         }
 
         private void HandlerPinsOnCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
@@ -1011,7 +989,7 @@ namespace Mapsui.UI.Forms
         private void AddLayers()
         {
             // Add MapView layers
-            _mapControl.Map.Layers.AddMultiple(new [] { _mapDrawableLayer, _mapPinLayer, _mapCalloutLayer, MyLocationLayer });
+            _mapControl.Map.Layers.Add(_mapDrawableLayer, _mapPinLayer, _mapCalloutLayer, MyLocationLayer);
         }
 
         /// <summary>
@@ -1020,19 +998,7 @@ namespace Mapsui.UI.Forms
         private void RemoveLayers()
         {
             // Remove MapView layers
-            _mapControl.Map.Layers.RemoveMultiple(new[] { MyLocationLayer, _mapCalloutLayer, _mapPinLayer, _mapDrawableLayer });
-        }
-
-        /// <summary>
-        /// Re-adds all layers that MapView uses
-        /// </summary>
-        private void ReAddLayers()
-        {
-            // Remove MapView layers
-            RemoveLayers();
-
-            // Readd them, so that they always on top
-            AddLayers();
+            _mapControl.Map.Layers.Remove(MyLocationLayer, _mapCalloutLayer, _mapPinLayer, _mapDrawableLayer);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


charlenni:
Thank you for looking into this problem. It is only needed, because there is no order in the layer collection of Mapsui. So each time a layer is added, the Forms MapView has to remove it layers and add it again.

Isn't it possible two shorten the code? There are many functions, that are double and only the selection of the lists are different.

Sebastian1989101:
@charlenni did some refactoring. The first version was just the mess I used personally in my iOS apps since a few weeks to prevent the lag as good as I could. 

About the MapView: I do not use it personally but does this not cause a memory leak by registering all the events but never unregister from them?

Mapsui/Layers/LayerCollection.cs
```diff
@@ -15,10 +15,18 @@ public class LayerCollection : IEnumerable<ILayer>
         public delegate void LayerAddedEventHandler(ILayer layer);
         public delegate void LayerMovedEventHandler(ILayer layer);
 
+        public delegate void MultipleLayersRemovedEventHandler(IEnumerable<ILayer> layers);
```
inforithmics, line 18: I think a generic Event Delegate should be used here

https://docs.microsoft.com/en-us/dotnet/api/system.eventhandler-1?view=net-5.0

so the first parameter is the sender, and an Inherited EventArgs is added, where the layersRemoved, layersAdded are Properties.

So adding Additional Information doesn't break the Interafce

inforithmics: Sorry für the misunderstanding. I ment that if you want to add  additional parameters in the future it won't break\change the Event Signature. 

Yes changing it to the regular Event Layout would bei fine with me.

Sebastian1989101: @inforithmics ah that makes sense. In thise case the other question would be if any event in the LayerCollection should change to have a uniform structure of the events. This however may break some users code if they already use the existing events. Because the events that affected single layers are already in there and only the events for multi layer edits are new.

inforithmics: I would leave the existing Events as is. If they need to change for good reason then the standart event structure should be used.

Mapsui/Layers/LayerCollection.cs
```diff
@@ -15,10 +15,18 @@ public class LayerCollection : IEnumerable<ILayer>
...
+        public delegate void MultipleLayersAddedEventHandler(IEnumerable<ILayer> layers);
+        public delegate void MultipleLayersModifiedEventHandler(IEnumerable<ILayer> layersRemoved, IEnumerable<ILayer> layersAdded);
+
         public event LayerRemovedEventHandler LayerRemoved;
         public event LayerAddedEventHandler LayerAdded;
         public event LayerMovedEventHandler LayerMoved;
 
+        public event MultipleLayersRemovedEventHandler MultipleLayersRemoved;
+        public event MultipleLayersAddedEventHandler MultipleLayersAdded;
+        public event MultipleLayersModifiedEventHandler MultipleLayersModified;
```
inforithmics, line 28: Wouldn't it suffice to have only One Event MultipleLayersModified with the properties showing what was added and what was removed.

Mapsui.UI.Forms/MapView.cs
```diff
@@ -999,10 +1020,19 @@ private void AddLayers()
...
-            _mapControl.Map.Layers.Remove(_mapDrawableLayer);
+            _mapControl.Map.Layers.RemoveMultiple(new[] { MyLocationLayer, _mapCalloutLayer, _mapPinLayer, _mapDrawableLayer });
+        }
+
+        /// <summary>
+        /// Re-adds all layers that MapView uses
+        /// </summary>
+        private void ReAddLayers()
+        {
+            // Remove MapView layers
```
inforithmics, line 1031: Shouldn't Here not the ModifyMultiple be used. So only one change is triggered instead of two.

Sebastian1989101: This is just a wrapper to re-add the layers of the Forms Map. The content of this methods was previously just in the event call. Otherwise each event would need to remove and add the layers to the forms map like it was on the HandlerLayerChanged(ILayer layer) method before in this class. So this method was created in form of the refactoring that @charlenni suggested for this change.

Mapsui/Layers/LayerCollection.cs
Sebastian1989101: Yes, I would also prefer a single event with a EventArgs implementation used. However this would be different from the already existing methods. So either this will have two different event implementaiton layouts or each action has it's own event as it is right now. I think this is mostly something about personal preference. @pauldendulk added the single events long ago, so maybe there is a reason why it was done this way?

Mapsui.UI.Forms/MapView.cs
inforithmics: So that is good for me.

inforithmics:
I made a Performance Measurement and this patch Improved the Performance from.

DisplayMapRenderLoop

to

DisplayLayerRemove

So roughly 2.5 Times (This was tested against the Render Loop Patch, For this Measurement I merged the two patches)

Mapsui/Layers/LayerCollection.cs
pauldendulk: @Sebastian1989101 @inforithmics I am not sure if I add the multiple event handlers or if it was based on some other existing component. It basically followed the customary pattern that was used back then (and maybe still is). 

 I am a big fan of reducing the amount of code. What about replacing all methods with just one? A single layer change could be sent as an array with one entry. Perhaps the arguments should also allow added and removed layers in one method. Perhaps the argument should have Added, Remove and Modified fields.

This is something that should not go in the upcoming release though, which should be due in a few weeks. Perhaps adding the methods like @Sebastian1989101 suggested is the best solution for now. It solves a actual performance problem. Some people will be working with the version for years.

A altered version could be committed shortly after the release.

Also we should take into account @charlenni's remark. A layer order should be added which should be independent from the order in the array.

pauldendulk:
@inforithmics How did you do the performance test?

inforithmics:
I normally use Jetbrains dotTrace. 
But you can although use the Visual Studio Profiler.
https://docs.microsoft.com/en-us/visualstudio/profiling/profiling-feature-tour?view=vs-2019

Mapsui/Layers/LayerCollection.cs
Sebastian1989101: @pauldendulk not the multiple but the event handlers for the single events. Thats why I have oriented the code to the same structure. 

Reducing the code is a good point. I got a suggestion from a friend on how to make all the new code work without adding new methods, which I have done before. I will make a new commit to this which will reduce this further without breaking any existing API for users of the current version of Mapsui.

Sebastian1989101:
@inforithmics @pauldendulk I merged the Sinlge/Multi methods into one method with a params parameter. This way there are no API breaks while at the same time extending the existing stuff to be capable of taking more input at once. I also changed the events by removing all Mutli-Change events and just using one with a EventArgs parameter. Within the Mapsui Komponent this event is now used on no longer the single calls while at the same time the single events still exists for the users of Mapsui who already use them.

inforithmics:
Fantastic pull request. The Code looks now much cleaner and the interface supports now changes in the Layers handling like avoiding removing and readding them.

inforithmics (approved):

Mapsui/Layers/LayerCollectionChangedEventArgs.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using System.Collections.Generic;
+
+namespace Mapsui.Layers
+{
+    public class LayerCollectionChangedEventArgs : EventArgs
```
pauldendulk, line 6: ah yes!, in line with what I suggested here https://github.com/Mapsui/Mapsui/pull/1185#discussion_r701715168

Mapsui/Layers/LayerCollection.cs
```diff
@@ -68,58 +72,117 @@ public void CopyTo(ILayer[] array, int arrayIndex)
...
+                copy.AddRange(layers);
+
+            _layers = new ConcurrentQueue<ILayer>(copy);
+            foreach (var layer in layers)
+                OnLayerAdded(layer);
+
+            OnChanged(layers, null);
+        }
+
+        public bool Remove(params ILayer[] layers)
```
pauldendulk, line 114: This replaces the existing ```public bool Remove(ILayer layer)```. There is no need to introduce breaking changes in this phase. Let's keep all the existing methods and add an extra overloads for the new methods.

pauldendulk (changes requested):

Mapsui/Layers/LayerCollection.cs
Sebastian1989101: Would this really count as breaking change? Because it still works 1:1 as before. It can still be called with a single Layer like `Remove(new Layer());`. Thats why all unit tests are still working without any change to them.

pauldendulk: ah, because of the params keyword you could also enter a single layer. Nice solution!

Mapsui.UI.Forms/MapView.cs
```diff
@@ -999,10 +998,7 @@ private void AddLayers()
         private void RemoveLayers()
         {
             // Remove MapView layers
-            _mapControl.Map.Layers.Remove(MyLocationLayer);
-            _mapControl.Map.Layers.Remove(_mapCalloutLayer);
-            _mapControl.Map.Layers.Remove(_mapPinLayer);
-            _mapControl.Map.Layers.Remove(_mapDrawableLayer);
+            _mapControl.Map.Layers.Remove(MyLocationLayer, _mapCalloutLayer, _mapPinLayer, _mapDrawableLayer);
```
pauldendulk, line 1001: And here you can add multiple layers without wrapping them into an array. Another advantage of the params keyword.

pauldendulk: The only disadvantage is that you cannot pass in a List of IEnumerable without casting it to an array. If that is the primary use case that might be a reason to go for something else.

Sebastian1989101: @pauldendulk the only way to make all kinds of enumerables work, is with a IEnumerable overload for each of those methods. I mean it would not add much code if it is prefered to give the user every option. So the question remains is, should I add those overloads?

pauldendulk: It is not clear if it is actually used that way. So, for now I would say there is no reason to add them.

pauldendulk (approved):
This is smart solution, adds needed functionality with little code.

PR #1186, opened 2021-09-02
Allow custom user agent

pauldendulk:
Why: Because the default user agents can get blocked by osm

### DIFF
--- a/Mapsui/Utilities/OpenStreetMap.cs
+++ b/Mapsui/Utilities/OpenStreetMap.cs
@@ -10,12 +10,12 @@ namespace Mapsui.Utilities
         private static readonly BruTile.Attribution OpenStreetMapAttribution = new BruTile.Attribution(
             "© OpenStreetMap contributors", "https://www.openstreetmap.org/copyright");
 
-        public static TileLayer CreateTileLayer()
+        public static TileLayer CreateTileLayer(string userAgent = DefaultUserAgent)
         {
-            return new TileLayer(CreateTileSource()) { Name = "OpenStreetMap" };
+            return new TileLayer(CreateTileSource(userAgent)) { Name = "OpenStreetMap" };
         }
 
-        private static HttpTileSource CreateTileSource(string userAgent = DefaultUserAgent)
+        private static HttpTileSource CreateTileSource(string userAgent)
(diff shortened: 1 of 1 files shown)


pauldendulk:
Context: The layer sends a user-agent at every request. The previous user-agent used in BruTile got blocked by osm at some point. This blocking probably happens because someone used the BruTile library to bulk load tiles. This is not allowed by the osm service. As a consequence every app that is using that user-agent will not receive tiles anymore. This could also happen to the default user-agent of Mapsui. So it should become common practice to specify your own user-agent if you create an app.

Mapsui/Utilities/OpenStreetMap.cs
```diff
@@ -14,12 +15,12 @@ public static TileLayer CreateTileLayer()
             return new TileLayer(CreateTileSource()) { Name = "OpenStreetMap" };
```
inforithmics, line 15: Shouldn't there be a parameter here with the userAgent.
or the Alternative would be that the DefaultUserAgent is not a constant but a Static Property that can be set from outside

Mapsui/Utilities/OpenStreetMap.cs
```diff
@@ -14,12 +15,12 @@ public static TileLayer CreateTileLayer()
...
         }
 
-        private static HttpTileSource CreateTileSource()
+        private static HttpTileSource CreateTileSource(string userAgent = DefaultUserAgent)
```
inforithmics, line 18: I don't think that this can be called externally because it is private static ..

pauldendulk: Ah, you are right. Too hurried. I pushed an update.

inforithmics (approved):

janusw:
I'm also ok with the changes here. I don't use the `OpenStreetMap` class, though, but directly rely on the BruTile functionality for OSM, where a custom user agent is already supported:
```
ILayer l = new TileLayer(KnownTileSources.Create(KnownTileSource.OpenStreetMap, userAgent: "myUA", minZoomLevel: minZL, maxZoomLevel: maxZL));
```
Feel free to merge this, Paul!

PR #1187, opened 2021-09-03
Added a render loop instead of always calling InvalidateSurface()

janusw:
The render loop is called each 16 ms and checks, if something has changed (needs redraw or an animation is running). If a redraw is already running, the frame is dropped.

This is a rebase of @charlenni's PR #1092 to current master. There were some conflicts in `Mapsui.UI.Forms/MapControl.cs` affecting:
* PaintSurface
* RefreshGraphics
* OnTimerCallback

I fixed all of those.

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -1,3 +1,4 @@
+using Mapsui.Geometries.Utilities;
 using Mapsui.Rendering;
 using Mapsui.UI.Utils;
 using SkiaSharp;
@@ -5,11 +6,12 @@ using SkiaSharp.Views.Forms;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
 using System.Linq;
-using Mapsui.Geometries.Utilities;
-using Xamarin.Forms;
+using System.Threading;
 using System.Threading.Tasks;
-using Mapsui.Utilities;
+using Xamarin.Forms;
 
 namespace Mapsui.UI.Forms
 {
@@ -60,6 +62,16 @@ namespace Mapsui.UI.Forms
         private int _numOfTaps = 0;
         private readonly FlingTracker _flingTracker = new FlingTracker();
         private Geometries.Point _previousCenter;
+        private bool _sizeChanged = false;
+
+        // Timer for rendering loop
+        System.Threading.Timer _timer;
+        // Values for drawing loop
+        readonly Stopwatch _stopWatch = new Stopwatch();
+        double _fpsAverage = 0.0;
+        const double _fpsWanted = 60.0;
+        int _fpsCount = 0;
+        object _lockObj = new object();
 
         /// <summary>
         /// Saver for angle before last pinch movement
@@ -77,6 +89,11 @@ namespace Mapsui.UI.Forms
         {
             Initialize();
         }
+
+        /// <summary>
+        /// Flag for redrawing view
+        /// </summary>
+        public bool NeedsRedraw { get; set; } = true;
 
         public float ScreenWidth => (float)Width;
 
@@ -129,13 +146,16 @@ namespace Mapsui.UI.Forms
             Map = new Map();
             BackgroundColor = Color.White;
 
+            // Create timer for redrawing
+            _timer = new System.Threading.Timer(OnTimerCallback, null, TimeSpan.FromMilliseconds(1000.0 / _fpsWanted), TimeSpan.FromMilliseconds(1000.0 / _fpsWanted));
+
             _initialized = true;
         }
 
         private void OnSizeChanged(object sender, EventArgs e)
         {
             _touches.Clear();
-            SetViewportSize();
+            _sizeChanged = true;
         }
 
         private async void OnTouch(object sender, SKTouchEventArgs e)
@@ -306,13 +326,32 @@ namespace Mapsui.UI.Forms
 
         void PaintSurface(SKCanvas canvas)
         {
-            if (PixelDensity <= 0) return;
+            if (_sizeChanged)
+            {
+                SetViewportSize();
+                _sizeChanged = false;
+            }
 
-            Navigator.UpdateAnimations();
+            bool lockTaken = false;
 
-            canvas.Scale(PixelDensity, PixelDensity);
+            try
+            {
+                Monitor.TryEnter(_lockObj, ref lockTaken);
 
-            _renderer.Render(canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
+                if (lockTaken)
+                {
+                    canvas.Scale(PixelDensity, PixelDensity);
+
+                    _renderer.Render(canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
+
+                    NeedsRedraw = false;
+                }
+            }
+            finally
+            {
+                if (lockTaken)
+                    Monitor.Exit(_lockObj);
+            }
         }
 
         private Geometries.Point GetScreenPosition(SKPoint point)
@@ -322,7 +361,7 @@ namespace Mapsui.UI.Forms
 
         public void RefreshGraphics()
         {
-            _invalidate();
+            NeedsRedraw = true;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


Mapsui.UI.Forms/MapControl.cs
```diff
@@ -60,6 +62,16 @@ public TouchEvent(long id, Geometries.Point screenPosition, long tick)
...
         private readonly FlingTracker _flingTracker = new FlingTracker();
         private Geometries.Point _previousCenter;
+        private bool _sizeChanged = false;
+
+        // Timer for rendering loop
+        System.Threading.Timer _timer;
+        // Values for drawing loop
+        readonly Stopwatch _stopWatch = new Stopwatch();
+        double _fpsAverage = 0.0;
+        const double _fpsWanted = 60.0;
```
inforithmics, line 72: Maybe this could be changed to be a property, so that the programmer can change it if he wants. For example there are Devices with more than 60 Fps Screen refresh rate. But for maps without animation this could be slower.

Mapsui.UI.Forms/MapControl.cs
```diff
@@ -794,5 +833,42 @@ private float GetPixelDensity()
...
+                return;
+
+            // Calculate an averaged fps
+            _fpsAverage += fps;
+            _fpsCount++;
+
+            if (_fpsCount == 20)
+            {
+                fps = _fpsAverage / _fpsCount;
+                Debug.WriteLine($"FPS {fps.ToString("N3", CultureInfo.InvariantCulture)}");
```
inforithmics, line 863: Good Information to see if the Rendering is struggeling.

inforithmics (approved):

inforithmics:
As said in https://github.com/Mapsui/Mapsui/pull/1092 This patch improves Responsivity a lot.  
I think that is a step into the direction of seperating the display and processing of data. The display needs to be responsive and tied to fps, the processing could be done in async Tasks in the background. I checked the animation samples and they still work. And even don't use a lot of CPU time.

Mapsui.UI.Forms/MapControl.cs
```diff
@@ -306,13 +326,32 @@ void OnPaintSurface(object sender, SKPaintSurfaceEventArgs args)
...
+            {
+                SetViewportSize();
+                _sizeChanged = false;
+            }
+
+            bool lockTaken = false;
+
+            try
+            {
+                Monitor.TryEnter(_lockObj, ref lockTaken);
```
pauldendulk, line 339: A lock is usually used to synchronize calls from different threads but the render loop should only be called from a single thread, and as far as I know it it is right now. I assume calling it from different threads will probably cause crashes.

inforithmics: I think the Timer Tick is not guaranted to be on the same thread. So if an elapsed timer tick is still in progress, it needs to guard against it. Not issuing the redraw twice.

https://stackoverflow.com/questions/1435876/do-c-sharp-timers-elapse-on-a-separate-thread

pauldendulk: There is too much to discuss here :(

pauldendulk:
There is a lot to discuss with regard to these changes. Some general things to keep in mind:
- When you call InvalidateSurface (which has different names on different platforms) you do not call the render loop explicitly but you only set a 'dirty flag'. You tell the engine that an update is need. It is up to the engine to do rendering or not. This is good pattern which is meant to prevent too many render calls. 
- If the dirty flag is false there should be no update. Mapsui is used for gps tracking on mobile devices where people sometimes spend a day walking. It is important that the battery is not drained by needless rendering.
- If there rendering takes longer than the interval the second iteration should be postponed. So we don't have stacked up render updates.
- It is probably good that have some interval in between render iteration to allow other work. However, it could be important to make sure the render thread is not locked.
- So far, we depended on the render loop logic of the platform. We should take more control of this. Yet, we can not ignore the platform dependent render loop logic. 
- It is hard (impossible) to stick to a strict render loop update cycle. There will be variations. If we use animations we can not pre-calculate the updates hoping that they will align with the render loop cycle. Animation updates need to be calculated from the render loop itself.

pauldendulk (changes requested):
No specific requests yet. This just needs more discussion. See my remark in the comments.

inforithmics (changes requested):
See comments

pauldendulk:
> See comments

I was referring to my general comments. 
https://github.com/Mapsui/Mapsui/pull/1187#issuecomment-912927660

charlenni:
Ok, I try to explain the idea behind the code.

First: This isn't a render loop. The name is wrong. It is a "collect render requests" loop. 

In the moment, each change in Mapsui leads to a RefreshGraphics and a redraw of the screen. If you do many short things, the map is drawn again and again. This shouldn't happen. So we have to collect the drawing request.

On the other side, we want to have animations for different things like fling or animated my location features. This animations should updated regularly. At best with 60 FPS. Not more.

So, why not combine this two thing. Take a timer, set it to an interval of 16 ms (1000 ms / 60) and Check in the Tick, if the screen should be updated, because an animation changed the screen or any function of Mapsui requested an update in the last 16 ms. If yes, set the dirty flag, so that the system knows, that the screen should be redrawn. Nothing else does this loop. Does it replace the system render loop? No. Does it render anything by its own? No. Could it be made in another way? Yes. We could count redraw request from Mapsui and call the screen redraw after a given amount of them (ok, bad solution).

I assume, that this is the easiest way to achieve a reduction of redraws of the screen.

What should be changed?
1. A property for FPS
2. Move the loop from Mapsui.UI.Forms to the shared code

We could discuss about the lock. I assume, that the code runs very fast and we don't need one. But let's handle it the save way.

janusw:
> I assume, that this is the easiest way to achieve a reduction of redraws of the screen.

Yes, both @inforithmics  and me confirmed that the modifications improve performance a lot (in cases where it was bad before). It looks like this is the right way to go.
 
> What should be changed?
> 
>     1. A property for FPS
> 
>     2. Move the loop from Mapsui.UI.Forms to the shared code

@charlenni Are you willing and able to take this on soon? Or should I look into it?

Btw, despite the performance improvements, I was still able to find cases where the drawing is so complex that 16ms are apparently not sufficient (on mobile devices) and the app was practically freezing. So maybe the "auto-throttling" of the frame rate does not really work yet? Or it just needs some further tuning? I'll investigate a bit more ...

janusw:
I just rebased the branch to current master (in order to take advantage of #1175 etc).

charlenni:
@janusw I could do this, if you don't need it tomorrow. This PR was for discussion, is 6 month old and at the beginning no one liked it 😉

Regarding the performance and the 16 ms: we could introduce a flag for the draw process. With this flag the loop could check, if the screen is still redrawn. If this is the case, it drops the InvalidateSurface this time, but tries again next time it is called. It could be, that we then draw the same map twice (check is at the end of the draw process).

charlenni:
@pauldendulk There are much more issues to improve the performance. This is only a first step. But I think, that this is a big step in the right direction: 
1. Decouple redrawing from changes of map data
2. Synchronize animations with drawing

janusw:
> @janusw I could do this, if you don't need it tomorrow.

I don't. But if it's not possible for you to make a move within a week or so, I'd probably try to make some progress myself. That's why I'm asking.


> This PR was for discussion, is 6 month old and at the beginning no one liked it 😉

Well, I'm definitely a bit angry at myself for not trying it earlier. Possibly I was a bit put off by the way you communicated it, and I was also constantly dealing with other bugs in Mapsui.

With most other stuff out of the way, performance issues are public enemy number one for me right now 😜

charlenni:
I thought about this. @pauldendulk likes a discussion about this change. So it would be useless to change anything, before this isn't discussed to an end.

My thoughts about this:
- MapControl is the Part of Mapsui which is responsible to display the map on the screen. It handles all input and output.
- Map contains all data about the map. But why is there a ZoomLock and PanLock? Should go to Viewport.
- Viewport is the part, which holds information about the part of the map data that should be displayed on screen and in which resolution
- Navigator is the part, that could change the viewport 

Now it is clear, that MapControl is responsible to draw the map. But when should this be done? When the map data or the viewport change or the OS says, that there should be a redraw of the control. How often do we need an update for map data or viewport change? Around each 16 ms. Should be enough. Shorter times are not visible. Could we change anything on the frequence of OS updates? No.

When we now call all 16 ms a function, what should it check? First it should check, if there was a change of the map data or the viewport. If not, the map on the screen is ok. But if it changed, the next to check is, if the map draws itself in this moment. If yes, it is useless to start a redraw. So drop it and wait to the next call of the function. If no, we could start to redraw the map by saying the OS, that the control is Invalide and should be redrawn. So the IS could decide, which is the correct moment for this.

What else should be done? The timer should be stopped, if the control isn't visible any longer. For this we need two new functions, that could handle appear and disappear of the control on the screen. This is useful to save CPU time and power.

So what do we need:
1. A flag for the drawing process in progress like "drawing" (changes handled by a lock)
2. A flag for indicating that the control needs a redraw, because something changed like "refresh"
3. A timer, that is called all 16 ms with a tick function, that could invalidate the control. Perhaps the time between the ticks should be changeable by a property.
4. A function, which could be called when the control appears on screen or a long running process is stopped
5. A function, which could be called when the control disappears from screen or a long running process is started
6. All this should be handled in the common part of MapControl, so there is no need for any special implementation

Did I missed something? Is there anything else that we have to consider? Any problems with this solution?

BTW I think, it should be possible to draw a complete map in around 16 ms. All map controls (like Google Maps, MapBox or GuruMaps) could do this. If this is not possible, we have to think about, what's going wrong with Mapsui.

Mapsui.UI.Forms/MapControl.cs
inforithmics: The idea of @charlenni https://github.com/Mapsui/Mapsui/pull/1187#issuecomment-912968407 would solve this.

inforithmics: So the lock should go to the timer Tick

janusw:
Closing this one in favor of #1192.

Issue #1188, opened 2021-09-03
Map navigates to wrong area


bertt:
Sometimes when doing a very quick sequence of mouse down/mouse move/mouse up the map navigates to a complete other area (could be countries away). I've seen this behaviour on IPhone but it's not very easy to reproduce.

Sample:

https://user-images.githubusercontent.com/538812/132014787-9a411da8-f8bc-446b-8a4a-d6a7035b35c9.MP4

Suggestion: Make it possible to disable map animation (fly to?)

charlenni:
@bertt Thank you bringing this up. But which part of Mapsui do you use? UI.iOS or UI.Forms?

Do you think, that the view thinks it is a double click?

bertt:
Hi yes Xamarin.Forms

janusw:
I have also seen behavior like this with UI.Forms occasionally.

> Do you think, that the view thinks it is a double click?

Don't think so (why would a double-click take you to a totally different area?). Maybe rather an out-of-control swiping / panning operation?

@bertt Are you using the MapView class, or just the MapControl?

bertt:
This issue is fixed in other map navigation issue by @pauldendulk , see https://github.com/Mapsui/Mapsui/issues/1183#issuecomment-912627184

I'll close this issue.

Issue #1183, opened 2021-08-27
IProvider - GetFeaturesInView not called when navigating map


bertt:
I'm drawing a grid using a layer with a custom IProvider as datasource. Sometimes when navigating around (like panning the map) the function ' GetFeaturesInView' of the IProvider is not called (or very late), resulting in the grid not being visualized.

Demo:

![GetFeaturesInView](https://user-images.githubusercontent.com/538812/131153289-6c049c64-a90d-4b58-adc8-267e309c10c6.gif)

I've made a Xamarin.Forms project to reproduce, see project 'MemoryProvider' (branch bug/getfeaturesinview_not_called/) in the [samples/Mapsui.Samples.MemoryProvider folder](https://github.com/Mapsui/Mapsui/tree/bug/getfeaturesinview_not_called/Samples/Mapsui.Samples.MemoryProvider/MemoryProvider/). For the custom IProvider see [GridMemoryProvider.cs](https://github.com/Mapsui/Mapsui/blob/bug/getfeaturesinview_not_called/Samples/Mapsui.Samples.MemoryProvider/MemoryProvider/MemoryProvider/GridMemoryProvider.cs)

So far I've tested on UWP.

pauldendulk:
I can reproduce the problem. It seems it never occurs when you drag, wait and do mouse up, and it does occur sometimes if you do the mouse up while dragging. In that case the map has some inertia and is sliding. So probably there is some problem with sliding logic. 

Is this what you see also?

bertt:
yes it happens with a quick sequence of mouse down. mouse move, mouse up

pauldendulk:
I think that the critical part is that you do the mouse up while moving the mouse.

pauldendulk:
I pushed a MapView.UseFling property (set to false) to your branch. I could not reproduce your problem with it. Perhaps you can also check if this solves your other problem.

pauldendulk:
see: https://github.com/Mapsui/Mapsui/blob/bug/getfeaturesinview_not_called/Mapsui.UI.Forms/MapControl.cs#L143

bertt:
tested, works much better :-) But it seems the MapView.Info event is now broken?

pauldendulk:
Apparently the scope was too broad, was confused by the comment, how the new push:

bertt:
yep, Info event does work again 👍

pauldendulk:
[...]

PR #1189, opened 2021-09-04
Add HasFling property to forms

pauldendulk:
Two issues have been reported which could be resolved by disabling fling with the HasFling property added in this PR.
https://github.com/Mapsui/Mapsui/issues/1183
https://github.com/Mapsui/Mapsui/issues/1188

This PR adds a property to disable fling which is a sufficient solution for the user reposting the issue. With fling enabled these issues remain however.

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -9,7 +9,6 @@ using System.Linq;
 using Mapsui.Geometries.Utilities;
 using Xamarin.Forms;
 using System.Threading.Tasks;
-using Mapsui.Utilities;
 
 namespace Mapsui.UI.Forms
 {
@@ -89,6 +88,7 @@ namespace Mapsui.UI.Forms
         public ISymbolCache SymbolCache => _renderer.SymbolCache;
 
         public bool UseDoubleTap = true;
+        public bool UseFling = true;
 
         public void Initialize()
         {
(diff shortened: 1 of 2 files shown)


charlenni:
As described in my other comment: could be solved in another way. But this will work too.

What I'm missing is the same flag in MapView. Because MapView doesn't inherit MapControl, you have to add a wrapper around this flag.

pauldendulk:
@charlenni ah, an needs to be accessible of course. That needs to be addressed.

bertt (approved):

charlenni:
I thought of this PR this night and think, that we have some problems with it.

We now solved the problem of the missing FlingEnd event by saying "don't use fling anymore", but only on Mapsui.Forms. The problem still exists for Mapsui.iOS, Mapsui.Android, Mapsui.WPF and so on. If you want to solve it this way, you have to add such a flag in all MapControls.

The other way would be to change the FlingFinal in Navigator.cs from

        private void FlingFinal(AnimationEntry entry)
        {
            // Nothing to do
        }

to

        private void FlingFinal(AnimationEntry entry)
        {
            _viewport.SetCenter(_viewport.Center.X, _viewport.Center.Y);

            Navigated?.Invoke(this, ChangeType.Discrete);
        }

This will trigger an update at the end of the fling. With this, you could use the fling correct with all MapControls.

I made a PR for this (#1198), so it would be nice, if @bertt could test it, if it solves #1183.

PR #1192, opened 2021-09-05
Invalidate loop

charlenni:
A first shoot to use an invalidate loop to redraw MapControl. For this, more things are moved into the common MapControl. The interval of the loop could be changed and the loop could be started and stopped, if the control vanishes from screen.
`RefreshGraphics()` now only set a flag.

Should replace #1187

### DIFF
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -459,6 +459,11 @@ namespace Mapsui.UI.Forms
         internal IMapControl MapControl => _mapControl;
 
         /// <summary>
+        /// Update interval for invalidation timer in ms
+        /// </summary>
+        public int UpdateInterval => _mapControl.UpdateInterval;
+
+        /// <summary>
         /// IMapControl
         /// </summary>
 
@@ -527,6 +532,29 @@ namespace Mapsui.UI.Forms
         }
 
         #endregion
+
+        /// <summary>
+        /// Start updates for control
+        /// </summary>
+        /// <remarks>
+        /// When this function is called, the control is redrawn if needed
+        /// </remarks>
+        public void StartUpdates(bool refresh)
+        {
+            _mapControl.StartUpdates(refresh);
+        }
+
+        /// <summary>
+        /// Stop updates for control
+        /// </summary>
+        /// <remarks>
+        /// When this function is called, the control stops to redraw itself, 
+        /// even if it is needed
+        /// </remarks>
+        public void StopUpdates()
+        {
+            _mapControl.StopUpdates();
+        }
 
         internal void AddCallout(Callout callout)
         {
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -461,7 +461,14 @@ namespace Mapsui.UI.Forms
         /// <summary>
         /// Update interval for invalidation timer in ms
         /// </summary>
-        public int UpdateInterval => _mapControl.UpdateInterval;
+        public int UpdateInterval
+        {
+            get => _mapControl.UpdateInterval;
+            set
+            {
+                _mapControl.UpdateInterval = value;
+            }
+        }
 
         /// <summary>
         /// IMapControl
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -539,7 +539,7 @@ namespace Mapsui.UI.Forms
         /// <remarks>
         /// When this function is called, the control is redrawn if needed
         /// </remarks>
-        public void StartUpdates(bool refresh)
+        public void StartUpdates(bool refresh = true)
         {
             _mapControl.StartUpdates(refresh);
         }
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -556,6 +556,17 @@ namespace Mapsui.UI.Forms
             _mapControl.StopUpdates();
         }
 
+        /// <summary>
+        /// Force a update of control
+        /// </summary>
+        /// <remarks>
+        /// When this function is called, the control draws itself once 
+        /// </remarks>
+        public void ForceUpdate()
+        {
+            _mapControl.ForceUpdate();
+        }
+
         internal void AddCallout(Callout callout)
         {
             if (!_callouts.Contains(callout))
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -40,6 +40,19 @@ namespace Mapsui.UI.Wpf
             Map = new Map();
         }
 
+        void CommonPaintControl(object canvas)
+        {
+            if (Renderer == null) 
+                return;
+            if (_map == null) 
+                return;
+            if (!Viewport.HasSize) 
+                return;
+
+            Navigator.UpdateAnimations();
+            Renderer.Render(canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
+        }
+
         /// <summary>
         /// After how many degrees start rotation to take place
         /// </summary>
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -28,6 +28,17 @@ namespace Mapsui.UI.Wpf
     {
         private Map _map;
         private double _unSnapRotationDegrees;
+        // Flag indicating if a drawing process is running
+        private bool _drawing = false;
+        // Flag indicating if a new drawing process should start
+        private bool _refresh = false;
+        // Action to call for a redraw of the control
+        private Action _invalidate;
+
+       void CommonInitialize()
+        {
+            Map = new Map();
+        }
 
         /// <summary>
         /// After how many degrees start rotation to take place
@@ -186,6 +197,11 @@ namespace Mapsui.UI.Wpf
         {
             RefreshData(changeType);
             RefreshGraphics();
+        }
+
+        public void RefreshGraphics()
+        {
+            _refresh = true;
         }
 
         private void MapDataChanged(object sender, DataChangedEventArgs e)
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -34,14 +34,25 @@ namespace Mapsui.UI.Wpf
         private bool _refresh = false;
         // Action to call for a redraw of the control
         private Action _invalidate;
+        // Timer for loop to invalidating the control
+        private System.Threading.Timer _invalidateTimer;
+        // Interval between two calls of the invalidate function in ms
+        private int _updateInterval = 16;
 
        void CommonInitialize()
         {
+            // Create map
             Map = new Map();
+            // Create timer for invalidating the control
+            _invalidateTimer = new((state) => InvalidateTimerCallback(state), null, System.Threading.Timeout.Infinite, 16);
+            // Start the invalidation timer
+            StartUpdates();
         }
 
-        void CommonPaintControl(object canvas)
+        void CommonDrawControl(object canvas)
         {
+            if (_drawing)
+                return;
             if (Renderer == null) 
                 return;
             if (_map == null) 
@@ -49,10 +60,62 @@ namespace Mapsui.UI.Wpf
             if (!Viewport.HasSize) 
                 return;
 
+            // Start drawing
+            _drawing = true;
+            // All requested updates up to this points will be handled by this redraw
+            _refresh = false;
             Navigator.UpdateAnimations();
             Renderer.Render(canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
+            // End drawing
+            _drawing = false;
         }
 
+        void InvalidateTimerCallback(object state)
+        {
+            if (_drawing || !_refresh)
+                return;
+
+            _invalidate();
+        }
+
+        /// <summary>
+        /// Start updates for control
+        /// </summary>
+        /// <remarks>
+        /// When this function is called, the control is redrawn if needed
+        /// </remarks>
+        public void StartUpdates()
+        {
+            _invalidateTimer.Change(0, _updateInterval);
+        }
+
+        /// <summary>
+        /// Stop updates for control
+        /// </summary>
+        /// <remarks>
+        /// When this function is called, the control stops to redraw itself, 
+        /// even if it is needed
+        /// </remarks>
+        public void StopUpdates()
+        {
+            _invalidateTimer.Change(System.Threading.Timeout.Infinite, System.Threading.Timeout.Infinite);
+        }
+
+        /// <summary>
+        /// Interval between two redraws of the MapControl in ms
+        /// </summary>
+        public int UpdateInterval
+        {
+            get => _updateInterval;
+            set
+            {
+                if (_updateInterval != value)
+                {
+                    _updateInterval = value;
+                    StartUpdates();
+                }
+            }
+        }
         /// <summary>
         /// After how many degrees start rotation to take place
         /// </summary>
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -103,6 +103,17 @@ namespace Mapsui.UI.Wpf
         }
 
         /// <summary>
+        /// Force a update of control
+        /// </summary>
+        /// <remarks>
+        /// When this function is called, the control draws itself once 
+        /// </remarks>
+        public void ForceUpdate()
+        {
+            _invalidate?.Invoke();
+        }
+
+        /// <summary>
         /// Interval between two redraws of the MapControl in ms
         /// </summary>
         public int UpdateInterval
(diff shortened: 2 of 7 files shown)


janusw:
Thanks for the update! Seems the CI is not quite happy with it, though:

```
Error: MapControl.cs(59,13): error CS0103: The name 'InternalInitialize' does not exist in the current context [/Users/runner/work/Mapsui/Mapsui/Mapsui.UI.Android/Mapsui.UI.Android.csproj]
Error: MapControl.cs(66,13): error CS0103: The name 'InternalInitialize' does not exist in the current context [/Users/runner/work/Mapsui/Mapsui/Mapsui.UI.Android/Mapsui.UI.Android.csproj]
Error: MapControl.cs(159,13): error CS0103: The name 'CommonPaintControl' does not exist in the current context [/Users/runner/work/Mapsui/Mapsui/Mapsui.UI.Android/Mapsui.UI.Android.csproj]
```

janusw:
And also:

```
Mapsui.UI.Shared\MapControl.cs(278,21): error CS0111: Type 'MapControl' already defines a member called 'RefreshGraphics' with the same parameter types [D:\a\Mapsui\Mapsui\Mapsui.UI.Uwp\Mapsui.UI.Uwp.csproj]
```

charlenni:
So, I hope now, that the basic problems are solved.

Should it be possible in the XAML renderer of WPF to check for the `PixelDensity`, then some more code could be moved to `CommonDrawControl()`.

In the current code, we don't distinguish between rendering and drawing. I think, that these are two different processes. But this is another issue.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -28,7 +28,94 @@ public partial class MapControl : INotifyPropertyChanged
...
+        }
 
+        /// <summary>
+        /// Stop updates for control
+        /// </summary>
+        /// <remarks>
+        /// When this function is called, the control stops to redraw itself, 
+        /// even if it is needed
+        /// </remarks>
+        public void StopUpdates()
```
inforithmics, line 99: Good method for static maps, saves rendering Time and Electricity

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -28,7 +28,94 @@ public partial class MapControl : INotifyPropertyChanged
...
+            Map = new Map();
+            // Create timer for invalidating the control
+            _invalidateTimer = new((state) => InvalidateTimerCallback(state), null, System.Threading.Timeout.Infinite, 16);
+            // Start the invalidation timer
+            StartUpdates();
+        }
+
+        void CommonDrawControl(object canvas)
+        {
+            if (_drawing)
```
inforithmics, line 54: good condition if the previous rendering is still in progress

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -28,7 +28,94 @@ public partial class MapControl : INotifyPropertyChanged
...
+        /// </remarks>
+        public void StopUpdates()
+        {
+            _invalidateTimer.Change(System.Threading.Timeout.Infinite, System.Threading.Timeout.Infinite);
+        }
+
+        /// <summary>
+        /// Interval between two redraws of the MapControl in ms
+        /// </summary>
+        public int UpdateInterval
```
inforithmics, line 107: What happens if someone enters a negative number?

Mapsui.UI.Wpf/MapControl.cs
```diff
@@ -47,14 +47,23 @@ public partial class MapControl : Grid, IMapControl
...
         public MapControl()
         {
+            CommonInitialize();
+            Initialize();
+        }
+
+        void Initialize()
+        {
+            _invalidate = () => {
+                if (Dispatcher.CheckAccess()) InvalidateCanvas();
```
inforithmics, line 57: This guarantes that only a single thread is executed (the uithread of which there is only one)

Mapsui.UI.Shared/MapControl.cs
charlenni: Added a check.

Mapsui.UI.Wpf/MapControl.cs
charlenni: Was the same before. I didn't change anything in this part of the code.

Mapsui.UI.Shared/MapControl.cs
charlenni: I'm not sure, if we need a lock object. But I assume, that it is so fast, that there are no problems.

Mapsui.UI.Shared/MapControl.cs
charlenni: Do you think we need access to `_refresh`, `_drawing` or `_invalidate()` from outside?

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -28,7 +28,98 @@ public partial class MapControl : INotifyPropertyChanged
...
+            // End drawing
+            _drawing = false;
+        }
+
+        void InvalidateTimerCallback(object state)
+        {
+            if (_drawing || !_refresh)
+                return;
+
+            _invalidate();
```
inforithmics, line 78: I tried out the branch and here it throws a null reference exception, Becuase ComminInitializer is first initialized and then the initializer.
in the Common Initializer the Timer is initialized and starts firing. But the _invalidate() is initialized in the Initialized so it can happen that a exception occours.

Mapsui.UI.Shared/MapControl.cs
inforithmics: I cannot think of a use case yet.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -28,7 +28,98 @@ public partial class MapControl : INotifyPropertyChanged
...
+            if (_map == null) 
+                return;
+            if (!Viewport.HasSize) 
+                return;
+
+            // Start drawing
+            _drawing = true;
+            // All requested updates up to this points will be handled by this redraw
+            _refresh = false;
+            Navigator.UpdateAnimations();
```
inforithmics, line 67: Maybe here could be the time measuresd and when it takes more than 16ms An Debug.Writline took more than (16ms) or a Trance.Writline(took more than 16 ms).

inforithmics:
I measured the Map Display time

From

image

To

image

so roughly a 400% performance improvement

Mapsui.UI.Shared/MapControl.cs
charlenni: Solved

Mapsui.UI.Shared/MapControl.cs
charlenni: This could be made at this point, but is not part of this PR. The next step would be to improve the performance of Mapsui. Up to now, we want to get rid of unneeded redrawing.

charlenni:
Is it possible to see the column headers? Would be nice. Thank you.

charlenni:
Is it possible to get the number of calls for `Mapsui.UI.Forms.MapControl.PaintSurface()`? In the first image it is around 4919.

inforithmics:
I updated the Performance comment

charlenni:
Thank you for the update.

That is strange. `PaintSurface` is called 13 times, while `VisibleFeatureIterator` is called 17 times. It seems, that the later one is also called from outside of the draw process. Have to look into this.

inforithmics:
The additional Renderings come from a RenderToBitmap for Rasterizing.

image

janusw:
> For `UpdateInterval`, I'm missing a corresponding property in the MapView, otherwise it cannot be set when using a MapView (same for `StartUpdates` and `StopUpdates`).

I added these in my fork: https://github.com/janusw/Mapsui/commits/InvalidateLoop

Feel free to include it into this PR if you agree!

charlenni:
This is the answer on the last comment of @pauldendulk on #1092, which is closed.

Yes, we have to do this and that. And yes, there are hundreds of ways to improve the performance. But all this isn‘t discussed here.

This discussion is about reducing the redraws of the screen. It is about the fact, that in the moment each change of anything triggers a redraw of the screen. To get rid of this, we have to collect such redraws. But what is the correct way to do this? This approach waits a given amount of milliseconds and than, if there is something to draw, updates the screen.

So the question is:
1. Do we accept the need of such a collection?
2. Do we think, this approach could work?
3. Do we have a better approach for this?

janusw:
> This discussion is about reducing the redraws of the screen. It is about the fact, that in the moment each change of anything triggers a redraw of the screen. To get rid of this, we have to collect such redraws. But what is the correct way to do this? This approach waits a given amount of milliseconds and than, if there is something to draw, updates the screen.
> 
> So the question is:
> 
>     1. Do we accept the need of such a collection?

From my point of view: Clearly YES!

The current scheme of redrawing the map everytime something changes simply does not scale well to complex scenarios. I can show you some where the changes here make a hell of a difference.

As an example, think of a map with a large number of pins (order of magnitude: hundreds or thousands), each of which is updated regularly. The current Mapsui fails badly on such a scenario (completely unusable, in particular on iOS), and this branch gives phenomenal improvements.

It seems @inforithmics has more relevant use cases.

>     2. Do we think, this approach could work?

It sounds very reasonable to me (in theory), and also in practice I can confirm that it greatly improves the responsiveness of the map for scenarios like the one described above. So yes, it works!


>     3. Do we have a better approach for this?

I don't have a much better idea right now. Having a configurable update interval is a good idea, and also the methods `StartUpdates` and `StopUpdates` are pretty useful, since they avoid wasting time on complex rendering if the map is not being displayed at all.


IMHO we should clearly go ahead with this. Further performance improvements can certainly be added later, but should not be subject of this PR.

charlenni:
BTW I assume that this PR nothing change at the FPS. It stops the rate at around 60 FPS (1000 ms/16 ms). But in the moment the FPS is limited by the time `CommonDrawControl()` needs. So, if we want to increase the FPS, we have to improve this part. This PR is only for reactiveness of the control.

charlenni:
And if I interpret the performance diagram of @inforithmics correct, then we could draw the map 13 times in a second, which leads to 13 FPS, correct?

Opposite to the last proposal, this hasn't a calculation for FPS. This should go in the next step into `CommonDrawControl()`. Should be the next PR.

charlenni:
Perhaps a `ForceUpdate()` could be helpful in scenarios, where you want to have a static map that should be redrawn after the initial configuration. It invalidates the screen once.

janusw:
> Perhaps a `ForceUpdate()` could be helpful in scenarios, where you want to have a static map that should be redrawn after the initial configuration. It invalidates the screen once.

Might possibly be useful and is trivial to implement. (I personally don't have a use case for this, though.)

@charlenni Could you please push your MapView changes (or merge mine from https://github.com/janusw/Mapsui/commit/bfe05ed93848d28947c34fdb3aec099dacd3578f), please?

charlenni:
@janusw Sorry for the delay. I'm still on vacation, and it isn't so easy to find a WiFi to upload the changes ;-)

charlenni:
@inforithmics If you now compare the performance of Mapsui, you should post the values for `CommonDrawControl()`. That are the times, that the control needs to draw its content. The `Time`and `Calls`. First divided by the second will give a mean drawing time. Correct?

janusw:
> @janusw Sorry for the delay. I'm still on vacation, and it isn't so easy to find a WiFi to upload the changes ;-)

Nevermind (also: thanks!). You were posting all days, so I thought you were available. Enjoy your vacation!

charlenni:
All possible with the smartphone, but no PR changes.

Mapsui.UI.Forms/MapView.cs
```diff
@@ -528,6 +533,29 @@ public Geometries.Point ToPixels(Geometries.Point coordinateInDeviceIndependentU
 
         #endregion
 
+        /// <summary>
+        /// Start updates for control
+        /// </summary>
+        /// <remarks>
+        /// When this function is called, the control is redrawn if needed
+        /// </remarks>
+        public void StartUpdates(bool refresh)
```
janusw, line 542: The parameter should probably have a default value here (as in the corresponding MapControl method).

charlenni:
Did I said that I hate the MapView construction? All this wrappers. Perhaps I should make the buttons as widgets.

janusw:
> Did I said that I hate the MapView construction? All this wrappers.

Certainly not very elegant. Wasn't there the idea to make MapView inherit from MapControl? Is that feasible? At least it would remove the need for these kinds of wrappers ...

Mapsui.UI.Forms/MapView.cs
```diff
@@ -458,6 +458,11 @@ public INavigator Navigator
         /// </summary>
         internal IMapControl MapControl => _mapControl;
 
+        /// <summary>
+        /// Update interval for invalidation timer in ms
+        /// </summary>
+        public int UpdateInterval => _mapControl.UpdateInterval;
```
janusw, line 464: This is not the proper syntax apparently. When I try to assign to MapView.UpdateInterval, I get:
`Error CS0200: Property or indexer 'MapView.UpdateInterval' cannot be assigned to -- it is read only (CS0200)`

janusw (changes requested):

Mapsui.UI.Forms/MapView.cs
charlenni: Hopefully solved

janusw: Think so.

janusw (approved):
I am happy with the PR at this point. I gives huge performance improvements for some of my use cases. Thanks for the great work, @charlenni !

If @pauldendulk and @inforithmics are ok with it as well, I hope we can merge this very soon.

inforithmics (approved):

janusw:
Alright, we have two approvals here. @pauldendulk Do you have any further comments, or can I merge this?

pauldendulk:
> Alright, we have two approvals here. @pauldendulk Do you have any further comments, or can I merge this?

I'll would like to have a look at it later this week.

janusw:
> I'll would like to have a look at it later this week.

Sure, take your time! (I have to admit that this issue is kind of urgent on my side, but of course I understand that you want to have a proper look at it.)

I guess this PR would be a good topic for a meeting on Friday (if we manage to have one this week) ...

pauldendulk:
> Should it be possible in the XAML renderer of WPF to check for the PixelDensity, then some more code could be moved to CommonDrawControl().

There is a method to get the PixelDensity for WPF, so I think this should be possible. Or are you saying there is problem with this code? (see Mapsui.UI.Wpf.MapControl.GetPixelDensity())

It would be nice if it could be moved but perhaps an exception would occur on ```canvas.Scale(PixelDensity, PixelDensity);``` if it is zero.

pauldendulk:
Is there a general way in which this could be tested? Which project did you run? Is there a sample for this? Of not there should be (not part of this PR). Perhaps based on @bertt's [sample](https://github.com/Mapsui/Mapsui/tree/bug/getfeaturesinview_not_called/Samples/Mapsui.Samples.MemoryProvider/MemoryProvider/).

charlenni:
How should this be made testable? BTW did we tested the way before?

charlenni:
To be correct: it is a visible control like a button. So how to test it? There is a tool from Xamarin, where you could test such things in many, many devices. But it costs.

pauldendulk:
I was merely asking which project I should run in order to see the performance problem so I could see how it is affected by the patch. 

But on the topic. It is possible to unit test the individual parts. A invalidate loop could be in een separate class that could be tested. This is hard because it is a process over time, but it still possible. That fact that we did not have much tests in the past does not mean we should not add it in the future.

pauldendulk:
I will merge this PR.

I had some hesitation because the underlying problem is in the platform and was thinking about a solution that is closer to the platform. 

All these platforms have a method called Invalidate or SetNeedsDisplay. These methods have names like these because they should not do a direct call to the OnPaint method. They should set a flag to like the '_refresh' flag in this PR. And ideally the platform implementation should also have a way to cap the framerate. Apparently this is not what is happening.

A solution I was thinking of was to fix this in the OnPaint method. Here we could use a Thread.Sleep to cap the framerate. In that way it is also easier to avoid overlap in calls (if the paint itself takes more than 16 ms).  A problem with Thread.Sleep is however that it locks the thread, which could be problematic as well. Not something I want to investigate now.

This PR solves a problem which probably hits many users. Thanks a lot for this contribution @charlenni!

charlenni:
You can use any sample that uses a MapControl on any platform. Because it is the same for all platforms. It seems that the most problems are on iOS. Use a sample that changes the map often.

charlenni:
If we make a solution for each platform, then the code is getting complicated. I thought about a method, so we could handle this problem on all platforms in the same way.

We could have made it on Android totally different. It seems, that Android could handle this by itself. But than we had to do it for the Forms or Maui platform also separate for each platform, what needs many #if in the code.

So I chooses a way, which is the same on all platforms, make it better on some of them and doesn't make it worth on others.

If we find better solution in the future, it isn't a problem to change this.

BTW it is problematic in the OnPaint event, because you don't know, if the redraw is called because something on the map has changed which could wait or if the control reappears on the screen and needs an immediate redraw.

pauldendulk:
> If we make a solution for each platform, then the code is getting complicated. I thought about a method, so we could handle this problem on all platforms in the same way.

I imagined a generic solution with code placed in a shared class which is called from all implementations like the CommonDrawControl. But perhaps you are right and we should take platform differences into account. With that in mind I really would like to have more statistics.

PR #1195, opened 2021-09-07
update Topten.RichTextKit to 0.4.145

janusw:


janusw:
Rebased to current master and fixed conflicts with #1194.

janusw:
Note: Apart from simply updating the RichTextKit version here, I have also added a dependency on it in `Mapsui.Forms.nuspec`.

I'm not sure if this is strictly required or not. At least RichTextKit is used in Mapsui.UI.Forms. In principle there is already an implicit dependency on RichTextKit via Mapsui.Forms -> Mapsui -> RichTextKit. This might already be sufficient. However, SkiaSharp (for example) is also used as an explicit dependency in Mapsui *and* Mapsui.Forms (*and* Mapsui.Native).

pauldendulk:
> Note: Apart from simply updating the RichTextKit version here, I have also added a dependency on it in `Mapsui.Forms.nuspec`.
> 
> I'm not sure if this is strictly required or not. At least RichTextKit is used in Mapsui.UI.Forms. In principle there is already an implicit dependency on RichTextKit via Mapsui.Forms -> Mapsui -> RichTextKit. This might already be sufficient. However, SkiaSharp (for example) is also used as an explicit dependency in Mapsui _and_ Mapsui.Forms (_and_ Mapsui.Native).

Not sure what's the best policy here. In general I would go for minimal. So, if not needed then don't add. A motivation to add it explicitly is if it used expliciet in the project. So if in Mapsui.Forms we do things with RichTextKit itself that might be a reason to add it.

pauldendulk:
I can remove the RichTextKit project reference from Mapsui.Forms and everything still compiles, so it is not used in there. So perhaps we should not add the nuget dependency. 

There one 'using Topten.RichTextKit;' but it is unnecessary.

You say:
> However, SkiaSharp (for example) is also used as an explicit dependency in Mapsui and Mapsui.Forms (and Mapsui.Native).

I don't see what you are referring to. The mapsui.forms.nuspec only refers to 'SkiaSharp.Views.Forms'.

charlenni:
RichTextKit is used for Callouts. They move sometime from Forms to the native part, so that all UI could use them. That is the reason, why Mapsui.Forms no longer need a reference.

janusw:
> I can remove the RichTextKit project reference from Mapsui.Forms and everything still compiles, so it is not used in there. So perhaps we should not add the nuget dependency.
> 
> There one 'using Topten.RichTextKit;' but it is unnecessary.

Ah, right. I removed that one, and also removed the package reference to RichTextKit from Mapsui.UI.Forms, so that adding the nuspec reference is indeed not necessary (I removed that commit from the branch and rebased to fix conflicts).


> > However, SkiaSharp (for example) is also used as an explicit dependency in Mapsui and Mapsui.Forms (and Mapsui.Native).
> 
> I don't see what you are referring to. The mapsui.forms.nuspec only refers to 'SkiaSharp.Views.Forms'.

Right. Sorry for the confusion. (Also this was slightly cleanup up in #1196.)

charlenni:
Thank you for cleaning up the code 👍

pauldendulk (approved):
Great, thanks

PR #1196, opened 2021-09-08
Mapsui.nuspec: remove dependency on SkiaSharp.Views

janusw:
* SkiaSharp.Views is only used in the Mapsui.UI.* parts
  (which are contained in the Mapsui.Native and Mapsui.Forms packages)
* this avoids nuget warnings like this, when referencing the Mapsui package from another project:
  Package 'SkiaSharp.Views 2.80.3' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework '.NETStandard,Version=v2.0'. This package may not be fully compatible with your project.
* it is a fixup for PR #1175


pauldendulk (approved):
ah, cleaning up. I love it.

PR #1197, opened 2021-09-09
Update the documentation regarding the new Mapsui.Native package

janusw:
The quickstart guides for WPF & Android now require the Mapsui.Native package instead of the Mapsui base package (the latter previously also included the native UI parts).


pauldendulk:
Your first commits adds a lot of documents. Looking at a few it seems that they should have there in the first place. A lot of documents are missing in the current api on github (they used to be there). Do you know what is happening?

Even Mapsui core is missing
image

Publishing the new api docs is supposed to be an automatic process.

pauldendulk:
Discussed this with @janusw. It seems the files added in the PR were supposed to be here in the first place. Apparently something went wrong in the past building the docs. Could be on a local machine or could be on the build server.

pauldendulk (approved):
Great you also thought about the docs.

PR #1199, opened 2021-09-11
release 3.0.0-rc.2

janusw:
We talked about tagging a new (pre-)release after all the recent fixes. This updates the version number. After it is merged, we only need to:
* set a tag
* upload the nupkgs to nuget.org

I may be able to do the first one (not sure if I have sufficient permissions), but the uploading probably needs to be done by Paul.

### DIFF
--- a/Mapsui.Common.props
+++ b/Mapsui.Common.props
@@ -3,7 +3,7 @@
   
   <!--- Package information, Version -->
   <PropertyGroup>
-    <Version>3.0.0-alpha.4</Version>
+    <Version>3.0.0-rc.2</Version>
     <NeutralLanguage>en</NeutralLanguage>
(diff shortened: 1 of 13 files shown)


pauldendulk:
I think this PR can be closed. 

The versions that are changed should be set by the build script and it always worked like that before but apparently something broke after alpha.4. I have this fixed now in my local branch. I will create a PR soon. Perhaps I will admin-merge this myself because I want to get a release out. I wanted to do that yesterday but ran into this problem with the version tag and then ran into quite some old stuff that needed to get fixed or updated.

The git tag is also set by the script. 
https://github.com/Mapsui/Mapsui/blob/b25c683b627f670697887b1e66565aa3b3f4a504/Scripts/buildpackpush.bat#L8

PR #1202, opened 2021-09-11
Rework build tooling

pauldendulk:
Context:
While preparing a release I ran into a build issue, the versionupdater did not build properly because it was .net framework. I converted it to .net core and other changes followed.


pauldendulk:
Things done:
- Removed unused updateversionnumber.proj
- Removed Mapsui.Common.props and moved the version info to Directory.Build.props. Removed all references of Mapsui.Common.props. Advantages: We have just one file for common csproj info. We do not have to add the Mapsui.Common.props reference to new files (it was actually missing a quite a few at the moment).
- Updated some old references to Mapsui.Mac files to Mapsui.Native files.
- In the Mapsui.Native.nuspec removed the references to the dlls that should come from the Mapsui.nuget.
- Converted the versionupdater to .net core and completely rewrote it. Now it updates the Mapsui.Common.props version, and also still the assemblyinfo files.
- Build versionupdater in bin folder so these build files are ignored by git. Deleted the old binaries (and accompanying) files.

PR #1203, opened 2021-09-11
Remove unnecessary package refs

janusw:
Pure cleanup: This PR removes several unnecessary package references from the sample projects (these packages are not being used anywhere).

### DIFF
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
@@ -122,8 +122,8 @@
     <PackageReference Include="SkiaSharp.Views.Forms">
       <Version>2.80.3</Version>
     </PackageReference>
-    <PackageReference Include="SQLitePCLRaw.bundle_green">
-      <Version>1.1.14</Version>
+    <PackageReference Include="sqlite-net-pcl">
+      <Version>1.6.292</Version>
     </PackageReference>
     <PackageReference Include="Svg.Skia">
       <Version>0.5.7.1</Version>
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
@@ -104,9 +104,6 @@
     <PackageReference Include="BruTile.MbTiles">
       <Version>4.0.0</Version>
     </PackageReference>
-    <PackageReference Include="HarfBuzzSharp">
-      <Version>2.6.1.7</Version>
-    </PackageReference>
     <PackageReference Include="Itinero">
       <Version>1.5.0</Version>
     </PackageReference>
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
@@ -123,7 +123,7 @@
       <Version>2.80.3</Version>
     </PackageReference>
     <PackageReference Include="sqlite-net-pcl">
-      <Version>1.6.292</Version>
+      <Version>1.8.116</Version>
     </PackageReference>
     <PackageReference Include="Svg.Skia">
       <Version>0.5.7.1</Version>
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
@@ -104,9 +104,6 @@
     <PackageReference Include="BruTile.MbTiles">
       <Version>4.0.0</Version>
     </PackageReference>
-    <PackageReference Include="ConcurrentHashSet">
-      <Version>1.1.0</Version>
-    </PackageReference>
     <PackageReference Include="HarfBuzzSharp">
       <Version>2.6.1.7</Version>
     </PackageReference>
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Mac/Mapsui.Samples.Forms.Mac.csproj
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Mac/Mapsui.Samples.Forms.Mac.csproj
@@ -130,8 +130,8 @@
     <PackageReference Include="SkiaSharp.Views.Forms">
       <Version>2.80.3</Version>
     </PackageReference>
-    <PackageReference Include="SQLitePCLRaw.bundle_green">
-      <Version>1.1.14</Version>
+    <PackageReference Include="sqlite-net-pcl">
+      <Version>1.6.292</Version>
     </PackageReference>
     <PackageReference Include="Svg.Skia">
       <Version>0.5.7.1</Version>
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Mac/Mapsui.Samples.Forms.Mac.csproj
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Mac/Mapsui.Samples.Forms.Mac.csproj
@@ -115,9 +115,6 @@
     <PackageReference Include="BruTile.MbTiles">
       <Version>4.0.0</Version>
     </PackageReference>
-    <PackageReference Include="ConcurrentHashSet">
-      <Version>1.1.0</Version>
-    </PackageReference>
     <PackageReference Include="HarfBuzzSharp">
       <Version>2.6.1.7</Version>
     </PackageReference>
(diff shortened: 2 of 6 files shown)


pauldendulk:
Four nuget packages are removed:
- HarfBuzzSharp
- ConcurrentHashSet
- SQLitePCLRaw.bundle_green
- sqlite-net-pcl - still works on 


How did you determine that the package could be removed? Did you use a tool? Or just looked at the references?

It is not always easy to see how a package is needed. 
- In some cases a package you use could depend on another package. When removing that other package things could still compile but fail at runtime.
- In some cases a platform dependent reference needs to be added because the package uses a 'bait and switch' pattern. 
- In some cases a package needs to added to resolve a conflict with package versions of uses packages/project. In that case the buildserver would spot the problem.

**HarfBuzzSharp**. I think this came with work on callout but is not used now.

**ConcurrentHashSet**: Not sure where this came from but if things compile I think it will be okay. 

- SQLitePCLRaw.bundle_green and sqlite-net-pcl. In the past I needed to add those because they worked with a bait and switch pattern. So in order to get things to work an a specific platform the reference had to be added to the 'head' project. Did you test of loading a MBTiles file still works on iOS? (I assume it is in the samples but I am not sure).

janusw:
> Four nuget packages are removed:
> 
>     * HarfBuzzSharp
> 
>     * ConcurrentHashSet
> 
>     * SQLitePCLRaw.bundle_green
> 
>     * sqlite-net-pcl - still works on
> 
> 
> How did you determine that the package could be removed? Did you use a tool? Or just looked at the references?

I did not use a tool here, but checked manually (grepping for references in the sources). All these packages were showing updates in the nuget package manager (there are more seemingly unused packages which I did not touch).

> It is not always easy to see how a package is needed.
> 
>     * In some cases a package you use could depend on another package. When removing that other package things could still compile but fail at runtime.

If a package depends on another one, such a transitive dependency should work without having to add the second package explicitly (unless there is a bug somewhere that you need to work around).

>     * In some cases a platform dependent reference needs to be added because the package uses a 'bait and switch' pattern.

I don't understand what a 'bait and switch' pattern means in our context.

>     * In some cases a package needs to added to resolve a conflict with package versions of uses packages/project. In that case the buildserver would spot the problem.

Obviously everything compiles and links fine. None of these packages seems to be related to version-conflicts. I did not check for runtime problems.

charlenni:
One short comment: HarfBuzzSharp is needed by RichTextKit. It calculates the positions of characters with a given font. So it should come with RichTextKit. RichTextKit was introduced by me for the Callout class. It draws the text for callouts, so it is possible to have different fonts and styles in a Callout.

Perhaps we should introduce this later in LabelStyle too.

pauldendulk:
> One short comment: HarfBuzzSharp is needed by RichTextKit. It calculates the positions of characters with a given font. So it should come with RichTextKit. RichTextKit was introduced by me for the Callout class. It draws the text for callouts, so it is possible to have different fonts and styles in a Callout.

Okay, so removing this would break things. How could we have detected this?

> Perhaps we should introduce this later in LabelStyle too.

Makes sense.

charlenni:
Next comment: ConcurrentHashSet is now integrated in Mapsui.Utilities in `ConcurrentCollections`. The comment says, that there are problems with the Nuget package, so this only collection was copied. It is used in `WritableLayer.cs` and `TileFetchDispatcher.cs`. So I think, it was introduced by @pauldendulk.

charlenni:
Next comment: `SQLitePCLRaw.bundle_green` is used by the WPF samples to connect to a MBTiles file (samples for WPF and Forms.WPF). We don't use SQLite outside the samples. SQLite is perhaps part of BruTile?

pauldendulk:
> Next comment: ConcurrentHashSet is now integrated in Mapsui.Utilities in ConcurrentCollections. The comment says, that there are problems with the Nuget package, so this only collection was copied. It is used in WritableLayer.cs and TileFetchDispatcher.cs. So I think, it was introduced by @pauldendulk.

I can remember adding the local copy, but not the nuget reference, but it could be mine originally.

pauldendulk:
> Next comment: SQLitePCLRaw.bundle_green is used by the WPF samples to connect to a MBTiles file (samples for WPF and Forms.WPF). We don't use SQLite outside the samples. SQLite is perhaps part of BruTile?

It is an indirect reference of BruTile.MbTiles. 
- BruTile.MbTiles: https://www.nuget.org/packages/BruTile.MbTiles/
- References: https://www.nuget.org/packages/sqlite-net-pcl/
- References: https://www.nuget.org/packages/SQLitePCLRaw.bundle_green/

Updating BruTile.MbTiles will probably add it again, but more important, I think it is needed for the MbTiles sample if it indeed uses *bait and switch*. 

Bait and switch works by adding a dummy dll that can be accessed from a .netstandard dll. But the app only works if you also add a reference to the actual implementation in your head project. During compilation the dll in the head will have priority over the dummy and everything will work fine in your application. But you do need to add that platform dependent implementation to the head project because if you communicate with the dummy things will fail. You will only notice during runtime.

MbTiles is important functionality. I had many questions about it in the past. It is used for offline support.

pauldendulk:
The ConcurrentHashSet in the Android sample come with a commit called [Update Packages](https://github.com/Mapsui/Mapsui/commit/368b19b55f624acfca0a901867e0d2fc4c6bda61) from charlenni. I assume it was really just a package update and one of the many packages had a reference to ConcurrentHashSet. It could be that it is added again if we update again. It could be the reference has been removed by that package in the meantime. Hard to manage things like that.

pauldendulk:
> It could be that it is added again if we update again.

But this could be tested of course. Do an update of all package in the android sample and see if the package ConcurrentHashSet is added again.

pauldendulk:
Here Matthew Leibowitz the maintainer of SkiaSharp write about bait and switch:
https://github.com/mattleibow/CreatingPlatformPlugins#bait-and-switch

pauldendulk (changes requested):
About HarfBuzzSharp. This comes indirectly via 
RichTextKit https://www.nuget.org/packages/Topten.RichTextKit/
and then https://www.nuget.org/packages/SkiaSharp.HarfBuzz/
So, this one should stay. You could test if running nuget update on the branch re-adds HarfBuzzSharp.

About ConcurrentHashSet. It is most likely added by in an indirect reference through some other nuget. You could test this by running nuget update on the branch. If it is not added again assume that it can be removed. If it is added it might be worth to find out where it comes from so we are aware of how things work.

About SQLitePCLRaw.bundle_green. This one comes indirectly with sqlite-net-pcl. If we need sqlite-net-pcl it should stay otherwise it can be removed.

About sqlite-net-pcl. I think we need that one because it is a bait and switch nuget. You could test that by running the mbtiles sample in the app and see if it shows the map.

janusw:
> About HarfBuzzSharp. This comes indirectly via
> RichTextKit https://www.nuget.org/packages/Topten.RichTextKit/
> and then https://www.nuget.org/packages/SkiaSharp.HarfBuzz/
> So, this one should stay. You could test if running nuget update on the branch re-adds HarfBuzzSharp.
> 
> About ConcurrentHashSet. It is most likely added by in an indirect reference through some other nuget. You could test this by running nuget update on the branch. If it is not added again assume that it can be removed. If it is added it might be worth to find out where it comes from so we are aware of how things work.

I don't fully understand these comments. If a dependency comes in indirectly via another package, it should not be necessary to reference it directly, right? It should be pulled in implicitly via the other package, and updating nugets should not add a direct reference to that dependency. Or am I missing anything here?

> About SQLitePCLRaw.bundle_green. This one comes indirectly with sqlite-net-pcl. If we need sqlite-net-pcl it should stay otherwise it can be removed.
> 
> About sqlite-net-pcl. I think we need that one because it is a bait and switch nuget. You could test that by running the mbtiles sample in the app and see if it shows the map.

I will check if I find anything that is broken with this PR.

pauldendulk:
> I don't fully understand these comments. If a dependency comes in indirectly via another package, it should not be necessary to reference it directly, right? It should be pulled in implicitly via the other package, and updating nugets should not add a direct reference to that dependency. Or am I missing anything here?

To show you that it does work like that I just created a test project to find out that it doesn't work like that 😲 ... not anymore at least.

After adding BruTile.MbTiles to a .net core WPF app:

image

After adding BruTile.MbTiles to a .net framework WPF app:

image

So, those package have been added indirectly in the past indirectly, and have at some point been upgraded to the new csproj format where they were not needed anymore.

Bait and switch or package version conflict resolution might still be a reason to add a package explicitly.

janusw:
> After adding BruTile.MbTiles to a .net framework WPF app:
> 
> image

I don't know the old .NET framework world that well, but this is obviously a reason to hate it :/

janusw:
> > About sqlite-net-pcl. I think we need that one because it is a bait and switch nuget. You could test that by running the mbtiles sample in the app and see if it shows the map.
> 
> I will check if I find anything that is broken with this PR.

Indeed the MbTiles sample throws sqlite-related exceptions on this branch ...

`System.TypeInitializationException: The type initializer for 'SQLite.SQLiteConnection' threw an exception.`

janusw:
> Indeed the MbTiles sample throws sqlite-related exceptions on this branch ...

Since removing the sqlite-* packages is obviously not a good idea, I have discarded the corresponding commits and instead do the following now:
* replace SQLitePCLRaw.bundle_green by sqlite-net-pcl (to have more consistency among the sample apps)
* update sqlite-net-pcl to the latest version

But I'm sticking to the removal of ConcurrentHashSet and HarfBuzzSharp.

@pauldendulk I hope this PR is better with these changes ...

pauldendulk (approved):

PR #1204, opened 2021-09-11
Removed a bug, where default value of UseFling of MapView wasn't changed

charlenni:
Removed a bug, where the default value of UseFling of MapView wasn't correct. So the UseFling of the underlying MapControl isn't changed to false.


pauldendulk (approved):
ah, missed that.

PR #1205, opened 2021-09-11
Added a stopwatch for measuring drawing time of control

charlenni:
@pauldendulk Do you want the measuring of drawing time in this way? Do we need a property for the last elapsed time? Do we need a moving average? Do we need anything else?

### DIFF
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -38,6 +38,8 @@ namespace Mapsui.UI.Wpf
         private System.Threading.Timer _invalidateTimer;
         // Interval between two calls of the invalidate function in ms
         private int _updateInterval = 16;
+        // Stopwatch for drawing
+        private System.Diagnostics.Stopwatch _stopwatch = new System.Diagnostics.Stopwatch();
 
        void CommonInitialize()
         {
@@ -62,10 +64,14 @@ namespace Mapsui.UI.Wpf
 
             // Start drawing
             _drawing = true;
-            // All requested updates up to this points will be handled by this redraw
+            // Start stopwatch
+            _stopwatch.Restart();
+            // All requested updates up to this point will be handled by this redraw
             _refresh = false;
             Navigator.UpdateAnimations();
             Renderer.Render(canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
+            _stopwatch.Stop();
+            System.Diagnostics.Debug.WriteLine($"Time for drawing control [ms]: {_stopwatch.Elapsed.TotalMilliseconds}");
             // End drawing
             _drawing = false;
         }
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -66,19 +66,19 @@ namespace Mapsui.UI.Wpf
 
             // Start drawing
             _drawing = true;
-#if DEBUG
+
             // Start stopwatch before updating animations and drawing control
             _stopwatch.Restart();
-#endif
+
             // All requested updates up to this point will be handled by this redraw
             _refresh = false;
             Navigator.UpdateAnimations();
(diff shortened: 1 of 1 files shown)


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -62,10 +64,14 @@ void CommonDrawControl(object canvas)
 
             // Start drawing
             _drawing = true;
-            // All requested updates up to this points will be handled by this redraw
+            // Start stopwatch
```
inforithmics, line 67: This code could be conditional compiled with an #if Debug

charlenni:
@inforithmics Do you thought about it in this way?

inforithmics (approved):

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -64,14 +66,18 @@ void CommonDrawControl(object canvas)
...
             // Start stopwatch
             _stopwatch.Restart();
+#endif
             // All requested updates up to this point will be handled by this redraw
             _refresh = false;
             Navigator.UpdateAnimations();
             Renderer.Render(canvas, new Viewport(Viewport), _map.Layers, _map.Widgets, _map.BackColor);
+#if DEBUG
             _stopwatch.Stop();
             System.Diagnostics.Debug.WriteLine($"Time for drawing control [ms]: {_stopwatch.Elapsed.TotalMilliseconds}");
```
pauldendulk, line 79: Perhaps it would be better to log to Mapsui's central logger: https://mapsui.com/documentation/how-to-work-with-mapsui.html?q=logger. This way the user can decide where to log it. Could also be into a control into their app.

Mapsui.UI.Shared/MapControl.cs
pauldendulk: With a conditional statement it will be only available for developers. For diagnosing the problem of users this may also be useful. How about a static boolean property?

Mapsui.UI.Shared/MapControl.cs
charlenni: Perfect comment. Logger is the right place for such things. Thank you.

Mapsui.UI.Shared/MapControl.cs
charlenni: With the changes of the output to the logger (see below), we add 3 additional lines of code: restart of stopwatch, stop of stopwatch and invoke of the logger delegate. Do you assume, that this is very time-consuming? If we use a static boolean, we have two if statements too.

So at the end, we have 3 possible way to make this:
1. Do it each time in Debug and Release
2. Do it only, if a special flag for MapControl is set
3. Do it only for the Debug part of the Mapsui package
Which way we should go? My preference is the first one.

pauldendulk: Okay, just do it each time. 

My guesses wrt performance:
- A boolean check is very fast, only in rare situations this is noticeable. So we could do it, but is it worth it? Without the check the code is simpler. That's also important.
- Writing to de Debug.WriteLine can be significantly slow. In a recent performance tests I did (in another project) writing to log at DEBUG level was the main bottleneck. But this is not done anymore after your last commit.
- In the current code I think constructing the string (and garbage collecting it) is the biggest thing. But compared to everything else done each render iteration I think it can be ignored.

charlenni: Done

pauldendulk (approved):

pauldendulk:
> @pauldendulk Do you want the measuring of drawing time in this way? Do we need a property for the last elapsed time? Do we need a moving average? Do we need anything else?

A late response in a more general sense. As a future feature I would like to have statistics like this printed in the map. On top of everything else. A static boolean property could be used to turn it on or of. An app could add a way to turn it on during runtime. So our users can easily what is going on, even on a deployed app.

A running average will probably be useful to keep the numbers a bit more stable. Frames per second (FPS) would be another interesting statistic. Time in between renderings as well.

PR #1206, opened 2021-09-11
Update getting-started-wpf.md

pauldendulk:
This is a fix for 2.0. This is what happens when user do not have the 'include pre-releases' check set.


pauldendulk:
For v3 it will be different. But I also noticed there is actually a problem with the nugets of v3 because the xaml renderer dll is missing. This needs a new release. Also, I think that WPF should be in a separate package. The 'Native' phrase stems mostly from 'native mobile' in the context of Xamarin development. Also, I think it is a totally different user group.

janusw:
> But I also noticed there is actually a problem with the nugets of v3 because the xaml renderer dll is missing.

... because you recently removed it in dadd2989f69ce081c9fb22d469ccc4242446081c ;)

I think that commit was really a bit premature and not reviewed by anybody. Not good practice :/

> This needs a new release. Also, I think that WPF should be in a separate package. The 'Native' phrase stems mostly from 'native mobile' in the context of Xamarin development. Also, I think it is a totally different user group.

Well, yeah, but all of that is really a different topic, not related at all to getting-started-wpf, so let's please move it to a different issue.

pauldendulk:
>> But I also noticed there is actually a problem with the nugets of v3 because the xaml renderer dll is missing.

> ... because you recently removed it in dadd298 ;)

No, that is a misunderstanding. This is a different issue. The dlls that are in core should not be added in the Native package as well. The problem was that Mapsui.Redering.Xaml dll was removed from the Mapsui core, while Mapsui.Rendering.Skia was kept in there, that was somewhat surprising to me. It could be added back in there, but since it only builds on net48 I will move it to the new Wpf package. 

> Well, yeah, but all of that is really a different topic, not related at all to getting-started-wpf, so let's please move it to a different issue.

As you can see it is not part of the PR. I will create a new one for it.

docfx/documentation/getting-started-wpf.md
```diff
@@ -7,22 +7,22 @@ Start a new WPF application in Visual Studio.
 ### Step 2
 In the package manager console type:
 ```
-PM> Install-Package Mapsui.Native
+PM> Install-Package Mapsui
```
janusw, line 10: Why did you actually approve my PR #1197, if you're reverting it now?

So then the idea would be to apply the v3 changes to the docu once a final 3.0.0 release has been published?

Maybe it would be good to have separate documentation for v2 and v3 side by side? This would also be useful for the API docu ...

pauldendulk: Sorry about that. There is a problem with building the documentation on my machine and when I run my build script it automatically pushes those changes. This is something I need to fix.

janusw:
> > Well, yeah, but all of that is really a different topic, not related at all to getting-started-wpf, so let's please move it to a different issue.
> 
> As you can see it is not part of the PR. I will create a new one for it.

This may need some discussion before implementing anything. See #1207. Please add further comments on this topic there. This PR is about the getting-started guide.

janusw:
> > > But I also noticed there is actually a problem with the nugets of v3 because the xaml renderer dll is missing.
> 
> > ... because you recently removed it in [dadd298](https://github.com/Mapsui/Mapsui/commit/dadd2989f69ce081c9fb22d469ccc4242446081c) ;)
> 
> No, that is a misunderstanding. This is a different issue. The dlls that are in core should not be added in the Native package as well. The problem was that Mapsui.Redering.Xaml dll was removed from the Mapsui core, while Mapsui.Rendering.Skia was kept in there, that was somewhat surprising to me.

That was done in #1175 (intentionally on my side) and you approved it. The reason why I did it like this was that Rendering.Xaml is only used for WPF (which part of Mapsui.Native), but not for Mapsui.Forms. That's why I did not put it into the Mapsui base package (in constrast to Rendering.Skia), but into Mapsui.Native.


> It could be added back in there, but since it only builds on net48 I will move it to the new Wpf package.

With the current package structure, it belongs into Mapsui.Native IMHO. Once we have a Mapsui.WPF, yes, it should go in there.

docfx/documentation/getting-started-wpf.md
janusw: I can build the documentation and commit it to your branch, if you want.

Also I don't really have a problem with any of the modifications here, but if you're rolling back the change to `getting-started-wpf.md`, you should also do the same with `getting-started-android.md`.

janusw (changes requested):

docfx/documentation/getting-started-wpf.md
pauldendulk: oh, you refer to the '.Native' part. I can revert that for now, but  I think it will be temporary.

pauldendulk: @janusw I re-added the '.Native' because the purpose of this MR is to fix the user-agent problem which I think is a pretty big issue. The getting started is the first thing people try and it simply does not work. If I would see this I would move on to another project right away. 

Then again, keeping the '.Native' does not make that much sense either, with the current release there is no '.Native', with the rc.2 release it does not work, and my proposal is to change it in rc.2.

pauldendulk: @janusw In summary, I think this is small change which can be merged.

janusw: Yes, feel free to merge 👍

janusw (approved):

janusw:
Rebased to master, re-generated the docs, then merged.

Issue #1207, opened 2021-09-11
Split up the Mapsui.Native package


janusw:
Proposed by @pauldendulk in #1206:

> Also, I think that WPF should be in a separate package. The 'Native' phrase stems mostly from 'native mobile' in the context of Xamarin development. Also, I think it is a totally different user group.

There were some comments about this before (see #1034), which were not really picked up in the discussion of #1175. I see two options for the split-up:

1. Split Mapsui.Native into Mapsui.Win (UWP & WPF) and Mapsui.Mobile (iOS & Android). Or ...
2. Split it into four different packages: Mapsui.UWP, Mapsui.WPF, Mapsui.iOS and Mapsui.Android.

The latter provides clearer separation of the platforms, but there were some concerns about inflation of the number of packages.

pauldendulk:
My proposal was to split off a Mapsui.Wpf nuspec. 

Motivation:
- WPF is used independently of the other platforms. I certainly don't see WPF and UWP used together, they are both Windows, but if you build one you do not build the other. I don't like the idea of combining them in one package.
- I expect that WPF will be in use for longer period, so it will have it's own timeline. While the other platforms will be replaced more and more by Xamarin.Forms and MAUI.
- I could make sense to split the packages up even further to also add separate UWP, iOS ad Android packages. That also means adding 3 other nuspecs that need to be maintained. I think that is a big investment especially when it is not clear how long will will be maintaining them. If the nuget download number are low we may stop supporting them for the next release.

pauldendulk:
We also split off other platforms in a later PR. For instance, it would perhaps make sense to move UWP to a separate package because I don't think this is often used in combination with iOS and Android. But then again, I wonder if that is worth the investment.

pauldendulk:
A general remark. I understand that you want to move towards a more formal process. That's okay. We can move there step by step, but at the moment there is a release out that is not functioning. So there is some urgency here.

pauldendulk:
I added a PR: https://github.com/Mapsui/Mapsui/pull/1208

charlenni:
I would say a Mapsui.WPF, a Mapsui.Mobile (UWP, iOS, Android) and a Mapsui.Forms is ok. Later there could be a Mapsui.Maui.

janusw:
> I would say a Mapsui.WPF, a Mapsui.Mobile (UWP, iOS, Android) and a Mapsui.Forms is ok. Later there could be a Mapsui.Maui.

[...]

PR #1208, opened 2021-09-11
Move WPF to separate nuspec

pauldendulk:
See discussion here:
https://github.com/Mapsui/Mapsui/issues/1207


NuSpec/Mapsui.Wpf.nuspec
```diff
@@ -0,0 +1,34 @@
...
+    </description>
+    <tags>map maps mapping geo gis osm</tags>
+    <copyright>© Mapsui Contributers</copyright>
+    <language>en-US</language>
+    <dependencies>
+      <!-- [ or ] is including, ( or ) is excluding -->
+      <group targetFramework="net48">
+        <dependency id="Mapsui" version="$version$"/>
+        <dependency id="BruTile" version="[4.0.0,5.0.0)"/>
+        <dependency id="Newtonsoft.json" version="[11.0.1,)"/>
```
janusw, line 22: I think the dependencies need to be cleaned up here: Newtonsoft.Json, RichTextKit and Svg.Skia do not seem to be required (at least).

NuSpec/Mapsui.Wpf.nuspec
```diff
@@ -0,0 +1,34 @@
...
+  <metadata>
+    <id>Mapsui.Wpf</id>
+    <version>$version$</version>
+    <title>Mapsui</title>
+    <authors>Paul den Dulk</authors>
+    <owners>Paul den Dulk</owners>
+    <projectUrl>https://github.com/Mapsui/Mapsui</projectUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>
+      Mapsui - Library for mapping (native UI for different platforms)
```
janusw, line 12: The description should be adjusted ("UI for WPF" or similar) ...

janusw (changes requested):

janusw:
In general I'm okay with this, but I think we should go further with the separation (only splitting off WPF but leaving the rest together seems inconsistent).

If you're in a hurry with this, Paul, it can be merged when the minor comments above are fixed. But I would vote for continuing to separate packages in a follow-up PR.

We probably do not want to change the packaging scheme for every minor release, so we should really agree on a scheme, implement it and then make a final release 3.0.

janusw:
And one more comment: It's a bit of a shame that this sort of problem is not caught by any of the automatic CI builds.

Can we improve this situation somehow (without too much effort)?

pauldendulk:
> And one more comment: It's a bit of a shame that this sort of problem is not caught by any of the automatic CI builds.
> Can we improve this situation somehow (without too much effort)?

This is an advanced test. I don't have this in my other projects. I assume others do something like this, perhaps we can learn from their workflow. There might be tools for it. 

What we could do is create a script that updates some code to the [latest nuget](https://mapsui.com/documentation/nuget-of-latest-build.html?q=nuget) and runs some tests. It will be easier to test the core Mapsui project. For UI projects it will be harder. We could build the 'getting started' projects and at least test if it builds. It would catch the problem we just saw.

janusw:
> What we could do is create a script that updates some code to the [latest nuget](https://mapsui.com/documentation/nuget-of-latest-build.html?q=nuget) and runs some tests. It will be easier to test the core Mapsui project. For UI projects it will be harder. We could build the 'getting started' projects and at least test if it builds. It would catch the problem we just saw.

Well, I don't think we need any advanced UI tests for the start. Simply having the 'getting started' guides as VS projects and building them (without project references to Mapsui core, but using the latest nupkgs) should be enough. (Also there is already some value in having getting-started projects, instead of just describing how to set them up.)

Would be nice to do that sometime, but it should not be part of this PR. Please just address my review comments above, then this can be merged. Or do you want me to take over? This is the most urgent PR we have right now.

pauldendulk:
> Would be nice to do that sometime, but it should not be part of this PR. Please just address my review comments above, then this can be merged. Or do you want me to take over? This is the most urgent PR we have right now.

The thing is you made me think about the alternatives. Perhaps we should jump to a solution with [nugetizer](https://www.nuget.org/packages/nugetizer). I will look at this the coming days.

janusw:
> > Would be nice to do that sometime, but it should not be part of this PR. Please just address my review comments above, then this can be merged. Or do you want me to take over? This is the most urgent PR we have right now.
> 
> The thing is you made me think about the alternatives. Perhaps we should jump to a solution with [nugetizer]

I don't really know nugetizer and I only had a quick look at it, but from what I understand it seems like this mightly only help us get rid of the nuspec files (not sure if its worth to go there). In any case, we still have to decide how to cut our packages.

If you're not planning to go ahead with this PR in a timely manner, I guess we should at least fix Mapsui.Native.nuspec, because right now it is simply broken.

janusw:
> If you're not planning to go ahead with this PR in a timely manner, I guess we should at least fix Mapsui.Native.nuspec, because right now it is simply broken.

See #1214.

NuSpec/Mapsui.Wpf.nuspec
pauldendulk: Updated https://github.com/Mapsui/Mapsui/pull/1208/commits/bffc01ef07db36432231a8a937c8a7d3f240dbdb

NuSpec/Mapsui.Wpf.nuspec
pauldendulk: Updated https://github.com/Mapsui/Mapsui/pull/1208/commits/eabeab3a1cd444459457106828fc96ef53faa659

pauldendulk:
@janusw I processed your remark. You can review again.

janusw (approved):
LGTM. Thanks!

PR #1214, opened 2021-09-15
fix Mapsui.Native.nuspec

janusw:
* by re-adding Mapsui.Rendering.Xaml.dll
  (which was accidentally removed in dadd2989f69ce081c9fb22d469ccc4242446081c)


pauldendulk:
@janusw I have split up this file, so I think this PR has become irrelevant and could be close. Do you agree?

janusw:
Sure.

Issue #1211, opened 2021-09-13
Empty map when following UWP Getting Started


bertt:
When following the UWP Getting Started (https://github.com/Mapsui/Mapsui/blob/feature/add-getting-started-uwp/docfx/documentation/getting-started-uwp.md) using Mapsui.Native 3.0 rc 2 I get an empty (white) map.

There is a compiler warning about Svg.Skia:

```
Warning		Found conflicts between different versions of "Svg.Skia" that could not be resolved.
There was a conflict between "Svg.Skia, Version=0.5.0.0, Culture=neutral, PublicKeyToken=dafe96fe6c845a74" and "Svg.Skia, Version=0.5.7.1, Culture=neutral, PublicKeyToken=dafe96fe6c845a74".
    "Svg.Skia, Version=0.5.0.0, Culture=neutral, PublicKeyToken=dafe96fe6c845a74" was chosen because it was primary and "Svg.Skia, Version=0.5.7.1, Culture=neutral, PublicKeyToken=dafe96fe6c845a74" was not.
    References which depend on "Svg.Skia, Version=0.5.0.0, Culture=neutral, PublicKeyToken=dafe96fe6c845a74" [C:\Users\btemm\.nuget\packages\svg.skia\0.5.0\lib\netstandard2.0\Svg.Skia.dll].
        C:\Users\btemm\.nuget\packages\svg.skia\0.5.0\lib\netstandard2.0\Svg.Skia.dll
          Project file item includes which caused reference "C:\Users\btemm\.nuget\packages\svg.skia\0.5.0\lib\netstandard2.0\Svg.Skia.dll".
            C:\Users\btemm\.nuget\packages\svg.skia\0.5.0\lib\netstandard2.0\Svg.Skia.dll
    References which depend on "Svg.Skia, Version=0.5.7.1, Culture=neutral, PublicKeyToken=dafe96fe6c845a74" [].
        C:\Users\btemm\.nuget\packages\mapsui\3.0.0-rc.2\lib\netstandard2.0\Mapsui.Rendering.Skia.dll
          Project file item includes which caused reference "C:\Users\btemm\.nuget\packages\mapsui\3.0.0-rc.2\lib\netstandard2.0\Mapsui.Rendering.Skia.dll".
            C:\Users\btemm\.nuget\packages\mapsui\3.0.0-rc.2\lib\netstandard2.0\Mapsui.Rendering.Skia.dll
[...]

PR #1215, opened 2021-09-16
update the dependencies on Svg.Skia in the nuspec files

janusw:
* using the latest version
* matching the package refs in Mapsui.Rendering.Skia etc
* this is supposed to fix issue #1211

### DIFF
--- a/NuSpec/Mapsui.Native.nuspec
+++ b/NuSpec/Mapsui.Native.nuspec
@@ -20,7 +20,7 @@
         <dependency id="BruTile" version="[4.0.0,5.0.0)"/>
         <dependency id="Newtonsoft.json" version="[11.0.1,)"/>
         <dependency id="SkiaSharp" version="[2.80.3,3.0.0)"/>
-        <dependency id="Svg.Skia" version="[0.5.0,0.6)"/>
+        <dependency id="Svg.Skia" version="[0.5.7.1,0.6)"/>
         <dependency id="SkiaSharp.Views" version="[2.80.3,3.0.0)"/>
         <dependency id="Topten.RichTextKit" version="[0.4.145,0.5)"/>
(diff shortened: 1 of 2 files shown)


pauldendulk:
General comment: This is a common problem and this is the kind of maintenance I was referring to in an earlier comment. If the nuget is directly generated from the csproj this problem is mitigated.

pauldendulk (commented):
Svg.Skia is a packagereference to Mapsui, not to the Mapsui.Native project, so I would think it should not be in Mapsui.Native at all. I comes indirectly from the Mapsui package. Perhaps I am wrong, but then I would like to know why it is needed.

pauldendulk:
Another general comment. When we upgrade to a minor or patch of some package there is usually a reason for that, It fixes a bug of some sort. So the users should also get this upgraded version when installing the package. If the the lower limit of the version is not upgraded they will get the lowest version installed. So, we should always keep the lower version in sync with the installed version in the project (or generate from the csproj).

janusw:
> Svg.Skia is a packagereference to Mapsui, not to the Mapsui.Native project, so I would think it should not be in Mapsui.Native at all.

True. I am removing it in #1214 (along with two other deps of the same kind).

janusw:
> So, we should always keep the lower version in sync with the installed version in the project (or generate from the csproj).

I was not sure about this, because I thought it might be better to keep compatibility with lower versions (if possible), in case someone has a reason to stick with the lower version.
But yes, it seems like nuget always uses the lowest possible version, and I agree that it's probably better to bump the nuspec dep together with the package ref then. (Will do this for future updates.)

pauldendulk (approved):
Upgrade is necessary. Mapsiu.Native will be removed after this MR, thats no problem.

PR #1217, opened 2021-09-17
Downgrade sqlite to 1.6.292

pauldendulk:
I saw the upgrade in [this PR](https://github.com/Mapsui/Mapsui/pull/1203) and wondered if that could cause a problem and it did. The WPF sample throw an exception when trying to open mbtiles. Apparently there were api changes between 1.6 and 1.8. This is not according to semver but happens a lot.

BruTile.MbTiles was compiled against 1.6.292. The nuget refers to that version:
https://www.nuget.org/packages/BruTile.MbTiles/

### DIFF
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Android/Mapsui.Samples.Forms.Android.csproj
@@ -123,7 +123,7 @@
       <Version>2.80.3</Version>
     </PackageReference>
     <PackageReference Include="sqlite-net-pcl">
-      <Version>1.8.116</Version>
+      <Version>1.6.292</Version>
     </PackageReference>
     <PackageReference Include="Svg.Skia">
       <Version>0.5.7.1</Version>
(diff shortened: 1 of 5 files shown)


pauldendulk:
In general I think we should upgrade a package if we have a specific reason for it. There are many packages that are not compatible with each other, sometimes because of api changes sometimes because of bugs in higher or lower versions. In the javascript world this is even a bigger problem. Sometimes it is hard to find any set of packages that works for all needed features.

Note, a 'specific reason' can be that everyone else is using version x, so if we do too that makes it easier for our users. So upgrading now and then is a good also. 

And if we upgrade lets do it after a major release, not while in the release candidate (RC) phase.

janusw:
> I saw the upgrade in [this PR](https://github.com/Mapsui/Mapsui/pull/1203) and wondered if that could cause a problem and it did.

Maybe you should be a bit stricter in your reviews. If a PR breaks the CI build and you wonder whether it could cause problem, the appropriate reaction would be to utter your concerns, do further checks or request further checks from the author.

In fact I saw the CI breakage and planned to look into it. But before I managed to do that, you just merged it.

janusw (approved):
Approved. This is simply a revert of b8b267a29b7b38f4fc7835d7fd43c2cfe6a4f378, which should have been rejected in the first place.

PR #1219, opened 2021-09-17
Add Mapsui.iOS.nuspec

pauldendulk:
Added the same references as in the csproj (this is what happens if you generate the nuget from the csproj). Not completely sure if this is enough because Mapsui.Samples.iOS has a truck load of nugets that could be important because of 'bait and switch' or 'version resolution'. This is not something I will test.

### DIFF
--- a/NuSpec/Mapsui.iOS.nuspec
+++ b/NuSpec/Mapsui.iOS.nuspec
@@ -1,0 +1,29 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd">
+  <metadata>
+    <id>Mapsui.iOS</id>
+    <version>$version$</version>
+    <title>Mapsui</title>
+    <authors>Paul den Dulk</authors>
+    <owners>Paul den Dulk</owners>
+    <projectUrl>https://github.com/Mapsui/Mapsui</projectUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>
+      Xamarin iOS map components based on the Mapsui library
+    </description>
+    <tags>map maps mapping geo gis osm</tags>
+    <copyright>© Mapsui Contributors</copyright>
+    <language>en-US</language>
+    <dependencies>
+      <!-- [ or ] is including, ( or ) is excluding -->
+      <group targetFramework="Xamarin.iOS">
+        <dependency id="Mapsui" version="$version$"/>
+        <dependency id="SkiaSharp.Views" version="[2.80.3,3.0.0)"/>
+        <dependency id="System.Memory" version="[4.5.4,5.0.0]"/>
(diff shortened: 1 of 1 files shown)


janusw:
> Added the same references as in the csproj (this is what happens if you generate the nuget from the csproj).

I wonder why the reference to System.Memory is in there. It is not being used explicitly in the code. Maybe it can be removed?


> Not completely sure if this is enough because Mapsui.Samples.iOS has a truck load of nugets that could be important because of 'bait and switch' or 'version resolution'.

1) I don't think any of the package refs of Mapsui.Samples.iOS are relevant for Mapsui.UI.iOS.
2) I think most of the refs of Mapsui.Samples.iOS are unnecessary after all.

> This is not something I will test.

It is hard to test at all, because the build scripts are not adjusted to the new package yet.

janusw (changes requested):
When moving the Mapsui.UI.iOS parts to a new package, they also should be removed from Mapsui.Native.nuspec.

pauldendulk:
> When moving the Mapsui.UI.iOS parts to a new package, they also should be removed from Mapsui.Native.nuspec.

As discussed in another thread Mapsui.Native will be deprecated after it is replace with separate nuspecs. All other nuspecs are now submitted as PRs so it is of no use to update Mapsui.Native.

pauldendulk:
> I wonder why the reference to System.Memory is in there. It is not being used explicitly in the code. Maybe it can be removed?

This PR creates a nuspec for the current csproj, which is using System.Memory. Weather the csproj should use System.Memory is a separate issue. It should not block this PR. 

A possible reason why it was once added may be that is solves a version conflict of System.Memory. Similar to this https://github.com/Mapsui/Mapsui/issues/1093. But this is something we should not spend time on. This whole platform might be deprecated soon.

pauldendulk:
> It is hard to test at all, because the build scripts are not adjusted to the new package yet.

This is another PR.

janusw:
> > It is hard to test at all, because the build scripts are not adjusted to the new package yet.
> 
> This is another PR.

Well ok, IMHO all of these change should ideally have been part of a single PR, but since you did it like this now, I guess this one is ok.

janusw (approved):

janusw:
... only one more questions that just came to my mind ...

> Added the same references as in the csproj (this is what happens if you generate the nuget from the csproj).

If this is anyway just generated from the csproj, do we even still need an explicit nuspec at all?

pauldendulk:
> If this is anyway just generated from the csproj, do we even still need an explicit nuspec at all?

It is currently not generated from the csproj, but this is just the procedure that I am following. The logic is that if this works in our sample projects it should work in our nuget packages. 

I would prefer to generate the nugets from the csproj though. This is what I am doing in another project making use of nuget.build.packaging and I am looking for something like this to use for Mapsui.

PR #1222, opened 2021-09-17
Remove mac sample

pauldendulk:
Because it breaks the appveyor build and I don't know how to fix it. It is also not possible to run the Mac solution from the main project on windows. The mapsui.mac.sln could be used for that. 

It would perhaps be handy to have the map project as part of mapsui.sln so that nuget packages get upgraded all in one.


pauldendulk:
oh, I now get the same error in Visual Studio. First time. Perhaps it is a VS update. I updated mine a week ago but if the Mac project does not change it would not rebuild my projects.

janusw:
Totally disagree. This is certainly not the right thing to do here.

I think the errors that you refer to are the following:
```
Mapsui.Samples.Forms.Mac\AppDelegate.cs(35,13): error CS0012: The type 'Func<,>' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.
Mapsui.Samples.Forms.Mac\AppDelegate.cs(35,27): error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.
```

If you look closely, they first appeared in b8b267a29b7b38f4fc7835d7fd43c2cfe6a4f378. So I think it should be sufficient to revert this commit, as you proposed in #1217. This PR should be closed.

janusw:
The build errors are gone after #1217 has been merged. Closing.

pauldendulk:
@janusw oh, I had not made the connection with the broken build, but this is indeed the problem. The error seemed quite unrelated, an sqlite upgrade cause the build to fail for the mac project, obscure.

The reason I tested is that I saw the sql upgrade and wondered if that could break something so I tested it myself.

pauldendulk:
@janusw btw, why did you upgrade sqlite, was I a conscious decision or just a mistake?

janusw:
> @janusw btw, why did you upgrade sqlite, was I a conscious decision or just a mistake?

That was a conscious decision, or let's say an experiment (but not a successful one, obviously).

PR #1223, opened 2021-09-17
These multi target wpf head projects confused the command line compiler

pauldendulk:
And there is no real need for it.


pauldendulk:
This change is now part of another PR.

PR #1224, opened 2021-09-17
Feature/update build scripts for new nuspecs

pauldendulk:
Updated the build scripts for the new nuspecs.
Also removed the multi targetting of two samples because this could cause random fails of the build script, and has no real purpose.


pauldendulk:
The build server should succeed ones the nuspec PR are merged. Let's wait for that.

janusw:
> Updated the build scripts for the new nuspecs.

This looks perfectly fine.

However, the Mac scripts should be modified as well (the iOS and Android packages can and should be built on Mac). I can add and test that if you want.

> Also removed the multi targetting of two samples because this could cause random fails of the build script, and has no real purpose.

I don't really understand this part (and I'm not comfortable reviewing it): What is the origin of the issues? Why do you say the multi-targeting has no real purpose? And why does it need to be included here (it's totally unrelated to the build scripts and nuspecs)?

janusw:
> > Also removed the multi targetting of two samples because this could cause random fails of the build script, and has no real purpose.
> 
> Why do you say the multi-targeting has no real purpose? 

Apparently Mapsui.UI.Wpf has the same multi-targettting, and people might want to run it with .NET Core, no?

pauldendulk:
> However, the Mac scripts should be modified as well (the iOS and Android packages can and should be built on Mac). I can add and test that if you want.

Yes, please, I prefer not to test on mac.

> I don't really understand this part (and I'm not comfortable reviewing it): What is the origin of the issues? Why do you say the multi-targeting has no real purpose? And why does it need to be included here (it's totally unrelated to the build scripts and nuspecs)?

It is related in the sense that if you test the build scripts they fail (sometimes) because of the multi-targetting. I usually fix this by doing a manual rebuild from visual studio, but this should not be needed.

> Apparently Mapsui.UI.Wpf has the same multi-targettting, and people might want to run it with .NET Core, no?

I don't actually know how you could do that from visual studio, but it is probably possible, but even if you run it you will see an application that is functionally equivalent. It was added as part of a PR that adds multi-targetting for dlls that are in the nuget, which is in itself useful.

It maybe useful in some sense but it blocks a primary part of our flow, and if we do not use it and users don't ask for it I think we should remove it. If there is a request to add it again then we should require a solution that is compatible with our build scripts.

janusw:
> > However, the Mac scripts should be modified as well (the iOS and Android packages can and should be built on Mac). I can add and test that if you want.
> 
> Yes, please, I prefer not to test on mac.

Alright, I have rebased the branch to current master and added a commit that updates the Mac build script.

janusw:
I'm happy with the build-script changes at this point (these parts could be merged), but I'm still not fully convinced about removing the multi-targeting ...

> Also removed the multi targetting of two samples because this could cause random fails of the build script

How do these fails actually look like? And where can one observe them in the CI?

And also: If one has to do this, shouldn't one rather use .NET Core for the samples and remove the old .NET Framework 4.8 instead? (This would be more future-proof IMHO.)

pauldendulk:
> How do these fails actually look like? And where can one observe them in the CI?

I don't see it in CI. Just on my local machine. It appears to be caused by manual builds in Visual Studio. Perhaps it randomly selects one of the targets of the head project.

> And also: If one has to do this, shouldn't one rather use .NET Core for the samples and remove the old .NET Framework 4.8 instead? (This would be more future-proof IMHO.)

Ah yes! That was my though when I worked on this yesterday. That is the right thing to do.

janusw:
> > And also: If one has to do this, shouldn't one rather use .NET Core for the samples and remove the old .NET Framework 4.8 instead? (This would be more future-proof IMHO.)
> 
> Ah yes! That was my though when I worked on this yesterday. That is the right thing to do.

Ok, then, I guess this I can agree to (even if the problem only occurs on your machine ;)

bertt (changes requested):
don't need net4.8 here

pauldendulk:
> don't need net4.8 here

Adjusted to netcoreapp3.1. Can be reviewed again.

janusw (approved):
This looks ok now. Thanks!

janusw:
> This looks ok now.

@bertt, do you agree? Your change request is blocking the merge right now ...

bertt (approved):

PR #1225, opened 2021-09-18
Fix desktop samples

pauldendulk:
Why: The desktop samples where not loaded in the wpf application
because the LoadAssembly method was inside a sample that was
moved.
Note: This is to prepare for cleanup of the desktop samples.


pauldendulk:
To clarify. All the samples are loaded by reflection. It goes through the assemblies and looks voor classes with the ISample interface. But when an project reference is added the assembly is not loaded. This surprised me but apparently there is some kind of optimization done by the compiler that removes assemblies that are not actually used. The workaround is to add a random method that does nothing and call it from the main app.

PR #1226, opened 2021-09-18
Added ButtonWidget as replacement for SvgButton in MapView

charlenni:
Added a ButtonWidget, so that SvgButton in MapView could replace. Advantage is, that for ButtonWidget not layout is needed. Is one step in direction to inherit MapView from MapControl in Forms. Goal is to get rid of many wrappers in MapView.


janusw:
> Is one step in direction to inherit MapView from MapControl in Forms. Goal is to get rid of many wrappers in MapView.

I definitely like that plan. I also like the fact that the Forms-only SvgButton is replaced by a ButtonWidget that lives in the Mapsui base library. That helps to make the code more generic. 👍 

I also did some light testing of the new buttons via Mapsui.Samples.Forms on Android, iOS and Mac. The buttons seem to work well on all three platforms (from a user perspective). Haven't looked at the actual code yet ...

janusw (commented):
One thing that is apparently still missing is a Xaml renderer implementation. Do you plan to still add this, or can we live without it?

janusw:
Another thing that I like is the fact that the button size and placement is controlled by MapView properties (instead of being hard-coded as before) :+1:

janusw:
Functionality-wise there is a bit of overlap/duplication of the new ButtonWidget (and its usage in MapView) with the existing ZoomInOutWidget. In principle one only needs one of the two (and the ButtonWidget is more general, of course). There are even samples where both of them are shown (like the "Add Polyline" sample etc). Would it be worth to get rid of the ZoomInOutWidget? (This could of course be done in a follow-up PR, if necessary, I just want to discuss if it makes sense.)

charlenni:
Yes, the ZoomInOutWidget could be dropped. It was an example, what is doable with widgets, but I don‘t know, if someone uses it. Perhaps a deprecated message would be useful.

Another thing is the VerticalAlignment and HorizontalAlignment of ButtonWidget. It isn‘t used for placing the button. Only the Envelop is used. I would like to get rid of, but it isn‘t possible, because it is declared in IWidget/Widget as public.

Further, the tap on the buttons could be in the not used corners of the buttons. If you tap there, a click is registered, even if there isn‘t something. Not so nice, but much easier to do. BTW the GetInfo... doesn‘t check for hitting widgets.

Properties for the pictures of the buttons would be nice.

charlenni:
Regarding the Xaml renderer: I don‘t know, if I should do this. Xaml is not my world. I try to look into this, but will not invest to much time.

janusw:
> Regarding the Xaml renderer: I don‘t know, if I should do this. Xaml is not my world.

Personally I also don't care too much about the Xaml renderer. I've never used it and probably never will. I'm sure Skia is the future, and IIRC @pauldendulk mentioned that he'd like to remove (or deprecate) it after 3.0.

I guess our options here are:
1) Leave the ButtonWidget without Xaml rendering. After all it is a new component that is only used in MapView so far (i.e. the Forms world where the Xaml renderer is not relevant).
2) Add a Xaml-renderer implementation for ButtonWidget (hoping that is not too much work). Maybe @pauldendulk could help here?
3) Defer this PR after 3.0, when the Xaml renderer will be removed anyway.

pauldendulk:
>> Regarding the Xaml renderer: I don‘t know, if I should do this. Xaml is not my world.

> Personally I also don't care too much about the Xaml renderer. I've never used it and probably never will. I'm sure Skia is the future, and IIRC @pauldendulk mentioned that he'd like to remove (or deprecate) it after 3.0.

Removing the xaml renderer will be the first thing to do after this release. Earlier in the project we already decided to accept that some things don't work in xaml. So your option 1 is acceptable.

pauldendulk (approved):
I can not do an proper review of this PR. There are so many things to do and and Forms is one of the things I try to ignore. The widget renderer solution has been used before and works well. 

Since @janusw comments are positive and xaml rendering can be ignore I will approve and merge this PR so that it can go into rc.3.

PR #1228, opened 2021-09-19
Remove fields from csproj that are in nuspec

pauldendulk:
Some of these were added by me in an attempt to build nuget from csproj but not the ones in Mapsui.Forms.csproj.

### DIFF
--- a/Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj
+++ b/Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj
@@ -19,8 +19,6 @@
 
   <PropertyGroup>
     <DebugType>portable</DebugType>
-    <PackageId>Mapsui</PackageId>
-    <Description>Mapsui</Description>
   </PropertyGroup>
 
 </Project>
--- a/Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
+++ b/Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
@@ -1,15 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <PackageVersion>2.0</PackageVersion>
-    <Authors>Mapsui Contributors</Authors>
-    <PackageIconUrl>https://github.com/Mapsui/Mapsui/blob/master/logo/png/icon.png</PackageIconUrl>
     <NeutralLanguage>en</NeutralLanguage>
-    <PackageProjectUrl>https://github.com/Mapsui/Mapsui</PackageProjectUrl>
-    <Summary>Xamarin.Forms implementation of Mapsui</Summary>
-    <PackageTags>Mapsui Xamarin.Forms map maps mapping</PackageTags>
-    <Description>Xamarin.Forms implementation of Mapsui</Description>
-    <PackageId>Mapsui.Forms</PackageId>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
     <DefineConstants>TRACE;__FORMS__;DEBUG;NETSTANDARD1_3</DefineConstants>
--- a/Mapsui/Mapsui.csproj
+++ b/Mapsui/Mapsui.csproj
@@ -18,9 +18,6 @@
   </ItemGroup>
   <PropertyGroup>
     <DebugType>portable</DebugType>
-    <PackageId>NotMapsui</PackageId>
-    <PackageVersion>1.0.0</PackageVersion>
-    <Description>Mapsui</Description>
   </PropertyGroup>
 
 </Project>


Mapsui/Mapsui.csproj
```diff
@@ -18,9 +18,6 @@
   </ItemGroup>
   <PropertyGroup>
     <DebugType>portable</DebugType>
-    <PackageId>NotMapsui</PackageId>
```
janusw, line 21: Yes, this should be removed. The Mapsui base nuget cannot be built from the Mapsui csproj alone, and I have stumbled about the 'NotMapsui' before and found it very confusing.

Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj
```diff
@@ -19,8 +19,6 @@
 
   <PropertyGroup>
     <DebugType>portable</DebugType>
-    <PackageId>Mapsui</PackageId>
```
janusw, line 22: Ok with me as well (same arguments as for Mapsui.csproj).

Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
```diff
@@ -1,15 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
-    <PackageVersion>2.0</PackageVersion>
```
janusw, line 4: Here I'm not so sure. The Mapsui.Forms package can in principle be built from the csproj (in contrast to Mapsui and Mapsui.Rendering.Skai), so why not leave it in?

pauldendulk: Yes, you can argue about this. 

Argument to remove: Duplication is a bad thing in general. It is more maintenance, more likely that one will be updated while the other is not, more code that we need to manage. I went through the code and it made me think. Is this relevant for the nuget package or not? It is an extra burden, a small one, but still.

Argument to keep: You can build a nuget package from the csproj and from VS solution menu. That can be of practical use, but only if this is actually used to build manual nuget packages. I do not. I create packages to test the nuspec, so don't use the csproj. I assigned this PR to @charlenni because perhaps he uses it.

janusw: Agreed. Both arguments are valid. I tend to lean towards 'keep', both I don't have a strong opinion on it. @charlenni, what you do you think?

One could also think about removing Mapsui.Forms.nuspec and make the scripts build the package via the csproj. Only downside is that the build procedure is slightly inconsistent then among Mapsui and Mapsui.Forms.

pauldendulk:
> One could also think about removing Mapsui.Forms.nuspec and make the scripts build the package via the csproj. Only downside is that the build procedure is slightly inconsistent then among Mapsui and Mapsui.Forms.

But it will not have the correct dependency on the Mapsui nuget, unless you fiddle with the script. But yes, we should have something like that at some point. In that case the should be removed nuspec, but where are not there yet today.

Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
charlenni: Sorry, I'm out of this discussion, because I don't know anything about the things you discuss here ;-). Perhaps someone else has more knowledge about this.

pauldendulk: ah, so that means this is not here because you need it, so my vote would to be to remove it.

charlenni: I never touched one of the nuspec or csproj files by myself.

janusw:
> > One could also think about removing Mapsui.Forms.nuspec and make the scripts build the package via the csproj. Only downside is that the build procedure is slightly inconsistent then among Mapsui and Mapsui.Forms.
> 
> But it will not have the correct dependency on the Mapsui nuget, unless you fiddle with the script.

Ah, ok. If there can be issues with the dependency on the Mapsui base package, then I guess the Mapsui.Forms part is also fine with me. (Once we switch to csproj packaging, this can easily be added again.)

janusw (approved):

sailman342:
Hi all of you,

I see you work hard on Mapsui those times, congratulations :-)

I gave up using Mapsui due to build problems with dependencies I could not solve, having plenty of build complains and not knowing how to solve them. For instance NewtonSoft was a regular case !

I have plenty of time to develop my App so it was not of major importance for me and I went to Bing .... but I would like one day to come back to Mapsui given the power of the available interface !

Since some time I am working with .Net 6 preview and Visual Studio 2022 preview. The management of the dependencies is totally different and solves many issues ! 

image


As you can see, the Nuget dependencies are not at the top level anymore, it seems they are embedded within the Mapsui package, they don't appear either in the project file now ..

image


This solved a lot of dependency problems I had in other areas as well . 

Knowing Janus is working on Mac and that it will be fully supported in the next versions, including Maui, I am wondering if it could help you to move directly to .Net 6 and Visual Studio 2002 which are very clos to release ?

Kindest regards to the team and many thanks to all of you for your efforts

 (signed a newbie)

pauldendulk:
@sailman342 I think this improvement is primarily because of the use of PackageReference. The indirect dependencies are not added to the project anymore. This was changed in Mapsui with a MR of @inforithmics https://github.com/Mapsui/Mapsui/pull/971.

There will still be an indirect dependency on newtonsoftjson though. Unfortunate that this dependency causes so much problems because it is used by only small parts of the code. We should refactor those parts out to separate assemblies.

sailman342:
Thanks a lot :-)

PR #1229, opened 2021-09-19
Fix some spelling errors in comments

pauldendulk:
Why: Some of these kept distracting me. I did a bit more
using resharper 'spelling issues', but not all. We should
do a full scan with it.


janusw (approved):
Ok with me. Thanks for the cleanup!

janusw:
Ah, wait, I have another one. I'll just add it here if you don't mind ...

PR #1237, opened 2021-09-23
Downgrade skiasharp to 2.80.2

pauldendulk:
Not sure at all if this solves #1216 but if we have such confusing issues it is better to have our dependencies and indirect dependencies in sync.

### DIFF
--- a/Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj
+++ b/Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SkiaSharp" Version="2.80.3" />
+    <PackageReference Include="SkiaSharp" Version="2.80.2" />
     <PackageReference Include="Svg.Skia" Version="0.5.7.1" />
     <PackageReference Include="Topten.RichTextKit" Version="0.4.145" />
   </ItemGroup>
--- a/Mapsui.UI.Android/Mapsui.UI.Android.csproj
+++ b/Mapsui.UI.Android/Mapsui.UI.Android.csproj
@@ -79,7 +79,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="SkiaSharp.Views">
-      <Version>2.80.3</Version>
+      <Version>2.80.2</Version>
     </PackageReference>
     <PackageReference Include="System.Memory">
       <Version>4.5.4</Version>
--- a/Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
+++ b/Mapsui.UI.Forms/Mapsui.UI.Forms.csproj
@@ -13,8 +13,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="SkiaSharp" Version="2.80.3" />
-    <PackageReference Include="SkiaSharp.Views.Forms" Version="2.80.3" />
+    <PackageReference Include="SkiaSharp" Version="2.80.2" />
+    <PackageReference Include="SkiaSharp.Views.Forms" Version="2.80.2" />
     <PackageReference Include="Xamarin.Forms" Version="5.0.0.2083" />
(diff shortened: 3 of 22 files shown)


janusw:
> Not sure at all if this solves #1216

AFAICS it does not. What would help there would be a downgrade of (the dependency on) Svg.Skia.


> but if we have such confusing issues it is better to have our dependencies and indirect dependencies in sync.

I think you refer to the indirect dependency of Svg.Skia on SkiaSharp here. If I look at https://www.nuget.org/packages/Svg.Skia/, then I see that version 0.5.7.1 (which we use in Mapsui) requires SkiaSharp >= 2.80.2. So ideally it should also work with 2.80.3 (provided that no relevant API changes occurred, which should not happen in such a minor version change). Of course one should check that this is actually the case, but so far I'm not aware of any problems in this direction.

In summary, I'm a bit reluctant to do this downgrade, if it does not fix an actual problem. I'm fine with downgrading the minimum required version in the nuspec to 2.80.2 again, but the package refs in the library should rather use 2.80.3, if there are no API differences between the two.

In general, I think it would be useful to cover as much as possible of such potential package-update issues with unit tests. In an ideal world, running the unit tests alone would be sufficient to decide whether one can update a package to a particular version or not.

pauldendulk:
> > Not sure at all if this solves #1216
> 
> AFAICS it does not. What would help there would be a downgrade of (the dependency on) Svg.Skia.

The error is a MethodNotFoundException in a call from richtextkit to skiasharp. I don't see how Svg.Skia could be involved here. It has a dependency on skiasharp, but how should this cause this problem? Is the wrong dll copied?  However, I have no alternative explanation either. Perhaps it is some bug in the Xamarin linker that has random effects. 
 
> > but if we have such confusing issues it is better to have our dependencies and indirect dependencies in sync.
> 
> I think you refer to the indirect dependency of Svg.Skia on SkiaSharp here. If I look at https://www.nuget.org/packages/Svg.Skia/, then I see that version 0.5.7.1 (which we use in Mapsui) requires SkiaSharp >= 2.80.2. So ideally it should also work with 2.80.3 (provided that no relevant API changes occurred, which should not happen in such a minor version change). Of course one should check that this is actually the case, but so far I'm not aware of any problems in this direction.

I am fully aware of how semver is supposed to work. I also have enough experience  to know that this is not how it works in practice.  Every version of skiasharp has bugs somewhere, and that is one of the better libraries. RichTextKit and Svg.Skia have been developed with 2.80.2, so lets use that version.

The >= 2.80.2 thing is just what everyone is doing (as is Mapsui) to allow the user some flexibility in case they really need to upgrade. If you clone the Svg.Skia repository you can see they are using 2.80.2, anything above is never tested in Svg.Skia.

> In summary, I'm a bit reluctant to do this downgrade, if it does not fix an actual problem. I'm fine with downgrading the minimum required version in the nuspec to 2.80.2 again, but the package refs in the library should rather use 2.80.3, if there are no API differences between the two.

No, the entire idea of keeping things in sync is that our project and our users get the same dlls.

> In general, I think it would be useful to cover as much as possible of such potential package-update issues with unit tests. In an ideal world, running the unit tests alone would be sufficient to decide whether one can update a package to a particular version or not.

To test the nuget packages themselves we need some more advanced tests as we discussed earlier. Possible but not easy.

I would like to release a version of Mapsui with SkiaSharp 2.80.2 today.

janusw:
Well, ok, I don't have the feeling that it will help with anything, but if you're dead set on downgrading, I certainly won't stand in the way :)

pauldendulk:
> Well, ok, I don't have the feeling that it will help with anything, but if you're dead set on downgrading, I certainly won't stand in the way :)

I don't expect it to solve the MissingMethodException issue but there is a small chance that it will. For me it is important to exclude one possible cause. Another reason I want to release is that the new version will have a minimal version in the nuspec for Svg.Skia. Also not a likely cause of the problem but just not right. Users now run with 0.5.0.0 while we develop with 0.5.7.1.

I see that there is an inconsistency in the Xamarin.Forms reference as well. I will look at that next.

PR #1238, opened 2021-09-24
Feature/xamarin forms down to 4.6.0.1180

pauldendulk:
@charlenni @janusw The Xamarin.Forms version was not in sync with the dependency that means our users get a different version of Xamarin.Forms than what we compile and test against. A possible cause of problems.


pauldendulk:
@charlenni @janusw I see that the version of [SkiaSharp.Views.Forms](https://www.nuget.org/packages/SkiaSharp.Views.Forms/2.80.2) that we use depends on Xamarin.Forms 4.6.0.772. To keep everything in sync we should downgrade the version in our solution (no change in nuspec). This would be better. Is there a reason not to?

pauldendulk:
Actually I just pushed another commit to downgrade our solution version. Is there an objection against this?

pauldendulk:
The build server has problems with the downgrade. I am going for the upgrade.

PR #1241, opened 2021-09-24
Add netcoreapp3.1 to wpf nuspec

pauldendulk:


pauldendulk:
Merging because I want to test with nuget of latest build

PR #1245, opened 2021-09-25
Update getting-started-xamarin-forms.md

bertt:


pauldendulk:
@bertt I like keeping the sample super simple and in sync with the other samples, but the additional stuff might be useful as well. The transformation is important for the scalebar widget. Perhaps this could all be in a separate 'widgets' tutorial. Charlenni wrote the original, what do you think @charlenni?

pauldendulk:
If we make any changes here let's do that after we use compiled code samples. See https://github.com/Mapsui/Mapsui/issues/1409

Issue #1125, opened 2021-05-08
WinUI MapControl Replacement


robloo:
# Implement a WinUI Version of the Mapsui MapControl

The future of the UWP platform as a whole is questionable. However, Microsoft is continuing to support the UI layer with WinUI 3.0 which should be released and ready in November 2021. That release is missing a key control: MapControl. Mapsui is identified as one of the most promising alternatives to fill that feature gap. 

Below is a tentative list of things that need to be implemented or fixed before Mapsui seems viable in production apps. This is focused on the current UWP version of the control which seems less maintained than others. Consider this issue more of an epic and I can branch-out separate issues as needed.

## Open Issues

**High priority**

 - [ ] Make a WinUI version of the MapControl using the UWP control as a base. ~Possibly deprecate the UWP control at this point.~
 - [ ] UWP control doesn't work with Mapsui 3.0 at all. The app crashes. (potentially #978 )
 - [ ] Invalidation and reloading of tiles isn't working correctly after panning. Panning too far will show tiles in a very pixelated low resolution.
 - [ ] The Skia dependencies conflict with apps that use the latest SkiaSharp. With Mapsui 2.0.3 there are warnings such as:

   > Detected package version outside of dependency constraint: Mapsui 2.0.3 requires SkiaSharp.Views (>= 1.68.2.1 && < 1.69.0) but version SkiaSharp.Views 2.80.2 was resolved.

   This should be fixed with Mapsui v3 but the version requirements really should never be this constrained in the future. Generally, >= to a version only and not put a <= minor update.
 - [ ] ~It's possible to zoom out so far that the map goes into an error state. It appears to go outside of the viewport when zooming back in.~
       - Should be restricted in the Viewport by the application.

**Lower priority**

 - [ ] World wrapping #518. This is necessary for the user experience.
 - [ ] Prevent panning past the edges of the map when the zoom factor allows such a restriction. This ensures the map will always be 100% visible (along with world-wrapping).
 - [ ] Add a Pins collection to greatly simplify showing pins on a map

Updated as of Mapsui 2.0.3

## Additional Comments

I can put resources on fixing the high priority items. Would like to get directional feedback though and discuss how each should be addressed first.

charlenni:
If I see it correct, then there isn't a SkiaSharp version for WinUI. So it would be difficult to add a WinUI version to MapsUI.

robloo:
@charlenni

> If I see it correct, then there isn't a SkiaSharp version for WinUI.

[...]

PR #1246, opened 2021-09-25
WinUI (Work in progress)

inforithmics:
Initial WinUI Mapsui applied the changes mentioned in

https://github.com/Mapsui/Mapsui/issues/1125

### DIFF
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -18,6 +18,8 @@ namespace Mapsui.UI.Android
 namespace Mapsui.UI.iOS
 #elif __UWP__
 namespace Mapsui.UI.Uwp
+#elif __WINUI__
+namespace Mapsui.UI.WinUI
 #elif __FORMS__
 namespace Mapsui.UI.Forms
 #else
--- a/Mapsui.UI.Uwp/ExtensionMethods/PointExtensions.cs
+++ b/Mapsui.UI.Uwp/ExtensionMethods/PointExtensions.cs
@@ -1,6 +1,10 @@
 ﻿using Windows.Foundation;
 
+#if __WINUI__
+namespace Mapsui.UI.WinUI
+#else
 namespace Mapsui.UI.Uwp
+#endif
 {
     public static class PointExtensions
     {
--- a/Mapsui.UI.Uwp/MapControl.cs
+++ b/Mapsui.UI.Uwp/MapControl.cs
@@ -21,6 +21,17 @@ using Windows.Devices.Sensors;
 using Windows.Foundation;
 using Windows.Graphics.Display;
 using Windows.System;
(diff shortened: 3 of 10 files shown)


inforithmics:
I made new WinUI csproj for the samples and the Mapsui.UI Control where it made sense I used a link to the uwp files and conditional compiled it with __WINUI__

Mapsui.Geometries/Mapsui.Geometries.csproj
```diff
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <Configurations>Release;Debug</Configurations>
+    <RuntimeIdentifiers>win10-x86;win10-x64;win10-arm64</RuntimeIdentifiers>
```
pauldendulk, line 5: Are the RuntimeIdentifiers necessary for netstandard projects?

inforithmics: They are a workaround so that it compiles. Because Mapsui.UI.WinUI searches them in the references Projects. Without out this it gives a compilation error that it cannot find the runtime win10-x64 ...

inforithmics: This was the Error Message.

C:\Program Files\dotnet\sdk\5.0.400\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(241,5): error NETSDK1047: Assets file 'C:\projects\mapsui\Mapsui.Geometries\obj\project.assets.json' doesn't have a target for 'netstandard2.0/win10-x86'. Ensure that restore has run and that you have included 'netstandard2.0' in the TargetFrameworks for your project. You may also need to include 'win10-x86' in your project's RuntimeIdentifiers. 

It could be that upgrading the sdk to 6.0.0-rc.1 could solve this problem but I don't know for sure.

pauldendulk:
When I update master after this merge and start Mapsui.Samples.WinUI I get this exception:

System.Reflection.TargetInvocationException: 'Exception has been thrown by the target of an invocation.'
Inner Exception:
COMException: Class not registered (0x80040154 (REGDB_E_CLASSNOTREG))

@inforithmics Can you start that project on your machine?

pauldendulk:
oh, I can start the package. No idea how this should work. Why can we set the Mapsui.Samples.WinUI project as startup project while it can not run? What function does it have?

inforithmics:
Certain dependencies need to be installed so that it works. 

See here for instructions.
https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/set-up-your-development-environment?tabs=stable

Unfortunatly there is not yet a single WinUI Development checkbox in the Visual Studio Installer.

pauldendulk:
Perhaps there should be a readme.md in the project to refer this page.

Issue #1190, opened 2021-09-04
Proposal: Add an additional layout to Mapsui.UI.Forms


charlenni:
MapView is a convenient wrapper around Mapsui.UI.MapControl. All important properties of MapControl are wrapped by MapView. The reason for this is, that MapView add buttons to to the view. This buttons are placed in a RelativeLayout with the MapControl.

If MapControl would have this RekativeLayout by its own, then it would be possible to derive MapView from MapControl and all this wrappers could be dropped. With this, the Content of MapControl would be a RelativeLayout, that contains a SK...View.

Any comments on this? Other suggestions?

janusw:
I definitely like the idea of deriving MapView from MapControl! This would avoid a lot of unnecessary wrapping code.

On minor comment: Looking at the MapView code, it seems like it actually uses an AbsoluteLayout, not a RelativeLayout. But that does not really affect this discussion, of course.

janusw:
And yeah, if it is necessary for making MapView inherit from MapControl, I don't see a problem with moving the layout into the MapControl.

Of course it cannot go into the generic part of MapControl, but only into the Mapsui.Forms implementation, right?

charlenni:
You are right.

Another possibility would be to create the buttons as widgets. Then there is no need for any layout.

charlenni:
That will be the plan. MapView then will inherit all from MapControl and add some convinence functions (pins and so on).

PR #1247, opened 2021-09-26
Maui

inforithmics:


pauldendulk:
@inforithmics If we start on MAUI I would like to make a fresh start and not port from the existing MapView.

inforithmics:
No Problem, I only do this to see how far i can get and if the Build server will build it. So more an explorative pull request. Which I don't know if it will work.

charlenni:
@inforithmics I think, that using `ToNative()` all over and only add one function `ToForms()` and `ToMaui()` which call this `ToNative()` would be easier. So many of the `#if`/`#else` would be dropped.

Mapsui.UI.Forms/Objects/Callout.cs
```diff
@@ -58,7 +58,11 @@ public class Callout : BindableObject, IFeatureProvider
         /// <summary>
         /// Bindable property for the <see cref="Anchor"/> property
         /// </summary>
-        public static readonly BindableProperty AnchorProperty = BindableProperty.Create(nameof(Anchor), typeof(Point), typeof(MapView), default(Point));
+#if __MAUI__
+        public static readonly BindableProperty AnchorProperty = BindableProperty.Create(nameof(Anchor), typeof(Microsoft.Maui.Graphics.Point), typeof(MapView), default(Microsoft.Maui.Graphics.Point));
+#else
+        public static readonly BindableProperty AnchorProperty = BindableProperty.Create(nameof(Anchor), typeof(Xamarin.Forms.Point), typeof(MapView), default(Xamarin.Forms.Point));
+#endif
```
charlenni, line 65: Wouldn't it better to make a 

    #if __MAUI__
        using Point = Microsoft.Maui.Graphics.Point
    #else
        using Point = Xamarin.Forms.Point
    #endif

instead of changing this on each occurrence?

inforithmics: Good Point

inforithmics:
Currently the Build won't pass because .NET 6 rc1 is not installed on the Build Server images. (So maybe the images will be upgraded next month with .NET 6 rc1, because it has a go live license).

Mapsui.UI.Forms/Extensions/ColorExtensions.cs
```diff
@@ -7,19 +11,33 @@ public static class ColorExtensions
         /// </summary>
         /// <param name="color">Color in Mapsui format</param>
         /// <returns>Color in Xamarin.Forms.Maps format</returns>
+#if __MAUI__
```
charlenni, line 14: Add a native function to do this and use this than all over in the source code.

    #if __MAUL__
        public static Microsoft.Maui.Graphics.Color ToMaui(this Styles.Color color)
        {
            return new Microsoft.Maui.Graphics.Color(color.R / 255.0f, color.G / 255.0f, color.B / 255.0f, color.A / 255.0f);
        }

        public static Microsoft.Maui.Graphics.Color ToNative(this Styles.Color color)
        {
            return new Microsoft.Maui.Graphics.Color(color.R / 255.0f, color.G / 255.0f, color.B / 255.0f, color.A / 255.0f);
        }
    #else
        public static Xamarin.Forms.Color ToForms(this Styles.Color color)
        {
            return new Xamarin.Forms.Color(color.R / 255.0, color.G / 255.0, color.B / 255.0, color.A / 255.0);
        }

        public static Xamarin.Forms.Color ToNative(this Styles.Color color)
        {
            return new Xamarin.Forms.Color(color.R / 255.0, color.G / 255.0, color.B / 255.0, color.A / 255.0);
        }
    #endif

With this, you don't need a `#if`/`#endif` in the code. You could always use `ToNative()`.

Mapsui.UI.Forms/MapView.cs
```diff
@@ -856,8 +873,11 @@ private void HandlerInfo(object sender, MapInfoEventArgs e)
                     SelectedPin = clickedPin;
 
                     SelectedPinChanged?.Invoke(this, new SelectedPinChangedEventArgs(SelectedPin));
-
+#if __MAUI__
```
charlenni, line 876: Here you then need only a 

    var pinArgs = new PinClickedEventArgs(clickedPin, _mapControl.Viewport.ScreenToWorld(e.MapInfo.ScreenPosition).ToNative(), e.NumTaps);

without any `#if`/`#else`/`#endif`.

inforithmics: Should be done now

Mapsui.UI.Forms/Extensions/ColorExtensions.cs
inforithmics: Although done

inforithmics:
Starting the Sample I get an exception of a missing Microsoft.UI.Xaml.Controls.ElementAnimator on Windows have to investigate that.

inforithmics:
Maui support is only availabe in the next SkiaSharp version 2.88.0.
Maui Windows Support is dependent on WinAppSdk which is currently in preview1.
And it compiles currently only in Visual Studio 2022
It compiles in Visual Studio 2019 with "Use Preview of the .NET Sdks"
So I think I try it out when the next round of updates is released.

inforithmics:
It generates nuget packages and the build succeed, but somehow mapsui isn't displayed in the Sample project, I have to investigate that.

pauldendulk:
What do you mean with this?:
> mapsui isn't displayed in the Sample project

pauldendulk:
I mentioned above that I would like to make a fresh start with MAUI. The MapView has a lot of code and a lot of that code is platform specific. This has all been developed by charlenni, initially outside of the Mapsui repository. When it was integrated into the Mapsui repository the plan was to move functionality gradually into the MapControl. This never happened.

The MapControl design (with the shared MapControl project) is intended to create a component that has as much platform independent code as possible. The MapControl should also allow extensions of platform specific functionality (like platform specific popups) but the idea was to allow this without integrating too much platform specific code into the MapControl. 

For MAUI my intention is to start with the MapControl and then add the current MapView functionality step by step. Make it easy to add pins by adding the needed layers and adding the method to add them, but those would be platform independent. 

I can see how it is easy to move forward from the current MapView codebase and I assume @charlenni would prefer to continue from that code base because he is very familiar with it. But for me it is important to make a fresh start and move forward in smaller steps.

inforithmics:
Yes I agree with you, but for this there ist already a proposal to improve the Situation. https://github.com/Mapsui/Mapsui/issues/1190
With this pull request I wanted to Find out where the Problems are and possible Solutions. What I'm surprised the most that the Build Server can Build it. But unfortunatly Maui ist still a work in Progress, Ranking the wrong dependencies it missing some Hauses it to not Funktion. And things that workend in Xamarin Forms need workarounds

inforithmics:
The Samples "Work" now: Following Maui/Skia Problems. 
1. Skia Gpu is currently not working. 
    Workaround: MapControl.UseGPU = false; 
2. OnSizeChanged Event is not Firing:
    Workaround: private void SizeChangedWorkaround()

I guess these things will be fixed until Release of Maui. I will try with next Skia / Maui Preview

inforithmics:
No Problem I'll fix the conflicting files. Because Maui is based on Forms with includes and conditional compilation. It is not so much work to keep the Pull request up to date. I think the earlist this can be merged is when the new Skia Sharp Version is released and when Visual Studio 2022 is official released. Which is at the earliest in November.

appveyor.yml
```diff
@@ -19,7 +19,7 @@ nuget:
   disable_publish_on_pr: true
 install:
 - ps: Invoke-WebRequest 'https://dot.net/v1/dotnet-install.ps1' -OutFile 'dotnet-install.ps1';
-- ps: ./dotnet-install.ps1 -InstallDir '~/.dotnet' -Version '6.0.100-rc.2.21505.57';
+- ps: ./dotnet-install.ps1 -InstallDir '~/.dotnet' -Version '6.0.100';
```
pauldendulk: 🎉

pauldendulk:
The migrations might take a while. All that time you need to merge master back into your branch. If there is a way to merge your branch into master without breaking the build we should go for it. Perhaps the project should explicitly not be built somehow. A separate solution with MAUI might also be an option, although my renames and such would not reach the MAUI project either. Not sure what a good option is.

inforithmics:
@pauldendulk I thought about it, and have following idea. The problem with merging comes mainly from Mapsui.Forms changes. because I use conditional compilation to compile the same code for forms and Maui. Now my proposal is to make a second pull request where only this changes with the conditional compilation are included. So merging is then almost painless.

inforithmics:
I think I'm going with idea 2. I'll introduce a Mapsui.Maui.sln which works with Visual Studio 2022 und the Mapsui.sln will still work with Visual Studio 2019.

inforithmics:
When this builds this can be merged, I made Mapsui.Maui.sln project and a mauibuild: Github task.

pauldendulk:
> When this builds this can be merged, I made Mapsui.Maui.sln project and a mauibuild: Github task.

What should I install to build this?

inforithmics:
For Mapsui.Maui.sln You need to install Visual Studio 2022 Preview.
     https://visualstudio.microsoft.com/de/vs/preview/ With following settings

image

Then you need

image

Sometimes Mapsui.Samples.Maui doesn't build reliable so I simply disable the other targets and let it run on Windows.

Commenting out this in the Project File. (Needs Visual Studio Restort to work).
<TargetFrameworks>net6.0-android;net6.0-ios;net6.0-maccatalyst</TargetFrameworks>

pauldendulk (approved):
This will probably be our most important target in a year.

Issue #1248, opened 2021-09-27
[XF.Forms] Null reference exception when I display the map with 3.0.0-rc.4


lanarchyste:
When I display the map on Android for a Xamarin.Forms project with the following code:
```
<mapsui:MapView
            x:Name="CustomerMap"
            Grid.Row="0"
            IsMyLocationButtonVisible="False"
            IsNorthingButtonVisible="True"
            IsZoomButtonVisible="True"
            MyLocationEnabled="False"
            MyLocationFollow="False"
            UniqueCallout="True"
            UseDoubleTap="True"
            VerticalOptions="FillAndExpand" />
```

I get the following error:
```
System.NullReferenceException: Object reference not set to an instance of an object
   at Xamarin.Forms.Xaml.ValueNode.Accept (Xamarin.Forms.Xaml.IXamlNodeVisitor visitor, Xamarin.Forms.Xaml.INode parentNode) [0x00000] in <1009-27 15:26:08.231 E/AppCenterCrashes( 1535):   at Xamarin.Forms.Xaml.ApplyPropertiesVisitor.SetPropertyValue (System.Object xamlelement, Xamarin.Forms.Xaml.XmlName propertyName, System.Object value, System.Object rootElement, Xamarin.Forms.Xaml.INode node, Xamarin.Forms.Xaml.HydrationContext context, System.Xml.IXmlLineInfo lineInfo) [0x0007d] in <10f24b47010e4ddbac77d68824d70ca8>:0 
   at Xamarin.Forms.Xaml.ApplyPropertiesVisitor.Visit (Xamarin.Forms.Xaml.ValueNode node, Xamarin.Forms.Xaml.INode parentNode) [0x00090] in <10f24b47010e4ddbac77d68824d70ca8>:0 
   at Xamarin.Forms.Xaml.ElementNode.Accept (Xamarin.Forms.Xaml.IXamlNodeVisitor visitor, Xamarin.Forms.Xaml.INode parentNode) [0x00044] in <10f24b47010e4ddbac77d68824d70ca8>:0 
[...]

PR #1249, opened 2021-09-27
Removed a bug crashing MapViewControl

charlenni:
Removed a bug where Xaml code for visibility of buttons could crash the Forms MapView control.

See #1248.


pauldendulk (approved):

pauldendulk:
The appveyor build fails but I am going to ignore this for now. It looks like a problem with the net48 unit tests. They should be converted to net core.

PR #1250, opened 2021-09-27
[WiP]  Widget for drawing time

charlenni:
Add a widget, which draws a mean value for drawing time on the map


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -142,6 +143,21 @@ public int UpdateInterval
                 }
             }
         }
+
+        private double _drawingTime;
+
+        public double DrawingTime
+        {
+            get { return _drawingTime; }
```
pauldendulk, line 151: What about this warning? Apparently there is a View.DrawingTime already. Perhaps use a different name.
https://docs.microsoft.com/en-us/dotnet/api/android.views.view.drawingtime?view=xamarin-android-sdk-9

charlenni:
@pauldendulk I now refactored all. I add a `Performance` object, which holds the relevant data. If this isn't set in MapControl, no data is saved. This `Performance` object is then added to the `PerformanceWidget` (renamed it), which use the data to render it on the screen. Disadvantage is, that the second last drawing time is displayed.

I added also a `Count` property, so you could see, how often the screen was redrawn.

pauldendulk (approved):
Great feature, thanks!

charlenni:
@pauldendulk Oh, it was WiP. It wasn't ready. I made some more improvements. So how could I make a PR with them? I will try it.

pauldendulk:
> @pauldendulk Oh, it was WiP. It wasn't ready. I made some more improvements. So how could I make a PR with them? I will try it.

I installed the github WIP app https://github.com/apps/wip. Now the checks of the PR show it is a WIP if 'WIP' is used as the prefix.

PR #1252, opened 2021-09-28
iOS System Memory Workaround

inforithmics:
This should fix https://github.com/Mapsui/Mapsui/issues/1216

by including a target in Xamarin.Ios that creates this PackageReference

<PackageReference Include="System.Memory" Version="4.5.4" IncludeAssets="None" />

Avoiding that the Developer using Mapsui has to do it himself

### DIFF
--- a/NuSpec/Mapsui.Forms.nuspec
+++ b/NuSpec/Mapsui.Forms.nuspec
@@ -22,12 +22,19 @@
       <group targetFramework="netstandard2.0">
         <dependency id="Mapsui" version="$version$"/>
         <dependency id="Xamarin.Forms" version="[5.0.0.2083,6.0)" />
-        <dependency id="SkiaSharp.Views.Forms" version="[2.80.2,3.0.0)" />
+        <dependency id="SkiaSharp.Views.Forms" version="[2.80.2,3.0.0)" />		
+      </group>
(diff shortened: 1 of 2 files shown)


pauldendulk (approved):
Okay. A xamarinios1.0 group is added to we can treat it differently from the rest. A targets file was added. And this file is added to the lib/xamarinios1.0/ folder in the nuget. I never have done something like this before. Let's merge it. We could test with the nuget created by the build (after I fixed the appveyor build). Would be awesome if this fixed this nagging issue.

pauldendulk:
@inforithmics I am not setup for ios so can not properly test this. 

1) I did some research though. First build the MVR sample of dino. 
Result: The bin folder included the System.Memory nuget.

2) Then I added the  IncludeAssets="None" solution:
```xml
<PackageReference Include="System.Memory" IncludeAssets="None">
  <Version>4.5.4</Version>
 </PackageReference>
```
Result: The bin folder did not include the System.Memory nuget.

3) Then I removed the System.Memory nuget altogether. And added the nuget of latest build:
```console
PS> install-package mapsui -source https://ci.appveyor.com/nuget/mapsui -pre
```
Result: The bin folder did not include the System.Memory nuget.

So the new package has the same result as the IncludeAssets="None".

I assume this works and won't create another RC specifically for this.

PR #1253, opened 2021-09-28
Stopwatch widget

charlenni:
Added some more values to the performance widget


charlenni:
Ok, now you could follow last drawing time, mean from last x drawing times (in sample is set to 10, so the value below 10 draws isn't correct), min, max and count. If you click on the widget, the values are cleared.

Anything else we need?

charlenni:
@pauldendulk You added the first part. Would it be possible to look in the second part too? Thank you for your support.

pauldendulk (approved):
I'll merge it. Will see how it looks in the sample tomorrow.

charlenni:
> I'll merge it. 

Thank you.

> Will see how it looks in the sample tomorrow.

Should work in all samples, because it isn't a Forms version, but a native one. Perhaps we have to work a bit in the look or the data, which is displayed.

pauldendulk:
Had a quick look. Good idea to add max and min. I though that there should perhaps be way to reset the stats. Perhaps with a click on the dialog. Then I clicked on it. And the stats we reset! 😍

The stat that means the most to me is FPS (frames per second). The max possible FPS is directly derived from average, but we do some waiting. The FPS with the current loop is mostly fixed by the timer, but still I would like to see it since there can also be skips of frames, and those skips is where it gets interesting, that is where te problems are.

charlenni:
@pauldendulk But how to calculate the FPS? It couldn't go better than the timer values. That's obvious. But in the other direction, when say the drawing time is greater than 16 ms, the FPS goes down. So I assume

    FPS = Performance.Mean > UpdateInterval ? 1000 / UpdateInterval : 1000 / Performance.Mean

I would say, a counter for dropped frames would be better. That should be possible in the `CommonDrawControl`, if `_drawing` is true. When this is the case, we drop a frame, which means, that the FPS goes down.

pauldendulk:
@charlenni I did not see the problem when I wrote that comment. I saw you now added 'virtual' FPS, which is an okay workaround.

I think a better solution would need to calculate the FPS at the level of the render loop, like we do for the Mean.

PR #1255, opened 2021-09-30
Performance widget improvements

charlenni:
Adding frames per second to Performance.
Refactoring of PerformanceWidgetRenderer, so that the code is easier to read and the result on screen is better.

### DIFF
--- a/Mapsui/Utilities/Performance.cs
+++ b/Mapsui/Utilities/Performance.cs
@@ -5,12 +5,14 @@ namespace Mapsui.Utilities
 {
     public class Performance
     {
+        readonly int _maxValues;
+        readonly double[] _drawingTimes;
+
+        bool _turnaround;
         int _pos;
-        int _maxValues;
         int _count;
         double _min, _max;
         double _sum = 0;
-        public double[] _drawingTimes;
 
         public Performance(int maxValues = 20)
         {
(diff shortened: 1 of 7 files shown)


Mapsui/Utilities/Performance.cs
```diff
@@ -47,7 +48,17 @@ public Performance(int maxValues = 20)
         /// <summary>
         /// Mean value of all MaxValues drawing times
         /// </summary>
-        public double Mean => _sum / _maxValues;
+        public double Mean => _turnaround ? _sum / _maxValues : (_pos > 0 ? _sum / _pos : 0);
+
+        /// <summary>
+        /// Possible frames per second calculated from Mean
+        /// </summary>
+        public int FPS => Mean == 0 ? 0 : (int)(1000.0 / Mean);
```
pauldendulk, line 56: Just for my understanding. The FPS is value is now 'virtual' in the sense that we could draw with this FPS but we don't because of the render loop buffering.

pauldendulk (approved):
Thanks a lot. Refactoring reduces the code a lot. And great you added documentation.

Mapsui/Utilities/Performance.cs
charlenni: Yes, you are right. It shows the maximal possible FPS. With this, it is possible to adjust the timer value if you want, so that more than 60 FPS are possible.

PR #1256, opened 2021-10-01
Fix a problem with fetcher threads

pauldendulk:
All threads created would end in a wait state and would never be cleaned up.

I introduced this problem yesterday when working on the fetcher unit tests.


pauldendulk:
Confirmed this is fixed in the nuget-of-latest-build used in dino's mapsuidebug sample where I first noticed the problem.

PR #1257, opened 2021-10-01
Get callout to work in WPF

pauldendulk:
@charlenni  The callout functionality non-forms sample called 'Callout' did not function. It was like this for at least many months. Today I looked at it. I got it to work by looking at the way a callout is generated in Forms, but there is still a lot I don't understand. 

The sample as it was used a bitmap. A png was created. That png was generated correctly. It was added to the BitmapRegistry. In a later step this png was (or should?) be transformed to a SKImage. This did not work as it should but I am not sure how it was supposed to work. Why did we use a png here? Performance? Is a png for callout used in other samples?

I altered this sample to get it to work but perhaps it is better to repair the original sample.


pauldendulk:
Replaced by #1261

Issue #1109, opened 2021-04-05
Create Mapsui.Rendering.Uno to support Uno platform.


vkotov:
Is it possible to adapt Mapsui rendering subsystem to run on [Uno platform](http://platform.uno)?
On iOS, Android and macOS, the Uno Platform relies on the current Xamarin Native stack (not Xamarin.Forms). 
On WebAssembly, the Uno Platform relies directly on the Mono-WASM runtime, a part of.NET 5.
On Linux, Uno Platform relies on Skia to draw the UI on canvas.
Uno Platform acts a bridge for WinUI and UWP apps to run natively on iOS, macOS, Android, Linux and WebAssembly.
You can run your C# and XAML source code unmodified and compile it on a different platform yet have it behave the same at runtime.

tmijieux:
I think it is possible
I dont think it is required to create a "Mapsui.**Rendering**.Uno , *rendering* is the is the part of mapsui library that does the rendering but it is mostly not exposed in the final usable controls used in the different frameworks, basically anywhere either skiasharp or "xaml renderer" can run, we do not need to create a new "rendering" namespace rather a new Mapsui.UI.Uno namespace.What need to be done, is implementing a partial class MapControl that inherits controls for the target framework .
Skiasharp already has views for uno [here](https://github.com/mono/SkiaSharp/tree/main/source/SkiaSharp.Views.Uno), so you need to take example on any of the Mapsui.UI.XXXX.MapControl class and subclass these views and that'll probably works...

pauldendulk:
> I think it is possible
> I dont think it is required to create a "Mapsui.**Rendering**.Uno" , _rendering_ is the part of mapsui library that does the rendering but it is mostly not exposed in the final usable controls used in the different frameworks, basically anywhere either skiasharp or "xaml renderer" can run, we do not need to create a new "rendering" namespace rather a new Mapsui.UI.Uno namespace.
> 
> What need to be done, is implementing a partial class MapControl that inherits Views for the target framework .
> Skiasharp already has views for uno [here](https://github.com/mono/SkiaSharp/tree/main/source/SkiaSharp.Views.Uno), so you need to take example on any of the Mapsui.UI.XXXX.MapControl class and subclass these views and that'll probably works...

Yes, I assume that is correct. The Mapsui.Rendering.Skia is a .netstandard2.0 that I think can just be used as such in Mono WASM.

FrozDark:
[...]

PR #1259, opened 2021-10-03
[WIP] Uno

inforithmics:
Implementation of this
https://github.com/Mapsui/Mapsui/issues/1109

And parts of this.
https://github.com/unoplatform/uno/issues/4891


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -12,12 +12,12 @@
 using System.Runtime.CompilerServices;
 using Mapsui.Utilities;
 
-#if __ANDROID__
+#if __UWP__
```
pauldendulk, line 15: Do you not need a Mapsui.UI.Uno namespace? and __UNO__ define?

Mapsui.sln
```diff
@@ -1612,6 +1888,574 @@ Global
 		{C008562A-ED14-4EF0-BAD6-B7967EF9A9F3}.Release|x64.Build.0 = Release|Any CPU
 		{C008562A-ED14-4EF0-BAD6-B7967EF9A9F3}.Release|x86.ActiveCfg = Release|Any CPU
 		{C008562A-ED14-4EF0-BAD6-B7967EF9A9F3}.Release|x86.Build.0 = Release|Any CPU
+		{B78CDE0F-568F-4064-8B74-FCD920E22351}.Ad-Hoc|AnyCPU.ActiveCfg = Debug|Any CPU
```
pauldendulk, line 1891: This is probably generated by Visual Studio but still we need to get a hold on it somehow. This is the guid of UNOB78CDE0F-568F-4064-8B74-FCD920E22351, but it adds all configurations. The most of them we use only in one platform. I don't know exactly what would be the minimal configurations, perhaps only Debug|AnyCPU an Release|AnyCpu. Since we are starting to work on v4 this is good moment for a cleanup. Also for all the older platforms.

NuSpec/Mapsui.Uno.nuspec
```diff
@@ -0,0 +1,64 @@
...
+    <owners>Paul den Dulk</owners>
+    <projectUrl>https://github.com/Mapsui/Mapsui</projectUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>
+      Uno map components based on the Mapsui library
+    </description>
+    <tags>map maps mapping geo gis osm Uno</tags>
+    <language>en-US</language>
+    <dependencies>
+      <!-- [ or ] is including, ( or ) is excluding -->
```
pauldendulk, line 17: oh, now I realize that UNO supports all platforms like Xamarin.Forms. A heavy burden.

Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Droid/Resources/values/Styles.xml
```diff
@@ -0,0 +1,12 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
```
pauldendulk, line 1: And then there are sample for each separate platform as well. phewww

Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Shared/FeatureExtensions.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Mapsui.Providers;
+
+namespace Mapsui.Samples.Uno
+{
+    static class FeatureExtensions
+    {
+        public static string ToDisplayText(this IEnumerable<KeyValuePair<string, IEnumerable<IFeature>>> featureInfos)
```
pauldendulk, line 10: There is nothing platform specific for the UNO platform in here is there? I was thinking this could be a method used over all our samples. It is not a very nice solution though (I wrote it) but still if we want to improve it butter to do it in one place. 

Also think it would be worthwhile to invest more in skia rendered solutions. We could use the callouts instead. Or we could work on a skia info box. Similar to the callout but not a popup on  item, but a widget on a fix spot which shows info of the selected feature.

Mapsui.UI.Shared/MapControl.cs
inforithmics: Actually the define is UWP and the namespace is uwp too, It is a convetion in the Uno platform because it has the same names as UWP. I had do put the UWP define at the first place because else in the android or Macios Targets the uno namespace would have been Android or iOS

Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Droid/Resources/values/Styles.xml
inforithmics: In Maui they managed to put all platforms into the same project (Multitargeting). Maybe in the future this could although be possible for Uno.

Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Shared/FeatureExtensions.cs
inforithmics: I removed the duplicate FeatureExtensions and used the one in Samples.Common

Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Droid/Resources/values/Styles.xml
inforithmics: The Sample Projects are only place holders. I didn't change anything in them exception adding references.
the actual sample code lives in Mapsui.Samploes.Uno.Shared

Mapsui.UI.Shared/MapControl.cs
pauldendulk: Ah, I knew uno was based on uwp but I now understand they use the uwp define. And on ios and android both the uwp and ios/android defines are set. Learned something, thank you.

Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Droid/Resources/values/Styles.xml
pauldendulk: Nevertheless I think we should put a lot of effort in minimizing the amount of code we need. Do we need Uno specific sample as all (as in Mapsui.Samploes.Uno.Shared)?. Could we upgrade the mapcontrol and widgets so that we can handle everything with cross platform code. This might be some effort but I think we must invest in it at this point because we will have so many different versions of Mapsui soon. Perhaps even Blazor at some point.

inforithmics:
I currently tried only UWP Uno Implementation. The others I have to try if they work. So I don't expect to change anything in mapsui.uno itself but maybe in the samples head projects. But I can remove the WIP.

inforithmics:
I although hoped to have one sample Project for all platforms but there is nothing yet. Here  MAUI seems note advanced.

pauldendulk:
Going to fix the conflict after merge.

inforithmics:
Certain dependencies need to be installed so that it works. 

See here for instructions.
https://docs.microsoft.com/en-us/windows/apps/windows-app-sdk/set-up-your-development-environment?tabs=stable

Unfortunatly there is not yet a single WinUI Development checkbox in the Visual Studio Installer.

PR #1260, opened 2021-10-04
Map view update

charlenni:
Thank you, @pauldendulk and all the other contributors for version 3.0.0.

But development shouldn't stop. Here is the version of MapView that inherits directly from MapControl. It doesn't need any wrappers around properties, methods or events.

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -50,7 +50,7 @@ namespace Mapsui.UI.Forms
         /// </summary>
         private const int touchSlop = 8;
 
-        private bool _initialized = false;
+        protected bool _initialized = false;
         private double _innerRotation;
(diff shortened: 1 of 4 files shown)


Mapsui.UI.Forms/MapControl.cs
```diff
@@ -50,7 +50,7 @@ public TouchEvent(long id, Geometries.Point screenPosition, long tick)
         /// </summary>
         private const int touchSlop = 8;
 
-        private bool _initialized = false;
+        protected bool _initialized = false;
```
inforithmics, line 53: Maybe only a Readonly Protected Property could expose this

inforithmics (approved):

charlenni:
@inforithmics Thank you for looking into this.

Do you think, that we need something else for this PR?

inforithmics:
It is fine from my side, it's mostly removal of rendundant code which is good. So this should be merged.

PR #1261, opened 2021-10-08
Removed a bug where Callout wasn't visible in WPF sample anymore

charlenni:
This is instead of #1257

### DIFF
--- a/Mapsui/Styles/CalloutStyle.cs
+++ b/Mapsui/Styles/CalloutStyle.cs
@@ -46,6 +46,18 @@ namespace Mapsui.Rendering.Skia
         Right,
     }
 
+    /// <summary>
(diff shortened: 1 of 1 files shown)


pauldendulk:
ah, that is a simple fix. 

Still, the sample involves a lot of code. This is discouraging for users. I was looking into this because I would like to provide a very simple way for users to add feature info. Right now every platform needs to add some control to show the info. There should be some simple way to turn on feature info on touch/click.

One thing I want to do right now is add a new sample that does not create the bitmaps in the sample code. Why was this done in the first place? Performance I guess, but did we notice it was slow? If a single callout is shown at a time performance is not such a problem.

I am also thinking of a new style for showing feature info.

PR #1262, opened 2021-10-08
Feature/fixes for avalonia

pauldendulk:
- Zoom on mouse pointer (the bug was in the navigator, this also fixes it for uwp).
- Get slider to work (with a workaround).
- Move info display to the left (not pretty but the map does not move up, I think it is better).
- Add tests to samples (they are added and interestingly a transparent background is really transparent in Avalonia, I see visual studio as background).
- Remove UWP from namespace (a suggested change).
- Fix 'collection was modified' error which occurred during animations. This was a bug that would have affect in other places as well.
- Some typos in comments
- Resharper warnings and suggestions I ran into


pauldendulk:
So far I have been working with resharper at work and for private projects but I think editorconfig is the future so lets go for that.

PR #1263, opened 2021-10-09
Feature/add simple calllout sample

pauldendulk:
@charlenni 
- Added a 'CustomCalloutSample' sample under a new 'Info' category where all callout are off by default and toggled on/off by clicking them.
- Renamed the existing callout sample to CustomCalloutSample and moved it also under 'Info'.

@inforithmics In Avalonia a click caused a double event (turning it on and immediately off again). This was because both OnSingleTapped and MapControl_PointerPressed triggered the event. I removed the OnInfo from MapControl_PointerPressed but perhaps we should look into what the further purpose is of that method. In WPF is was important for dragging a rect to zoom into, but this is not implemented in Avalonia.

@charlenni and @inforithmics To make it so that this info sample works on all platforms I added an Info event to the Map. There is already a Info event on the MapControl. In the past there was only an Info event on the Map. This was moved because users that worked primarily with the MapControl found it confusing. For crossplatform code (and samples) it is better to have it on the Map. Since I think the cross platform goal is important I now made this change.

Image showing two point that were clicked in Avalonia:
image

### DIFF
--- a/Samples/Mapsui.Samples.Common/Maps/Callouts/SingleCalloutSample.cs
+++ b/Samples/Mapsui.Samples.Common/Maps/Callouts/SingleCalloutSample.cs
@@ -18,7 +18,7 @@ namespace Mapsui.Samples.Common.Maps.Callouts
 {
     public class SingleCalloutSample : ISample
     {
-        public string Name => "1 Simple Callout";
+        public string Name => "1 Single Callout";
         public string Category => "Info";
 
         public void Setup(IMapControl mapControl)


inforithmics:
The Info Event ist in my opinion a good idea. In Avalonia I think there are Partys implemented for an Info Text Sample. But I have to investigate this later. The removal og the double Event ist a good idea. I somehow duplicated the Event Handling.

charlenni:
I'm not very happy with moving an event to Map. For me the Map holds only information about the map data (like layers, projections and so on), but has nothing to do with the handling of events or displaying of the data. It should even possible to use a Map from different MapControls.

So why has Info to move to Map? We have a MapControl event Info in the shared part of MapControl, all platforms get this. Perhaps I didn't understand the idea behind this.

The new sample is good. It is easier to understand as the old one.

inforithmics:
I don't have a strong opinion in this maybe it would bei good to move mapview to shared and possible have platforms specific Implementation (partial class like Mapcontrol)

pauldendulk:
> I'm not very happy with moving an event to Map. For me the Map holds only information about the map data (like layers, projections and so on), but has nothing to do with the handling of events or displaying of the data. It should even possible to use a Map from different MapControls.
> 
> So why has Info to move to Map? We have a MapControl event Info in the shared part of MapControl, all platforms get this. Perhaps I didn't understand the idea behind this.

It is good to have a discussion about the Map.Info event. One reason not to like it is that it is now both on the MapControl and Map. From the perspective of a user who sees the MapControl as just another componenten like so many other UI components it is more logical to have the event on the MapControl, that is where you start looking first. Also with the MapControls shared code we don't have duplication for each platform. But the problem when writing the sample is that you do not have access to the MapControl. You could say this is just the problem of the samples but for many users it will make sense to define the Map in the platform independent shared part of their code. That may still be possible if users inject the IMapControl somehow in their shared code, but that is another hurdle to take. 

The way the sample could be set up now (with Map.Info) is nice. You can easily add the handling when defining the Map. It allows a lot of flexibiity. I think it will be useful for other scenarios as well. Below is the code. What is an alternative way to write such a solution? 

```csharp

        public static Map CreateMap()
        {
            var map = new Map();

            map.Layers.Add(CreatePointLayer());
            map.Home = n => n.NavigateTo(map.Layers[1].Envelope.Centroid, map.Resolutions[5]);
            map.Info += MapOnInfo;

            return map;
        }

        private static void MapOnInfo(object sender, MapInfoEventArgs e)
        {
            var calloutStyle = e.MapInfo.Feature?.Styles.Where(s => s is CalloutStyle).Cast<CalloutStyle>().FirstOrDefault();
            if (calloutStyle != null)
            {
                calloutStyle.Enabled = !calloutStyle.Enabled;
                e.MapInfo.Layer.DataHasChanged(); // To trigger a refresh of graphics.
            }
        }

```

pauldendulk:
@charlenni So how about this one? 

The summary: There is already a MapControl.Info event. This adds a Map.Info event. The question is why do we need two? I was not sure about this when I wrote this code. It just seemed handy at the local level of my code. After discussing this here and  thinking about my view is now that we do need two. 

The reason: They serve a different purpose. If you want to do something that is general for you application you can use the MapControl.Info event. If you want to do something that is specific to your Map (related to a specific layer in that Map) then use the Map.Info event. Of course if you never change the Map object it doesn't matter and you can choose any.

I had the impression from your response you assumed the implementation of the Info event callback was now part of Mapsui, this is not the case. Perhaps it is some other misconception. I have time today to have a video session to talk this through.

PR #1283, opened 2021-10-17
Mapsui.Mac.sln: remove Mapsui.Desktop.Tests

janusw:
* this project does not exist any more
* it was removed from Mapsui.sln already
* see commit bb28a9c7058ed36e4883916439636486660e3770


janusw:
Thanks for merging, @pauldendulk. Do you mind if I backport this to develop/3.0 as well? (This is also affected and I assume it will still receive bugfixes for a while.)

pauldendulk:
Yes, please.

btw, do you know an easy way to merge this change into develop/3.0? Either in gitlab or command line. It seems a very common thing to do, but am always improvising.

janusw:
> Yes, please.
> 
> btw, do you know an easy way to merge this change into develop/3.0? Either in gitlab or command line. It seems a very common thing to do, but am always improvising.

Well, yeah. I prefer to work with git on the command line. What I'd usually do in a case like this is:

```
git checkout develop/3.0
git cherry-pick -x a2a5437508cccdfaac999f5f6844daef9fde4a32
git push upstream develop/3.0
```

Just did that. (Note: My remotes are set up such that my fork is called 'origin' and the original Mapsui repo is called 'upstream'.)

PR #1304, opened 2021-10-29
Update render tests

pauldendulk:
Update the render tests.
- Add test for StackedLabelLayer.
- Move WriteableLayer sample, which was not tested, out of tests.
- Calculate Viewport based on Map definition and it's layer extend. This is simpler and looks better.
- In the Map.Home method navigate to Map extent.
- Add SmokeWhite background. This is useful to see how big an image is in some image viewers.


pauldendulk:
I also updated the documentation on the rendering tests:
https://mapsui.com/documentation/rendering-tests.html

PR #1305, opened 2021-10-29
Nullable reference fixes

inforithmics:
Some more Nullable Reference fixes

### DIFF
--- a/Tools/VersionUpdater/Program.cs
+++ b/Tools/VersionUpdater/Program.cs
@@ -11,7 +11,7 @@ namespace VersionUpdater
     public class Options
     {
         [Option('v', Required = true, HelpText = "Specifies the version to set in semver format")]
-        public string Version { get; set; }
+        public string Version { get; set; } = default!;
     }
 
     static class Program
--- a/Mapsui.Core/Layers/LayerCollection.cs
+++ b/Mapsui.Core/Layers/LayerCollection.cs
@@ -128,7 +128,7 @@ namespace Mapsui.Layers
             return success;
         }
 
-        public void Modify(IEnumerable<ILayer> layersToRemove, IEnumerable<ILayer> layersToAdd)
+        public void Modify(IEnumerable<ILayer>? layersToRemove, IEnumerable<ILayer>? layersToAdd)
         {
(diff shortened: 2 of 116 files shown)


Tools/VersionUpdater/Program.cs
```diff
@@ -11,7 +11,7 @@ namespace VersionUpdater
     public class Options
     {
         [Option('v', Required = true, HelpText = "Specifies the version to set in semver format")]
-        public string Version { get; set; }
+        public string Version { get; set; } = default!;
```
pauldendulk, line 14: What does default! mean?

pauldendulk (approved):
👍

pauldendulk:
Your strategy is now to just add ? where values are actually nullable? The real benefit will come when we can make more fields non-nullable, but perhaps this will take some more thought.

charlenni:
I thought about this too.

What can we do to improve this? Or is it ok, that we use nullable types? Should we use `is null` instead of `== null`?

Tools/VersionUpdater/Program.cs
inforithmics: = default means the field is assigned null (the default value of the field)
= default! means it is assigned null, but view the field as non null assigned.
Because the Version is required i guess always a value is assigned here from externally.

inforithmics:
is null and == null are sementically equal.

string? is the same as string except that the compiler adds an attribut to the Type that it could contain null. So using in a context where nullable is not enabled. The user sees only string.

So nullable Reference Types is basically only a Code Analysis Feature that tells the programmer to test for nullable when accessing potentially nullable variables.

so changing a type from 

`string test {get; set;}`

to

`string? test {get; set;}`

doesn't change the semantics of the programm it only influences the analyzing.

inforithmics:
@pauldendulk The long term goal could be to have not so many nullable fields to avoid the null checks by creating / setting Default values.

But I think this has to be decided on a case by case basis. But having the compiler now complaining about nullable fields allows to program in a null aware way.

pauldendulk:
> But having the compiler now complaining about nullable fields allows to program in a null aware way.

Definitely.

PR #1306, opened 2021-10-30
Format csprojs

pauldendulk:
Formatting csproj to be consistent. Now all have newlines between all elements that are directly in the root element.  I did this for all sdk format csprojs and ignored all old csprojs.

There are quite a few old csprojs. I think projects that hang on to the old format have no future and we should remove all those at some point.


pauldendulk:
In general I think we should work towards a situation where we understand everything in the csprojs. In the past the csprojs just grew more complex because we had to get things to compile and did not look at them because everything was autogenerated. With the new csproj we should regularly clean them up. If elements have no use anymore they should be removed, or perhaps simplified.

PR #1310, opened 2021-10-30
IProvider.GetExtent() to MRect

pauldendulk:


pauldendulk:
This also fixes a bug in forms where lat and lon order was wrong. This bug was introduced weeks ago.

PR #1312, opened 2021-10-30
Nullable fixes in skia rendering

inforithmics:
Made some more nullable fixes.

What I'm not sure is the new MRect type.

```
   public MRect(IEnumerable<MRect> rects)
        {
            Max = null;
            Min = null;

            foreach (var rect in rects)
            {
                Min ??= rect.Min.Clone();
                Max ??= rect.Max.Clone();

                Min.X = Math.Min(rect.Min.X, Min.X);
                Min.Y = Math.Min(rect.Min.Y, Min.Y);
                Max.X = Math.Max(rect.Max.X, Max.X);
                Max.Y = Math.Max(rect.Max.Y, Max.Y);
            }
        }
```

Max and Min are currently defined  not nullable.  changing that would require to change although

things like that

`public double MinX => Min.X;`

giviing a default when Min is not set.

`public double MinX => Min?.X ?? 0;`

### DIFF
--- a/Mapsui.Rendering.Skia/SkiaWidgets/ScaleBarWidgetRenderer.cs
+++ b/Mapsui.Rendering.Skia/SkiaWidgets/ScaleBarWidgetRenderer.cs
@@ -31,23 +31,23 @@ namespace Mapsui.Rendering.Skia.SkiaWidgets
             // Update paints with new values
             _paintScaleBar.Color = scaleBar.TextColor.ToSkia(layerOpacity);
             _paintScaleBar.StrokeWidth = scaleBar.StrokeWidth * scaleBar.Scale;
-            _paintScaleBarStroke.Color = scaleBar.Halo.ToSkia(layerOpacity);
+            _paintScaleBarStroke!.Color = scaleBar.Halo.ToSkia(layerOpacity);
             _paintScaleBarStroke.StrokeWidth = scaleBar.StrokeWidthHalo * scaleBar.Scale;
-            _paintScaleText.Color = scaleBar.TextColor.ToSkia(layerOpacity);
+            _paintScaleText!.Color = scaleBar.TextColor.ToSkia(layerOpacity);
             _paintScaleText.StrokeWidth = scaleBar.StrokeWidth * scaleBar.Scale;
             _paintScaleText.Typeface = SKTypeface.FromFamilyName(scaleBar.Font.FontFamily,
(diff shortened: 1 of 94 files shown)


pauldendulk:
We can design the MRect to our own liking.

Perhaps Min and Max should be non-nullable. I don't see a scenario right now why they should be nullable. If we make it non-nullable it should perhaps not have a default constructor. We could also initialize in the constructor at 0,0 0,0 but I can not think of situations where that would be useful A 0,0 0,0 rect is likely to cause exceptions. So I would prefer to remove it.

The default constructor is currently used in two placed. 
- One is a completely unused field. Callout.MRect. So that is just a good riddance. 
- The other is in the ScaleBarWidget. This is only relevant if there are no 'points' and I think an exception would be more appropriate in that case. The user will only be confused if the ScaleBarWidget has no size at all.

inforithmics:
I'm for Min and Max to be non nullable

When rects.Cout() == null probably an exception should be thrown.
` public MRect(IEnumerable<MRect> rects)`

pauldendulk:
> I'm for Min and Max to be non nullable

And also removing the default constructor?

> When rects.Count() == 0 probably an exception should be thrown. ` public MRect(IEnumerable<MRect> rects)`

Yes, that also makes sense.

inforithmics:
Yes I would remove the Default Constructor:

Mapsui.Rendering.Skia/SkiaWidgets/ScaleBarWidgetRenderer.cs
```diff
@@ -128,7 +128,7 @@ public class ScaleBarWidgetRenderer : ISkiaWidgetRenderer
                 canvas.DrawText(scaleBarText2, posX2, posY2 - textSize2.Top, _paintScaleTextStroke);
                 canvas.DrawText(scaleBarText2, posX2, posY2 - textSize2.Top, _paintScaleText);
 
-                envelop = envelop.Join(new MRect(posX2, posY2, posX2 + textSize2.Width, posY2 + textSize2.Height));
+                envelop = envelop?.Join(new MRect(posX2, posY2, posX2 + textSize2.Width, posY2 + textSize2.Height));
             }
 
             scaleBar.Envelope = envelop;
```
pauldendulk, line 134: scaleBar.Envelope could now be null if the points are null. Better to throw an exception when there are no points. A scaleBar needs a position. And perhaps the nullable code can prevent null points altogether.

pauldendulk (approved):

PR #1324, opened 2021-11-01
Remove xcopy extensionmethods becasuse it did not exist

pauldendulk:
This xcopy broke the build, so I removed it. 

In general I think we should try to avoid solutions like this because they are hard to maintain. If more folders are added they should all be xcopied? Could this not have been solved with a shared project?

Also I think it is time to deprecate UWP. It can be replaced with WinUI. How is that in Uno? It has not fully moved to WinUI? Maybe my question does not make sense be I don't follow the Uno news at all, so I really know nothing about it. Could we wait for a WinUI Uno version before we release it?

Also also, I want to get rid of the old csproj format.

### DIFF
--- a/Samples/Mapsui.Samples.Uwp/Mapsui.Samples.Uwp.csproj
+++ b/Samples/Mapsui.Samples.Uwp/Mapsui.Samples.Uwp.csproj
@@ -194,8 +194,7 @@
   </PropertyGroup>
   <Import Project="$(MSBuildExtensionsPath)\Microsoft\WindowsXaml\v$(VisualStudioVersion)\Microsoft.Windows.UI.Xaml.CSharp.targets" />
   <PropertyGroup>
-    <PreBuildEvent>xcopy ..\..\..\MainPage.* ..\..\..\..\Mapsui.Samples.Uno\Mapsui.Samples.Uno.Shared /Y
-xcopy ..\..\..\ExtensionMethods\*.* ..\..\..\..\Mapsui.Samples.Uno\Mapsui.Samples.Uno.Shared /Y</PreBuildEvent>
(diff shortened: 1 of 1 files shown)


inforithmics:
I will make a pull request with a shared project.

Regarding Uno and UWP There are two Uno Version Uno.UWP and Uno.WinUI. But unfortunatly WinUI is still in Preview.

pauldendulk:
> Regarding Uno and UWP There are two Uno Version Uno.UWP and Uno.WinUI. But unfortunatly WinUI is still in Preview.

@inforithmics I have all kinds of vague build errors with Uno. I now work with the Uno projects unloaded. With WinUI about to replace Uwp I think we should not invest in UWP anymore and the v4 release should use a WinUI version of Uno. Do you know when the WinUI version is expected? Where could I track this info?

inforithmics:
I will try to fix up the Uno build with a pull request.
And I will make a pull request that introduces Project Filters for better development.

https://docs.microsoft.com/en-us/visualstudio/ide/filtered-solutions?view=vs-2019#:~:text=To%20improve%20performance%20when%20opening,and%20enables%20more%20focused%20review.

regarding Uno.WinUI this needs .NET 6 will be released in November. But yes I'm absolutly in favor of basing Mapsui.Uno
on Mapsui.WinUI. 
But I currently try not to use to much unsupported components. Because as I have discovered in my MAUI pull request in Preview packages almost every iteration of a preview breaks things (Thoguh they fix things too).

Regarding WinUI Releases see this page.

Preview is the next WinUI Version. (Has name change of nuget packages).
0.8.* is the current offical supported version (Is Project Reunion).

https://github.com/microsoft/WindowsAppSDK/releases

pauldendulk:
AppVeyor builds correctly so there is probably no real problem with the Uno build, just something not right on my machine, or perhaps things get corrupted because of the wild refactoring I do at the moment. So don't put any effort into this for me, I can work in the current situation.

PR #1333, opened 2021-11-03
Rename Feature to GeometryFeature and MemoryLayer to GeometryMemoryLayer

pauldendulk:
This PR contains two renames and nothing else. It is in a separate PR so that other PRs become more readable.


pauldendulk:
Other PRs will introduce geometryless alternatives.

PR #1334, opened 2021-11-03
Feature/work on features

pauldendulk:
@inforithmics This PR stems from the nullable changes. FetchInfo.Extent was nullable but this made no sense, no fetch is needed when the extent is null. Since it also makes no sense to initialize the Extent with 0,0 0,0 in the constructor FetchInfo is now fully immutable. All values have to be set through the constructor. This is the right way to go. More non-nullable, more immutable. I am also interested in C# 9 records.

### DIFF
--- a/Mapsui.Core/Layers/FetchInfo.cs
+++ b/Mapsui.Core/Layers/FetchInfo.cs
@@ -2,19 +2,25 @@
 {
     public class FetchInfo
     {
-        public FetchInfo() { }
+        public FetchInfo(MRect extent, double resolution, string? crs = null, ChangeType changeType = ChangeType.Discrete)
+        {
+            Extent = extent;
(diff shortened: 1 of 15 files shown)


pauldendulk:
@inforithmics btw IFeature.BoundingBox should perhaps also be non-nullable. 

Also, Perhaps we should rename all Envelope and BoundingBox fields to Extent.

inforithmics:
Good idea.  With my annotation I only showed the current State. But for
better nullability handling changes need to be made like initializing only
with the Constructor. (Or using Immutability, Init only Properties ...)

pauldendulk:
Yes I understand. Step one is to make it explicit. Step two is to avoid nullability. Good you mention init only properties, perhaps they would have been a better solution here. At least it would have resulted in less changes.

PR #1342, opened 2021-11-05
Feature/remove mapsui geometry step by step

pauldendulk:


pauldendulk:
Just created this pull request. When reviewing it myself perhaps it is not a good idea.

pros:
- It is not an abbreviation
- It is not that much longer than Rect
- Rectangle is itself not a really long term that makes reading a sentence harder.

cons:
- Rect is shorter and it is a heavily used in the code base. I have a general rule that you could use abbreviations if they are used a lot for better readability of the code base.
- **In WPF, UWP,  Avalonia and Skia they use Rect.**

Especially that last argument weights heavy. Let's do just as everyone else is doing. Not merging this.

PR #1347, opened 2021-11-05
Remove lat lon constructor from Position

pauldendulk:
@charlenni I noticed the Position had a lat, lon constructor. I had intended Mapsui to always used lon/lat (consistent with x/y) order of coordinates. This PR removes the constructor and replaces it with init properties. The Position is still readonly and there is no order of parameters at all this way.

### DIFF
--- a/Mapsui.UI.Forms/Extensions/PositionExtensions.cs
+++ b/Mapsui.UI.Forms/Extensions/PositionExtensions.cs
@@ -11,7 +11,7 @@
         {
             var result = Projection.SphericalMercator.ToLonLat(point.X, point.Y);
 
-            return new Position(result.lat, result.lon);
+            return new Position { Longitude = result.lon, Latitude = result.lat };
         }
     }
 }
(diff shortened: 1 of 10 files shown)


charlenni:
There are two parts in the Xamarin Forms implementation: MapControl and MapView.

MapControl has the same behavior as all other MapControls. Not more, not less than them.

MapView mimics the behavior of the map control of Xamarin Forms, so that it could replace it. It uses Mapsui, but has nothing to do with Mapsui. Many of the objects, that are used in MapView, aren‘t used by any other part of Mapsui. In this category Position belongs. It mimics the handling of coordinates in Xamarin Forms or in Google Maps. They always in lat/long format.

So I would say, if you couldn’t bring Google to use coordinates in long/lat format, than you should let Position in the old way.

pauldendulk:
This PR is not about changing the order of lat/lon at google or international standards. It is about the order within our code. At some point in our logic we have to move the lat/lon into an x/y coordinate system. For example, in the shapefile sample the source data and the layer data is in epsg:4326, but we draw the map with x/y coordinates that correspond to lon/lat. It would complicate things if we would try to do this any other way (it would be possible by using a different transformation matrix). The best option would be to have a consistent order throughout our code.

That said. I see your point. The Xamarin.Forms.Maps Position indeed has a constructor with lat/lon order. It makes sense to keep the Mapsui version one on one, so I will close this PR.

pauldendulk:
To clarify it a bit more. A way to deal with gps coordinates is to create a epsg:4326 data source (as in the projection point sample). You can add gps coordinates to that source without the need for projection. The projection is up to Mapsui. If you add a Point to such a data source it would take a lon/lat order in the constructor. (while the Position takes lat/lon).

PR #1351, opened 2021-11-06
Rename ILayer.Envelope to ILayer.Extent

pauldendulk:


pauldendulk:
IProvider.GetExtent() is derived from the collective IFeature.Extent and ILayer.Envelope is directly derived from IProvider.GetExtent() so better to call it ILayer.Extent. 

Map.Envelope is the next logical step.

Don't think IWidget.Envelope should be renamed. This something different.

charlenni:
What is the difference of ILayer.Extent, IProvider.Extent, Map.Extent and IWidget.Envelope? Has the last one a special function? Or describes it something different?

pauldendulk:
- IFeature.Extent is the extent of the geometry (or Raster, or MRect) in the feature. Effectively this is the extent of all its vertices (although curved lines could be an exception).
- IProvider.GetExtent() is directly derived from that. It is the combined extent of all the features within the provider
- ILayer.Extent returns the combined extent of all features within the layer or the IProvider.GetExtent() if it has a provider.
- Map.Extent returns the combined extent of all layers.

IWidget.Envelop is the position where the Widget is placed. It is not related to features. Also an important difference is that it is in Mapsui Screen coordinates while the Extents are in World coordinates.

For IWidget field names we should perhaps look at WPF or Avalonia for inspiration.

Issue #1360, opened 2021-11-07
What to make nullable and what not?


pauldendulk:
### How to go about nullability?

With the nullability checks added we are forced to think more about nullability. How I look at it now:
- It is better for things to be non-nullable because consuming classes don't need to check for null.
- For non-nullable you need to either 
  - Initialize the field.
  - Initialize the property.
  - Initialize from constructor arguments.
- Setting some default value seems a simple solution but if you do so you will need a default value that makes sense and often there is no such default. For instance, if a feature has no geometry what is the default Extent? An Extent of 0,0,0,0 is problematic. Suppose you have a layer with features in the range 100,100,200,200. If you add one empty feature with 0,0,0,0 the extent would become 0,0,200,200. So my conclusion is that Feature.Extent should either be nullable or all feature should have a constructor that enforces that the geometry is passed in and a realistic Extent is set. I now propose to do the latter.
- With init properties you still need a default value because the init property does not have to be set by the user. If there is no sensible default you will still need a constructor.
- Some fields have no sensible default and will have to be null.

### About some specific fields

- IFeature.Extent: **not** nullable. Set in constructor.
- IFeature.Geometry/Rect/Point/Raster: **not** nullable. Set in constructor.
- IProvider.GetExtents(): nullable. Unfortunately there could be no features only a null value would make sense.
- ILayer.Extent: nullable. Because derived from provider.
- IMap.Extent: nullable. Because derived from layers.
- IViewport.Envelope: Not sure. I think nullable. You could choose some default but if this is based on non existing data they may result in weird behavior. A zero Width Envelope could cause problems (like divide by zero). Some random value could cause fetching of irrelevant data. Note, this problem may only be relevant during startup while data is loaded but this is an important and often problematic part which we need to get right.
- INavigator.NavigateTo(MRect). Not sure. If it does not allow null the user will need to check. But if the user navigates to a null value (and nothing) happens this may also cause hard to track problems. I now think it should be not-nullable. 

There is a lot more to think about but these were just a few things I encounted.

pauldendulk:
This comment describes my current ideas about which fields should be nullable (or not). I intend to update this comment as my ideas change or expand.

### Nullability guidelines


| Kind | Nullability | Remark |
|---|---|---|
| IFeature.Extent | not-null | Set in constructor. |
| IFeature.Geometry/Rect/etc | not- null | Set in constructor. |
[...]

PR #1363, opened 2021-11-07
Fixed All Nullable Warnings fixed in Mapsui.Geometries Mapsui.Rendering.Skia

inforithmics:
Fixed all nullable warnings in Mapsui Geometries

### DIFF
--- a/Mapsui.Core/Features/RectFeature.cs
+++ b/Mapsui.Core/Features/RectFeature.cs
@@ -2,7 +2,7 @@
 {
     public class RectFeature : BaseFeature, IFeature
     {
-        public MRect? Rect { get; set; }
-        public MRect? Extent => Rect;
+        public MRect Rect { get; set; }
+        public MRect Extent => Rect;
     }
 }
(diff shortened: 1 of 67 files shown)


inforithmics:
Tried to adhere to this.

https://github.com/Mapsui/Mapsui/issues/1360

Mapsui.Core/Features/RectFeature.cs
```diff
@@ -2,7 +2,7 @@
 {
     public class RectFeature : BaseFeature, IFeature
     {
-        public MRect? Rect { get; set; }
-        public MRect? Extent => Rect;
+        public MRect Rect { get; set; } = default!;
```
pauldendulk, line 5: The default for MRect is null but the MRect is not nullable. I would expect a warning here.... but perhaps there is one, because this PR does not fix the warnings for Mapsui.Core.

pauldendulk (approved):
This is wonderful! Next we could look into ways to allow more non-nullability.

Issue #1338, opened 2021-11-04
Map is not refreshing correctly


ndevoecht:
Hello,

I'm using a really simple map with a custom button "locate user". When pressing this button, the map should display a pin on the user position and center on him. 

Problem is: when pressing the button, the pin start moving from 0,0 to x,y. The map doesn't refresh itself, so I need to press the button multiple times. Each time, the pin moves a bit closer to its real position. How can I make this instantaneous ?

Thanks already!

Mapsui: v2.0.3 (same issue in v3.0.0-alpha.1 - I can't update to v3.0.x because it requires Xamarin.Forms 5)
Mapsui.Forms: v2.0.3
SkiaSharp: v2.80.2
Xamarin.Forms: v4.8.0.1451

[map.zip](https://github.com/Mapsui/Mapsui/files/7474199/map.zip)

charlenni:
I assume, that this is the normal behavior of GPS coordinates on a real device. If you activate the GPS the first time, it tries to deliver a first position. This could be from last time. Then it tries to make the position better and better. If you use an emulator, this shouldn't happen.

The only thing you could do is to activate the GPS very early in your code, so that it has much time to get an accurate position.

ndevoecht:
Hey, thanks for the reply. I forgot to say the position returned by the GPS is correct from the start. The pin is set to the correct location. I just don't understand why it travels.

charlenni:
Sounds like you are using an animated navigator. Try to use `mapView.Navigator.CenterOn(sphericalMercatorCoordinate, 0);`. Does this solve your problem?

BTW. `NavigateTo()` change center and resolution at one time.

ndevoecht:
Hey @charlenni, thanks, I've simplified my code!

I found my issue: MyLocationLayer.UpdateMyLocation has a parameter animated which is true by default. Passing it as false fix my issue.

Thanks for your help!

janusw:
> I found my issue: MyLocationLayer.UpdateMyLocation has a parameter animated which is true by default. Passing it as false fix my issue.

Actually I once had the same problem (and found the same fix).

It seems like `animated=true` is a bad default, since apparently it does not work well in all situations. Maybe it would be worth to change it, and use the more robust unanimated update by default, instead of the more fancy but partly-problematic animation. Anyone who really wants the animation should specify that explicitly.

janusw:
[...]

PR #1367, opened 2021-11-08
MyLocationLayer: change default value of 'animated' parameter in update methods

janusw:
It is changed to unanimated updates by default, since the animation can be problematic or misleading in some situations (see e.g. #1338). Cf the rationale in https://github.com/Mapsui/Mapsui/issues/1338#issuecomment-961239334.

### DIFF
--- a/Mapsui.UI.Forms/Objects/MyLocationLayer.cs
+++ b/Mapsui.UI.Forms/Objects/MyLocationLayer.cs
@@ -173,7 +173,7 @@ namespace Mapsui.UI.Objects
         /// Updates my location
         /// </summary>
         /// <param name="newLocation">New location</param>
-        public void UpdateMyLocation(Position newLocation, bool animated = true)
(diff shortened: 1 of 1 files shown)


pauldendulk (approved):
Thanks for taking case of this. Should this also be merged into develop/3.x?

janusw:
> Should this also be merged into develop/3.x?

It's a change of API (a small one, though). Therefore I did not plan to backport it to 3.x, but I can do it if you think it's a good idea.

pauldendulk:
Good to think about the breaking change. Are there currently scenarios where the animation is a useful default?

janusw:
> Good to think about the breaking change. Are there currently scenarios where the animation is a useful default?

There are certainly scenarios where the animation is useful, I just don't think it's a good default value. Anyone who wants the animation can still select it manually.

Since this is more of a change of behavior than an actual bugfix, I'd rather not backport it to 3.x.

pauldendulk:
ok!

PR #1369, opened 2021-11-10
Adding borders to labels

### LINKED ISSUE #1366: More options in LabelStyle?
Thanks for a great library.

I'm trying to make a tool tip for features on the map by adding an additional layer with a label. It is working well so far. To make it look like a normal tool tip I would need sharp corners and a border. This isn't supported by the current `LabelStyle` but would be quite easy to implement in the `LabelRenderer`.

Would there be an interest in a merge request where I add `Rounding`, `BorderThickness` and `BorderColor` to the `LabelStyle` and `LabelRenderer`.

If there's an interest I would create merge requests both for `develop/3.0.0` and `master`.

eriove:
Closes #1366


pauldendulk (approved):
Great,  Thanks a lot.

PR #1383, opened 2021-11-13
Nullable fixes

inforithmics:
Some more nullable fixes mostly in the sample projects


Mapsui.Core/INavigator.cs
```diff
@@ -17,7 +17,7 @@ public interface INavigator
         /// <param name="scaleMethod">Scale method to use to determine the resolution</param>
         /// <param name="duration">Duration of animation in milliseconds</param>
         /// <param name="easing">The type of easing function used to transform from begin tot end state</param>
-        void NavigateTo(MRect extent, ScaleMethod scaleMethod = ScaleMethod.Fit, long duration = 0, Easing? easing = default);
+        void NavigateTo(MRect? extent, ScaleMethod scaleMethod = ScaleMethod.Fit, long duration = 0, Easing? easing = default);
```
pauldendulk, line 20: In a future version this should perhaps be non-nullable. We will need to avoid possible null calls. Perhaps the Home method should only be called if Extent is not null and send the Extent as separate parameter and not as part of the IViewport.Extent (because the IViewport.Extent can be null during startup).

Mapsui.Core/Layers/AnimatedFeatures.cs
```diff
@@ -135,19 +135,19 @@ private static void InterpolateAnimatedPosition(IEnumerable<AnimatedItem> items,
...
-            Debug.WriteLine("DateGps: " + item.Feature["DateGps"]);
-            Debug.WriteLine("DateReceived: " + item.Feature["DateReceived"]);
-            Debug.WriteLine("Longitude: " + item.Feature["Longitude"]);
-            Debug.WriteLine("Latitude: " + item.Feature["Latitude"]);
-
-            Debug.WriteLine("X: " + item.CurrentPoint.X);
-            Debug.WriteLine("Y: " + item.CurrentPoint.Y);
-            Debug.WriteLine("Previous X: " + item.PreviousPoint.X);
-            Debug.WriteLine("Previous Y: " + item.PreviousPoint.Y);
+            Debug.WriteLine("Trackee: " + item.Feature?["Trackee"]);
```
pauldendulk, line 138: If feature is null this will write eight lines of useless information. Better to check earlier.

But then again. This whole logging should be removed anyway.

Mapsui.Core/MRaster.cs
```diff
@@ -5,13 +5,13 @@ namespace Mapsui
 {
     public class MRaster : MRect
     {
-        public MRaster(MRaster raster) : base(raster.Min.X, raster.Min.Y, raster.Max.X, raster.Max.Y)
+        public MRaster(MRaster? raster) : base(raster.Min.X, raster.Min.Y, raster.Max.X, raster.Max.Y)
```
pauldendulk, line 8: I am still on the fence about this one. It complicates a lot of code of we allow the data of the feature to be null. 

Somewhere else I said it was complicated because a Geometry can be Empty, but perhaps we should consider that corrupt data.

Mapsui.Core/Providers/IProvider.cs
```diff
@@ -28,7 +28,7 @@ public interface IProvider<out T> where T : IFeature
         /// <summary>
         /// The spatial reference ID (CRS)
         /// </summary>
-        string CRS { get; set; }
+        string? CRS { get; set; }
```
pauldendulk, line 31: CRS null too? Do you think that is a good idea?

Mapsui.Core/Styles/LabelStyle.cs
```diff
@@ -142,7 +142,7 @@ public LabelStyle(LabelStyle labelStyle)
         /// <summary>
         /// The background color of the label. Set to transparent brush or null if background isn't needed
         /// </summary>
-        public Brush BackColor { get; set; } // todo: rename
+        public Brush? BackColor { get; set; } // todo: rename
```
pauldendulk, line 145: Here too everything null?

pauldendulk (changes requested):
All these changes are only in one direction, making everything nullable. While we would like to go in the other direction. Code will become simpler if we do not allow null and throw early if users do. Now the code if full of line like `if (variable = null) return;` This will prevent the warnings in visual studio but could result in behavior that users do not understand. Users write some code and then nothing happens and you see nothing. That is hard to debug. Perhaps it is not possible to make fields non-nullable, but at least we should investigate the possibility and perhaps rewrite some code to make it possible. If we remove all these warnings in our code by adding null checks everywhere I am afraid we will lose an opportunity to actually make it better.

I am not sure how to go forward with this. I think there should be more discussion about individual changes. Perhaps we should create more issues to discuss it.

inforithmics:
I'll improve this pullrequest with the Suggestions.
The method I used ist that I'll fix the warnings for example a property ist not initialized in the constructor and therefore nullable. Or a property is Set to null and therfore nullable too. This is the reason why I fixed the nullable warnings in the samples first because it demonstrates the usage of the library.

Mapsui.Core/Providers/IProvider.cs
inforithmics: No I don't think this is a good idea. I'll research where this is Set or not Set.

Mapsui.Core/Styles/LabelStyle.cs
inforithmics: Should I'll allow Setting to null and choose the transparent back color when Setting it to null.

Mapsui.Core/MRaster.cs
inforithmics: I'll Set it to non nullable and will fix it are the call sites and throw an exception if it is still null

Mapsui.Core/Layers/AnimatedFeatures.cs
inforithmics: Good point

Mapsui.Core/INavigator.cs
inforithmics: I Made this null to avoid to Check for null in the calling sites. But good point

pauldendulk:
It's difficult. I really would like to make more non-nullable but every time I think deeper I start doubting if it is possible, or should be null. Perhaps I should not be picky in this phase and just accept the choices you make. Every single choice could take a lot of attention.

pauldendulk (approved):

PR #1393, opened 2021-11-27
Fix Maui build

inforithmics:
Pull request to Fix Maui Build GitHub action.


pauldendulk (approved):
Sorry for breaking the build. I should be able to build the MAUI sln myself, I should just work with that one.

PR #1394, opened 2021-11-27
Nullabe Mapsui.Core Fixes

inforithmics:
I fixed the Nullable Warnings in Core

### DIFF
--- a/Mapsui.Core/Fetcher/FeatureFetcher.cs
+++ b/Mapsui.Core/Fetcher/FeatureFetcher.cs
@@ -18,7 +18,7 @@ namespace Mapsui.Fetcher
         public FeatureFetcher(FetchInfo fetchInfo, IProvider<IFeature> provider, DataArrivedDelegate dataArrived, long timeOfRequest = default)
         {
             _dataArrived = dataArrived;
-            var biggerBox = fetchInfo.Extent.Grow(
+            var biggerBox = fetchInfo.Extent?.Grow(
                 SymbolStyle.DefaultWidth * 2 * fetchInfo.Resolution,
                 SymbolStyle.DefaultHeight * 2 * fetchInfo.Resolution);
             _fetchInfo = new FetchInfo(biggerBox, fetchInfo.Resolution, fetchInfo.CRS, fetchInfo.ChangeType);
--- a/Mapsui.Core/Fetcher/FeatureFetcher.cs
+++ b/Mapsui.Core/Fetcher/FeatureFetcher.cs
@@ -31,7 +31,7 @@ namespace Mapsui.Fetcher
         {
             lock (_provider)
             {
-                var features = _provider.GetFeatures(_fetchInfo).ToList() ?? new List<IFeature>();
+                var features = _provider.GetFeatures(_fetchInfo)?.ToList() ?? new List<IFeature>();
                 _dataArrived.Invoke(features, _timeOfRequest);
             }
         }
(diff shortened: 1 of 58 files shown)


Mapsui.Core/Projections/CrsHelper.cs
```diff
@@ -36,18 +37,20 @@ public static int ToEpsgCode(string crs)
...
+            if (crs == null)
+                throw new ArgumentException(nameof(crs));
             if (crs.StartsWith(EpsgPrefix)) return CrsType.Epgs;
             if (crs.StartsWith(EsriStringPrefix)) return CrsType.EsriString;
             if (crs.StartsWith(Proj4StringPrefix)) return CrsType.Proj4String;
             throw new Exception($"crs not recognized: '{crs}'");
         }
 
-        public static bool IsCrsProvided(string fromCRS, string toCRS)
+        public static bool IsCrsProvided([NotNullWhen(true)] string? fromCRS, [NotNullWhen(true)] string? toCRS)
```
pauldendulk, line 48: ah, that is a nice solution.

Mapsui.Core/Layers/FetchInfo.cs
```diff
@@ -2,7 +2,7 @@
 {
     public class FetchInfo
     {
-        public FetchInfo(MRect extent, double resolution, string? crs = null, ChangeType changeType = ChangeType.Discrete)
+        public FetchInfo(MRect? extent, double resolution, string? crs = null, ChangeType changeType = ChangeType.Discrete)
```
pauldendulk, line 5: I don't think we should let the FetchInfo extent be nullable. In general we should try to look for ways to make things non-nullable. That would make the code simpler. Unfortunately this is often hard and I had to accept defeat in several situations. The reason that we can't make things non-nullable is that we often do not control the data, like the features in a provider, of the vertices in a polygon. But in this case we can. We initiate the fetch requests from our own code. It is no use to fetch data of the extent is null, nothing would be returned. So in this case we should check if the Viewport.Extent is null and if so create no request at all. 

In the guidelines here https://github.com/Mapsui/Mapsui/issues/1360#issuecomment-962825664 this would fall in the category of 'arguments', mentioned in the second bullit.

Was there a specific reason you made this nullable?

inforithmics: I saw code that tested for null of this property. So I assumed that it could be null, but I will change it to non nullable

Mapsui.Core/Projections/CrsHelper.cs
inforithmics: List of nullable analyse Attributes.
[!](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis)

Mapsui.Core/Fetcher/FeatureFetcher.cs
```diff
@@ -31,7 +31,7 @@ public void FetchOnThread()
         {
             lock (_provider)
             {
-                var features = _provider.GetFeatures(_fetchInfo).ToList();
+                var features = _provider.GetFeatures(_fetchInfo).ToList() ?? new List<IFeature>();
```
pauldendulk, line 34: Ah good,  we can return an empty list here. And perhaps in a next step IProvider.GetFeatures should also return an empty list.

pauldendulk (approved):
Fabulous 🌈

PR #1396, opened 2021-11-28
add way to examplify leaks on xf samples on phone idiom

tmijieux:
https://github.com/Mapsui/Mapsui/pull/1390#issuecomment-981063664

### DIFF
--- a/Mapsui.UI.Forms/MapView.cs
+++ b/Mapsui.UI.Forms/MapView.cs
@@ -863,5 +863,12 @@ namespace Mapsui.UI.Forms
                 MyLocationLayer?.Dispose();
             }
         }
+
+        public void Reset()
+        {
+            Pins.Clear();
+            Drawables.Clear();
+            HideCallouts();
+        }
     }
 }
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/App.xaml.cs
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/App.xaml.cs
@@ -15,9 +15,9 @@ namespace Mapsui.Samples.Forms
             Logger.LogDelegate += LogMethod;
 
             if (Device.Idiom == TargetIdiom.Phone)
-                MainPage = new NavigationPage(new Mapsui.Samples.Forms.MainPage());
+                MainPage = new NavigationPage(new MainPage());
             else
-                MainPage = new Mapsui.Samples.Forms.MainPageLarge();
+                MainPage = new MainPageLarge();
         }
 
         protected override void OnStart()
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/Leaks.cs
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/Leaks.cs
@@ -1,0 +1,55 @@
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace Mapsui.Samples.Forms
+{
+    public static class Refs
+    {
+        private static ConcurrentQueue<WeakReference> _refs = new();
+
+        public static void AddRef(object p)
+        {
+            _refs.Enqueue(new WeakReference(p));
+        }
+
+        private static void Collect()
+        {
+            GC.Collect(GC.MaxGeneration, GCCollectionMode.Forced);
+            GC.WaitForPendingFinalizers();
+            GC.Collect(GC.MaxGeneration, GCCollectionMode.Forced);
+        }
+
+        public static string Inspect()
+        {
+            Collect();
+            ConcurrentQueue<WeakReference> remainingRefs = new();
+
+            Dictionary<string, int> count = new();
+            foreach (var myref in _refs)
+            {
+                object obj = myref.Target;
+                if (obj != null)
+                {
+                    string k = obj.GetType().FullName;
+                    if (count.TryGetValue(k, out int v))
+                    {
+                        count[k] = v + 1;
+                    }
+                    else
+                    {
+                        count[k] = 1;
+                    }
+                    remainingRefs.Enqueue(myref);
+                }
+            }
+            string report = "Inspect refs:\n\n";
+            foreach (var kv in count)
+            {
+                report += $"{kv.Key} = {kv.Value}\n";
+            }
+            _refs = remainingRefs;
+            return report;
+        }
+    }
+}
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/LeaksPage.xaml
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/LeaksPage.xaml
@@ -1,0 +1,16 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<ContentPage
+  xmlns="http://xamarin.com/schemas/2014/forms"
+  xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+  x:Class="Mapsui.Samples.Forms.LeaksPage"
+  Title="Leaks">
+  <StackLayout Padding="20">
+    <StackLayout Orientation="Horizontal">
+      <Label Text="Enable disposing of mapview on samples: "/>
+      <CheckBox x:Name="enableDisposeCheckbox" HorizontalOptions="EndAndExpand"/>
+    </StackLayout>
+    <Label HorizontalOptions="Center"
+         VerticalOptions="Center"
+         x:Name="label"/>
+  </StackLayout>
+</ContentPage>
--- a/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPage.xaml.cs
+++ b/Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPage.xaml.cs
@@ -16,24 +16,52 @@ namespace Mapsui.Samples.Forms
         public MainPage()
         {
             InitializeComponent();
+            SetupToolbar();
 
             allSamples = AllSamples.GetSamples();
 
-            var categories = allSamples.Select(s => s.Category).Distinct().OrderBy(c => c);
-            picker.ItemsSource = categories.ToList<string>();
+            var categories = allSamples.Select(s => s.Category).Distinct().OrderBy(c => c).ToList();
+            categories.Insert(0, "All");
+            picker.ItemsSource = categories;
             picker.SelectedIndexChanged += PickerSelectedIndexChanged;
-            picker.SelectedItem = "Forms";
+            picker.SelectedItem = "All";
         }
 
         private void FillListWithSamples()
         {
             var selectedCategory = picker.SelectedItem?.ToString() ?? "";
-            listView.ItemsSource = allSamples.Where(s => s.Category == selectedCategory).Select(x => x.Name);
+            if (selectedCategory == "All")
+            {
+                listView.ItemsSource = allSamples.Select(x => x.Name);
+            }
+            else
+            {
+                listView.ItemsSource = allSamples
+                    .Where(s => s.Category == selectedCategory)
+                    .Select(x => x.Name);
+            }
         }
 
         private void PickerSelectedIndexChanged(object sender, EventArgs e)
         {
             FillListWithSamples();
+        }
+
+        private void LeakAction_Clicked(object sender, EventArgs e)
+        {
+            string report = Refs.Inspect();
+            NavigateToPage(new LeaksPage(report));
+        }
+
+        private void SetupToolbar()
+        {
+            var leakButton = new ToolbarItem
+            {
+                Text = "Leaks",
+                Order = ToolbarItemOrder.Secondary
+            };
+            leakButton.Clicked += LeakAction_Clicked;
+            ToolbarItems.Add(leakButton);
         }
 
         void OnSelection(object sender, SelectedItemChangedEventArgs e)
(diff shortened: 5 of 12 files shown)


inforithmics (commented):
Good catch the Android.Designer.cs File is generated and doesn't need to be included in git.

inforithmics (approved):
What I like about this pull request only changes in the samples are needed to have to test for leaks. So the mapsui nuget works as bevor.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPage.xaml.cs
```diff
@@ -16,26 +16,54 @@ public partial class MainPage : ContentPage
...
             allSamples = AllSamples.GetSamples();
 
-            var categories = allSamples.Select(s => s.Category).Distinct().OrderBy(c => c);
-            picker.ItemsSource = categories.ToList<string>();
+            var categories = allSamples.Select(s => s.Category).Distinct().OrderBy(c => c).ToList();
+            categories.Insert(0, "All");
+            picker.ItemsSource = categories;
             picker.SelectedIndexChanged += PickerSelectedIndexChanged;
-            picker.SelectedItem = "Forms";
+            picker.SelectedItem = "All";
```
pauldendulk, line 27: What effect has this change and why was it not 'All' before?

pauldendulk:
In your comment you mention a [Test] but not in the PR. Why is that?

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPage.xaml.cs
charlenni: I assume, it was my selection, because I tested at this time only Forms tests.

pauldendulk:
@tmijieux I scrolled through the code. I am not very familiar with this part of the code. What I miss when reviewing is that I don't easily see why things have changed. Perhaps it would help if you wrote some code comments as part of for instances the Leaks.cs file. Not the obligatory comments explaining what could be read from the code but the general purpose. What are you trying to achieve and what is the purpose of a certain class, or certain method.

@charlenni Any thoughts? You are most familiar with this part.

pauldendulk:
@tmijieux And there is a conflict now, please 'git pull origin master'.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPage.xaml.cs
```diff
@@ -16,26 +16,54 @@ public partial class MainPage : ContentPage
...
         }
 
         private void FillListWithSamples()
         {
             var selectedCategory = picker.SelectedItem?.ToString() ?? "";
-            listView.ItemsSource = allSamples.Where(s => s.Category == selectedCategory).Select(x => x.Name);
+            if (selectedCategory == "All")
+            {
+                listView.ItemsSource = allSamples.Select(x => x.Name);
```
tmijieux, line 35: the "all" category is "pseudo"-category that i added that show all samples at once, because the screen is much larger than what there is to display in a single category and when playing with the samples app it is just tedious to have a switch category all the times, it just add a lot more clicks to the experience and it just slow me downs a lot. (Plus the default apparence for the category picker does not make it obvious at all that it is actually a picker)

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPage.xaml.cs
```diff
@@ -16,26 +16,54 @@ public partial class MainPage : ContentPage
...
+        }
+
+        private void SetupToolbar()
+        {
+            var leakButton = new ToolbarItem
+            {
+                Text = "Leaks",
+                Order = ToolbarItemOrder.Secondary
+            };
+            leakButton.Clicked += LeakAction_Clicked;
```
tmijieux, line 63: this add a button in the contextual menu of the app navigation to go to the new "LeaksPage"

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPage.xaml.cs
```diff
@@ -47,12 +75,24 @@ void OnSelection(object sender, SelectedItemChangedEventArgs e)
...
+            NavigateToPage(new MapPage(sample, clicker));
 
             listView.SelectedItem = null;
         }
+
+        public static async void NavigateToPage(Page page)
+        {
+            try
+            {
+                await ((NavigationPage)Application.Current.MainPage).PushAsync(page);
```
tmijieux, line 90: since the PushAsync method is async i just wrapped it in a try catch block in a async void method, because if a async method throws without being awaited this can leaks memory and this is basically undefined behaviour.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPageLarge.xaml
```diff
@@ -1,8 +1,9 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
-			 xmlns:mapsui="clr-namespace:Mapsui.UI.Forms;assembly=Mapsui.UI.Forms"
-             x:Class="Mapsui.Samples.Forms.MainPageLarge">
+	     xmlns:mapsui="clr-namespace:Mapsui.UI.Forms;assembly=Mapsui.UI.Forms"
+             x:Class="Mapsui.Samples.Forms.MainPageLarge"
+             Title="All samples">
```
tmijieux, line 6: here i added a title to the page because the place where the titles of the app is supposed to appear in the navigation bar was empty

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPageLarge.xaml.cs
```diff
@@ -102,6 +102,7 @@ void OnSelection(object sender, SelectedItemChangedEventArgs e)
 
             if (sample != null)
             {
+                mapView.Reset();
```
tmijieux, line 105: this is another issue i found with the samples, basically on the UWP part on windows, each time you switch samples all the pins/ drawable and callout from previous samples remains forever, I just added this 'Reset()' method to remove all pins drawable and callout from the mapsuiforms:MapView and start with a fresh sample

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MapPage.xaml.cs
```diff
@@ -22,9 +24,13 @@ public MapPage()
             InitializeComponent();
         }
 
-        public MapPage(Action<IMapControl> setup, Func<MapView?, MapClickedEventArgs, bool>? c = null)
+        public MapPage(ISample sample, Func<MapView?, MapClickedEventArgs, bool>? c = null)
         {
             InitializeComponent();
+            Refs.AddRef(this);
+            Refs.AddRef(mapView);
```
tmijieux, line 31: Here i call my helper class to track the instance of the MapPage and the MapView inside my helper and detect if they are leaked

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MapPage.xaml.cs
```diff
@@ -22,9 +24,13 @@ public MapPage()
             InitializeComponent();
         }
 
-        public MapPage(Action<IMapControl> setup, Func<MapView?, MapClickedEventArgs, bool>? c = null)
+        public MapPage(ISample sample, Func<MapView?, MapClickedEventArgs, bool>? c = null)
```
tmijieux, line 27: here i pass the whole ISample instead of the setup function just to allow to set the page title just below. (the title display on the top of the app in the navigation bar)
This way we can immediately see on what sample we are, before we add to remember on what sample we choose, for a person with large finger on a small phone screen it can be difficult to say if you tapped the right place on the screen.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MapPage.xaml.cs
```diff
@@ -60,6 +66,24 @@ protected override void OnAppearing()
...
+            mapView.Info -= MapView_Info;
+
+            Compass.ReadingChanged -= Compass_ReadingChanged;
+            CrossGeolocator.Current.PositionChanged -= MyLocationPositionChanged;
+            CrossGeolocator.Current.PositionError -= MyLocationPositionError;
+
+            if (LeaksPage.DisposeMapView)
+            {
+                mapView.Dispose();
+            }
```
tmijieux, line 84: this is called when the page disappear. In the particular case of this app,this happens only when the page is removed from the navigation stack (since we never push another page on top of this one)

here i dispose of the mapview with a condition, we can toggle the `LeaksPage.DisposeMapView` boolean on the new `LeakPage` to check that if it is false , the page is really leaked , we can check that in the LeaksPage .


This whole stuff about the leakpage and this condition i am actually not willing to keep in the samples because I think the samples is supposed to look exactly like what a person using mapsui is willing to write his/her own app but this is just some kind of artifact designed to prove to you (the developers) that the leaks are reals.

Mapsui.UI.Forms/MapView.cs
```diff
@@ -843,5 +843,12 @@ private ButtonWidget CreateButton(float x, float y, SKPicture picture, Action<ob
 
             return result;
         }
+
+        public void Reset()
+        {
+            Pins.Clear();
+            Drawables.Clear();
+            HideCallouts();
```
tmijieux, line 851: i added this method to remove all pins , callout and drawables in a mapsuiforms:MapView but the name of the method is not very good. I use this in the UWP forms sample because there is a bug where samples do not reset to a clean map when you switch samples.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/App.xaml.cs
```diff
@@ -15,9 +15,9 @@ public App()
             Logger.LogDelegate += LogMethod;
 
             if (Device.Idiom == TargetIdiom.Phone)
-                MainPage = new NavigationPage(new Mapsui.Samples.Forms.MainPage());
+                MainPage = new NavigationPage(new MainPage());
             else
-                MainPage = new Mapsui.Samples.Forms.MainPageLarge();
+                MainPage = new MainPageLarge();
```
tmijieux, line 20: the full namepaces here are just useless because the is the namespace we are currently in in this current file.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/Leaks.cs
```diff
@@ -0,0 +1,55 @@
...
+                    {
+                        count[k] = v + 1;
+                    }
+                    else
+                    {
+                        count[k] = 1;
+                    }
+                    remainingRefs.Enqueue(myref);
+                }
+            }
```
tmijieux, line 45: this is my helper class to detect leaks, it just keep weakreferences to chosen objects in a concurrentqueue and count the objects that are still alive at a chosen time. As i said in another comment, this is just to examplify that leaks happens if you dont dispose with the current implementations but i am not sure I actually want to keep this in a sample.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/LeaksPage.xaml
```diff
@@ -0,0 +1,16 @@
...
+  xmlns="http://xamarin.com/schemas/2014/forms"
+  xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
+  x:Class="Mapsui.Samples.Forms.LeaksPage"
+  Title="Leaks">
+  <StackLayout Padding="20">
+    <StackLayout Orientation="Horizontal">
+      <Label Text="Enable disposing of mapview on samples: "/>
+      <CheckBox x:Name="enableDisposeCheckbox" HorizontalOptions="EndAndExpand"/>
+    </StackLayout>
+    <Label HorizontalOptions="Center"
```
tmijieux, line 12: This is the new leakpage with a checkbox to enable or disable the disposing of the mapview on the sample page.
And it also show a report of the liveness state of the currently tracked objects.

tmijieux (commented):
@pauldendulk i did a self review to explain my code. Also i did not yet add automatic tests because I currently have leaks and native crash issues with my business app so i try to focus on this for now and i dont have too much time for mapsui right now.

EDIT: As i said in some of the comments below, i personnaly dont want to really want to merge this in this current state, 
I havent yet read the PR about removing the timers, i will try to understand it first and do a statement later.

In the mean time i put this PR in draft mode

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPageLarge.xaml.cs
charlenni: Good idea to do this

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MapPage.xaml.cs
charlenni: Good idea, but has nothing to do with your other changes. Here an extra PR would be better. But I know, if you work and find such things, you don't want to create a new PR :-)

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MapPage.xaml.cs
charlenni: You are right, that these things shouldn't go into the normal samples. For testing they are needed, but not as samples. Would it be possible to create a new sample project beside Samples.Forms for this?

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MapPage.xaml.cs
tmijieux: i think i will eventually remove the optional leak stuff and page and just rename this PR to minor improvements in samples most probably...

inforithmics:
1. I'd like to merge this because there some Code Cleanups for the examples. So could you fix the Merge Conflicits?

tmijieux:
> 1. I'd like to merge this because there some Code Cleanups for the examples. So could you fix the Merge Conflicts?

ok, should I just remove the "leak" stuff and leave only the other code cleanups ?

inforithmics:
In my opinions the leak Stuff can stay. Because it helps Find leaks in Android and iOS. You could remove the Fragt Status than this can be merged.

Issue #1395, opened 2021-11-28
Animation without Timers


pauldendulk:
### Background

This is an idea which has been in the back of my mind for a long time. I could not make time for it so far and expect this won't change soon. By creating this issue and having some discussion around it perhaps we could make some progress. Perhaps someone else could pick this up.

### Why No Timers?
- **Performance:** Timers consume a thread and thread starvation can cause serious performance problems. A discrete performance decrease can arise when virtual threads are used instead of CPU threads.
- **Memory Leaks:** Timers need to be disposed and if they don't get disposed cause memory leaks https://github.com/Mapsui/Mapsui/pull/1390
- **Growing number of timers:** We have various things that need to be animated, like the animated viewport or animated features. Animated features get triggered by a data source with it's own timer. Multiple animated layers will mean multiple timers.
- **Syncing:** We can avoid the problem of syncing the animation state with the render update loop frequency. This frequency can vary, so it is hard to know in advance what the state of an animation should be on the next draw operation.

### How?
- Only the render loop should have a frequent update. Perhaps we need a single timer here, but I think it is possible to remove that one too (Let the renderer call Invalidate and do a Thread.Sleep before the next update is due). 
- In the render loop before rendering call een UpdateAnimations method on the objects supporting animation. The implementations should keep track of the origin-time/position and the target-time/position, and the easy function. The UpdateAnimation method will calculate the current position based on that. 
- A variation on the above is that this calculation is done in the Layer.GetFeatures or Provider.GetFeatures methods. Perhaps this is better.
- We should not render when nothing has changed. Now we render based on the invalidate state. This is state that is triggered by an event. Perhaps we should be able to call invalidate with an end time.

### Alternative solutions
- Perhaps more efficient timers with thread reuse (Task.Run and other things in the async/await realm).
- A single global timer (or a more fancy schedular) where all classes can make use of.

charlenni:
I had made this before, but dropped it, because you changed the Navigator in your way.

With #1397 there is only one timer in Mapsui: the render loop timer.  With these changes, there is only on static `Animations` object, which collects all `AnimationEntry`s. The `UpdateAnimations` function is called from the render loop. If there are any entries in the animations list, then an invalidation of the screen is made.

PR #1397, opened 2021-11-29
Animation

charlenni:
Added a central Animations object, that handles all AnimationEntrys. Animations could now repeat (start them again, if the duration is expired).

See #1395


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -95,7 +94,78 @@ private void CommonDrawControl(object canvas)
 
         private void InvalidateTimerCallback(object? state)
         {
-            if (!_refresh)
+            // Check, if we have to redraw the screen, because a animation is running or a refresh is wished
+
+/* Unmerged change from project 'Mapsui.UI.Uwp'
```
pauldendulk, line 99: What is this comment section? Is this generated somehow?

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -95,7 +94,78 @@ private void CommonDrawControl(object canvas)
...
+            if (!Animation.UpdateAnimations() && !_refresh)
+*/
+
+/* Unmerged change from project 'Mapsui.UI.Uno (netstandard2.0)'
+Before:
+            if (!Animations.UpdateAnimations() && !_refresh)
+After:
+            if (!Animation.UpdateAnimations() && !_refresh)
+*/
+            if (!Utilities.Animation.UpdateAnimations() && !_refresh)
```
pauldendulk, line 168: Looks good.

Samples/Mapsui.Samples.Common/Maps/Special/AnimationsSample.cs
```diff
@@ -28,13 +28,7 @@ public static Map CreateMap(IMapControl mapControl)
...
             var animations = CreateAnimationsForSymbolStyle(style);
 
-            // todo: Introduce one Animation for the MapControl that could be reused here.
-            var animation = new Animation();
-            animation.Ticked += (s, e) => {
-                animation.UpdateAnimations();
-                mapControl.Refresh();
-            };
-            animation.Start(animations, 10000);
+            Animation.Start(animations, 10000);
```
pauldendulk, line 31: The changes here demonstrate why this is a good improvement. Much better if there is no need to add timers in user code.

pauldendulk, line 31: The changes here demonstrate why this is a good improvement. Much better if there is no need to add timers in user code.

Mapsui.Core/Utilities/Animation.cs
```diff
@@ -1,158 +1,152 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
-using System.Timers;
 
 namespace Mapsui.Utilities
 {
-    public class Animation : IDisposable
+    public static class Animation
```
pauldendulk, line 6: This is great improvement. No timers in Animation and IDisposable can be removed.

Mapsui.Core/Navigator.cs
```diff
@@ -126,7 +125,9 @@ public void NavigateTo(MPoint? center, double resolution, long duration = -1, Ea
                     return;
 
 
-                _animation.Start(animations, duration);
+                Animation.Start(animations, duration);
+
+                _lastAnimations = animations;
```
pauldendulk, line 130: At this point I don't have a clear picture of how this works. _lastAnimatios is a list of AnimationEntry. They are stopped through the Animation.Stop method, but need to be stored in the Navigator. I will have to step through it in code. I will merge first. Always hard for me to understand these diffs.

pauldendulk (approved):
Great, Thanks a lot!

tmijieux:
why putting the animations in global static variables?
This is still asking for leaks as you put callback referencing the navigator for tick and final in the animation entry ?
You still need to dispose the navigator and stop all navigator animations in the dispose call for this to not leak.
Here, you replaced the global reference holding the timer by a global references on the `_entries` variable

Also , you could have one animation container per mapcontrol, I have multiple mapcontrol in my app (not all visible at the same time though). In thsi code, couldn't an animation on one of the map trigger a redraw on the other ?

Why not just let the animations callback (Tick and Final) call RefreshGraphics() to require a redraw and have a code like that
 ```csharp
void InvalidateTimerCallback()
{
     _animations.Update();
     if (_needRedraw)
     {
         ....
     }
 }
 ```

charlenni:
@tmijieux You are right. Using static classes is the way Paul would do it.

To have a Animation object in each MapControl would be a good thing. I could change Animation in this direction, if Paul like to have it in this way.

Regarding your suggestion at the end: I would say, that it is enough, if one of the AnimationEntrys is running. The only thing, that I could think about is, that an animation wouldn't change anything in the map. And than the user should use another animation object.

pauldendulk:
> Why not just let the animations callback (Tick and Final) call RefreshGraphics() to require a redraw 

A disadvantage of this method that there is no guarantee that the redraw will happen immediately. You change the animation state, request a redraw, but since you do not know how long iterations take they can be out of sync. 

But more important is that in general I try to avoid a design where a possibly large number of individual features/animations/layers start calling updates or send notifications. I am fully aware that this the common design for controls in user interfaces, but that is another kind of application. A use interface does not have thousands of controls, and if there would be, you will notice a performance degradation. 

In the past I used WPF xaml render to draw the map. This made it also possible to use the WPF hit detection which we could use to determine which item was clicked. The performance of this design was terrible. One iteration lasted seconds.

Most users of Mapsui are UI developers themselves and are familiar with that kind of architecture, and are inclined to use that design in Mapsui as well, but this is just not appropriate here. We should go for flat render iterations and try to make them as fast as possible. All we need to know about an animation is if it is running or not. 

About the use of static. I am a big fan of stateless and prefer the use of pure functions and try to limit the use of stateful objects. What is worse than that is static state. Unfortunately Mapsui has static state in a few places. This is often problematic.

Mapsui.UI.Shared/MapControl.cs
charlenni: I didn't creat this comment.

Mapsui.Core/Navigator.cs
charlenni: Navigator should have the possibility to stop its own animations (fling gesture and than stop by hand). For this, Navigator saves the last started animation, so that it could stop them.

tmijieux:
The main idea of this pull request is to use the timer of the mapcontrol invalidate loop to update the animations. 
The purpose of animations is to update the state of map or the state of viewport (through navigator) so it makes perfect sense to me to store animations in either map or mapcontrol rather than store them globally.

I agree with you that stateless functionnal is better than stateful but here we have global state which is stateful + there still is a memory leak which was my main issue in the first place and the very reason I proposed to add the Dispose api in the first place.

(static Animation._entries -> AnimationEntry -> Ticked event handler-> Navigator -> Navigated event handler -> MapControl -> whole ui hierarchy through Parent property on Xamarin.Forms)
so even if api looks nice because it looks like a functional call, it is only in looks.

If we put the AnimationEntries in the mapcontrol we can still remove the IDisposable interface on both Navigator and Animation class (but we need to keep it on MapControl) and user of the library can create as many AnimationEntry as they want without worrying about the timer . To start the animations , you would have to write `myMapControlInstance.StartAnimation(myAnimationEntry)` instead of calling the static `Animations` class

Mapsui.UI.Shared/MapControl.cs
pauldendulk: But it is added as part of your merge request so you need to try to find out what happened.

Mapsui.Core/Navigator.cs
pauldendulk: Perhaps Navigator should own the AnimationEntries and Animation should have no array of AnimationEntries at all.

charlenni: But how should Animation than update the AnimationEntrys?

Xamarin.Forms has an implementation, where you could name the AnimationEntrys with a string. Perhaps this would be a way?

pauldendulk:
First of a meta remark. Discussions like these are hard. We do not know each other. It is easy to make wrong assumptions about what the other is thinking.

> The main idea of this pull request is to use the timer of the mapcontrol invalidate loop to update the animations. 

Yes, I like this merge requests because it removes the timer from the Animations class.

> The purpose of animations is to update the state of map or the state of viewport (through navigator) so it makes perfect sense to me to store animations in either map or mapcontrol rather than store them globally.

I think we agree. The next step is to get rid of the static AnimationEntry array in Animation. I already placed a remark questioning the way those were shared between the Navigator and the Animation. At that moment I was thinking to make the Navigator the owner of the AnimationEntries. The Animation could be stateless. Animation.UpdateAnimations could be called with the AnimationEntry array. But an alternative could be to add it to the MapControl, or I think it would be better use the Map. But those are still central places, maybe we should push ownership all the way to the component that is animated. All the options are still open here.

btw. Current master is alpha v4. A next major release of v4 will take some time. This is work in progress.
 
> If we put the AnimationEntries in the mapcontrol we can still remove the IDisposable interface on both Navigator and Animation class (but we need to keep it on MapControl) and user of the library can create as many AnimationEntry as they want without worrying about the timer . To start the animations , you would have to write `myMapControlInstance.StartAnimation(myAnimationEntry, delay)` instead of calling the static `Animations` class

This suggestion is about touches on the topic of triggering the animation. We can distinguish two topics in this discussion:
- Who is the owner of the AnimationEntries
- How to start/stop/update/trigger an animation.

My previous comment was related to that second point and I think I still have a lot of explaining to do regarding that. What I don't want here is initiative of the individual components that needs to be animated. It can be passive. This is better for performance and makes it easier to track the logic (I do not like events going in all kinds of directions). I will post more about this in another comment (have to bring to kids to school now).

charlenni:
Than lets collect, what is important for us.

- Animation should have one point to update all animations, so that it could be called from render loop
- Animation should be usable from all parts of the library and from the developer side
- All AnimationsEntrys should be collected in a central Animation, that belongs to MapControl
- All AnimationEntrys should be deleted, if the MapControl is disposed
- Animation is part of MapControl, not of Map (Map holds only data, but has nothing to do with rendering)
- Animation should have the possibility to be aborted from the owner
- If a animation is running, the screen needs a redraw. No additional check is made. 
- Perhaps AnimationEntrys should have a marker, so that they could be easiely found in the list
- Perhaps AnimationEntrys should have childs, so that all animations, which belong together, are sychronized

Any other comments?

tmijieux:
@pauldendulk 
> First of a meta remark. Discussions like these are hard. We do not know each other. It is easy to make wrong assumptions about what the other is thinking.
> btw. Current master is alpha v4. A next major release of v4 will take some time. This is work in progress.

Yes I am aware of that, noticed that some API changed compared to what is currently available in 3.0.2. I currently use mapsui 3.0.2 without the nuget but directly from source in my business project and I apply patch to make it works exactly the way i want. Please do not feel any pressure about releasing any time soon. I'm totally ok with that. I personnaly think mapsui is great piece of software 👍 . I like the fact that it is open source(which allows me to use it the way i do), not tied with the default alternative which the google map, and i basically read a large part of the code and think it is really good. Thank you for making this library 😄 . Since I have interest on it (my business depends on it),I tend to be sometimes too much direct and I have a strong opinion, and often sound harsh when I state what I think about what the best direction for software is. Do not hesitate to tell me to stop and ditch my ideas, if i'm am too much invading, I won't hold a grudge. 
Sorry for assuming some of your thoughts. :bow:

The reasons i am making PRs is I dont want to branch in a completely other direction than your main repo, So i am making PRs about the changes that I deemed necessary for the well being of my app and that I consider could be beneficial for general mapsui usage. 

I had a bad experience with memory leaks. It was not really mapsui fault, because i found many leaks in all my application some, coming from my own code, and some from other libraries as well. There is also a bug that triggered native crash in the garbage collector that was probably more likely in low memory condition, the bug was not reproducible on any of the developers test devices and suddeny appeared in production 😢 . I know i am excessively conscious about this leaks problems right now, but i think it can quickly become a deal breaker (hopefully it is not for us) so this is why I am insisting that care should be taken to avoid leaks. Like i said, I have no time constraints, so take all the time you need to design the best API ever !! 👍 😉

pauldendulk:
@tmijieux Thanks for your reply. My meta remark was mainly with the upcoming discussion about the animation design in mind. After years of open source work I still have not found a good way to have an effective discussion about more complicated topics. I often have ideas about a general direction without knowing the exact implementation yet. Often this leads to misunderstandings. Perhaps we should have more video calls.

The memory leaks are a real issue in the current release so by pointing this out you help me and you help other users. This just needs to be fixed, so thanks for taking initiative in this.

PR #1398, opened 2021-11-29
More nullable fixes

inforithmics:
Some nullable fixes in Mapsui Maui Samples


inforithmics:
I changed the Code so that 
` new RasterFeature((MRaster?)null)`
is now allowed

pauldendulk:
Auch! the nullable Raster is indeed a complication. Null is used here to indicate the tile does not exist and should not be fetched again. I would prefer to have non-nullalbe rasters but that will take some redesign.

I always had this idea that perhaps a raster should be a BoundingBox (now MRect) that has a style that contains the tile image. A raster would then be more like a point with a symbol style. When introducing MRect I thought about it again and I now think it makes more sense because MRect is not tied to the geometry hierarchy. If we go for that Style field could be set to null instead (or empty list if it is a list).

Mapsui.Core/Layers/ImageLayer.cs
```diff
@@ -155,7 +155,7 @@ private void StartNewFetch(FetchInfo fetchInfo)
             });
         }
 
-        private void DataArrived(IEnumerable<IFeature> arrivingFeatures, object? state)
+        private void DataArrived(IEnumerable<IFeature>? arrivingFeatures, object? state)
```
pauldendulk, line 158: This is an argument to a function so I would think the caller is responsible for passing in an empty list if there are no features. I will merge this in it's current state though.

Samples/Mapsui.Samples.Maui/Mapsui.Samples.Maui.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>net6.0-android;net6.0-ios;net6.0-maccatalyst</TargetFrameworks>
+		<!--<TargetFrameworks>net6.0-android;net6.0-ios;net6.0-maccatalyst</TargetFrameworks>-->
```
pauldendulk, line 4: Is this meant as temporary? If yes, should it be reverted. If no, should it be deleted?

Samples/Mapsui.Samples.Wpf.Editing/Editing/EditManager.cs
```diff
@@ -213,10 +213,10 @@ public bool TryDeleteVertex(MapInfo? mapInfo, double screenDistance)
         {
             if (mapInfo?.Feature is GeometryFeature geometryFeature)
             {
-                var vertexTouched = FindVertexTouched(mapInfo, geometryFeature.Geometry.MainVertices(), screenDistance);
+                var vertexTouched = FindVertexTouched(mapInfo, geometryFeature.Geometry?.MainVertices() ?? new List<Point>(), screenDistance);
```
pauldendulk, line 216: Perhaps it makes more sense to not call the method at all if there are no vertices to be found.

Samples/Mapsui.Samples.iOS/Main.cs
```diff
@@ -9,7 +10,7 @@ static void Main(string[] args)
         {
             // if you want to use a different Application Delegate class from "AppDelegate"
             // you can specify it here.
-            UIApplication.Main(args, null, "AppDelegate");
+            UIApplication.Main(args, (Type?)null, typeof(AppDelegate));
```
pauldendulk, line 13: Did you mean nameof instead of typeof? Perhaps typeof was intended but it would call another method. Did you test if this still works?

pauldendulk (approved):
This catches a lot of possible null reference exceptions. Please go through the remarks, there may be some things that you want to adjust in a later PR.

Samples/Mapsui.Samples.Maui/Mapsui.Samples.Maui.csproj
inforithmics: Yes it ist a temporary change. I will reenable it with the Next Maui Preview.

Samples/Mapsui.Samples.iOS/Main.cs
inforithmics: No the the string overloads are obsolete, so I used the Type overloads.

PR #1399, opened 2021-11-30
[WIP] Next Animation

charlenni:
@tmijieux Changed Animation from a static version to an instance version. Animation is created for MapControl. Disadvantage is, that Animation could only be used from objects, which have access to MapControl. Perhaps you have a better idea.

### DIFF
--- a/Mapsui.Core/Navigator.cs
+++ b/Mapsui.Core/Navigator.cs
@@ -10,6 +10,7 @@ namespace Mapsui
     {
         private readonly Map _map;
         private readonly IViewport _viewport;
+        private readonly Animation _animation;
         private double _rotationDelta;
         private List<AnimationEntry> _lastAnimations = new();
 
@@ -26,10 +27,11 @@ namespace Mapsui
 
         public EventHandler<ChangeType>? Navigated { get; set; }
 
-        public Navigator(Map map, IViewport viewport)
+        public Navigator(Map map, IViewport viewport, Animation animation)
         {
             _map = map;
             _viewport = viewport;
+            _animation = animation;
         }
 
         private void AnimationTimerTicked(object sender, AnimationEventArgs e)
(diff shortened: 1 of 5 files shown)


pauldendulk:
This is not what I had in mind. 

What I was thinking was to keep the Animation class static but remove this list of AnimationEntries. Call AnimationUpdate on those things that are themselves animated. Or better, don't call AnimationUpdate at all but leave it up to the animated item to return it's current animation state given the time of update. This is how the AnimatedFeatures work. There are some things in that part of the code that needs to be improved, but the way animation works is how I want it. You just call GetFeatures at a certain time. The class knows the current time and knows the start and end of the animation so that it can return a feature in the state that corresponds to the animation at that time.

To use the same mechanism for viewport animations and style animations will need more work. I will work on this, but there is little I can do the coming week, this is the busiest time of the year for me.

charlenni:
@pauldendulk But how should this work? You don't want to have a list of animations, but you want to call them from within the render loop. You want a static Animation class, but it should be different between different MapControls.

If you do it on Feature basis, then you have to check each Feature, if there is an animation running. Or you have somewhere a list, which Features are animated. And for navigation you need a different system.

But I have to wait to see your results.

pauldendulk:
> @pauldendulk But how should this work? You don't want to have a list of animations, but you want to call them from within the render loop. You want a static Animation class, but it should be different between different MapControls.

The static Animation class will have no static list of AnimationEntries. The AnimationEntries will be stored more locally. This could be: MapControl, Map, Layer, Feature. I currently playing with a solution where I store it in the Layer.
 
> If you do it on Feature basis, then you have to check each Feature, if there is an animation running. 

Yes, pushing it all the way down to the individual feature may be a step too far.

> Or you have somewhere a list, which Features are animated. And for navigation you need a different system.

Yes, navigation as it currently works does not fit in this logic. Perhaps animation should be inside the Viewport. Or we should change the parameters to the rendering loop.

PR #1401, opened 2021-12-02
Some steps towards a different animation design

pauldendulk:


Mapsui.Core/UI/ViewportLimiterKeepWithin.cs
```diff
@@ -38,9 +38,9 @@ public class ViewportLimiterKeepWithin : IViewportLimiter
             return new MinMax(mostZoomedOut, mostZoomedIn);
         }
 
-        public void Limit(IViewport viewport, IReadOnlyList<double> mapResolutions, MRect? mapEnvelope)
+        public void Limit(Viewport viewport, IReadOnlyList<double> mapResolutions, MRect? mapEnvelope)
         {
-            viewport.SetResolution(LimitResolution(viewport.Resolution, viewport.Width, viewport.Height, mapResolutions, mapEnvelope));
+            viewport.Resolution = LimitResolution(viewport.Resolution, viewport.Width, viewport.Height, mapResolutions, mapEnvelope);
```
inforithmics, line 43: Good improvement. Set Methods should always be a Property

Mapsui.Core/Utilities/TransformationAlgorithms.cs
```diff
@@ -0,0 +1,26 @@
+﻿namespace Mapsui.Utilities
+{
+    public static class TransformationAlgorithms
```
inforithmics, line 3: I like it when logic gets detached from implementations, So this could be tested easier.

Mapsui.Core/ViewportAnimations/FlingAnimation.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using System.Collections.Generic;
+using Mapsui.Extensions;
+using Mapsui.Utilities;
+
+namespace Mapsui.ViewportAnimations
+{
+    public static class FlingAnimation
```
inforithmics, line 8: Nice Improvement so the Animation is now at one place.

Mapsui.Core/ViewportAnimations/FlyToAnimation.cs
```diff
@@ -0,0 +1,77 @@
+﻿using System;
+using System.Collections.Generic;
+using Mapsui.Geometries;
+using Mapsui.Utilities;
+
+namespace Mapsui.ViewportAnimations
+{
+    public class FlyToAnimation
```
inforithmics, line 8: And Fly Animation is at one place too.

Mapsui.Core/Widgets/Zoom/ZoomInOutWidget.cs
```diff
@@ -153,11 +153,11 @@ public override bool HandleWidgetTouched(INavigator navigator, MPoint position)
             if (Orientation == Orientation.Vertical && position.Y < Envelope.MinY + Envelope.Height * 0.5 ||
                 Orientation == Orientation.Horizontal && position.X < Envelope.MinX + Envelope.Width * 0.5)
             {
-                navigator.ZoomIn();
+                navigator.ZoomIn(500);
```
inforithmics, line 156: Is there a different default in Zoom in then 500 (is it now faster?)

Mapsui.UI.Avalonia/MapControl.cs
```diff
@@ -51,7 +51,6 @@ private void Initialize()
 
             PointerWheelChanged += MapControlMouseWheel;
 
-            Tapped += OnSingleTapped;
```
inforithmics, line 54: Did this not work? Or is it only a Touch event and not a Mouse Event?.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -96,78 +96,15 @@ private void CommonDrawControl(object canvas)
 
         private void InvalidateTimerCallback(object? state)
         {
-            // Check, if we have to redraw the screen, because a animation is running or a refresh is wished
```
inforithmics, line 99: Good to remove this I think this is some merge Conflict left over.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -96,78 +96,15 @@ private void CommonDrawControl(object canvas)
...
-            */
-
-            /* Unmerged change from project 'Mapsui.UI.Uno (netstandard2.0)'
-            Before:
-                        if (!Animations.UpdateAnimations() && !_refresh)
-            After:
-                        if (!Animation.UpdateAnimations() && !_refresh)
-            */
-            if (!Utilities.Animation.UpdateAnimations() && !_refresh)
+            // Check, if we have to redraw the screen
```
inforithmics, line 99: Good so the animations are triggered by the Timer and are synchronous with the refresh rat.

Mapsui.UI.Wpf/MapControl.cs
```diff
@@ -174,9 +174,6 @@ private void RunOnUIThread(Action action)
 
         private void MapControlMouseLeftButtonDown(object sender, MouseButtonEventArgs e)
         {
-            // We have a new interaction with the screen, so stop all navigator animations
-            Navigator.StopRunningAnimation();
```
inforithmics, line 178: Comment to myself. So the animations are now in the Viewport. And because they are triggered by the Refresh Timer. There is no need to stop them.

Mapsui/Providers/Wms/WmsProvider.cs
```diff
@@ -483,14 +484,7 @@ public IEnumerable<IFeature> GetFeatures(FetchInfo fetchInfo)
...
 
-            var view = new Viewport
-            {
-                Resolution = fetchInfo.Resolution,
-                Center = fetchInfo.Extent.Centroid,
-                Width = (fetchInfo.Extent.Width / fetchInfo.Resolution),
-                Height = (fetchInfo.Extent.Height / fetchInfo.Resolution)
-            };
-            if (TryGetMap(view, out var raster))
+            if (TryGetMap(fetchInfo.ToViewport(), out var raster))
```
inforithmics, line 487: Is ToViewport a new extension method?.

Samples/Mapsui.Samples.Common/Maps/Animations/PointFeatureAnimationSamples.cs
```diff
@@ -10,17 +10,16 @@
...
 namespace Mapsui.Samples.Common.Maps.Special
 {
-    public class AnimatedPointsSample : ISample
+    public class PointFeatureAnimationSamples : ISample
     {
-        public string Name => "Animated point movement";
+        public string Name => "Point Feature Animation";
 
-        public string Category => "Special";
+        public string Category => "Animations";
```
inforithmics, line 22: Good to have now a Animations Category because this is now an important Feature of Mapsui.

Samples/Mapsui.Samples.Common/Maps/Animations/SymbolAnimationSample.cs
```diff
@@ -8,38 +8,39 @@
...
 
-            return new Layer("Points")
+            var layer = new Layer("Points")
             {
                 DataSource = new MemoryProvider<IFeature>(CreatePoints(style)),
             };
+            layer.SymbolStyle = style;
+
+            var animations = CreateAnimationsForSymbolStyle(style);
+            layer.Animations.Add(() => Animation.UpdateAnimations(style, animations));
```
inforithmics, line 40: So when the layer is removed the Animation is removed. So the handling of them is much easier now.

Samples/Mapsui.Samples.Common/Maps/Animations/ViewportCenterAndZoomAnimationSample.cs
```diff
@@ -0,0 +1,47 @@
...
+using Mapsui.Layers.Tiling;
+using Mapsui.UI;
+using Mapsui.Utilities;
+using Mapsui.Widgets;
+using Mapsui.Widgets.ScaleBar;
+using Mapsui.Widgets.Zoom;
+
+namespace Mapsui.Samples.Common.Maps
+{
+    public class ViewportCenterAndZoomAnimationSample : ISample
```
inforithmics, line 12: Good to have a sample for every feature.

Samples/Mapsui.Samples.Wpf/Window1.xaml.cs
```diff
@@ -77,7 +75,7 @@ private void FillComboBoxWithCategories()
                 CategoryComboBox.Items.Add(category);
             }
 
-            CategoryComboBox.SelectedIndex = 1;
+            CategoryComboBox.SelectedIndex = 0;
```
inforithmics, line 78: Was it before 1 based and is it now 0 based? Or was it in the first place the wrong Selected index?.

Tests/Mapsui.Tests/ViewportTests.cs
```diff
@@ -0,0 +1,22 @@
+﻿using NUnit.Framework;
+
+namespace Mapsui.Tests
+{
+    [TestFixture]
+    public class ViewportTests
+    {
+        [Test]
+        public void SetCenterTest()
```
inforithmics, line 9: It's good to have more Basic Tests.

Mapsui.Core/Animations/Animation.cs
```diff
@@ -11,14 +12,13 @@ public static class Animation
         /// <summary>
         /// List of all active animations
         /// </summary>
-        private static readonly List<AnimationEntry> _entries = new();
```
inforithmics, line 14: I like the removal of this static instance makes the Memory Managment easier.

inforithmics (approved):
Generally I like the improvements, it groups and seperates the Animation related Code better. Also it improves the extendibility of mapsui. And this is a Strong Feature of Mapsui that it is easy to extend.

Mapsui.Core/INavigator.cs
```diff
@@ -117,11 +117,6 @@ public interface INavigator : IDisposable
         /// <param name="maxDuration">Max duration of fling deceleration, changes based on total velocity></param>
         void FlingWith(double velocityX, double velocityY, long maxDuration);
 
-        /// <summary>
-        /// Stop any current animation if it is running
-        /// </summary>
-        void StopRunningAnimation();
-
-        void FlyTo(MPoint center, double maxResolution, long duration = 2000);
+        void FlyTo(MPoint center, double maxResolution, long duration = 500);
```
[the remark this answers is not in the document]
pauldendulk: I tried to make FlyTo more similar to other FlyTo animation I had seen and appreciated. Those were indeed faster. It is not a very meaningful change as part of this PR. I will revert it.

Mapsui.Core/Layers/RasterizingLayer.cs
```diff
@@ -205,7 +205,8 @@ public void ClearCache()
             return new Viewport
             {
                 Resolution = renderResolution,
-                Center = extent.Centroid,
+                CenterX = extent.Centroid.X,
```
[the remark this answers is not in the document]
pauldendulk: The Viewport state is now this set of fields:
_centerX
_centerY
_resolution
_rotation
_width
_height
Six doubles, a bit simpler this way.

Mapsui.Core/UI/IViewportLimiter.cs
```diff
@@ -16,11 +16,11 @@ public interface IViewportLimiter
         /// </summary>
         MinMax? ZoomLimits { get; set; }
 
-        void Limit(IViewport viewport, IReadOnlyList<double> mapResolutions, MRect? mapEnvelope);
+        void Limit(Viewport viewport, IReadOnlyList<double> mapResolutions, MRect? mapEnvelope);
```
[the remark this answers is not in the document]
pauldendulk: There was a need to Viewport instead of IViewport as a consequence of the changes I made, but unfortunately I don't remember what it was at this point. IViewport will not be removed, but there is a need for more changes to the limiter. Since animations are now managed in Viewport the limiter also need to be at that level. The limiter can not be set from the outside like this anymore.

Mapsui.Core/UI/ViewportLimiterKeepWithin.cs
pauldendulk: Well there is now a distinction between Set methods which always reset the running animation and properties that set by the animations themselves. They may be a better solutions for this, but I guess you understand the problem, we need to distinguish between updates from from the animations themselves and updates to the animations.

Mapsui.Core/Utilities/TransformationAlgorithms.cs
pauldendulk: This is a pure function. I like to have move code in pure functions. Pure functions are often very easy to test. You put something in and expect something to come out.

Mapsui.Core/Widgets/Zoom/ZoomInOutWidget.cs
pauldendulk: Before there was no animation here. Not sure why but there were still some glitches in animations, perhaps that is why the immediate solutions was chosen. Users should be able to configure this though so there is room for improvement.

Mapsui.UI.Avalonia/MapControl.cs
pauldendulk: It did work, as in that the tab got detected, also for mouse, but it did not take take into account dragging. So if you drag the map for panning this event was triggered on the mouse up. So what happened in the animations samples is that after you panned the map the animation was triggered to animate to the mouse up location. Now this function is replace in the MouseUp event were an IsClick is detected, I think we just need that in Mapsui and we can not depend on the framework for click detection. Avalonia is now a bit more similar to WPF.

Mapsui.UI.Shared/MapControl.cs
pauldendulk: Yes, that was an important goal of these changes. I still think there is a possibility to remove this render loop timer as well, and fix the problems this solves in the onDraw methods themselves. With that solution the synchronization will be even better.

Mapsui.UI.Wpf/MapControl.cs
pauldendulk: Any call to a Set method on the Viewport will reset the animation, so if the Viewport is correctly implemented it will become impossible to let two animations run at the same time.

Mapsui/Providers/Wms/WmsProvider.cs
pauldendulk: Introduced when FetchInfo was introduced. The fact that this is needed indicates there is room for improvement wrt my FetchInfo implementation.

Samples/Mapsui.Samples.Wpf/Window1.xaml.cs
pauldendulk: This is only to make testing easier for me. Should be reverted before the merge. ... or actually I think it should be 2, because the animations category is now inserted before.

Tests/Mapsui.Tests/ViewportTests.cs
pauldendulk: Yes, there should be much more of these. I have more tests in any other project that I work on.

Mapsui.Core/Animations/AnimationEntry.cs
```diff
@@ -2,18 +2,18 @@
 
 namespace Mapsui.Utilities
 {
-    public class AnimationEntry
+    public class AnimationEntry<T>
```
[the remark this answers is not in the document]
pauldendulk: This needs to be improved. T is the target of the animations. Perhaps the target of the animations needs to be removed altogether. It is a pointer which makes it harder to track what is happening when animations are updated. Of course the target needs to be tracked at some level, but perhaps AnimationEntry should only return the current could value. This involves some more changes on other levels, and I don't have the full picture clear yet.

Mapsui.Core/Animations/AnimationEntry.cs
```diff
@@ -61,47 +61,52 @@ public class AnimationEntry
         public bool Repeat { get; }
 
         /// <summary>
-        /// Time, where this AnimationEntry has started
+        /// Time, where this AnimationEntry<Viewport> has started
```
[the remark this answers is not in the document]
pauldendulk: This is mistake in my comments. Thanks for spotting it. I just removed it.

Mapsui.Core/IReadOnlyViewport.cs
```diff
@@ -93,24 +104,16 @@ public interface IReadOnlyViewport
         /// </summary>
         /// <param name="worldX">X coordinate in map units</param>
         /// <param name="worldY">Y coordinate in map units</param>
-        /// <returns>MPoint in screen pixels</returns>
-        MPoint WorldToScreenUnrotated(double worldX, double worldY); // todo: Get rid of this method
+        /// <returns>Tuple of x and y in screen coordintes</returns>
+        (double screenX, double screenY) WorldToScreenXY(double worldX, double worldY);
```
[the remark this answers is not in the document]
pauldendulk: Yes, I like to use more simple doubles and tuples allow this. This way you are not tied to any kind of Point type. In the projections I also moved that way and I like it there.

Mapsui.Core/IReadOnlyViewport.cs
```diff
@@ -93,24 +104,16 @@ public interface IReadOnlyViewport
...
-        /// </summary>
-        /// <remarks>
-        /// If viewport is rotated, this corner points are not horizontally or vertically
-        /// aligned.
-        /// </remarks>
-        MQuad? WindowExtent { get; }
+        /// <param name="worldX">X coordinate in map units</param>
+        /// <param name="worldY">Y coordinate in map units</param>
+        /// <returns>The x and y in screen pixels</returns>
+        (double screenX, double screenY) WorldToScreenUnrotated(double worldX, double worldY); // todo: Get rid of this method
```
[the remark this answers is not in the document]
pauldendulk: When called within a method they would be assigned to local variables so there the target should be lowercase, but then again you could also assign it to a property. So I really don't know what is should be, and am ready to accept any guideline on this.

Mapsui.Core/Layers/ILayer.cs
```diff
@@ -27,7 +28,7 @@ namespace Mapsui.Layers
     /// <summary>
     /// Interface for map layers
     /// </summary>
-    public interface ILayer : INotifyPropertyChanged, IDisposable
+    public interface ILayer : IAnimatable, INotifyPropertyChanged, IDisposable
```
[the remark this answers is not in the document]
pauldendulk: I have not thought a lot about this yet, but yes there may be more changes along these lines. Currently the InvalidateTimerCallback method updates when there are animation changes and when _refresh is true. In the first case the layer tells an update is needed, in the second case RefreshGraphics was called, an entirely different route, but in some cases for a similar reason. 

I was just struggling with updating dynamic data in the Animating Point Sample. There is still a timer in the sample and I want it removed. I have not found a nice way yet. In that sample there are two levels of mutation, 1 is the animation itself, the other is the event on which the source changes. The purpose here was to update GPS coordinates with an animation. I applied this feature tracking a fleet of trucks moving around the country.

Mapsui.Core/IReadOnlyViewport.cs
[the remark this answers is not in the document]
charlenni: Should it be double? SkiaSharp is using in the most parts floats.

Mapsui.Core/Animations/AnimationEntry.cs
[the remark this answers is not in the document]
charlenni: I do not understand, why AnimationEntry needs knowledge about what it is animating. It provides some information, that are used for the animation and two function pointers for the functions, that are executed while the animation is running and when the animation is at the end. There is no special animation for Viewport nor any other thing. If you want, you could animate anything you want

Mapsui.Core/UI/ViewportLimiterKeepWithin.cs
charlenni: I would make it the other way. Normal users are trained to used properties to set values. They should stop the animations. Changing values without stopping should work with Set functions.

Thinking about this: both are wrong. As I said before, Navigator should change values of Viewport, not the user direct. And so users coming in trouble. Why not set values of Viewport direct? And what does Navigator do?

Mapsui.UI.Shared/MapControl.cs
charlenni: Could you explain this a bit better? I don't understand, how you would transfer the decision, if a redraw is needed or not, into the normal draw methods.

Mapsui.Core/Widgets/Zoom/ZoomInOutWidget.cs
charlenni: It is a very old widget and I would drop it.

Mapsui.Core/IReadOnlyViewport.cs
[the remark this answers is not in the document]
pauldendulk: So far we use doubles almost everywhere, a choice made a long time ago. For performance it might be better to use floats. If we make a change there we should also look into the use of doubles in many other places (In the renderer, our own geometries MPoint, MRect). We should only make such a change if it is a significant performance improvement and to measure that we should have a performance test of the renderer (without UI) using BenchmarkDotNet.

Mapsui.Core/Animations/AnimationEntry.cs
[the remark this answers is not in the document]
pauldendulk: Adding the T is not an improvement. It was a step in a refactoring I did not finish. That is the downside of stopping this work right now 🙁

The context: The purpose was to improve the current implementation in master. There the Tick and Final methods work because they have a pointer to the target of the animation. So at the deepest level there is a method that could change anything anywhere in the application. That makes things hard to track. Also there is a lot of code replication in the Tick and Final methods. All of them do the same multiplication. They only need to be different because they have a different target. It is a better design to have the target close to the top of the hierarchy and to keep te core deterministic (pure functions). This is the idea of the 'functional core imperative shell' architecture, and in my opinion, this is just good design in general, regardless of the architecture you claim to use.

Issue #1400, opened 2021-12-02
Design proposal for animations


pauldendulk:

charlenni:
Ok, you want to go from a common approach to a specialized approach. This is much more work to do, and it is a little tricky to get. 

For this approach, the render loop has to know of all types, that could have an animation. All these have to be checked, if a redraw is needed, because updates of animations have to be synchronized with redraws. In this case, I would suggest, that only Layers and Navigator are checked for animations. Think about a vector map with, say, 10,000 features. All of them must be checked each 16 ms, if they have an animation. And this is only for one tile. And if you have to check the styles of them too, it requires even more effort.

Regarding the "Update animations when they are fetched for rendering": If you want to do this, you have to redraw the screen each 16 ms (you do it in your PR #1401 with the timer), because you don't know if an animation is running, or you have to check, if one of the fetched object has changed and then start a redraw. So go through the fetched features twice. Both ways are not ideal.

Regarding the "Different kinds of animations": You missed the widgets. And some more things, I couldn't know about at the moment. So we have to check in the render loop all things, that could need an update.

I would suggest an interface

    public interface IAnimatable
    {
      // Update animations and return true, if a redraw is needed
      bool UpdateAnimations(long ticks)
    }

Then we go through all Layers in Map, check, if they implement `IAnimatable` and call `UpdateAnimations(ticks)`. It returns a boolean, if a redraw is needed.

`Animation` then provides static functions for starting and stopping `AnimationEntry`s, which could be used by the implementation of `IAnimatable`. Also it provides a function `UpdateAnimation` which handles all, so that the developer hasn't to do this.

AnimationEntry.Final could work the way as you describe. There is kept an End object, that is normally used in the Final method to set the final state. Final is only called in two places: one is when the animation ends normal and one is when the animation is stopped (when `callFinal` is true).

pauldendulk:
The MR that you point to has the animations in the layers, so no problems there with iterating all the features and styles.

> Regarding the "Update animations when they are fetched for rendering": If you want to do this, you have to redraw the screen each 16 ms (you do it in your PR #1401 with the timer), because you don't know if an animation is running, or you have to check, if one of the fetched object has changed and then start a redraw. So go through the fetched features twice. Both ways are not ideal.

[...]

PR #1404, opened 2021-12-02
[WIP] Animatable

charlenni:
Made a new version of possible animation with implementing IAnimatable interface. It is possible to animate the Navigator, Layers, Features and Style.

@tmijieux Perhaps you could look into it, if the leaks are gone.


inforithmics:
Overall Iike the Design. But would it be possible to clean ip the pull request (without the checkins from others) maybe recreate the pull request and Cherry picknthe relevant checkins. Because I have to Reviere Checkin for checkin

pauldendulk:
@charlenni I will continue with my own  branch related to animation. This branch has something messed up with the diff which makes it hard to interpret. 

I looks like I will be using some UpdateAnimation method. But in my current plans it will not be a root interface that will be inherited by anything that could animation (so no check for an IAnimation interface). The alternative I had in mind is to have no UpdateAnimation at all, and just calculate at GetFeatures. That still may be a good option. Disadvantage: You still will need to check if an animation is running. And it makes it makes it harder to test. My plan now is to pass in a timerTick into the UpdateAnimation method. And also in the animation Start. Timer in are always problematic to test and in this way we can keep it out of the core functionality.

charlenni:
Sorry, but it's not about implementation details, it's about basic things. I don't see why Viewport is better for animations than Navigator. For me, it just doesn't make sense. Viewport indicates what I see, Navigator moves the section I see. With PR, the viewport itself moves. And for the strict separation between the tasks of Viewport and Navigator, I still lack the coherent justification. Saving the function to stop the animations in Navigator is a bit too little for me. Especially since this is also called in Viewport.

Maybe you can explain the advantage a bit more.

pauldendulk:
> Sorry, but it's not about implementation details, it's about basic things. I don't see why Viewport is better for animations than Navigator. For me, it just doesn't make sense. Viewport indicates what I see, Navigator moves the section I see. With PR, the viewport itself moves. And for the strict separation between the tasks of Viewport and Navigator, I still lack the coherent justification. Saving the function to stop the animations in Navigator is a bit too little for me. Especially since this is also called in Viewport.
> 
> Maybe you can explain the advantage a bit more.

See my comment here:
https://github.com/Mapsui/Mapsui/issues/1400#issuecomment-1003349079

Your comment here makes it seem as if Navigator could be the single source of truth for animations instead of Viewport.
- This would only possible if Viewport would be fully encapsulated. Which is not the case at all.
- Navigator in its previous state was far from the single source of truth for animation. This was not the case it all. Animation were managed from all kinds of places in the code.

The coming weeks I will spend the limited time I have primary on coding.

pauldendulk:
One more thing. We need to have online video sessions.

PR #1406, opened 2021-12-04
Disposable analyzer

inforithmics:
Added an disposable analyzer and fixing the disposing warnings
This is to fix Memory Leaks and Native Resource Leaks


Mapsui.Layers.Tiling/Layers/TileLayer.cs
```diff
@@ -34,7 +34,7 @@ namespace Mapsui.Layers
     /// <summary>
     /// Layer, which displays a map consisting of individual tiles
     /// </summary>
-    public class TileLayer : BaseLayer, IAsyncDataFetcher
+    public sealed class TileLayer : BaseLayer, IAsyncDataFetcher, IDisposable
```
pauldendulk, line 37: Is 'sealed' here also suggested by the Disposable analyser? If so, do you understand why?

Mapsui.Layers.Tiling/Layers/TileLayer.cs
```diff
@@ -149,7 +159,9 @@ private void TileFetchDispatcherOnDataChanged(object sender, DataChangedEventArg
         private RasterFeature? ToFeature(TileInfo tileInfo)
         {
             var tileData = _tileSource.GetTile(tileInfo);
+#pragma warning disable IDISP001
```
pauldendulk, line 162: So here we do not need to dispose because the RasterFeature is the owner. The question of ownership becomes an important one now. Perhaps there are also other ways to organize the ownership. Perhaps it is possible to avoid ownership at the feature level, if not we will basically have to dispose every single thing we create :(

Mapsui.UI.Android/MapControl.cs
```diff
@@ -323,7 +331,8 @@ private MPoint GetScreenPosition(MotionEvent motionEvent, View view)
         /// <returns></returns>
         private static MPoint GetScreenPositionInPixels(MotionEvent motionEvent, View view)
         {
-            return new PointF(motionEvent.GetX(0) - view.Left, motionEvent.GetY(0) - view.Top).ToMapsui();
+            using var pointF = new PointF(motionEvent.GetX(0) - view.Left, motionEvent.GetY(0) - view.Top);
```
pauldendulk, line 334: So even Androids PointF is disposable :(
We could create an MPoint here immediately.

Mapsui.UI.MAUI/Mapsui.UI.MAUI.csproj
```diff
@@ -1,9 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>net6.0;net6.0-android;net6.0-ios;net6.0-maccatalyst</TargetFrameworks>
-		<TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows')) and '$(MSBuildRuntimeType)' == 'Full'">$(TargetFrameworks);net6.0-windows10.0.19041</TargetFrameworks>
-		<UseMaui>true</UseMaui>
+		<TargetFrameworks>net6.0</TargetFrameworks>
```
pauldendulk, line 4: ah, good to see that this can now be simplified.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -327,7 +329,9 @@ public INavigator Navigator
                 if (_navigator != null)
                 {
                     _navigator.Navigated -= Navigated;
+#pragma warning disable IDISP007 // Don't dispose injected.
```
pauldendulk, line 332: Maybe this was incorrect. You can create a Navigator on the outside, assign it though the setter, but within here we dispose. Unclear ownership. I started programming in c++ and later Delphi. There the rule was that if a class creates a resource it should clean up as well. In some situations it was more complicated and you needed to use smart pointers, but I tried to avoid those situations and preferred the single-class-ownership principle.

Mapsui.UI.Uwp/MapControl.cs
```diff
@@ -287,5 +287,38 @@ private float GetPixelDensity()
 #endif
         }
 
+#if __ANDROID__
```
pauldendulk, line 290: pooh, this get so complicated.

pauldendulk: How did you end up with the code. This is mainly for Uno, isn't it? Is this from a sample? Or is this what you ended up with trying to satisfy the analyzer?

pauldendulk:
@inforithmics Impressive! The analyzer allows a more fundamental approach to the dispose problem. 

It is unfortunate this dispose is needed in so many places, but that is not the problem of this PR. We should look for ways in which we can limit the number of disposable classes.

btw, in general failing to call calling dispose does not immediately cause a memory leak. In most cases it will be called by the garbage collector at some point. For instance a pointer to a native byte array needs an IDispose because for the garbage collector it is not possible to see the size of the byte array, therefore it can not properly determine how urgently it should be disposed, but it will get disposed at some point.

The timer is an exception. It is not disposed at all. Something related to threading.

pauldendulk (approved):

pauldendulk:
Another thing we need to do is to add a unit test for a memory leak in the mapcontrol. Something like this: https://stackoverflow.com/a/24871092/85325

Mapsui.UI.Uwp/MapControl.cs
inforithmics: Yes this is from UNO because the Dispose is different depending on the platform, I wrote this myself. I hope that with UNO 4.0 this gets better.

Mapsui.Layers.Tiling/Layers/TileLayer.cs
inforithmics: Yes you are right the Raster Images could be cached so disposing them is not really ideal

PR #1407, opened 2021-12-05
Use Xamarin Essentials, Maui and Forms have now the same code to open ..

inforithmics:
Fix Obsolete warnings in Mapsui.Forms around opening an url. (Is now the same code as Mapsui.Maui) Using Xamarin Essentials.

in Mauis case its Maui.Essentials but they are the "same".


pauldendulk:
Great. This was need to fix te warning but we will probably will be using essentials in other ways as well. For instance to request the current position in a cross platform way.

pauldendulk (approved):

PR #1408, opened 2021-12-05
Make extent in NavigateTo not nullable

inforithmics:
Follow up of this comment

https://github.com/Mapsui/Mapsui/pull/1383#discussion_r748798092


charlenni:
My knowledge in this area isn't very good, so I could only look astonished on all this work going on.

But isn't it better to check in the Navigator, if the Extent is null and then do nothing? So you have to change it only at one place.

inforithmics:
Yes that's is how it is currently implemented. In the NavigatedTo Implementation it returns when the extent is null.  So I can live with both ways extent allow to be nullable or always to have to specify the extent.

inforithmics:
Here in the Nullable Design Document it is stated that NavigateTo should not be nullable.
https://github.com/Mapsui/Mapsui/issues/1360#issuecomment-962825664

pauldendulk:
@charlenni The idea is that it is better to avoid problems than to fix them. This is very general principle. For instance.
- With regard to errors it is better to detect a problem early and immediately throw an exception than it is to try to keep running in a corrupt state (fail fast). This is not always possible, especially with regard to things that can fail like network errors, but where you can you should.
- With regard to IDiposable you would rather create a class that does not need to be disposed instead of calling dispose on all disposable classes.

Nullability is by some language designers seen as a big mistake. Often we can not avoid it but if we can we should. Checking for null can avoid a crash but can also lead to surprising behavior. If nothing happens while it should this can be a problem that is hard to diagnose. Better to make it impossible and force the caller make sure it is not null. There is now still a null check in the caller of Navigator, but we will push nullability back step by step.

pauldendulk:
@inforithmics There is a conflict. I could have look at this tomorrow.

Mapsui.Core/Navigator.cs
```diff
@@ -62,7 +62,8 @@ public void NavigateTo(MRect? extent, ScaleMethod scaleMethod = ScaleMethod.Fit,
         /// <param name="easing">The type of easing function used to transform from begin tot end state</param>
         public void NavigateToFullEnvelope(ScaleMethod scaleMethod = ScaleMethod.Fill, long duration = -1, Easing? easing = default)
         {
-            NavigateTo(_map.Extent, scaleMethod, duration, easing);
+            if (_map.Extent != null)
```
pauldendulk, line 65: Now we should think of a way to push this further back. A problem here is that the navigator has a pointer to the Map. This is in general something to avoid. I know it annoyed my when I had to add it. I need to think about it.

Samples/Mapsui.Samples.Common/Maps/Projection/MultiPolygonProjectionSample.cs
```diff
@@ -41,7 +41,8 @@ public static Map CreateMap()
             };
             map.Layers.Add(OpenStreetMap.CreateTileLayer());
             map.Layers.Add(geometryLayer);
-            map.Home = n => n.NavigateTo(extent);
+            if (extent != null)
+                map.Home = n => n.NavigateTo(extent);
```
pauldendulk, line 45: Perhaps we should throw an exception right after extent is created if it is null.

pauldendulk: I don't know how the sample app behaves in that case. I think complete crash. That is not what we want.

Tests/Mapsui.Tests.Common/Maps/BitmapSymbolInCollectionSample.cs
```diff
@@ -32,7 +32,7 @@ public static Map CreateMap()
             var map = new Map
             {
                 BackColor = Color.FromString("WhiteSmoke"),
-                Home = n => n.NavigateTo(layer.Extent?.Grow(layer.Extent.Width * 2))
+                Home = n => n.NavigateTo(layer.Extent!.Grow(layer.Extent.Width * 2))
```
pauldendulk, line 35: Perhaps the ! is also a good solution for the cases above. It should be non-zero. And if not it can crash. 

In this case it may be better to create the extent before creating the map so it will crash when selecting the sample, not when trying to navigate.

pauldendulk (approved):
👏 
I am going to look if I can do something with my own remarks.

Tests/Mapsui.Tests.Common/Maps/BitmapSymbolInCollectionSample.cs
pauldendulk: Just tested the ! but it does not crash when assigning a null! value. It just tells the compiler it is not null, and lying in my case. So we need something else. We can always throw, but I would like to have a very compact way to throw.

Mapsui.Core/Navigator.cs
charlenni: As said before, Navigator shouldn't work with Map, but only with Viewport or MapControl. Map holds only data. So the Map.Home is at the wrong place, too.

charlenni:
@pauldendulk But isn't it possible, that the developer calls `NavigateTo()` with a null? We are not the only one, that use this function. Perhaps throwing an ArgumentNullException would better here?

pauldendulk:
@charlenni Where should the ArgumentException be thrown?

charlenni:
@pauldendulk In all public `NavigateTo()`, that expect an Extent.

pauldendulk:
@charlenni Looking at that code now, and although the interface in not nullable its implementation is still nullable. It should be not nullable as well, just following the interface, and I think it would make sense to throw an ArgumentNullException if the argument is not null. It that what you had in mind?

@inforithmics What do you think?

PR #1415, opened 2021-12-10
Alter some null checks

pauldendulk:
This MR is not very important. It is mainly a kind of thinking aloud about how to go about nullability. 

I checked all the calls to void NavigateTo(MRect) in the samples. What I do now is add the ! when doing the Grow. This will result in an crash at that point. And I think that is a good thing. The Home method is called later on, and that would make it harder to find the bug.

There should be some form of try catch around creating the sample map so we can show the error to the user.

### DIFF
--- a/Samples/Mapsui.Samples.Common/Maps/Navigation/KeepCenterInMapSample.cs
+++ b/Samples/Mapsui.Samples.Common/Maps/Navigation/KeepCenterInMapSample.cs
@@ -1,7 +1,6 @@
 ﻿using Mapsui.Layers.Tiling;
 using Mapsui.Projections;
 using Mapsui.UI;
-using Mapsui.Utilities;
 
 namespace Mapsui.Samples.Common.Maps.Navigation
 {
@@ -26,10 +25,10 @@ namespace Mapsui.Samples.Common.Maps.Navigation
             // Madagaskar is used. In such a scenario it makes sense to also limit
             // the top ZoomLimit.
 
-            map.Limiter.PanLimits = GetLimitsOfMadagaskar();
+            var extent = GetLimitsOfMadagaskar();
+            map.Limiter.PanLimits = extent;
             map.Limiter.ZoomLimits = new MinMax(0.15, 2500);
-            if (map.Limiter.PanLimits != null)
-                map.Home = n => n.NavigateTo(map.Limiter.PanLimits);
+            map.Home = n => n.NavigateTo(extent);
             return map;
         }
 
--- a/Samples/Mapsui.Samples.Common/Maps/Projection/MultiPolygonProjectionSample.cs
+++ b/Samples/Mapsui.Samples.Common/Maps/Projection/MultiPolygonProjectionSample.cs
@@ -33,7 +33,7 @@ namespace Mapsui.Samples.Common.Maps.Projection
             // the Map CRS.
 
             var geometryLayer = CreateWgs84Layer();
-            var extent = geometryLayer.Extent?.Grow(10000);
+            var extent = geometryLayer.Extent!.Grow(10000);
             var map = new Map
             {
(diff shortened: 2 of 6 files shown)


Samples/Mapsui.Samples.Common/Maps/Navigation/KeepCenterInMapSample.cs
```diff
@@ -26,10 +25,10 @@ public static Map CreateMap()
             // Madagaskar is used. In such a scenario it makes sense to also limit
             // the top ZoomLimit.
 
-            map.Limiter.PanLimits = GetLimitsOfMadagaskar();
+            var extent = GetLimitsOfMadagaskar();
```
pauldendulk, line 28: This one is different from the others because here it was possible to use a non nullable MRect.

pauldendulk:
@inforithmics What do you think of this?

Samples/Mapsui.Samples.Common/Maps/Projection/MultiPolygonProjectionSample.cs
```diff
@@ -33,16 +33,15 @@ public static Map CreateMap()
             // the Map CRS.
 
             var geometryLayer = CreateWgs84Layer();
-            var extent = geometryLayer.Extent?.Grow(10000);
+            var extent = geometryLayer.Extent!.Grow(10000);
```
inforithmics, line 36: Generally I would only add the ! operator when the Nullable Analyser isn't smart enough to figure out that something is not null.

So when your intention is to dissallow null for the extent in this Function then I would add a check at the beginning of the Function  like geometryLayer.Extent ?? throw new ArgumentException() or something similar. 
The Nullable Analyser is smart enough to detect that Extent is afterwards always non null. 

But Because this is sample Code you know that here the Extent is defined and the ! Operator is okay.

Samples/Mapsui.Samples.Common/Maps/Special/RasterizingLayerSample.cs
```diff
@@ -23,9 +23,8 @@ public static Map CreateMap(float pixelDensity)
             var map = new Map();
             map.Layers.Add(OpenStreetMap.CreateTileLayer());
             map.Layers.Add(new RasterizingLayer(CreateRandomPointLayer(), pixelDensity: pixelDensity));
-            var extent = map.Layers[1].Extent?.Grow(map.Layers[1].Extent!.Width * 0.1);
-            if (extent != null)
-                map.Home = n => n.NavigateTo(extent);
+            var extent = map.Layers[1].Extent!.Grow(map.Layers[1].Extent!.Width * 0.1);
```
inforithmics, line 26: Her it is okay to use the ! Operator because I think the samples have an extent defined.

inforithmics (approved):

Samples/Mapsui.Samples.Common/Maps/Projection/MultiPolygonProjectionSample.cs
pauldendulk: '?? throw new ArgumentException(nameOf(geometryLayer.Extent))' was the alternative I considered but I found it a bit verbose, especially because these were samples and it would distract from the main goal. But in general I think you are right. For these samples it is acceptable, but in general we should prefer '?? throw new ArgumentException()'.

Samples/Mapsui.Samples.Common/Maps/Special/RasterizingLayerSample.cs
pauldendulk: > Here it is okay to use the ! Operator because I think the samples have an extent defined.

Well they are all samples where *with the current code* we know the extent is not null. The code could change though, but I don't we should be able to trust the current code in our repository.

PR #1416, opened 2021-12-10
Enable C# 10 usage

inforithmics:


pauldendulk (approved):
Does this just work for all our projects? Well let's see.

inforithmics:
Yes it should work for all projects

PR #1417, opened 2021-12-10
FixNugetPackage

inforithmics:
Added missing Assemblies into Mapsui Nuget package. Maybe Later eache Assembly could have it's own Nuget package


pauldendulk (approved):
Great, some more changes will be needed after more changes are made. Mapsui will be split into more Layer projects and Mapsui.Core will be renamed to Mapsui.

PR #1423, opened 2021-12-17
Move tests to where they belong

pauldendulk:


pauldendulk:
@inforithmics The tests should be in a class that corresponds to the class they are testing, also following the same folder structure. They may be exceptions but here it was possible. It also pleasant to not have a Xamarin.Forms dependency in Mapsui.Tests.

There was one unexpected issue. The MapView test did not create a MapView but a Map. When I changed this I got an error' : 'You must call Xamarin.Forms.Forms.Init(); prior to using this property.' I did not look any further into this but [Ignored] it, also because the test would probably fail on the leak as well.

PR #1427, opened 2021-12-18
IDisposableAnalyzer Improvements

inforithmics:


Mapsui.sln
```diff
@@ -50,6 +50,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 		Mapsui.ncrunchsolution = Mapsui.ncrunchsolution
 		Mapsui.sln.DotSettings = Mapsui.sln.DotSettings
 		Nullable.cs = Nullable.cs
+		OwnershipAttributes.cs = OwnershipAttributes.cs
```
pauldendulk, line 53: What kind of file is OwnershipAttributes.cs?

pauldendulk (commented):
Scrolled through the code. I have do some work on the IDisposableAnalyser myself to get a better understanding.

pauldendulk (approved):

Mapsui.sln
inforithmics: I removed it, because I thought that the Analyzer would honor it, But it is only an intention marking attribute for documentation but is not for analyzation so I removed it again.
Comes from here.

https://github.com/DotNetAnalyzers/IDisposableAnalyzers/pull/130

When you think it is useful to state ones intention than I could readd it

pauldendulk: ok fine. Leave it like this.

PR #1428, opened 2021-12-18
Don't Build long compiling Samples in Release (Faster AppVeyor Builds)

inforithmics:


inforithmics:
The AppVeyor build is now 10 Min faster 14 minutes instead of 24 minutes

Mapsui.sln
```diff
@@ -1623,7 +1621,7 @@ Global
 		{28848162-20B7-46F1-8364-8B41B0393CD4}.Debug|x86.ActiveCfg = Debug|x86
 		{28848162-20B7-46F1-8364-8B41B0393CD4}.Debug|x86.Build.0 = Debug|x86
 		{28848162-20B7-46F1-8364-8B41B0393CD4}.Debug|x86.Deploy.0 = Debug|x86
-		{28848162-20B7-46F1-8364-8B41B0393CD4}.Release|AnyCPU.ActiveCfg = Release|x86		
+		{28848162-20B7-46F1-8364-8B41B0393CD4}.Release|AnyCPU.ActiveCfg = Release|ARM
```
pauldendulk, line 1624: Does this make it faster?

pauldendulk (approved):
Could you summarize what you have changed? I see iPhone removed. This will build in github actions? Some other changes?

inforithmics:
1. Yes I removed the IPhone Release builds (Appveyor is Windows)
2.  I removed all UWP Builds (Uno, UWP, WINUI) because the Release Build of UWP takes ages.

inforithmics:
When all Samples should be build, I could add an Debug Build for all Samples which is faster, because the Debug build is much faster for UWP than Release.

inforithmics:
Maybe to Clarify the nuget packages are still built only some Samples in Release aren't build.

Issue #860, opened 2020-03-07
Struggling with slow rendering of LineString [Forms]


tonyfonager:
As my maps in my hiking app gets bigger and more "complicated", I am struggling with the speed of the rendering of the maps in Mapsui.

I have a map with a LineString containing 70k points, and when it tries to render that MemoryLayer its like the entire app "hangs" for many, many seconds - I cannot move the map, I cannot click - nothing.

When the engine/background task has finished - the map can then be moved, although very slowly.

So the question is - am I doing something wrong, or have I just reached a limit on a map with 70k points on a linestring (its actually several LineStrings, with a total of 70k points).

Fun fact : If I move away from there area containing my 70k LineStrings, the map movement is back to normal speeds!

I have recorded a video of it here : https://youtu.be/MQvxEb5c2XY

We have had this discussion before briefly, but now it has come to a limit where its a showstopper for me and speeds are no way acceptable anymore!

(Im on newest beta 37 of MapsUI)

charlenni:
Would it be posible to get a small repo for this? Only a map with the LineString, so we could test it?

tonyfonager:
I cannot ship you my app project, but let me see if I can create a new simple repo with the same issues.

charlenni:
Would be enough. So we could play around and look, where the problems.

tonyfonager:
I have a repo ready now - as it contains "private data" from my app, I would prefer if I could send it to you privately, instead of sharing URL here ?

You can PM me with an email address!

pauldendulk:
pauldendulk@gmail.com

tonyfonager:
Its sent via email to you now @pauldendulk

pauldendulk:
Thanks! I could run the sample and reproduce the problem. A practical thing you could do right now is to wrap it in a RasterizingLayer. See the bottom two lines below:

```csharp
// create MemoryLayer with polylines
ml = new Layer("Routes");
ml.DataSource = mp;
ml.MinVisible = GetZoomResolution(20);
ml.MaxVisible = GetZoomResolution(8); // make the route layer visible from zoom level 8 -> 20
ml.Tag = "RoutesLayer";
ml.Style = null;
ml.IsMapInfoLayer = true;
var rasterizingLayer = new RasterizingLayer(ml);
map.Layers.Add(rasterizingLayer);    
```
[...]

PR #1429, opened 2021-12-19
(WIP) Rasterizing tiling layer

inforithmics:
A Rasterizing and Tiling Layer.


Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -186,7 +186,7 @@ private void Render(object canvas, IReadOnlyViewport viewport, IEnumerable<IWidg
             // todo: We will need to select on style instead of layer
 
             layers = layers
-                .Select(l => (l is RasterizingLayer rl) ? rl.ChildLayer : l)
+                .Select(l => (l is IChildLayer rl) ? rl.ChildLayer : l)
```
charlenni, line 189: Wouldn’t it be better to use `cL` instead of `rl`? Later no one gets, for what `rl` stands.

inforithmics: Good point I'll fix this.

pauldendulk:
@inforithmics Could you summarize how this works? I would guess it rasterizes vectors data for each tile in a tile schema?

pauldendulk:
I would like to check out this branch and step through it, but I will not have time for this this weekend.

inforithmics:
1. The principle how this works is that i Create a RasterizingTileProvider who rasters every Tile that is requested.
2. This makes it possible to use the TilingLayer to calculate the Tiles.
3.  Using the TileLayer as Basis Layer makes possible to To Implement TileCaching, and only Raster Tiles that come newly into View making the Rasterizing Tile Layer faster.
4. I have made a Rasterizing Tile Layer Sample (the same sample as the RasterizingLayer but with this Rasterizing Tile Layer.

Future Improvements.
1. Instead of Rasterizing to Bitmap I'll try rasterizing to SkImage and Serialize this. (The serialized format is a vector format)

inforithmics:
Rendering to skp (Skia File format works now) so this  pull request can now be reviewed.

inforithmics:
I think The changes I wanted to to in initial Checkin. I have done here.

Further Improvments would be to implement a Geometry reducing Layer But for this I would use NTS and that would be a differnt checkin.
So the question is how to proceed with this checkin. Should I make a self review or do you want do discuss this in a video call?

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -70,30 +72,64 @@ public MapRenderer()
...
+                            if (Render(viewport, layers, background, pixelDensity, surface?.Canvas)) return null;
+                            using var image = surface!.Snapshot();
+                            using var data = image.Encode(); // Default format is Png
+                            data.SaveTo(memoryStream);
+                        }
+
+                        break;
+                    case EStreamFormat.Skp:
+                        var picture = (SKPicture?)RenderToPicture(viewport, layers, background);
+                        picture?.Serialize(memoryStream);
```
inforithmics, line 127: serializes the Draw calls to a stream

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -102,6 +138,17 @@ public MapRenderer()
             }
         }
 
+        private bool Render(IReadOnlyViewport viewport, IEnumerable<ILayer> layers, Color? background, float pixelDensity,
```
inforithmics, line 141: to avoid code duplication between rendertoPicture and RenderToBitmapStream

Mapsui.UI.Android/Mapsui.UI.Android.csproj
```diff
@@ -83,7 +83,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="SkiaSharp.Views">
-      <Version>2.80.2</Version>
+      <Version>2.80.3</Version>
```
inforithmics, line 86: In the samples the Skia version wasn't upgraded, causing a Method not found exception when using the new Serializing Extension of SkiaSharp

inforithmics:
I made this Pull Request to Improve the Rendering Performance of Large Geometry Data. Because Rendering it on tiles allow Parts to be reused and cached which is not currently possible with Rasterizing Layer.
So thousends of Geometries can be rendered blazingly fast on mobile Devices.

And it was inspired by this comment.

https://github.com/Mapsui/Mapsui/issues/860#issuecomment-596090939

pauldendulk:
> 1. Instead of Rasterizing to Bitmap I'll try rasterizing to SkImage and Serialize this. (The serialized format is a vector format)

SkImage is platform dependent, at the moment we only support Skia but with the current architecture it would still be possible to support another platform. We want to rasterize to SKImage because we want to fastest possible solution, but we could keep thing platform independent by rearranging the code. No concrete suggestions right now. Just a note.

You say "SkImage and Serialize this". What do you mean. I can understand you would like to create an SkImage right away, but what does it mean to serialize that? Or are this two different options you mention?

pauldendulk:
Will MapInfo work on the RasterizingTileLayer? I think it does work on the RasterizingLayer, forgot how that was implemented.

charlenni:
> 1. Instead of Rasterizing to Bitmap I'll try rasterizing to SkImage and Serialize this. (The serialized format is a vector format)

That's also a part I don't understand. [SKImage](https://docs.microsoft.com/en-us/dotnet/api/SkiaSharp.SKImage?view=skiasharp-2.80.2) is "an abstraction for drawing a rectangle of pixels". Event if you serialize it, it is still a rectangle of pixels. What you want to do is to record all the drawing commands for a tile in a SKPicture and redraw this later, correct? I assume, that this is a little faster than drawing it each time, because of the overhead of converting a geometry from Mapsui to SKPath, which happens at the moment on each redraw.

inforithmics:
> That's also a part I don't understand. [SKImage](https://docs.microsoft.com/en-us/dotnet/api/SkiaSharp.SKImage?view=skiasharp-2.80.2) is "an abstraction for drawing a rectangle of pixels". Event if you serialize it, it is still a rectangle of pixels.

Sorry for using the wrong name i meant SkPicture instead of SkImage.  
SkPicture as stored the draw operations that are done for drawing an image and is therefore resolution independent. I implemented this.

ETileFormat.Picture // SkPicture
ETileFormat.Skp // Serialized SkPicture
ETileFormat.Png // Png Bitmap

Mapsui.Layers.Tiling/Layers/RasterizingTileLayer.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <param name="pixelDensity"></param>
+        /// <param name="minTiles">Minimum number of tiles to cache</param>
+        /// <param name="maxTiles">Maximum number of tiles to cache</param>
+        /// <param name="dataFetchStrategy">Strategy to get list of tiles for given extent</param>
+        /// <param name="renderFetchStrategy"></param>
+        /// <param name="minExtraTiles">Number of minimum extra tiles for memory cache</param>
+        /// <param name="maxExtraTiles">Number of maximum extra tiles for memory cache</param>
+        /// <param name="persistentCache">Persistent Cache</param>
+        /// <param name="tileFormat">Stream Format</param>
+        public RasterizingTileLayer(
```
pauldendulk, line 31: The most important things are done in the RasterizingTileProvider. I wonder if it would be possible to use a regular TileLayer instead of introducing a RasterizingTileLayer. What changes does a TileLayer need in order to support this?

Mapsui.Rendering.Skia/BitmapInfo.cs
```diff
@@ -8,7 +8,8 @@ public enum BitmapType
     {
         Bitmap,
         Svg,
-        Sprite
+        Sprite,
+        Picture
```
pauldendulk, line 12: This has grown into a weird class supporting different types like this with an enum and the casting in the fetch. There should be better alternatives. What comes to mind now is a IBitmap, or a better name would now be IImage (originally it was intended for actual raster bitmaps like png), or perhaps even more general IResource. And then have different implementations for the different types. Or just one implementation that takes a generic T as argument.

Mapsui.Layers.Tiling/Layers/RasterizingTileLayer.cs
```diff
@@ -0,0 +1,112 @@
...
+            double renderResolutionMultiplier = 1,
+            IRenderer? rasterizer = null,
+            float pixelDensity = 1,
+            int minTiles = 200,
+            int maxTiles = 300,
+            IDataFetchStrategy? dataFetchStrategy = null,
+            IRenderFetchStrategy? renderFetchStrategy = null,
+            int minExtraTiles = -1,
+            int maxExtraTiles = -1,
+            IPersistentCache<byte[]>? persistentCache = null,
```
pauldendulk, line 42: Persistent cache is intended to write images to file or mbtiles. It is fancy. If we have no test for it perhaps it should could better be left out.

pauldendulk (changes requested):
This is exciting. The RasterizingTileProvider immediately converts to a platform dependent format, which allows very fast rendering. It also feels very fast if I test it in WPF. 

This code builds on a lot of existing Mapsui code that could be improved. Which is why I have a  lot of question about the implementation. There may be improvements needed to the existing code as well. The more code is built on top of the older code, the harder it becomes to change the older code. So, I think we need to look at that now. The review has questions but few concrete solutions, I need to think about it. Still I think it is good to submit this review now because your feedback on my feedback might help us find a better solution.

When running the sample I see a rendering error in between levels when I use mouse wheel. Did you see it and look into it already?

charlenni:
Sorry for answering in this way. Didn't find how to react to each entry.

[IChildLayer](https://github.com/Mapsui/Mapsui/pull/1429#discussion_r788465678): For me, ISourceLayer is the better naming, because this layer is the source for all information of the rasterizing layer.

Another reaction to TileLayer: Many important parts of TileLayer are defined as private and not as protected. So it is impossible to inherit from TileLayer and change this (like dataFetchStrategy or renderFetchStrategy). So TileLayer could not be extended.

pauldendulk:
> Another reaction to TileLayer: Many important parts of TileLayer are defined as private and not as protected. So it is impossible to inherit from TileLayer and change this (like dataFetchStrategy or renderFetchStrategy). So TileLayer could not be extended.

By making it extendable we are going the route of object oriented design. There is the big class which does all kinds of things and if you extend you get all the functionality for free, and they you can modify it's behavior in the derived class, but to do that correctly you have to understand the implementation of the base class. I have worked like this, and have felt te pain. Better to split up our TileLayer in smaller pieces that can be composed in newly created classes. 'Composition over inheritance'.

charlenni:
But with defining them as private, TileLayer is an end point, if you want to make a special tile layer. You have to get the code and rewrite all. Let's make an example: if I want to have a special TileLayer with an own function to convert tile data into a feature, I have to rewrite the complete tile layer object, because I couldn't replace the function for this in the constructor. And there isn't any public variable or function to do this. So, this function could only be set while the constructor is called. That is my problem in this case.

pauldendulk:
> But with defining them as private, TileLayer is an end point, if you want to make a special tile layer. You have to get the code and rewrite all. Let's make an example: if I want to have a special TileLayer with an own function to convert tile data into a feature, I have to rewrite the complete tile layer object, because I couldn't replace the function for this in the constructor. And there isn't any public variable or function to do this. So, this function could only be set while the constructor is called. That is my problem in this case.

Okay, but we could look also into other ways to solve those problems. 

You say "So, this function could only be set while the constructor is called". Why is it in your scenario not possible (or not easy) to set the function in the constructor?

charlenni:
Because it is a private variable in TileLayer. So I have to say, that all, which want to use a VectorTileLayer have to set this function by themselves. And if they don't do it, the app crashes.

inforithmics:
> Sorry for answering in this way. Didn't find how to react to each entry.
> 
> [IChildLayer](https://github.com/Mapsui/Mapsui/pull/1429#discussion_r788465678): For me, ISourceLayer is the better naming, because this layer is the source for all information of the rasterizing layer.
> 
> Another reaction to TileLayer: Many important parts of TileLayer are defined as private and not as protected. So it is impossible to inherit from TileLayer and change this (like dataFetchStrategy or renderFetchStrategy). So TileLayer could not be extended.

I will rename the IChildLayer to ISourceLayer with the corresponding variables

Mapsui.Layers.Tiling/Provider/RasterizingTileProvider.cs
```diff
@@ -0,0 +1,84 @@
...
+using BruTile.Cache;
+using BruTile.Predefined;
+using Mapsui.Extensions;
+using Mapsui.Rendering;
+
+namespace Mapsui.Layers;
+
+public class RasterizingTileProvider : ITileSource
+{
+    private readonly ConcurrentStack<IRenderer> _rasterizingLayers = new();
```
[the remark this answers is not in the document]
inforithmics: It should Render in parallel, Have to check why it doesn't do it. Somehow the graphic is not properly refershed when zooming with the mousewheel, pressing the buttons works. So there seems to be a bug around refersh.

pauldendulk:
I am about to merge a PR where Mapsui.Core is moved back into Mapsui. This will cause breaking changes to this PR. I see there are three files modified in Mapsui.Core, so that is something you have to fix. This is unfortunate, but I think the migration to NTS should have priority.

inforithmics:
No Problem I'll fix the merge problems, The reason I put this into WIP is that I have 2 Bugs in Rasterizing TileLayer.

1. Bug SkPicture is not scaled properly.
2. Bug it seems Scrolling by mouse does not update Tiles correctly (happens with TileLayer too, because OSM background Tiles have not the right resolution).

inforithmics:
Somehow the Merge didnt' work it seems I have to make a new Pull request.

inforithmics:
This Pull request should be merged https://github.com/Mapsui/Mapsui/pull/1452

pauldendulk:
> This Pull request should be merged #1452

Does this mean this PR could be closed?

inforithmics:
Use the new pullrequest

PR #1434, opened 2021-12-21
Bitmap registry improvements

inforithmics:
Bitmap Registry Improvments so that I can directly load an Svg, Bitmap from Resource into the BitmapRegistry


pauldendulk:
Some earlier merges caused conflicts.

inforithmics:
I'll fix this.
The Many Pin Sample has now a Performance boost with this checkin, because we only Render to Skia now I could do an optimization that I had to revert erlier becuase of the WPF Backend.

see here
https://github.com/Mapsui/Mapsui/issues/1084#issuecomment-787033314

inforithmics:
Could this be merged because I could use the changes Herr for my Rasterizing Timing Branche. And I think the changes are lie Risk because I Added neue Methodd instead of changing existing. The Main Change is that an Image can be registered with a key allowing to only Register svg Symbols once.

Mapsui.Core/Styles/BitmapRegistry.cs
```diff
@@ -10,6 +10,7 @@ public class BitmapRegistry
     {
         private static BitmapRegistry? _instance;
         private readonly IDictionary<int, object> _register = new Dictionary<int, object>();
+        private readonly IDictionary<object, int> _lookup = new Dictionary<object, int>();
```
inforithmics, line 13: This introduces a key bitmap id lookup so that images could be referenced by a key, like an resource path a file path, a file name. To make it as general as possible I used an object here.

Mapsui.Core/Styles/BitmapRegistry.cs
```diff
@@ -22,14 +23,19 @@ public class BitmapRegistry
         /// Register a new bitmap
         /// </summary>
         /// <param name="bitmapData">Bitmap data to register</param>
+        /// <param name="key">key for accessing bitmap</param>
         /// <returns>Id of registered bitmap data</returns>
-        public int Register(object bitmapData)
+        public int Register(object bitmapData, object? key = null)
```
inforithmics, line 28: optional specify a key to avoid re registering the same object over and over.

Mapsui.Core/Styles/BitmapRegistry.cs
```diff
@@ -22,14 +23,19 @@ public class BitmapRegistry
...
         {
             CheckBitmapData(bitmapData);
 
             var id = _counter;
             _counter++;
             _register[id] = bitmapData;
+            if (key != null)
```
inforithmics, line 35: register key if one is provided.

Mapsui.Core/Styles/BitmapRegistry.cs
```diff
@@ -88,5 +94,14 @@ private void CheckBitmapData(object bitmapData)
                 }
             }
         }
+
+        /// <summary> Try Get Bitmap Id </summary>
```
inforithmics, line 98: search by key for an already registered bitmap

Mapsui.Core/Utilities/EmbeddedResourceLoader.cs
```diff
@@ -20,12 +21,32 @@ public static class EmbeddedResourceLoader
         public static Stream Load(string relativePathToEmbeddedResource, Type typeInAssemblyOfEmbeddedResource)
         {
             var assembly = typeInAssemblyOfEmbeddedResource.GetTypeInfo().Assembly;
-            var fullName = GetAssemblyName(assembly) + "." + relativePathToEmbeddedResource;
+            var fullName = assembly.GetFullName(relativePathToEmbeddedResource);
```
inforithmics, line 24: GetFullname is an extension method to simplyfy the loading of Images in resources.

Mapsui.Core/Utilities/EmbeddedResourceLoader.cs
```diff
@@ -20,12 +21,32 @@ public static class EmbeddedResourceLoader
...
             var result = assembly.GetManifestResourceStream(fullName);
             if (result == null) throw new Exception(ConstructExceptionMessage(relativePathToEmbeddedResource, assembly));
             return result;
         }
 
+        public static int LoadBitmapId(this Type typeInAssemblyOfEmbeddedResource, string relativePathToEmbeddedResource)
```
inforithmics, line 30: Extension method to load Images from resources. Mostly used in the samples for now. But could be used in programs that use mapsui. This allows it to only cache once an image in the Resources speeding up the drawing of Pins for example.

Mapsui.Rendering.Skia/BitmapHelper.cs
```diff
@@ -16,6 +16,16 @@ public static class BitmapHelper
             // which has all information. So we should have a SymbolImageRegistry in which we store a
             // SymbolImage. Which holds the type, data and other parameters.
 
+            if (bitmapStream is SKImage skBitmap)
```
inforithmics, line 19: Allow directly to load a Skia Bitmap image avoid serializing and loading of it.

Mapsui.Rendering.Skia/BitmapHelper.cs
```diff
@@ -16,6 +16,16 @@ public static class BitmapHelper
...
+            {
+                return new BitmapInfo { Bitmap = skBitmap, };
+            }
+
+            if (bitmapStream is SKPicture skPicture)
```
inforithmics, line 24: SkPicture is an Skia Sharp Class that saves every Draw Call, Making redrawing much faster. And Resolution independent. This is the Class that allows it to draw a Pin or a Geometry blacingly fast.

Mapsui.Rendering.Skia/ImageStyleRenderer.cs
```diff
@@ -35,6 +35,16 @@ internal class ImageStyleRenderer
                         (float)offsetX, (float)offsetY,
                         opacity: opacity, scale: (float)symbolStyle.SymbolScale);
                     break;
+                case BitmapType.Picture:
```
inforithmics, line 38: Handle Pictures for Displaying Svg Images as SkPictures fast. And without Pixel artifacts (Resolution independent).

Mapsui.Rendering.Skia/PictureRenderer.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System;
+using Mapsui.Styles;
+using SkiaSharp;
+
+namespace Mapsui.Rendering.Skia
+{
+    internal class PictureRenderer
```
inforithmics, line 7: Introduction of a Picture Renderer (Ported from ImageRenderer to Render SkPictures like SkImages). Because SkImage is Pixel based and SkPicture is Drawing based I didn't want it to fold into one class But I'm open to changing this.

Mapsui.Rendering.Skia/SvgHelper.cs
```diff
@@ -58,5 +61,20 @@ public static class SvgHelper
         {
             return str.LoadSvg()?.Picture;
         }
+
+
+        public static int LoadSvgId(this Type typeInAssemblyOfEmbeddedResource, string relativePathToEmbeddedResource)
```
inforithmics, line 66: Directly load an Svg Image int Bitmap registry. Used mostly for pins because they are static.

Mapsui.UI.Forms/Objects/MyLocationLayer.cs
```diff
@@ -109,35 +112,23 @@ public MyLocationLayer(MapView view)
...
             if (bitmapMovingId == -1)
             {
-                var bitmapMoving = Utilities.EmbeddedResourceLoader.Load($"Images.MyLocationMoving.svg", typeof(MyLocationLayer));
-
-                if (bitmapMoving != null)
-                {
-                    // Register bitmap
-                    bitmapMovingId = BitmapRegistry.Instance.Register(bitmapMoving);
-                }
+                var bitmapMoving = typeof(MyLocationLayer).LoadBitmapId(@"Images.MyLocationMoving.svg");
```
inforithmics, line 115: simplify the code and use the new extension methods

Samples/Mapsui.Samples.Common/Maps/Symbols/SvgSample.cs
```diff
@@ -56,36 +57,16 @@ private static IEnumerable<IFeature> CreateSvgFeatures(IEnumerable<MPoint> rando
...
         }
 
         private static SymbolStyle CreateSvgStyle(string embeddedResourcePath, double scale)
         {
-            var bitmapId = GetBitmapIdForEmbeddedResource(embeddedResourcePath);
+            var bitmapId = typeof(SvgSample).LoadSvgId(embeddedResourcePath);
             return new SymbolStyle { BitmapId = bitmapId, SymbolScale = scale, SymbolOffset = new Offset(0.0, 0.5, true) };
         }
-
-        private static int GetBitmapIdForEmbeddedResource(string imagePath)
```
inforithmics, line 71: remove this because this is now a mapsui feature.

Samples/Mapsui.Samples.Common/Maps/Symbols/SymbolsSample.cs
```diff
@@ -90,37 +89,29 @@ private static IEnumerable<IStyle> CreateDiverseStyles()
...
                 new SymbolStyle {SymbolScale = 1, SymbolOffset = new Offset(diameter, diameter)},
-                CreateBitmapStyle("Mapsui.Samples.Common.Images.ic_place_black_24dp.png", 0.7),
-                CreateBitmapStyle("Mapsui.Samples.Common.Images.ic_place_black_24dp.png", 0.8),
-                CreateBitmapStyle("Mapsui.Samples.Common.Images.ic_place_black_24dp.png", 0.9),
-                CreateBitmapStyle("Mapsui.Samples.Common.Images.ic_place_black_24dp.png", 1.0),
-                CreateSvgStyle("Mapsui.Samples.Common.Images.Pin.svg", 0.7),
-                CreateSvgStyle("Mapsui.Samples.Common.Images.Pin.svg", 0.8),
-                CreateSvgStyle("Mapsui.Samples.Common.Images.Ghostscript_Tiger.svg", 0.05),
-                CreateSvgStyle("Mapsui.Samples.Common.Images.Ghostscript_Tiger.svg", 0.1),
+                CreateBitmapStyle("Images.ic_place_black_24dp.png", 0.7),
```
inforithmics, line 92: Mapsui.Samples.Common is now taken from the Assembly Namespace. Avoiding duplicate Code.

Tests/Mapsui.Tests.Common/Maps/BitmapAtlasSample.cs
```diff
@@ -45,12 +46,12 @@ private static MemoryLayer CreateLayer()
...
         public static List<IFeature> CreateFeatures()
         {
-            var atlas = LoadBitmap("Mapsui.Tests.Common.Resources.Images.osm-liberty.png");
+            var atlas = typeof(BitmapAtlasSample).LoadBitmapId("Resources.Images.osm-liberty.png");
             var spriteAmusementPark15 = new Sprite(atlas, 106, 0, 21, 21, 1);
             var spriteClothingStore15 = new Sprite(atlas, 84, 106, 21, 21, 1);
             var spriteDentist15 = new Sprite(atlas, 147, 64, 21, 21, 1);
             var spritePedestrianPolygon = new Sprite(atlas, 0, 0, 64, 64, 1);
-            var svgTigerBitmapId = LoadBitmap("Mapsui.Tests.Common.Resources.Images.Ghostscript_Tiger.svg");
+            var svgTigerBitmapId = typeof(BitmapAtlasSample).LoadSvgId("Resources.Images.Ghostscript_Tiger.svg");
```
inforithmics, line 54: new way of loading a Bitmap from an assembly, making it more refactor resistent.

Tests/Mapsui.Tests.Common/Maps/BitmapAtlasSample.cs
```diff
@@ -80,13 +81,5 @@ public static List<IFeature> CreateFeatures()
                 }
             };
         }
-
-        private static int LoadBitmap(string bitmapPath)
```
inforithmics, line 84: is now a mapsui feature

inforithmics:
I made this Checkin to Reduce the need of Handling IDisposable Types and Simplify the BitmapRegistry Handling.

inforithmics:
This Pull request although adresses this Problem improving the Svg Pin Performance.

https://github.com/Mapsui/Mapsui/issues/1084

Mapsui.Rendering.Skia/BitmapInfo.cs
```diff
@@ -33,6 +34,22 @@ public class BitmapInfo
             }
         }
 
+        public SKPicture? Picture
+        {
```
pauldendulk, line 38: Somehow it feels wrong to have one class with multiple formats like this. I known it was already like this, but we should try to think about another solution at some point.

Mapsui.Rendering.Skia/ImageStyleRenderer.cs
pauldendulk: Is that different than it was before? Could you elaborate?

Mapsui.Rendering.Skia/PictureRenderer.cs
pauldendulk: I have to look at this stepping through it in the debugger. Which sample should I use?

Samples/Mapsui.Samples.Common.Desktop/ShapefileSample.cs
```diff
@@ -82,8 +83,7 @@ private static IThemeStyle CreateCityTheme()
             // Scaling city icons based on city population.
             // Cities below 1.000.000 gets the smallest symbol.
             // Cities with more than 5.000.000 the largest symbol.
-            var localAssembly = Assembly.GetAssembly(typeof(ShapefileSample));
-            var bitmapId = BitmapRegistry.Instance.Register(localAssembly?.GetManifestResourceStream("Mapsui.Samples.Common.Desktop.Images.icon.png") ?? throw new InvalidOperationException());
+            var bitmapId = typeof(ShapefileSample).LoadBitmapId(@"Images.icon.png");
```
pauldendulk, line 86: Simplifications like this really help to make it more readable!

Samples/Mapsui.Samples.Common.Desktop/ThemeStyleSample.cs
```diff
@@ -101,11 +100,11 @@ public static ILayer CreateCityHoverPoints()
 
         private static SymbolStyle CreateCityStyle()
         {
-            var imageStream = EmbeddedResourceLoader.Load("Images.location.png", typeof(GeodanOfficesSample));
+            var location = typeof(GeodanOfficesSample).LoadBitmapId("Images.location.png");
 
             return new SymbolStyle
             {
-                BitmapId = BitmapRegistry.Instance.Register(imageStream),
+                BitmapId = location,
```
pauldendulk, line 107: It is now so simple it could also be assigned directly.

Samples/Mapsui.Samples.Common/Maps/Geometries/PointsSample.cs
```diff
@@ -86,17 +86,9 @@ private static SymbolStyle CreateBitmapStyle()
             // For this sample we get the bitmap from an embedded resouce
             // but you could get the data stream from the web or anywhere
             // else.
-            var path = "Mapsui.Samples.Common.Images.home.png"; // Designed by Freepik http://www.freepik.com
-            var bitmapId = GetBitmapIdForEmbeddedResource(path);
+            var bitmapId = typeof(PointsSample).LoadBitmapId(@"Images.home.png"); // Designed by Freepik http://www.freepik.com
             var bitmapHeight = 176; // To set the offset correct we need to know the bitmap height
             return new SymbolStyle { BitmapId = bitmapId, SymbolScale = 0.20, SymbolOffset = new Offset(0, bitmapHeight * 0.5) };
         }
-
```
pauldendulk, line 94: Deleting code is the best thing you can do.

pauldendulk:
Some more remarks.

The simplification this brings and the code that can removed is just great!

I did a PR to move the extensions to an Extensions folders. As far as I know we did this consistently so far, and I do so in other projects. A disadvantage could be that when continuing to move more to extension methods they all become available when adding 'using Mapsui.Extensions'. So far I have not experienced this as a problem yet.

It turns out that reviewer PRs by commenting on the diffs is hard for me. Usually I have a lot more informal discussion about a PR, but that is not easy to arrange in open source. So, I try to get move involved by checking out the branch and contributing to the code. Of course you could have moved the extension methods yourself but I just want to have my hands on the code.

charlenni:
BitmapRegistry was the last time I changed it already a discussion. It has to be decided, what we want to reach with this type of class. Should it be a lightweight dictionary for objects or a evolved class for handling different types of image formats that know how to draw them on the screen.

pauldendulk:
> BitmapRegistry was the last time I changed it already a discussion. It has to be decided, what we want to reach with this type of class. Should it be a lightweight dictionary for objects or a evolved class for handling different types of image formats that know how to draw them on the screen.

Indeed. But it is not a decision that I can take without diving deep into the code and investigating the alternatives. It was intended as platform independent thing but we also want the fastest possible rendering. So, in the render loop it should not be necessary to convert to a platform dependent image. This is not something I want to spend time on right now.

pauldendulk:
@inforithmics Also, the comments to your own PR were very helpful, I think this is even necessary for me to do a proper review.

charlenni:
@pauldendulk We discussed this some years before ;-) : The best would be to convert the default drawing behavior of standard geometries to replaceable style renderers like the widgets use. With this, each one could replace the rendering of e.g. a LineString or Bitmap by an own function. Then the [`RenderFeature()`](https://github.com/Mapsui/Mapsui/blob/b8fc582dd10c4db07a58094318ef6cf742ce1dc3/Mapsui.Rendering.Skia/MapRenderer.cs#L148) function would be reduced to its first part. No special features to Mapsui at all.

Mapsui.Core/Utilities/EmbeddedResourceLoader.cs
inforithmics: Thanks I've merged it

Mapsui.Core/Styles/BitmapRegistry.cs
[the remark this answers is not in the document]
inforithmics: the key is now a string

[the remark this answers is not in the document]
inforithmics: I could change the BitmapId to a string but then other parts of Mapsui would changing to because they use the BitmapId (int) as Reference to a Bitmap/Image

Mapsui.Core/Styles/BitmapRegistry.cs
```diff
@@ -88,5 +94,14 @@ private void CheckBitmapData(object bitmapData)
...
+        /// <param name="key">key</param>
+        /// <param name="bitmapId">bitmap id</param>
+        /// <returns>true if found</returns>
+        public bool TryGetBitmapId(object key, out int bitmapId)
```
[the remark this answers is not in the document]
inforithmics: There is only one key per string key so it would overrwrite the existing bitmap id for a certain key.

I have added a TryGetMapId founction because Mapsui cares only about the Id for a Bitmap and not the bitmap itself.
The bitmap itself is loaded in the rendering part of mapsui

Samples/Mapsui.Samples.Common.Desktop/ShapefileSample.cs
inforithmics: For stepping through the PictureRenderer, i would use the Symbols Sample

inforithmics:
@charlenni  I found some comments in the code about a more intelligent Bitmap registry, But I wanted to keep this change as small as possible, because my goals with this was 
1. Reduce the creation of disposable objects.
2. Have the possibility to register the same objects only once.

inforithmics:
The idea of using replacable style renderers for Geomtries sounds intresting, this could be an area of investigation after the Nts migration.

inforithmics:
@pauldendulk I can change the BitmapId from an int to a string if wanted. But I wanted to minimize overal changes.

pauldendulk:
> @pauldendulk I can change the BitmapId from an int to a string if wanted.
> This would remove the second Dictionary.
> But I wanted to minimize overall changes.

hmm, yeah, but if we keep adding small changes without rewriting the base things will gradually get worse. There are also disadvantages of removing the second dictionary, because a string will be slower than the int. So let's keep it like it is now.

pauldendulk (approved):

PR #1435, opened 2021-12-23
Added Microsoft.VisualStudio.Threading.Analyzers

inforithmics:
Fixed warnings The major thing I changed was that

in IProvider 

GetFeatures was changed from
` IEnumerable<T> GetFeatures(FetchInfo fetchInfo);`
to
` IAsyncEnumerable<T> GetFeatures(FetchInfo fetchInfo);`

So that The Features can now really be loaded Asynchronous.


pauldendulk:
I will need much more time to review this. Perhaps this can be done in a video chat session. But it will not be soon before this is wrapped up.

In general all the analyzer based changes would need more more though. We now now have analyzers for:
- Nullability
- Disposability 
- and now for Threading

All three raise all kinds of questions that have architectural changes as a consequences and not just fixes to the code as it is. We need deeper discussions about all these. In some cases changes need to be carried through that I already had in mind years ago. For instance the data fetching could be completely rewritten.

The main thing I was working on was getting rid of Mapsui.Geometries. Several big steps have been taken to prepare for that. One other preparation is to reorganize the projects. Then I can get to actually replacing Mapsui.Geometries with NTS. This work has been interrupted because I felt the need to deal with the animation timers and to prevent making everything IDisposable. The main focus there should be on the unit test. Are the objects disposed? Why are they not disposed without calling Dispose. This is not just because of the IDisposable interfaces in SkiaSharp but because of things like timers and and pointers and eventhandlers. But this is a full discussion in itself.

I wil now wrap up the work I have done on animations and merge that. In Januari I will continue with the NTS migration.

inforithmics:
Actually the NTS rewrite ist more important than this. Actually it was an explorative pull request. To geht to know the Mapsui Architekture. So I'm going to close this pull request and Make eventually two pull request. First fixing small threading Bugs without Architectual changes Second Architectual changes Like asyncenumberable.

PR #1436, opened 2021-12-25
…Changed TileLayer to handle IFeature instead of RasterFeature

charlenni:
Changed TileLayer to handle IFeature instead of RasterFeature, so that it could be used for all other types of tiled data, not only raster data.

Only problem could occur in RenderFetchStrategy.cs, line 51, where is checked, if the returned Feature has a Raster that isn't null. Perhaps IFeature could have a HasData flag for this.

### DIFF
--- a/Mapsui.Layers.Tiling/Fetcher/TileFetchDispatcher.cs
+++ b/Mapsui.Layers.Tiling/Fetcher/TileFetchDispatcher.cs
@@ -17,18 +17,18 @@ namespace Mapsui.Fetcher
         private readonly object _lockRoot = new();
         private bool _busy;
         private bool _viewportIsModified;
-        private readonly ITileCache<RasterFeature?> _tileCache;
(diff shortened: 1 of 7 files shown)


inforithmics (commented):
This change is okay because feature.Raster was not allowed to be null in the first place (nullable annotation said it should be non null.

inforithmics:
I fixed the Unit Tests not compiling to see if they will work

charlenni:
@inforithmics Thank you for the help!

inforithmics:
I'd like to merge this and use it in my Rasterizing Tiling Layer implementation.

inforithmics (approved):

charlenni:
Thank you for merging this. I used it, to create a [VectorTileLayer](https://github.com/charlenni/VectorTileLayer). It is still a work in progress and far from perfect, but it is a start.

PR #1437, opened 2021-12-26
Upgraded from SkiaSharp 2.80.2 to 2.80.3

inforithmics:
I upgraded the stable SkiaSharp usages from Skia Sharp 2.80.2 to Skia Sharp 2.80.3
The reason is to use

SkPicture Deserialize and Serialize methods. I plan to use it in my Rasterizing Tiling Pull request.
https://github.com/Mapsui/Mapsui/pull/1429
see:
https://github.com/mono/SkiaSharp/issues/880

The Preview skia Sharp Usage stays at 2.88.0-preview.179
Uno, WinUI, Maui use this Version (Uno Samples had the wrong references and I fixed it here).


pauldendulk:
> Uno, WinUI, Maui use this Version (Uno Samples had the wrong references and I fixed it here).

How is that determined? By looking at what Uno itself is using?

> I upgraded the stable SkiaSharp usages from Skia Sharp 2.80.2 to Skia Sharp 2.80.3

The [Svg.Skia](https://www.nuget.org/packages/Svg.Skia/0.5.7.2) version we are using still depends on 2.80.2. This violates our guidlines https://mapsui.com/documentation/contributors-guidelines.html#keep-our-direct-and-indirect-dependencies-in-sync. The newest non-preview release of Svg.Skia uses another version but it is an 2.88 preview. It is is a complicated situation with no proper solution. Since we are in between releases this is okay. I think our next release should aim to release with 2.88.0. And if it is not released by then we go for a 2.88 preview, like Svg.Skia.

pauldendulk (approved):

inforithmics:
Yes the  beweist Uno Release use a Preview Version of skia sharp. Because there are some important bugfixes.

PR #1438, opened 2021-12-30
Updated Avalonia to skia 2.88.0-preview.179

inforithmics:
Updated Avalonia too to the Skia Sharp Preview Release


inforithmics:
The newest Avalonia Release uses to the Preview Version of skia Sharp. And I think because the next mapsui Version will be based on Skia.Sharp 2.88 This is a step torwards this.

pauldendulk (approved):

PR #1439, opened 2022-01-07
Threading Analyzer and Fixes without Interface changes

inforithmics:
Threading Analyzer and Fixes without Interface changes


inforithmics:
This pull request should be reviewed as a Unified Comparision Because of the introduction of try catch it is easier to see the differences.

image

pauldendulk:
> This pull request should be reviewed as a Unified Comparision with ignoring whitespace Because of the introduction of try catch it is easier to see the differences.
> 
> image

Thanks for the suggestion. I had not used that features yet.

Samples/Mapsui.Samples.Forms/Mapsui.Samples.Forms.Shared/MainPageLarge.xaml.cs
```diff
@@ -135,36 +136,51 @@ private void OnPinClicked(object sender, PinClickedEventArgs e)
...
+                            PauseLocationUpdatesAutomatically = true
+                        });
+
+                CrossGeolocator.Current.PositionChanged += MyLocationPositionChanged;
+                CrossGeolocator.Current.PositionError += MyLocationPositionError;
+
+            }
+            catch (Exception ex)
+            {
+                Logger.Log(LogLevel.Error, ex.Message, ex);
```
pauldendulk, line 164: Is there perhaps a more global way to catch such exceptions? Perhaps this is only possible at the app level, and not for a component like ours.

Samples/Mapsui.Samples.Wpf.Editing/MainWindow.xaml.cs
```diff
@@ -122,7 +122,7 @@ private void InitializeZoomSlider(IReadOnlyList<double> mapResolutions)
 
         private void LogMethod(LogLevel logLevel, string? message, Exception? exception)
         {
-            Dispatcher.Invoke(() => {
+            Dispatcher.BeginInvoke(() => {
```
pauldendulk, line 125: Why does the Threading Analyzer say this is better?

pauldendulk (approved):
Let's do this. 

Something I am worrying about is that we added many checks in recent PRs, these result in warning, and many of the warning of earlier tools are still in the code. The 'Files changes' tab of this PR consists for 1/3th of other warnings. Perhaps we should start using warningAsErrors to enforce all these warnings are fixed.

pauldendulk:
I just turned on warningAsErrors in my animation branch and got a warning from this analyzer. It solved it like assigning to an ignored result (why is that good) and had to explicitly add a TaskScheduler, but I totally don't know what I am doing:

```csharp
        private void OnNavigated(long duration, ChangeType changeType)
        {
            // Note. Instead of a delay it may also be possible to call Navigated immediately with the viewport state
            // that is the result of the animation.
            _ = Task.Delay((int)duration).ContinueWith(t => Navigated?.Invoke(this, changeType), TaskScheduler.FromCurrentSynchronizationContext());
        }
```

Issue #1445, opened 2022-01-19
Creating custom (derived) Geometry is impossible to clone.


Hylke-Atmos:
The way Geometry is setup in MapSui makes it impossible to create custom (derived) Geometry, because it fails in to be able to be cloned or copied in some cases.

This issue is twofold:
1. Even though the IGeometry interface (and therefor IGeometryCollection) specify a clone function it is impossible to use (in particular in the collection classes such as GeometryCollection). These classes keep an Collection of Geometry objects instead of IGeometry objects also the IGeometry.Clone() function returns a Geometry which I think doesn't help either (but more about that later). When calling the Clone function on a GeometryCollection it will always throw an Exception: "Clone() has not been implemented on derived datatype", which is partially true but mainly because all derived classes redefine Clone() (using new keyword) returning their own type (probably for easier use in some other parts of the library), but therefor will not be called from within the Geometry collection, because that is calling Geometry.Clone()!
2. There is an alternative workaround created by the use of the "Copy" extension method, however this method checks the class name against any known class type. This works fine for known types within the library, but if you decide to create your own derived class, this doesn't work at all, since it is not part of the library.

In case you are wondering why would someone want to create their own derived class. I attempted to add an "ArrowLineString" which is a derived class from MultiLineString, and is intended to be a line between 2 points but with 2 additional linestringgeometries for visualizing an Arrow to indicate the direction of the line.

From what I saw a large refactor of the whole Geometry setup would be needed to solve this problem and make the whole geometry data model more abstract.

I was wondering;
- If it was intended this way?
- What was the intention of the public Clone() function if it is not intended to be used outside the library?
- Why is Clone() being redefined in all derived classes with the result of not being able to call clone on the abstract base class Geometry?
- If I'm missing something?
- If there are others who are running into the same issue?
- If there are plans on refactoring this part?
- If it would be appreciated when I help refactoring this?

charlenni:
Sorry, that I couldn't say something to your problem, but I could say something about I would solve this.

Instead of using a new Geometry type use a custom style with a custom renderer and use the style for holding information about the arrow(s) and draw them by yourself.

Hylke-Atmos:
Thanks for the response and although it is indeed a workaround for the problem, I personally think there is a bigger issue here.

[...]

PR #1447, opened 2022-01-21
Making clone return IGeometry object, and avoid breaking inheritance calls

Hylke-Atmos:
As discussed in Issue #1445 the current Clone implementation breaks inheritance due to the redefinition of Clone in each derived class.
This prevents the GeometryCollection from being able to be cloned and prevents users from creating custom geometry.

These changes worked for me in getting it to work with a custom derived geometry.
Unfortunately I wasn't able to test it against the full project, since it fails to build on my Visual Studio setup (different versions or other dependencies that are causing issues, that I didn't bother to figure out).

### DIFF
--- a/Mapsui.Geometries/BoundingBox.cs
+++ b/Mapsui.Geometries/BoundingBox.cs
@@ -84,8 +84,8 @@ namespace Mapsui.Geometries
                     continue;
                 }
 
-                min ??= boundingBox.Min.Clone();
-                max ??= boundingBox.Max.Clone();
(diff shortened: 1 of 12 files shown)


pauldendulk:
Master now completely removed Mapsui.Geometries and replaced it with NTS (where there is a Geometry.Copy method) which makes this PR to master irrelevant. It is important that this works correctly. Please test your scenario with the to be released NTS packages. We have to look into it if there are still any problems.

PR #1449, opened 2022-01-22
Move all layers that depend on Geometies to Mapsui.GeometryLayers project

pauldendulk:


pauldendulk:
This made it possible to remove the reference to Mapsui.Geometries to the Mapsui project (and for Mapsui.Core the Mapsui.Geometries reference was removed months ago). So, it now becomes possible to add a Mapsui.Nts repository and start porting individual parts. 

I am not sure yet about the naming and the way the projects are sliced up. Another project with a reference to Mapsui.Geometries is Mapsui.Layers.ArcGIS. Mapsui.GeometryLayer contains no layers, but just providers. Perhaps the project should be called Mapsui.MapsuiGeometries and perhaps the ArcGIS stuff should be added to that same project.

I might also want to do some more cleanup on this project, nevertheless, we are not very close to starting the port itself.

Issue #1448, opened 2022-01-21
Separate rendering process in two parts


charlenni:
At the moment, we have a render process that is a one pass process. All features are converted each time when the map is drawn in their native counterparts (LineString is converted each time the map is drawn in a SKPath). This costs much time.

In other projects, this part is divided in two parts: rendering and drawing. In rendering the map is layouted, features are converted to their native counterparts, it is determined which parts are drawn. This is only needed once. After that, the feature could be drawn often to the screen.

So it would be good to have this distinguishing between these two parts: rendering and drawing.

pauldendulk:
The RasterizingTileProvider creates an SkPicture at the moment the file is fetched. In that case a two step solution would not add anything. Do you also see it like that, or do I mis anything?

Could you refer to an example project that has this distinction? 

> At the moment, we have a render process that is a one pass process. All features are converted each time when the map is drawn in their native counterparts (LineString is converted each time the map is drawn in a SKPath). This costs much time.

You say every time, but is it not just te first time it comes into view and after that the cached version in RenderedGeometry is used? Nevertheless, it should not happen at all during drawing.

pauldendulk:
I can not put much time into this discussion at the moment but think it is important so I just want to say a few things.

Googling on 'distinction drawing rendering graphics' does not produce a lot of results but I think it is good to distinguish between the two principles and to use distinguishing terminology. Perhaps painting could be used instead of drawing though.

**Rendering**
Create a platform specific resource.
```csharp
SKPath path = ToSKPath(feature, style);
```

**Drawing or Painting**
Draw/Paint the platform specific resource to the canvas.
```csharp
canvas.DrawPath(path, paint);
```

I have a slight preference for paint right now because draw is a more general term than paint. Also, we currently use draw 1271 times in our code, and paint 262 times (searching  without match case and without whole word in cs files). In skia the event is also called PaintSurface (but yes, the calls themselves are called '.DrawXxxx' so there is something to say for both). But if you (@charlenni) and @inforithmics prefer draw we just go for that. Other opinions also welcome.

### No rendering in the draw/paint loop

With that terminology in mind I can say I agree with the idea that there should be no rendering in the draw/paint loop. This could be in our design guidelines.

### How should it work?

[...]

PR #1453, opened 2022-01-24
Changing variables for TileLayer from private to protected

charlenni:
The reason is, that with these changes, it is possible to derive from TileLayer and use it not only for raster tiles, but for all other types of tile oriented features. The main things are set via arguments in the constructor and up to now, it isn't possible to replace this constructor, because all variables are private. When making them protected, this would be possible.

### DIFF
--- a/Mapsui.Layers.Tiling/Layers/TileLayer.cs
+++ b/Mapsui.Layers.Tiling/Layers/TileLayer.cs
@@ -18,7 +18,6 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
-using System.IO;
 using System.Linq;
 using BruTile;
 using BruTile.Cache;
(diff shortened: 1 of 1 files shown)


inforithmics (commented):
I would suggest not to make the fields protected. In my opinion it ist better to introduce protected properties for the fields

charlenni:
But than we have to handle the changes outside of the constructor.

inforithmics:
You can define a Property as protected string MyProperty {get;} and the this can be set in the Constructor. // So this is the readonly equivalent of private string readonly _myfield

pauldendulk:
Perhaps it is possible to decompose the now rather bulky TileLayer (with 8 constructor parameters) so that another tile layer could use it's components and hopefully not too much code duplication, without having to use inheritance.

inforithmics:
@charlenni Have you some pseudo Code or Code where I can see how this protected fields would be used?

charlenni:
@inforithmics I don't have a special part of code, but if you look into the [VectorTileLayers](https://github.com/charlenni/VectorTileLayers) project, you will see, that I doubled all the code in TileLayer. If I could change the fetchTileAsFeature in the constructor, this isn't needed.

pauldendulk:
I think we should make more fundamental changes to allow this scenario. This issue is relevant https://github.com/Mapsui/Mapsui/issues/1448

charlenni:
Thank you for coming back to this @inforithmics. I will look into this again, if I could solve the problems with your suggestion.

pauldendulk:
I will close this. I prefer an alternative solution. I intend to spend time on the [two step rendering](#1448 ) process in the coming period.

PR #1455, opened 2022-01-30
(WIP) iOS Android net 6.0

inforithmics:
Pull request for android and ios using .NET 6
Android works,
iOS I'm still testing. 
I have recreated Sample Projects


inforithmics:
Skia sharp for iOS net6.0 has a Bug which avoids compilation. Fix ist already Made for Preview 13.

inforithmics:
When the Tooling is ready i will make an .NET 6.0 iOs Pull request

PR #1456, opened 2022-01-30
Initial Android Net 6.0

inforithmics:
Because Net 6.0 iOS has some Tooling problems, I made a pull request only with the Android Net 6.0 Changes


.github/workflows/dotnet.yml
```diff
@@ -88,6 +88,10 @@ jobs:
         vs-prerelease: true
     - name: install maui
       run: dotnet workload install maui
+    - name: install maui-check
+      run: dotnet tool install -g Redth.Net.Maui.Check
```
pauldendulk, line 92: Fancy. There is so much to learn https://github.com/Redth/dotnet-maui-check. Thanks for introducing this.

Directory.Build.props
```diff
@@ -27,7 +27,7 @@
 
   <ItemGroup>
     <!--Specify Compiler for Mac Builds-->
-    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="4.0.1">
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="4.1.0-1.final">
```
pauldendulk, line 30: What a weird version number. According to semver the version is '4.1.0' and it has a pre-release tag that starts with a number, which is odd, '1.final'. You would think the 'final' indicates it is the final version, but if it really was the final version it would have no pre-release tag and was just called '4.1.0'. And they do this all the time https://www.nuget.org/packages/Microsoft.Net.Compilers.Toolset/4.1.0-1.final.

Mapsui.UI.Android/Properties/AssemblyInfo.cs
```diff
@@ -23,12 +18,6 @@
 //      Revision
 //
 
-
-
-[assembly: AssemblyVersion("3.0.0")]
-[assembly: AssemblyFileVersion("3.0.0")]
-
 // Add some common permissions, these can be removed if not needed
 [assembly: UsesPermission(Android.Manifest.Permission.Internet)]
```
pauldendulk, line 22: As far as I know the AssemblyInfo.cs files are outdated and should not be used with the new csproj format anymore. I do see permissions are set here. So perhaps these permissions need to be moved somewhere else. Suppose you edit the permissions in the US of the android project, where does VS put them?

pauldendulk (approved):

pauldendulk:
If you resolve the conflict I can merge ASAP.

Directory.Build.props
inforithmics: Yes it's actually a Prerelease Version. So it would be best to Wait for it to Release a proper 4.1.0 Version

Mapsui.UI.Android/Properties/AssemblyInfo.cs
inforithmics: I found this StackOverflow Article
https://stackoverflow.com/questions/39051759/how-can-i-tell-which-reference-is-adding-a-permission-in-xamarin-android
I tend to Keep the AssemblyInfo.cs because it is still Assembly Info. Without the Attributes sets by the new csproj File.

inforithmics:
Use again netcoreapp3.1 removed usewpf it should now work on macos

charlenni:
@inforithmics I installed VS 2022 and updated Mapsui. Now I get the following problems in the Maui part

grafik

Any idea, what I missed?

inforithmics:
1. dotnet workload install maui should be run in an Administrative Console

pauldendulk:
> 1. "dotnet workload install maui" should be run in an Administrative Console

Perhaps we could add that in een README.md file in the Mapsui.Samples.Maui project.

Tests/Mapsui.Tests/Mapsui.Tests.csproj
```diff
@@ -1,11 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFramework>net60</TargetFramework>
+		<TargetFramework>netcoreapp3.1</TargetFramework>
 		<OutputType>Library</OutputType>
 		<RestorePackages>true</RestorePackages>
-		<GenerateAssemblyInfo>false</GenerateAssemblyInfo>
-		<UseWPF>true</UseWPF>
```
pauldendulk, line 8: Really weird this this was in here. Perhaps related to the xaml renderer we once had.

PR #1466, opened 2022-02-05
Try to upload mapsui nuget packges to nuget repository

inforithmics:
Try to upload nuget packages to the nuget repository so that every pull request can be downloaded in a nuget feed.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -62,6 +62,10 @@ jobs:
       with:
         name: nupkg.mac
         path: Artifacts/*.nupkg
+    - name: Add Mapsui Nuget Repository
+      run: nuget source Add -Name "GitHub" -Source "https://nuget.pkg.github.com/mapsui/mapsui/index.json" -UserName mapsui
+    - name: Upload Mapsui Nuget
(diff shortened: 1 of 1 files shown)


inforithmics:
It seems because the checkin would upload the build NuGet Packages to [Github NuGet Package Registry](https://nuget.pkg.github.com/mapsui/mapsui/index.json)
It isn't executed

pauldendulk:
My understanding is that the objective is that every PR by everyone will result in a nuget package hosted in the Mapsui nuget repository. 

I am not convinced we should do this. Reasons:
- It is hosted in the Mapsui repository, so the Mapsui project has some responsibility in this but it has not been reviewed while everyone can use the nugets. It does not seem right.
- I don't see how this could work with respect to version numbers. Usually you would want to use the latest nuget. The version number determines which nuget is the latest. If nugets are pushed from master and from branches how should te version be determined?

Are there github repositories which do something like this?

inforithmics:
I thought about it and I agree. There are some repositories that make nightly builds of main available as a nuget Repository on myget. Bit this is not free so I wanted to test it if it works on github Repository and then decide if it is worth it. But I think my change would stop building pull requests on github actions. So I would have to introduce a 4 th task to generate the packages on approved checkins. So I'm closing this.

PR #1467, opened 2022-02-06
[3.0.x] Fix WFS sample

janusw:
This fixes the errors from #1464 on the branch `develop/3.0` and makes some minor improvements in the WFS sample. I can forward-port it also to master (will be done as follow-up).

The sample now shows opaque red WFS features covering South Tyrol (similar as it did before) on an OSM background map.

Notes:
* The sample code also includes a label layer, but this one is not visible (not sure if this was working before). Possibly the WFS service does not provide these labels any more, or we're using the wrong field.
* I still see the following error when running the sample via `Mapsui.Samples.Forms.Mac` (not sure where this comes from; and yes, I have a working Internet connection of course):
```
Mapsui.Samples.Forms.Mac.exe Error: 0 : An exception occured while parsing a polygon geometry: The request was aborted: The request was canceled.
Warning: A WebException occurred. Do you have internet?, System.Net.WebException: The request was aborted: The request was canceled.
  at System.Net.HttpWebRequest.RunWithTimeoutWorker[T] (System.Threading.Tasks.Task`1[TResult] workerTask, System.Int32 timeout, System.Action abort, System.Func`1[TResult] aborted, System.Threading.CancellationTokenSource cts) [0x000e8] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/System/System.Net/HttpWebRequest.cs:956 
  at System.Net.WebConnectionStream.Read (System.Byte[] buffer, System.Int32 offset, System.Int32 count) [0x0007e] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/System/System.Net/WebConnectionStream.cs:138 
  at System.Xml.XmlTextReaderImpl.ReadData () [0x00366] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/referencesource/System.Xml/System/Xml/Core/XmlTextReaderImpl.cs:3165 
  at System.Xml.XmlTextReaderImpl.ParseText (System.Int32& startPos, System.Int32& endPos, System.Int32& outOrChars) [0x00377] in /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/src/Xamarin.Mac/mcs/class/referencesource/System.Xml/System/Xml/Core/XmlTextReaderImpl.cs:5224 
[...]

PR #1469, opened 2022-02-08
[3.0.x] WFS sample: further fixes

janusw:
* follow-up to #1467
* make labels show up
* make WFS layer slightly transparent

### DIFF
--- a/Samples/Mapsui.Samples.Common/Maps/Data/WfsSample.cs
+++ b/Samples/Mapsui.Samples.Common/Maps/Data/WfsSample.cs
@@ -24,11 +24,11 @@ namespace Mapsui.Samples.Common.Maps.Data
             mapControl.Map = CreateMap();
         }
 
-        private const string wfsUri = "https://geoservices.buergernetz.bz.it/geoserver/ows";
+        private const string wfsUri = "https://geoservices1.civis.bz.it/geoserver/p_bz-AdministrativeUnits/ows";
         private const string crs = "EPSG:3857";  // originally: "EPSG:25832"
-        private const string layerName = "MAPS_DISTRICTS_VW";
-        private const string nsPrefix = "gvcc_maps";
-        private const string labelField = "DIST_DESCR_DE";
+        private const string layerName = "Districts";
+        private const string nsPrefix = "p_bz-AdministrativeUnits";
+        private const string labelField = "NAME_DE";
 
         public static Map CreateMap()
         {
@@ -40,7 +40,7 @@ namespace Mapsui.Samples.Common.Maps.Data
                 map.Layers.Add(CreateWfsLayer(provider));
                 map.Layers.Add(CreateLabelLayer(provider));
 
-                map.Home = n => n.NavigateTo(new Point(1250000.0, 5900000.0), map.Resolutions[10]);
+                map.Home = n => n.NavigateTo(new Point(1270000.0, 5880000.0), map.Resolutions[9]);
 
                 return map;
                 
--- a/Samples/Mapsui.Samples.Common/Maps/Data/WfsSample.cs
+++ b/Samples/Mapsui.Samples.Common/Maps/Data/WfsSample.cs
@@ -22,8 +22,6 @@ namespace Mapsui.Samples.Common.Maps.Data
         public void Setup(IMapControl mapControl)
         {
             mapControl.Map = CreateMap();
-            var bb = new BoundingBox(1100000.0, 5800000.0, 1400000.0, 6000000.0);
-            mapControl.Navigator.NavigateTo(bb);
         }
 
         private const string wfsUri = "https://geoservices.buergernetz.bz.it/geoserver/ows";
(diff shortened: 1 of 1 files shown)


Samples/Mapsui.Samples.Common/Maps/Data/WfsSample.cs
```diff
@@ -56,7 +56,7 @@ private static ILayer CreateWfsLayer(WFSProvider provider)
         {
             return new Layer(layerName)
             {
-                Style = new VectorStyle {Fill = new Brush {Color = Color.Red}},
+                Style = new VectorStyle { Fill = new Brush { Color = Color.FromArgb(192, 255, 0, 0) } },
```
inforithmics, line 59: Good improvement now you can see through

inforithmics (approved):

JBou:
Is it possible to make the labels visible without zooming in? Can you otherwise explain me why they are hidden after a specific zoom level. Is it even possible to control this?

janusw:
> Is it possible to make the labels visible without zooming in? Can you otherwise explain me why they are hidden after a specific zoom level. Is it even possible to control this?

I'm not sure, but I would assume the labels are hidden if the features become to small and/or the labels would be overlapping with each other. So it looks less like a bug and more like a feature to me ;)

I don't know if there is a way to control it, but we can check the code.

JBou:
No problem, I don't think it's a bug either ;)
Just wanted to point out that new users may not easily see the labels because they need to zoom, it may not be user-friendly. But it's only my opinion, if there was an easy fix we can fix it, otherwise we can just leave it as it is. It's only a sample, and if someone is interested, he willl find it.

janusw:
> Just wanted to point out that new users may not easily see the labels because they need to zoom, it may not be user-friendly.

Maybe an easy way to work around this problem would be to set the bounding box on startup such that the labels are visible by default?

janusw:
> > Just wanted to point out that new users may not easily see the labels because they need to zoom, it may not be user-friendly.
> 
> Maybe an easy way to work around this problem would be to set the bounding box on startup such that the labels are visible by default?

Did that in 0677086371443136a74478f24e4ca4a54ac69613, which makes the labels show up in the inital view on MacOS. Ok with you?

I'll note that setting the bounding box does not seem to work on Android and iOS, where the map starts up with a whole-world view. No idea why. However, I can see the labels on both platforms when zooming in.

Performance-wise it seems that Android is much better than iOS (and MacOS): The app freezes shortly when selecting the WFS sample, but after that the navigation of the map is quite performant (much more than on iOS). One should have a look into the performance issues later.

pauldendulk:
> I'll note that setting the bounding box does not seem to work on Android and iOS, where the map starts up with a whole-world view. No idea why. However, I can see the labels on both platforms when zooming in.

The MapControl.Map.Home method is intended for this scenario (Home may not be a good name. DefaultViewport might be better). There is a nasty problem here, when starting an app several things get initialized on distinct points in time. The size of the app is determined while the app is loaded, the screen sometimes resizes after it is shown for the first time, and the data is loaded asynchronously. At some point, when the screen size is determined and the relevant data is loaded you want to set the viewport. The idea is that the MapControl calls the Home method at that point. But even the Home method might not work perfectly.

pauldendulk:
> Performance-wise it seems that Android is much better than iOS (and MacOS): The app freezes shortly when selecting the WFS sample, but after that the navigation of the map is quite performant (much more than on iOS). One should have a look into the performance issues later.

This might be related to the FetchWorker and threads. I said something about this in the developer chat. This could be improved and I want to make an issue for it.

JBou:
As stated on [this page](https://geoportal.buergernetz.bz.it/geodatendienste.asp), they are currently moving to a new infrastructure.
I have found the following new wfs service with the same feature:

`https://geoservices1.civis.bz.it/geoserver/p_bz-AdministrativeUnits/ows?Service=WFS`
Namespace: `p_bz-AdministrativeUnits`
Layername: `Districts`
Label: `NAME_DE`

I can test it out tomorrow, I'm not at the pc atm.
Once they have moved, it should be more stable. Bit I don't know if this will be the final wfs service url. Otherwise we can also move to another service.

pauldendulk:
Is there no option in the xml parse logic to make it work with any namespace? 

btw, the implementation uses an older xml parser, XmlDocument is a newer one, perhaps it is easier to make it work with namespaces. It might be a fairly big rewrite though.

janusw:
> Is there no option in the xml parse logic to make it work with any namespace?

Well, for the namespace alone this could work in principle. But you anyway need to choose which of the many layers you want to show (which is determined by the combination of namespace and layer name).


> btw, the implementation uses an older xml parser, XmlDocument is a newer one, perhaps it is easier to make it work with namespaces. It might be a fairly big rewrite though.

Might be worth it. Right now the WFS implementation seems to be a bit shaky and unstable (throwing various errors and warnings in a non-deterministic way, and sometimes also parts of the rendering seem to be missing).

PR #1473, opened 2022-02-12
Fix WFS sample on master

### LINKED ISSUE #1464: WFS sample is not working.
De xml parser returns:
System.Exception: 'A service exception occured: Unknown namespace [p_bz-cadastre]'

As a consequence I can not test the NTS port for WFS. I do not want to dig into this right now so created this issue to keep track of it.

janusw:
This is a follow-up to #1467 and #1469. It rebases all their changes to master and thus fixes #1464 there (and makes some further improvements to the sample, see the discussion in the two PRs).


pauldendulk (approved):

PR #1476, opened 2022-02-12
License to MIT

pauldendulk:


pauldendulk:
Deleted the LGPL LICENSE.md and replaced it with an MIT LICENSE. This needs to be updated in some more places. Perhaps it should be added to all nuspecs. If we search the full repository there will probably be more places. Also the readme and docs need to be updated.

janusw:
> Also the readme and docs need to be updated.

I just updated the README, using a relative path instead of a link to master.

janusw:
> If we search the full repository there will probably be more places.

I fixed three more references to LICENSE.md that I found. Simply grepping for 'LGPL' or 'LICENSE.md' gives no more hits at this point.

janusw (approved):
All of this is ok with me.

Directory.Build.props
```diff
@@ -4,7 +4,7 @@
     <Version>3.0.0</Version>
     <NeutralLanguage>en</NeutralLanguage>
     <Copyright>Copyright © Mapsui Developers 2018</Copyright>
-    <PackageLicenseUrl>https://raw.githubusercontent.com/Mapsui/Mapsui/master/LICENSE.md</PackageLicenseUrl>
```
dotMorten, line 7: For standard licenses use the expression instead:     `<PackageLicenseExpression>mit</PackageLicenseExpression>`

Mapsui.Layers.Tiling/Layers/TileLayer.cs
```diff
@@ -1,20 +1,3 @@
-// Copyright 2008 - Paul den Dulk (Geodan)
-// 
-// This file is part of SharpMap.
-// Mapsui is free software; you can redistribute it and/or modify
-// it under the terms of the GNU Lesser General Public License as published by
-// the Free Software Foundation; either version 2 of the License, or
-// (at your option) any later version.
-//
```
dotMorten, line 8: Even with changing source, I still think there should be license headers and origin of copied source code

Directory.Build.props
janusw: See 445a52a5448472d1f7489d0f0caa0f39c155c5dc

Mapsui.Layers.Tiling/Layers/TileLayer.cs
pauldendulk: @dotMorten I updated the headers. The license header is now part .editorconfig so with code cleanup it is easy to add it to all the files. For the purpose of this review I only did it for the changed files. Do you think this solution is okay?

Mapsui.Layers.Tiling/Fetcher/DataFetchStrategy.cs
```diff
@@ -1,19 +1,8 @@
...
-// SharpMap is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU Lesser General Public License for more details.
-
-// You should have received a copy of the GNU Lesser General Public License
-// along with SharpMap; if not, write to the Free Software
-// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA 
+// Copyright (c) The Mapsui authors.
+// The Mapsui authors licened this file under the MIT license.
```
JBou, line 2: There is a spelling mistake: licened

pauldendulk: oops!

pauldendulk: Pushed a fix. Thank you very much!

PR #1483, opened 2022-02-13
Fixed WFS PropertyName using POST requests

JBou:
This fixes the following error:

`System.Exception: A service exception occured: Requested property: PPOL_ID,p_bz-Cadastre:PPOL_CCAT_ID,p_bz-Cadastre:PPOL_ISTAT is not available for p_bz-Cadastre:ParcelsPolygons.  The possible propertyName values are: [PPOL_ID, PPOL_CCAT_ID, PPOL_ISTAT]`

When using GET requests you need to join the [PropertyNames with commas](https://docs.geoserver.org/latest/en/user/services/wfs/reference.html#getfeature), but when using POST request, you need to pass every property by his own xml element.
It may even work when **not using namespaces** and joining them with commas, because as you can see above in the error response, the first property has the namespace removed, only the second and third still have it. But I think better than removing the namespaces in the request is passing every propertyName as single xml element.

Without this fix WFS with POST requests is not working when using multiple LABELS, and I need POST requests in my current project, because the query parameter is too long when using GET requests (using a filter with many entries) and I get an error from the server:
`System.Net.WebException: Der Remoteserver hat einen Fehler zurückgegeben: (414) Request-URI Too Long.`

I will make a follow up PR to master branch afterwards.

### DIFF
--- a/Mapsui/Providers/Wfs/Utilities/WFS_1_0_0_TextResources.cs
+++ b/Mapsui/Providers/Wfs/Utilities/WFS_1_0_0_TextResources.cs
@@ -103,9 +103,11 @@ namespace Mapsui.Providers.Wfs.Utilities
                     xWriter.WriteAttributeString("typeName", qualification + featureTypeInfo.Name);
                     xWriter.WriteAttributeString("srsName", ProjectionHelper.EpsgPrefix + featureTypeInfo.SRID);
                     xWriter.WriteElementString("PropertyName", qualification + featureTypeInfo.Geometry.GeometryName);
-                    if (!labelProperties.All(string.IsNullOrWhiteSpace))
-                        xWriter.WriteElementString("PropertyName", string.Join(",", 
-                            labelProperties.Where(x => !string.IsNullOrWhiteSpace(x)).Select(lbl => qualification + lbl)));
+                    foreach (var labelProperty in labelProperties.Where(labelProperty =>
+                                 !string.IsNullOrWhiteSpace(labelProperty)))
+                    {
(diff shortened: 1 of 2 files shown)


janusw (commented):
Looks good at first sight. Since you probably have a test setup (on which you discovered this bug), maybe it makes sense to mould it into a unit test?

Apparently there already is a class `Mapsui.Tests.Wfs.WfsProviderTests`, however it is still empty! Would be the perfect place for such a test (which could simply trigger a suitable WFS request and check that no exceptions occur).

pauldendulk:
Good idea to add tests! Our coverage is way to low.

Unit tests should not do web requests, but I may be possible to store the xml response of the web request as data and load it as a Stream and feed that tot the WfsProvider in a unit test. Perhaps the provider needs to be adapted to allow a stream as input. I use this testing technique in BruTile a lot for parsing WMS and Tile Layers capabilities.

The again, since the correct functioning of our sample does actually depend on live services it does make sense to test this. This would have also helped tackling we just solved. Such tests should be in a separate test project however. They are more like api test. Perhaps called Mapsui.Samples.Api.Tests or Mapsui.Samples.Service.Tests,

pauldendulk:
A test would have been nice but lets just merge, its a bugfix.

JBou:
I will try to add a test when I get some time, I'm only a hobby programmer and have not much time left the weekends. But anyway, thanks for merging, it fixes the bug in the meantime 👍 
Should I make another PR to merge this into the main branch?

pauldendulk:
> Should I make another PR to merge this into the main branch?

Yes, please. 

If you decide to add a test then just add it to master.

Issue #1143, opened 2021-06-08
MyLocationLayer: add an optional callout/label


janusw:
It would be useful to have a configurable callout-like functionality for `MyLocationLayer`, e.g. to show the coordinates of the user's position (or other information).

Technically, the problem seems to be that the `Callout` class requires a `Pin` to work with, and the location icon in `MyLocationLayer` is not a `Pin` (but a `Feature`). Possible solutions could include:
* Making the location icon in `MyLocationLayer` an actual `Pin`.
* Extend the `Callout` class to work not only with a `Pin`, but also with a `Feature`.
* Create a new class ('FeatureCallout'?) that works with a `Feature`.
* ...?

I have not checked if any of this is feasible at all (and if yes, which one is preferable).

pauldendulk:
In the long run a solution that does not depend on Xamarin.Forms will be the better solution.

janusw:
In principle one could simply use a GeometryFeature with a LabelStyle (cf. the LabelsSample).

The only problem I can see with this approach: It would be great to be able to toggle the label by clicking on the MyLocation icon, but a GeometryFeature as such is not really clickable, is it?

pauldendulk:
> but a GeometryFeature as such is not really clickable, is it?

Yes, you can use the MapInfo for this. An example of how this is done is in the 'Info' | 'Single Callout' sample. Click on a dot. Perhaps this mechanism could be improved, but I think it is much better to develop in this direction. The dependencies on the platforms always cause problems in the long run, and this 'native' support would also work in other platforms such as Avalonia, Uno and WinUI. Also, more platform dependent code results in more maintenance, and more bugs (because we do not test extensively on every platform).

janusw:
> > but a GeometryFeature as such is not really clickable, is it?
> 
> Yes, you can use the MapInfo for this. An example of how this is done is in the 'Info' | 'Single Callout' sample. Click on a dot. Perhaps this mechanism could be improved, but I think it is much better to develop in this direction.

Thanks for the comments, Paul, but I'm afraid you might slightly misunderstand the scope of what I'm proposing:
1. I don't want to display information outside of the map (this is what the sample does that you mention, and I can already achieve that). I want information to be displayed in a label on the map itself, next to the user's location.
2. I would like a solution that works with `MapView.MyLocationLayer`. This is a nice standard facility to show the user location. I want to enhance it a bit.

[...]

PR #1486, opened 2022-02-14
MyLocationLayer: add an optional label next to the symbol

janusw:
This PR implements the feature suggested in #1143: It adds an optional label to the MyLocationLayer, next to the actual location symbol. The label text can be chosen freely, so it can for example be used to display the coordinates of the user's location (or any other information). The label is hidden by default and can be shown or hidden as desired. Also clicks on the MyLocation symbol can be caught now, e.g. in order to toggle the label when clicked.

I extended Mapsui.Samples.Forms to display a label with coordinates when the MyLocation symbol is clicked, so anyone can test the feature in the sample apps.

Any kind of feedback welcome ...

### DIFF
--- a/Mapsui.UI.Forms/Objects/MyLocationLayer.cs
+++ b/Mapsui.UI.Forms/Objects/MyLocationLayer.cs
@@ -35,6 +35,7 @@ namespace Mapsui.UI.Objects
         private readonly MapView _mapView;
         private readonly GeometryFeature _feature;
         private readonly GeometryFeature _featureDir;
+        private readonly GeometryFeature _featureLabel;
 
         private static int _bitmapMovingId = -1;
         private static int _bitmapStillId = -1;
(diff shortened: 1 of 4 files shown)


janusw:
> I saw, you made a new Feature for the direction, but normally it should be enough, to add another style. So the user location would be one Feature with three styles: one for the user location itself, one for the direction and one for an info text/callout.

I'll start with this one: Very good point! (Note that I'm still on the learning curve in Mapsui's guts, and don't know the API very well yet.)

I rebased the branch to current master and added a commit that fixes this.

janusw:
> So, what is more common: a label with a white background following your position or a callout to see what is your current position?

Actually a callout style might indeed make more sense here. It looks nicer and offers more possibilites.

However, my first attempt to use a callout style somehow failed. I'll try again later.


> Both are styles, so why not expose a InfoStyle, which could be set by the user if the default isn't good enough?

Hm, I'm not sure if it's necessary to expose the complete style. We don't do that with the other two styles in the MyLocationLayer either. The user obtains full control then, but also more opportunity to break stuff  ;)

Let me think about this ...

janusw:
> > So, what is more common: a label with a white background following your position or a callout to see what is your current position?
> 
> Actually a callout style might indeed make more sense here. It looks nicer and offers more possibilites.
> 
> However, my first attempt to use a callout style somehow failed. I'll try again later.

I suspect this might be related to #1496. Any hints on this are very welcome ...

janusw:
> > So, what is more common: a label with a white background following your position or a callout to see what is your current position?
> 
> Actually a callout style might indeed make more sense here. It looks nicer and offers more possibilites.
> 
> However, my first attempt to use a callout style somehow failed. I'll try again later.

Alright, this works now. I'm switching to CalloutStyle in dad21d76d21a171c24148479fdaca5dfe6cff0d5.

> > Both are styles, so why not expose a InfoStyle, which could be set by the user if the default isn't good enough?
> 
> Hm, I'm not sure if it's necessary to expose the complete style. We don't do that with the other two styles in the MyLocationLayer either. The user obtains full control then, but also more opportunity to break stuff ;)

Actually I'd rather stay with a simple interface for now, at the expense of full flexibility (unless you have a strong preference to have a configurable style). If anyone using this needs more flexibility, we can still extend it later.

From my side this PR is ready now. I hope I have taken care of all your comments, @charlenni. Ok for master like this?

Mapsui.UI.Forms/Objects/MyLocationLayer.cs
```diff
@@ -89,6 +91,44 @@ public bool IsMoving
         /// <value>Scale of symbol</value>
         public double Scale { get; set; } = 1.0;
 
+        /// <summary>
+        /// The text that is displayed in the MyLocation callout.
+        /// (can contains line breaks).
```
charlenni, line 96: Seems to be a typo: contain instead of contains

charlenni:
For me it looks good. And yes, I think we could assume, that developers with special needs create their own special MapControl, which could handle all, what they need.

janusw:
> For me it looks good.

Cool 😃, is that an approval?

(Btw, I'm not sure where the GitHub Actions failures come from, but they occur also on master ...)

charlenni (approved):
Yes, that is an approval. My first one, if it works.

charlenni:
Yes, the failing action is because of UWP. Got the same error with my PR.

janusw:
> Yes, that is an approval. My first one, if it works.

It does 🙂, thank you. I just merged it.

Issue #1489, opened 2022-02-14
MapView: ordering of layers


janusw:
In the Mapsui.UI.Forms.MapView, the ordering of layers is the following:
1. user layers (base maps + overlays)
2. drawables
3. pins
4. callouts
5. mylocation

From my point of view, there are two issues with this:
1. Mylocation is on top of everything else (possibly hiding drawables, pins & callouts, which I find kind of unnatural)
2. The ordering cannot be changed (any more), since the original ordering is restored in `MapView.HandleLayers`.

Before v3, changing the ordering was possible, and in fact I used to do this via `Map.Layers.Move`. Since #1185 was merged, this does not work any more. The ordering is always restored in `MapView.HandleLayers`.

There are two possible solutions that would fix this for me:
1. Change the default ordering so that mylocation is below the drawables, pins & callouts.
2. Have a way to change the default ordering.

Which one would be preferable?

charlenni:
Short explanation, why this happens: MapView is a convenience layer around MapControl. It should mimic the behavior of Xamarin.Forms original map library. This said, I have to create the extra layers before all other layers and check, that they on top. If this isn‘t done, the next layer (I assume a map tile layer) would cover this layers. Mapsui couldn‘t handle this, because it doesn‘t know anything about special layer ordering.

For me, the MyLocation should on top. You look from above on the map. The highest is yourself. The pins lay on the ground. But that‘s only for me. It seems, you see it different.

pauldendulk:
I think we can improve layer ordering before the release of v4 (but not before the first beta, the beta series can be long). The Mapview now has some weird workarounds because Mapsui lacks a better way of ordering the layers. This also affects performance. Here is a proposal: https://github.com/Mapsui/Mapsui/issues/1491

janusw:
> For me, the MyLocation should on top. You look from above on the map. The highest is yourself. The pins lay on the ground. But that‘s only for me. It seems, you see it different.

Ok, so it seems we disagree on the 'correct' ordering (which is not too surprising, I guess 😄). That means we need a way to allow custom ordering ...

janusw:
> I think we can improve layer ordering before the release of v4

I hope so :)

> The Mapview now has some weird workarounds because Mapsui lacks a better way of ordering the layers. This also affects performance. Here is a proposal: #1491

[...]

PR #1494, opened 2022-02-15
MapView: allow custom layer ordering

janusw:
* if one of the 'internal' layers is moved to another position explicitly,
  we refrain from  restoring the original order
* this is sort of a quick fix for #1489, in the long run it might be cleaner to introduce layer groups


charlenni (approved):
As we discussed before, this changes should be ok.

janusw:
> As we discussed before, this changes should be ok.

Thanks. I also verified once more with the AppVeyor nupkgs that it indeed fixes the problem in my application. Thus I'm merging it.

Issue #1503, opened 2022-02-17
Callout: anchor wrong when rotating with map (again!)


janusw:
This is sort of a reincarnation of #1151, with exactly the same phenomenology ...

When using a `Pin` with `Callout`, both of which have `RotateWithMap=true`, the callout anchor is placed incorrectly when the map is rotated off the default orientation, observable e.g. in the "Add Pin Sample" (PinSample.cs).

This has been working with 3.0.2 (was fixed in #1179), but seems to be broken again now in 4.0.0-beta.1.

inforithmics:
Can you test my Pull Request. This seems to fix the Problem

pauldendulk:
I do not see this problem in the Info | Custom Callout sample.

![image](https://user-images.githubusercontent.com/963462/154631730-27bb61a2-f1ce-48d9-9af0-0bb3819895b4.png)

How could I reproduce this?

inforithmics:
You can reproduce this with add pin sample. Adding a pin an then rotating the map and deploying it to an Android Phone for example.

pauldendulk:
So it seems to be more like a mono compiler bug. Perhaps it is solved with MAUI en .NET 6. Do you know if in .NET 6 mono is replaced with Microsofts own runtime?

pauldendulk:
I was about to mention that we should have regression tests for this, but if this is only on Android it will not be easy to test that way. @janusw Do you also see the problem on Mac? If so, it would help to create regression tests.

charlenni:
@pauldendulk You don‘t see it in your Test, because there the Offset is 0/0. You could only see the effect, if you set Offset to another value.

janusw:
> @janusw Do you also see the problem on Mac? 

Yes, I also see it on Mac.

pauldendulk:
> @pauldendulk You don‘t see it in your Test, because there the Offset is 0/0. You could only see the effect, if you set Offset to another value.

Ah!, so we should add that to our samples as well.

pauldendulk:
> > @janusw Do you also see the problem on Mac?
> 
> Yes, I also see it on Mac.

With any of the existing samples or only in your app?

janusw:
> > @pauldendulk You don‘t see it in your Test, because there the Offset is 0/0. You could only see the effect, if you set Offset to another value.
> 
> Ah!, so we should add that to our samples as well.

We have that already (in "Forms -> Add Pin Sample").

janusw:
[...]

PR #1504, opened 2022-02-17
Fixed Callout Pin Rotation

inforithmics:
Fixing this https://github.com/Mapsui/Mapsui/issues/1503

### DIFF
--- a/Mapsui.Rendering.Skia/PictureRenderer.cs
+++ b/Mapsui.Rendering.Skia/PictureRenderer.cs
@@ -16,8 +16,8 @@ namespace Mapsui.Rendering.Skia
 
             var scaleX = rect.Width / picture.CullRect.Width;
             var scaleY = rect.Height / picture.CullRect.Height;
-            var matrix = SKMatrix.CreateTranslation(rect.Left, rect.Top);
-            matrix.PostConcat(SKMatrix.CreateScale(scaleX, scaleY));
+            var matrix = SKMatrix.CreateTranslation(rect.Left / scaleX, rect.Top / scaleY);
+            matrix = matrix.PostConcat(SKMatrix.CreateScale(scaleX, scaleY));
 
             canvas.DrawPicture(picture, ref matrix, skPaint);
             if (dispose)
--- a/Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
+++ b/Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
@@ -42,8 +42,10 @@ namespace Mapsui.Rendering.Skia
             if (viewport.Rotation != 0 && calloutStyle.SymbolOffsetRotatesWithMap)
             {
                 var mapRotation = viewport.Rotation / 180.0 * Math.PI;
-                symbolOffsetX = (float)(Math.Cos(mapRotation) * symbolOffsetX - Math.Sin(mapRotation) * symbolOffsetY);
-                symbolOffsetY = (float)(Math.Sin(mapRotation) * symbolOffsetX + Math.Cos(mapRotation) * symbolOffsetY);
+                var x1 = Math.Cos(mapRotation) * symbolOffsetX - Math.Sin(mapRotation) * symbolOffsetY;
+                var y1 = Math.Sin(mapRotation) * symbolOffsetX + Math.Cos(mapRotation) * symbolOffsetY;
+                symbolOffsetX = (float)x1;
+                symbolOffsetY = (float)y1;
             }
 
             // Save state of the canvas, so we could move and rotate the canvas


Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
```diff
@@ -42,8 +42,10 @@ public class CalloutStyleRenderer : SymbolStyle
             if (viewport.Rotation != 0 && calloutStyle.SymbolOffsetRotatesWithMap)
             {
                 var mapRotation = viewport.Rotation / 180.0 * Math.PI;
-                symbolOffsetX = (float)(Math.Cos(mapRotation) * symbolOffsetX - Math.Sin(mapRotation) * symbolOffsetY);
-                symbolOffsetY = (float)(Math.Sin(mapRotation) * symbolOffsetX + Math.Cos(mapRotation) * symbolOffsetY);
+                var x1 = Math.Cos(mapRotation) * symbolOffsetX - Math.Sin(mapRotation) * symbolOffsetY;
+                var y1 = Math.Sin(mapRotation) * symbolOffsetX + Math.Cos(mapRotation) * symbolOffsetY;
+                symbolOffsetX = (float)x1;
+                symbolOffsetY = (float)y1;
```
inforithmics, line 48: This change is semantically the same as the original Code, But makes a big difference on screen.  This is the original Code piece and this works. Maybe @charlenni has more insight here.  For me it seems a Numeric-Compiler-Precision Problem.

Mapsui.Rendering.Skia/PictureRenderer.cs
```diff
@@ -16,8 +16,8 @@ public static void Draw(SKCanvas canvas, SKPicture picture, SKRect rect, float l
 
             var scaleX = rect.Width / picture.CullRect.Width;
             var scaleY = rect.Height / picture.CullRect.Height;
-            var matrix = SKMatrix.CreateTranslation(rect.Left, rect.Top);
```
inforithmics, line 19: This is a Bug I discovered in my TilingRasterizing Pull request the Picture didn't scale properly.

inforithmics:
When 
https://github.com/Mapsui/Mapsui/pull/1502

goes in this should become green

Mapsui.Rendering.Skia/PictureRenderer.cs
pauldendulk: This seems a serious bug. Good to fix it right away, but for my understanding, this is not This is not a cause of the rotation original problem, right?

Mapsui.Rendering.Skia/PictureRenderer.cs
```diff
@@ -16,8 +16,8 @@ public static void Draw(SKCanvas canvas, SKPicture picture, SKRect rect, float l
...
-            matrix.PostConcat(SKMatrix.CreateScale(scaleX, scaleY));
+            var matrix = SKMatrix.CreateTranslation(rect.Left / scaleX, rect.Top / scaleY);
+            matrix = matrix.PostConcat(SKMatrix.CreateScale(scaleX, scaleY));
```
pauldendulk, line 20: This too. Basically this code had no effect before this fix. Where in the samples can I see it's effect?

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
pauldendulk: This is very strange. Everything I know says it should be the same. I introduced this bug and I would introduce it again if I saw this code later. I am not a big fan of code comments, because they get outdated easily (better to use a more readable code style) but this a situation where code comments are really needed.

pauldendulk (commented):
I would like to be able to reproduce this problem so I can understand how it is fixed.

inforithmics:
I'll add an Code Comment.

Mapsui.Rendering.Skia/PictureRenderer.cs
inforithmics: I found this Bug in my Tile Rendering Pull request when I Rendered SkPictures. I'll look if there are Pictures scaled in mapsui. If they are then I will make a separate pull request

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
janusw: I can confirm that this PR fixes the problem. Pretty strange that this change would make any difference ...

janusw (changes requested):
The CalloutStyleRenderer part is certainly ok with me (ideally with a comment).

The PictureRenderer part should probably stay out of this, or go into a separate commit at least, since it seems to be completely unrelated.

Thanks for taking care of this so quickly!

janusw:
> When #1502
> 
> goes in this should become green

It is in master, so this PR should be rebased to master ...

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
charlenni: Looking into it, it is totally clear, that this lines are totally different (had to think about it some time, but sounds better, to say it is totally clear  ;-) ).

The existing is 

    a = a - b
    b = a + b

The first line changes `a`, so that it isn't the same as before, when you go to the second line. It is like you do a swap with

    a = b
    b = a

In our case, `symbolOffsetX` is replaced by a new value, but used in the next line as if it has the old value.

I hope, now it is clear, what the problem was.

janusw: Ah, sure 🙈. Totally missed that. So it's not about numeric conversion at all, but there is indeed a good reason for the temporary variables!

janusw: So maybe it would be a good idea to clean this up a little: Represent the offset as an `MPoint` and use `MPoint.Rotate` for the rotation?

janusw: I'm trying this in #1508, which is an alternative to this PR. Any feedback?

PR #1505, opened 2022-02-18
Remove Mapsui.Maui.sln

inforithmics:


inforithmics:
I fixed the Maui/Windows Project and it works now with Visual Studio 2022 (without Preview)

inforithmics:
Pro Tip for running in Visual Studio 2022 and in Android Emulator, Hot Reload has to be disabled or else it won't start.

pauldendulk:
> I fixed the Maui/Windows Project and it works now with Visual Studio 2022 (without Preview)

Does this mean you fixed the problem you mentioned in your earlier comment? This one:

> The mapsui.ui.Maui project is now in the Mapsui.sln This can be opened and compiled in Visual Studio 2022 But for execution and Testing there is still Visual Studio 2022 Preview needed.

inforithmics:
Yes it works now with Visual Studio 2022 17.1 with disabled Hot Reload.

Mapsui.UI.Uwp/Mapsui.UI.Uwp.csproj
```diff
@@ -125,7 +125,7 @@
       <Version>6.2.12</Version>
     </PackageReference>
     <PackageReference Include="SkiaSharp.Views">
-      <Version>2.80.2</Version>
+      <Version>2.80.3</Version>
```
pauldendulk, line 128: So for views we do not use 2.88.0-preview.209. Is this for a specific reason?

pauldendulk (approved):
Wonderful that Maui can now be part of Mapsui.sln. This simplifies a lot.

Mapsui.UI.Uwp/Mapsui.UI.Uwp.csproj
inforithmics: Currently the new platforms use 2.88.0-preview.209 (Uno, Maui, Avalonia, WinUI) and the Stable Platforms (Forms, Android, UWP, iOS) use 2.80.3. If you want I can move all platforms to 2.88.0-preview.209

This change was done because I changed in the pull request that updated to 2.80.3 this platforms somehow where not updated. I think it makes no sense to have 3 skia sharp versions. One stable and one unstable for the platforms that need the new skia Sharp Preview. But it would be simpler to have only one Skia Sharp Preview Version.

pauldendulk: Yes, I think we could use 2.88.0-preview.209 everywhere. I expect we will be in beta for a while.

inforithmics: Then I will do a Pull Request for that.

PR #1507, opened 2022-02-18
Add correct ToString() for basic types

charlenni:
Add correct ToString() for basic types, so it is possible to hover over while debugging to see the values without expanding the content.


pauldendulk (approved):
Very useful!

PR #1509, opened 2022-02-18
extend the v4 upgrade guide

janusw:
* with another case I encountered


pauldendulk (approved):
Thanks again.

btw, Many changes were made for which the namespace was not updated. We should go through this before the final release. It will result in a lot of changes.

PR #1511, opened 2022-02-18
RasterizingTilingProviderV3

inforithmics:
I tried to minimize changes. 

RasterizingTilingProvider, Bugfixes and Sample


Mapsui/Navigator.cs
```diff
@@ -89,9 +91,10 @@ public void ZoomTo(double resolution, long duration = 0, Easing? easing = defaul
         public void ZoomTo(double resolution, MPoint centerOfZoom, long duration = 0, Easing? easing = default)
         {
             var (worldCenterOfZoomX, worldCenterOfZoomY) = _viewport.ScreenToWorldXY(centerOfZoom.X, centerOfZoom.Y);
-            _viewport.SetAnimations(ZoomAroundLocationAnimation.Create(_viewport, worldCenterOfZoomX, worldCenterOfZoomY, resolution,
-                _viewport.CenterX, _viewport.CenterY, _viewport.Resolution, duration));
-            OnNavigated(duration, ChangeType.Discrete);
+            var animationEntries = ZoomAroundLocationAnimation.Create(_viewport, worldCenterOfZoomX, worldCenterOfZoomY, resolution,
+                _viewport.CenterX, _viewport.CenterY, _viewport.Resolution, duration);
+            AddFinalAction(animationEntries, () => OnNavigated(ChangeType.Discrete));
```
pauldendulk, line 96: Why is AddFinalAction needed? What goes wrong in the RasterizingTileLayer in the old implementation?

inforithmics:
Because of this. It says it triggers the Navigated Event after the duration of the Animation. When the animation does take a little longer then the Navigated is invoked before the navigation is finished (This happens on mobile Platforms which are slower). By adding this call at the Finish Action of the Animation this is called at the End of the animation making the the Call really when the animation is finished and the it works.
_ = Task.Delay((int)duration).ContinueWith(_ => Navigated?.Invoke(this, changeType));

Visually you can see this that after Zooming with the Mouse some Tiles still display the Previous Tile (You can see this when zooming in Some Tiles are blocky) With the Rasterising Tile Layer you see this that the Shapes are of different Size.

pauldendulk:
> Because of this.
> 
> _ = Task.Delay((int)duration).ContinueWith(_ => Navigated?.Invoke(this, changeType));
> 
> It says it triggers the Navigated Event after the duration of the Animation. When the animation does take a little longer then the Navigated is invoked before the navigation is finished (This happens on mobile Platforms which are slower). By adding this call at the Finish Action of the Animation this is called at the End of the animation making the the Call really when the animation is finished and the it works.
> 
> Visually you can see this that after Zooming with the Mouse some Tiles still display the Previous Tile (You can see this when zooming in Some Tiles are blocky (OpenStreetMap)) With the Rasterising Tile Layer you see this that the Shapes are of different Size (MemoryLayer).

Thanks, I needed this clarification. This is a problem that needs some kind of solution. I would prefer to make the animations fully passive, but as long as there is no proper solution we will have to use this.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -185,7 +185,7 @@ private void Render(object canvas, IReadOnlyViewport viewport, IEnumerable<IWidg
             // todo: We will need to select on style instead of layer
 
             layers = layers
-                .Select(l => (l is RasterizingLayer rl) ? rl.ChildLayer : l)
+                .Select(l => (l is ISourceLayer sl) ? sl.SourceLayer : l)
```
pauldendulk, line 188: If a layer extents ISourceLayer I expect it to be a source layer but in this case it has a source layer. I would prefer IHasSourceLayer here. But this is not blocking.

Mapsui/Layers/IDataSourceLayer.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Mapsui.Providers;
+
+namespace Mapsui.Layers;
+
+public interface ILayerDataSource
```
pauldendulk, line 8: A simple interface with just one method. I think we should do that more. Perhaps also split our existing ILayer interface up further. Perhaps RefreshData should be part of this interface.

Similar to my remark above, I would prefer IHasDataSource.

Samples/Mapsui.Samples.Common.Desktop/Cache/SqlitePersistentCache.cs
```diff
@@ -0,0 +1,68 @@
...
+using BruTile.Cache;
+using SQLite;
+
+namespace Mapsui.Samples.Common.Desktop.Cache;
+
+public class SqlitePersistentCache : IPersistentCache<byte[]>
+{
+    private readonly string _file;
+
+    public SqlitePersistentCache(string name)
```
pauldendulk, line 12: SqlitePersistentCache, interesting. I think caching is important. Perhaps something like this could be part of the nuget. 

See also this:
"Cache Tile downloads locally according to HTTP Expiry Header, alternatively a minimum of 7 days."
https://github.com/Mapsui/Mapsui/discussions/1463

pauldendulk (approved):
Thanks a lot!

I would like to have this changed: Use IHasDataSource and IHasSourceLayer. Otherwise the expectation is that it **is** a DataSource/TileSource.

Issue #1115, opened 2021-04-16
Nupkg size has blown up recently


janusw:
The size of the Mapsui nupkg file built by AppVeyor used to be around 800kB, but recently is has grown strongly to about 15MB. I investigated a bit and found ...

* [Mapsui.3.0.0-build.1430.nupkg](https://ci.appveyor.com/project/pauldendulk/mapsui/builds/38191756/artifacts) to be the last small one (built from 65ef14ddad3576420ad00f4802795715cb9ebc44)
* [Mapsui.3.0.0-build.1433.nupkg](https://ci.appveyor.com/project/pauldendulk/mapsui/builds/38257237/artifacts) to be the first large one (built from b4109abc892fe3890122170ed27630d443cde12e)

However, @charlenni's commit b4109abc892fe3890122170ed27630d443cde12e looks absolutely innocent to me. I have no idea why it would change the size of the nupkg so drastically. So I suspect something with the AppVeyor build setup may have changed around that time. @pauldendulk, do you have an idea what's going on there?

janusw:
>     * [Mapsui.3.0.0-build.1430.nupkg](https://ci.appveyor.com/project/pauldendulk/mapsui/builds/38191756/artifacts) to be the last small one (built from [65ef14d](https://github.com/Mapsui/Mapsui/commit/65ef14ddad3576420ad00f4802795715cb9ebc44))
> 
>     * [Mapsui.3.0.0-build.1433.nupkg](https://ci.appveyor.com/project/pauldendulk/mapsui/builds/38257237/artifacts) to be the first large one (built from [b4109ab](https://github.com/Mapsui/Mapsui/commit/b4109abc892fe3890122170ed27630d443cde12e))

This assessment was actually based on the order of commits to master, but checking the order of builds shows that the change happened between these two:
* [Mapsui.3.0.0-build.1431.nupkg](https://ci.appveyor.com/project/pauldendulk/mapsui/builds/38216429/artifacts) is still small (from March 14)
* [Mapsui.3.0.0-build.1432.nupkg](https://ci.appveyor.com/project/pauldendulk/mapsui/builds/38233743/artifacts) is the first large one (from March 15)

I suspect this is either related to a change of the AppVeyor build configuration, or an update of tooling/libraries on the AppVeyor build server ...

janusw:
It also looks like the size of any app package built from the latest versions is blown up by almost 15 MB (but interestingly only on Android, not on iOS). Still, this is pretty bad :/

[...]

PR #1513, opened 2022-02-19
Updated to Skia Sharp 2.88.0-preview.209

inforithmics:


janusw (commented):
What's the motivation for this? Does it fix an actual problem? Is it a good idea to use a preview version instead of a stable release?

inforithmics:
This version of Skia sharp has for example fixed the nuget package size increase for Android.

inforithmics:
Currently all the mapsui platforms that need it have this preview version and zhe others have 2.80.3. with my pending pull request ios and android will go to the preview version and only wpgf and forms stay at 2.80.2 so it is more a unification pull request to have all platforms on the same version.

janusw:
> This version of Skia sharp has for example fixed the nuget package size increase for Android.

Ah, very good :) That is #1115.

However, it seems we're not even building a Mapsui.Android package any more (at least not in the AppVeyor builds ;)

pauldendulk:
> However, it seems we're not even building a Mapsui.Android package any more (at least not in the AppVeyor builds ;)

Xamarin.Android and Xamarin.iOS will be added again in 4.0.0-beta.1. The old project format caused problems which I assume will be fixed in the new format.

janusw:
> > However, it seems we're not even building a Mapsui.Android package any more (at least not in the AppVeyor builds ;)
> 
> Xamarin.Android and Xamarin.iOS will be added again in 4.0.0-beta.1. 

But 4.0.0-beta.1 has been released already! Do you mean 4.0.0-beta.2?

pauldendulk:
> But 4.0.0-beta.1 has been released already! Do you mean 4.0.0-beta.2?

Yes, sorry you are right.

pauldendulk (approved):

PR #1508, opened 2022-02-18
fix rotation of callout anchor

janusw:
* fixes #1503
* and at the same time cleans up the code a bit
  - use an MPoint for the symbol offset, so that we can use its rotate method
  - add an conversion method Offset.ToPoint
  - add a variant of MPoint.Rotate that otates around the origin

PR #1517, opened 2022-02-22
MyLocationLayer: make sure that callout rotates properly with map

janusw:
* so that not only the callout itself rotates with the map,
  but also the symbol offset does
* this is sort of a follow-up to #1486 and #1508


pauldendulk (approved):
Thanks!

Would be nice to have a rendering test for this.

Issue #1520, opened 2022-02-24
System.Exception: coordinates can not be length 0


janusw:
This is a new exception that I see after porting my application from Mapsui 3.0.2 to current master (4.0.0-beta.1 and later builds):

```
System.Exception: coordinates can not be length 0
  at Mapsui.Nts.Extensions.CoordinateExtensions.ToLineString (System.Collections.Generic.IEnumerable`1[T] coordinates) [0x00008] in <c7779d0ef6d54298a3f6f385b1af7b78>:0 
  at Mapsui.UI.Forms.Polyline.OnPropertyChanged (System.String propertyName) [0x00053] in <4f5c7a74bf3e4ef8afe3a3bd33642daf>:0 
  at Mapsui.UI.Forms.Polyline.OnCollectionChanged (System.Object sender, System.Collections.Specialized.NotifyCollectionChangedEventArgs e) [0x00000] in <4f5c7a74bf3e4ef8afe3a3bd33642daf>:0 
  at System.Collections.ObjectModel.ObservableCollection`1[T].OnCollectionChanged (System.Collections.Specialized.NotifyCollectionChangedEventArgs e) [0x00018] in /Users/builder/jenkins/workspace/archive-mono/2020-02/android/release/external/corefx/src/System.ObjectModel/src/System/Collections/ObjectModel/ObservableCollection.cs:263 
  at System.Collections.ObjectModel.ObservableRangeCollection`1[T].OnCollectionChanged (System.Collections.Specialized.NotifyCollectionChangedEventArgs e) [0x00015] in <4f5c7a74bf3e4ef8afe3a3bd33642daf>:0 
  at System.Collections.ObjectModel.ObservableCollection`1[T].OnCollectionReset () [0x00000] in /Users/builder/jenkins/workspace/archive-mono/2020-02/android/release/external/corefx/src/System.ObjectModel/src/System/Collections/ObjectModel/ObservableCollection.cs:362 
  at System.Collections.ObjectModel.ObservableCollection`1[T].ClearItems () [0x00018] in /Users/builder/jenkins/workspace/archive-mono/2020-02/android/release/external/corefx/src/System.ObjectModel/src/System/Collections/ObjectModel/ObservableCollection.cs:166 
  at System.Collections.ObjectModel.ObservableRangeCollection`1[T].ClearItems () [0x0000f] in <4f5c7a74bf3e4ef8afe3a3bd33642daf>:0 
[...]

PR #1521, opened 2022-02-24
CoordinateExtensions.ToLineString: remove exception

janusw:
* in order to fix issue #1520


janusw:
Using the nupkgs from this PR, I have verified that simply removing the exception (as done here) indeed fixes #1520 for me.

pauldendulk (approved):

PR #1522, opened 2022-02-25
Directed symbols

pauldendulk:
This is work related to a project I am working on. 

- It simplifies the sample, which can be further simplified
- It adds an IDynamic interface which allows Provider to notify the listener. I am not sure if this is a good solution.
- It adds symbols that point in the direction that they are moving.

image


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -496,13 +496,13 @@ private void OnViewportSizeInitialized()
...
-            if (Viewport.Extent != null)
-            {
-                if (Viewport.Extent == null)
-                    return;
-                var fetchInfo = new FetchInfo(Viewport.Extent, Viewport.Resolution, Map?.CRS, changeType);
-                _map?.RefreshData(fetchInfo);
-            }
+            if (Viewport.Extent == null)
+                return;
+            if (Viewport.Extent.GetArea() <= 0)
```
pauldendulk, line 501: Apparently RefreshData is called when width and height are null. I added a check but perhaps it should not be called at all.

Mapsui/Layers/IDynamic.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) The Mapsui authors.
+// The Mapsui authors licensed this file under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using Mapsui.Fetcher;
+
+namespace Mapsui.Layers;
+
+public interface IDynamic
```
pauldendulk, line 9: This is identical to what is now in ILayer, perhaps ILayer should itself be derived from IDynamic, or should be removed from ILayer and should only be implemented by certain layers if needed.

Mapsui/Utilities/Algorithms.cs
```diff
@@ -35,7 +38,12 @@ public static double Distance(MPoint a, MPoint b)
         /// <returns>Returns the angle in radians</returns>
         public static double DegreesToRadians(double degrees)
         {
-            return degrees * Math.PI / 180.0;
+            return degrees * TO_RADIANS;
+        }
+
+        public static double RadiansToDegrees(double radians)
```
pauldendulk, line 44: Surprised we did not have this one yet.

Samples/Mapsui.Samples.Common/Mapsui.Samples.Common.csproj
```diff
@@ -14,6 +14,7 @@
     <None Remove="EmbeddedResources\torrejon-de-ardoz.mbtiles" />
     <None Remove="EmbeddedResources\world.mbtiles" />
     <None Remove="Images\arrow.png" />
+    <None Remove="Images\arrow.svg" />
```
pauldendulk, line 17: Added an arrow svg to the samples. Perhaps we want to put a selection of symbols in Mapsui core to make it easy for app developers to use them.

inforithmics:
Was this tested with the RasterizingLayer (would this work with the dynamic changes?)

inforithmics:
Generally I like the general idea. And it improves the Interfaces. I generally like to have one specific Interface for one functionality it gives more freedom for extensibility without the need to implement thigs that are not necessary.

pauldendulk:
> Was this tested with the RasterizingLayer (would this work with the dynamic changes?)

You could mean two things:
1. Was this tested with a RasterizingLayer as background. Answer: No, but I see no reason why this would not work.
2. Was this tested with the features within a RasterizingLayer. Answer; No, I expect this this will not work. You should only use rasterizing if there are too many feature to get proper performance when not rasterizing. A RasterizingLayer will make the updates slower and the animations would not be smooth. I think animations should always be direct render updates.

pauldendulk:
> Generally I like the general idea. And it improves the Interfaces. I generally like to have one specific Interface for one functionality it gives more freedom for extensibility without the need to implement thigs that are not necessary.

Yes, perhaps we should work more in that direction, also with respect to our current interfaces like ILayer and IProvider. Split them up. And don't extent interfaces from other interfaces. Just extend classes from multiple interfaces.

inforithmics:
> > Was this tested with the RasterizingLayer (would this work with the dynamic changes?)
> 
> You could mean two things:
> 
> 1. Was this tested with a RasterizingLayer as background. Answer: No, but I see no reason why this would not work.
> 2. Was this tested with the features within a RasterizingLayer. Answer; No, I expect this this will not work. You should only use rasterizing if there are too many feature to get proper performance when not rasterizing. A RasterizingLayer will make the updates slower and the animations would not be smooth. I think animations should always be direct render updates.

I meant option 2. The reason I asked if this interface could be used to for updates that occour less often. Than a life direction update. But it was only curiosity and not a critique.

Samples/Mapsui.Samples.Common/GeoData/RandomPointGenerator.cs
```diff
@@ -21,9 +22,10 @@ private static IEnumerable<PointFeature> CreateFeatures(IEnumerable<MPoint> rand
             return randomPoints.Select(p => new PointFeature(p) { ["Label"] = counter++.ToString() });
         }
 
-        public static IEnumerable<MPoint> GenerateRandomPoints(MRect? envelope, int count = 25, int seed = 192)
+        public static IEnumerable<MPoint> GenerateRandomPoints(MRect? envelope, int count, Random? random = null)
         {
-            _random = new Random(seed);
+            if (random == null)
+                random = new Random(192);
```
inforithmics, line 28: Good to have a seed makes the data reproducible for testing purposes.

Samples/Mapsui.Samples.Common/Mapsui.Samples.Common.csproj
inforithmics: I think this is a good idea too.

inforithmics (approved):
I'll have a look why the Build fails for this

inforithmics:
The Error is

Error parsing the nested project section in solution file. A project with the GUID "{D410C36D-6F9D-4C0C-AEA3-40BC436C4894}" is listed as being nested under project "{969DB691-49FB-4785-98F4-2112CC1BA7C4}", but does not exist in the solution

Seems the merge somehow introduced a solution file error.l

Mapsui.sln
```diff
@@ -193,6 +193,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Mapsui.Samples.Maui", "Samp
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Mapsui.UI.MAUI", "Mapsui.UI.MAUI\Mapsui.UI.MAUI.csproj", "{4E6DDCEC-E178-41A8-B2D4-2E74E93F58EC}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Mapsui.Nts.Tests", "Tests\Mapsui.Nts.Tests\Mapsui.Nts.Tests.csproj", "{1711EFD5-C1A5-441A-97D8-4002F084E9D7}"
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Mapsui.Samples.Eto", "Samples\Mapsui.Samples.Eto\Mapsui.Samples.Eto.csproj", "{D410C36D-6F9D-4C0C-AEA3-40BC436C4894}"
```
inforithmics, line 197: Here there is a missing EndProject. Without this it fails to run on Appveyor.

This should be replaced

```
Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Mapsui.Nts.Tests", "Tests\Mapsui.Nts.Tests\Mapsui.Nts.Tests.csproj", "{1711EFD5-C1A5-441A-97D8-4002F084E9D7}"
```
with this (add EndProjectLine)

```
Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Mapsui.Nts.Tests", "Tests\Mapsui.Nts.Tests\Mapsui.Nts.Tests.csproj", "{1711EFD5-C1A5-441A-97D8-4002F084E9D7}"
EndProject
```

pauldendulk:
> I meant option 2. The reason I asked if this interface could be used to for updates that occur less often, than a life direction update. But it was only curiosity and not a critique.

ah, so it may be useful to update a single tile if necessary, and not so much for animations.

Mapsui/Layers/IDynamic.cs
pauldendulk: Yes, Perhaps the base classes could also be split up.

Mapsui.sln
pauldendulk: Thanks a lot. That would have been time consuming for me to figure out.

PR #1525, opened 2022-02-27
Updated Uno Sample to net6

inforithmics:


inforithmics:
Uno uses now net6.0 Target and the new Multitargeting Infrastructure of Maui. All Unos Sample Projects now compile. 
The Uno Wasm Sample now works except the Wms Background Layers seem not to work. Have to Debug that. (So Mapsui is now "working" in the Browser too).

inforithmics:
The reason that The tiles not work in webassembly is following Code.

return _httpClient.GetByteArrayAsync(arg).ConfigureAwait(false).GetAwaiter().GetResult();

It should be an async call with await instead of getawaiter.GetResult. Beccause getAwaiter().GetResult() is not supported iin the WebAssembly Runtime.

So this would be a bigger change First in Brutile and than in mapsui

pauldendulk:
@inforithmics Sorry for causing conflicts. I hope to catch up on the MRs the coming week and keep up in the follow weeks so that there will be less merge requests.

pauldendulk (approved):
Thanks. Please see if it is possible to remove the commented out code.

PR #1526, opened 2022-02-27
Port to Eto.Forms

rafntor:
As discussed here https://github.com/Mapsui/Mapsui/discussions/1524

Please have a look and let me know if anything missing
(nuspec and CI is missing and should probably be done seperatly?)


inforithmics:
Had to approve the full build run.

inforithmics:
A nuspec file would be great. You can look into the Nuspec folder to see how it should be made. (Copy an existing one for example)

inforithmics:
For CI integration should I make a pull request on your branch or change it in this pull request?

rafntor:
Either way is fine for CI - just updating directly is probably easiest for you (?)

Mapsui.UI.Eto/Extensions/Extensions.cs
```diff
@@ -0,0 +1,12 @@
+﻿
+namespace Mapsui.UI.Eto.Extensions
+{
+    using global::Eto.Drawing;
+    public static class Extensions
+    {
+        public static MPoint ToMapsui(this PointF point)
```
inforithmics, line 7: Is this an Eto Point? (Intrestingly every platform has it's own Points)

Mapsui.UI.Eto/MapControl.cs
```diff
@@ -0,0 +1,201 @@
...
+            ControlInitialize();
+        }
+        private void ControlInitialize()
+        {
+            _invalidate = () => RunOnUIThread(Invalidate);
+
+            // Mapsui.Rendering.Skia use Mapsui.Nts where GetDbaseLanguageDriver need encoding providers
+            System.Text.Encoding.RegisterProvider(System.Text.CodePagesEncodingProvider.Instance);
+
+            LoadComplete += MapControlLoaded;
```
inforithmics, line 28: In the Disposing part this Event Handlers should be derigstered to avoid Memory leaks

Mapsui.UI.Eto/MapControl.cs
```diff
@@ -0,0 +1,201 @@
...
+        }
+        private float GetPixelDensity()
+        {
+            var center = PointToScreen(Location + Size / 2);
+
+            return Screen.FromPoint(center).LogicalPixelSize;
+        }
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
```
inforithmics, line 189: Deregister here the Event Handlers.

for example

`LoadComplete -= MapControlLoaded`

Mapsui.UI.Eto/Mapsui.UI.Eto.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <TargetFramework>netstandard2.0</TargetFramework>
```
inforithmics, line 4: Is Eto Form only .net Standard compatible or does it work on .Net 6.0 too (Because all .Net is moving into that direction).

Samples/Mapsui.Samples.Eto/MainForm.eto.cs
```diff
@@ -0,0 +1,47 @@
...
+			this.Title = "My Eto Form";
+			this.MinimumSize = new Size(200, 150);
+			this.Size = this.MinimumSize * 4;
+			this.Padding = 10;
+
+			this.Content = new StackLayout
+			{
+				Items =
+				{
+					"Hello World!",
```
inforithmics, line 20: Is this Hello World visible? Maybe it can be removed. Seems for me some Template code. (Maybe Hello Mapsui or Hello Eto)

inforithmics (approved):

inforithmics:
I'll merge this if the LangVersion and the Dispose (DeRegister Events is handled)

Mapsui.UI.Eto/Extensions/Extensions.cs
rafntor: It is yes ;-)

Mapsui.UI.Eto/MapControl.cs
rafntor: Hmm, I looked at the existing MapControl-classes and used same approach as those, is this really an issue? https://stackoverflow.com/questions/17399991/should-i-always-disconnect-event-handlers-in-the-dispose-method

inforithmics: I thought about it. And I think this isn't necessary because the event handler and Event live in the same class.

rafntor:
I think I adressed the things you pointed out (except the eventhandler deregistering which I am not sure is needed)
Also added nuspec (not tested) ; hopefully I got it right, please double-check that it looks OK

inforithmics:
Merging this now

inforithmics:
path to eto nuget package

https://github.com/Mapsui/Mapsui/suites/5479049914/artifacts/174767453

rafntor:
Thanks for the support ;-)

PR #1527, opened 2022-02-28
(WIP) Use ProjNet for better Projection Support

inforithmics:


inforithmics:
The main changes I made is I added a Mapsui.Extensions Project (For additional Dependencies like ProjNet which is LGPL migt change to MIT) See discussion here.

https://github.com/NetTopologySuite/ProjNet4GeoAPI/issues/18

This Pull Request supports all transformations of SRID.csv and Spherical Mercator used as default.

Mapsui/Projections/DefaultProjectionFactory.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+
+namespace Mapsui.Projections
+{
+    public static class DefaultProjectionFactory
+    {
+        static DefaultProjectionFactory()
+        {
+            Create = () => throw new Exception("No method to create a renderer was registered");
```
charlenni, line 9: Shouldn't there a message about a projection and not a renderer?

inforithmics: Copy paste error. But I think I should use here the Default of new Projection(),

charlenni:
Are you sure, that Proj4GeoAPI is the right library? They asked for help 3 days ago, because they don't have the time to maintain it anymore. The last version is from 2019 and also the discussion about the license is 2 years old. So I don't think, that they will change it from LGPL to another version.

pauldendulk:
> Are you sure, that Proj4GeoAPI is the right library? They asked for help 3 days ago, because they don't have the time to maintain it anymore. The last version is from 2019 and also the discussion about the license is 2 years old. So I don't think, that they will change it from LGPL to another version.

hmm, The last time I checked Proj4GeoAPI seemed the best option but this does not sound good. Is there any other option? 

If it will be LGPL we can not use it as part of Mapsui. It will be possible to create a Mapsui adapator for it in a separate git repository. Or perhaps even as part of Proj4GeoAPI itself.

inforithmics:
I made some more research and https://github.com/DotSpatial/DotSpatial/tree/master/Source/DotSpatial.Projections seems to be 
MIT and has more projections. I think I know a workaround for referenceing a .NET Framework library in a .Net Standard library. I expect it to work without problems because there shouldn't be any suprises.

There seems to be a Standard compilation (but I think it is best to use the official one)
https://www.nuget.org/packages/DotSpatial.Projections.NetStandard/

inforithmics:
Seems I do at the moment Einsteinian programming. (Program something and then throwing it away and than Programming it better the second time).

pauldendulk:
> Seems I do at the moment Einsteinian programming. (Program something and then throwing it away and than Programming it better the second time).

😂🤣😂 all I have time for is briefly glance at the comments and in the meantime there is a whole new PR again.

PR #1528, opened 2022-02-28
Use DotSpatial.Projection for Projection Improvements

inforithmics:


Mapsui/Providers/ProjectingProvider.cs
```diff
@@ -15,7 +15,7 @@ public class ProjectingProvider : IProvider<IFeature>
         public ProjectingProvider(IProvider<IFeature> provider, IProjection? projection = null)
         {
             _provider = provider;
-            _projection = projection ?? new Projection();
+            _projection = projection ?? ProjectionDefaults.Projection;
```
pauldendulk, line 18: We should try to reduce our use of static fields. Usually things like this would be solved with dependency injection. Perhaps we should start using some form of dependency injection. This nuget is what I use in my backend projects [Microsoft.Extensions.DependencyInjection.Abstractions](https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection.Abstractions) but I am not sure how that interoperates with what is used in MAUI (could use exactly the same) and for instance Avalonia (could use something different).

It could help with a lot of our other design issues as well.

Tests/Mapsui.Tests/Mapsui.Tests.csproj
```diff
@@ -24,6 +24,7 @@
 		<ProjectReference Include="..\..\Mapsui.Tiling\Mapsui.Tiling.csproj" />
 		<ProjectReference Include="..\..\Mapsui.Rendering.Skia\Mapsui.Rendering.Skia.csproj" />
 		<ProjectReference Include="..\..\Mapsui\Mapsui.csproj" />
+		<ProjectReference Include="..\..\Samples\Mapsui.Samples.Common.Desktop\Mapsui.Samples.Common.Desktop.csproj" />
```
pauldendulk, line 27: Why is Mapsui.Samples.Common.Desktop added? 

We should go the other way. Spit it up more. Mapsui.Tests should only have a dependency on Mapsui and nothing else (exception external dependencies that help with testing). The fact that it is not at the moment is a consequence of my refactoring, which is basically not finished yet. There should be a Mapsui.Tiling.Tests with a dependency on Mapsui.Tiling and there should be a Mapsui.Nts.Tests with a dependency on Mapsui.Nts. Mapsui.Rendering.Skia already has a test project.

Mapsui.Nts/Mapsui.Nts.csproj
```diff
@@ -1,11 +1,20 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
```
pauldendulk, line 7: The Mapsui.Nts.dll is in the core Mapsui nuget so now it will also have a dependency on DotSpatial.Projects. I would like to avoid that but not sure how.

pauldendulk (commented):
Thanks a lot for the MR. Better projection support has been on the wish list for a long time. I need to think about the extra dependency though.

inforithmics:
A possibility would be to make mapsui extensions nuget package/project and but the dependency and code there. So if someone doesn't need this functionality the dependency wouldn't be added.

Tests/Mapsui.Tests/Projections/ProjectionTests.cs
```diff
@@ -62,5 +67,43 @@ public void CoordinateProjectionTest()
                 Assert.AreNotEqual(coordinates[i].Y, projectedCoordinates[i].Y);
             }
         }
+
+        [Test]
+        public void CoordinateNtsProjectionTest()
```
pauldendulk, line 72: Thanks for adding unit tests!

Tests/Mapsui.Tests/Projections/ProjectionTests.cs
```diff
@@ -62,5 +67,43 @@ public void CoordinateProjectionTest()
...
+                Assert.AreNotEqual(coordinates[i].X, projectedCoordinates[i].X);
+                Assert.AreNotEqual(coordinates[i].Y, projectedCoordinates[i].Y);
+            }
+        }
+
+        [Test]
+        public void ShapeFileReadTest()
+        {
+            // arrange
+            DotSpatialProjection.Init();
```
pauldendulk, line 98: I could not really guess what was happening in the Init method. Perhaps it is better not to have an Init method and just use the code like this:

        ProjectionDefaults.Projection = new DotSpatialProjection();
        ProjectionDefaults.CrsFromEsri = CrsFromEsri;

And Perhaps CrsFromEsri should just be set in ProjectionDefaults itself. There is no other option right now, so why not.

Mapsui.Extensions/Projections/DotSpatialProjection.cs
```diff
@@ -0,0 +1,231 @@
...
+
+        if (fromCoordinateSystem == null || toCoordinateSystem == null)
+        {
+            return null;
+        }
+
+        return (fromCoordinateSystem, toCoordinateSystem);
+    }
+
+    public static string CrsFromEsri(string esriString)
```
pauldendulk, line 206: Perhaps this method should be in it's own class. It is not tied to the projection. It is related to projections but not itself a projection and CrsFromEsri could be used in combination with another projection.

Mapsui.Extensions/Mapsui.Extensions.csproj
```diff
@@ -5,12 +5,22 @@
...
+    <ProjectReference Include="..\Mapsui.Nts\Mapsui.Nts.csproj" />
     <ProjectReference Include="..\Mapsui\Mapsui.csproj" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="BruTile" Version="4.0.0" />
-    <PackageReference Include="sqlite-net-pcl" Version="1.6.292"/>
+    <PackageReference Include="sqlite-net-pcl" Version="1.6.292" />
+    <PackageReference Include="DotSpatial.Projections.NetStandard" Version="1.0.0" />
+    <!--<PackageReference Include="DotSpatial.Projections" Version="3.0.0" IncludeAssets="None" ExcludeAssets="All" PrivateAssets="None" GeneratePathProperty="true" NoWarn="NU1202;NU1701" />-->
```
pauldendulk, line 16: Can these commented out parts be removed.

pauldendulk (changes requested):
Looks great! Just a few comments you could look at.

Tests/Mapsui.Tests/Projections/ProjectionTests.cs
pauldendulk: And if the CrsFromEsri is set globally perhaps you need only the first line.

Mapsui/Providers/ProjectingProvider.cs
inforithmics: I reverted this change, dependencies should come with the constructor so they are not dependent on any Dependency Injection framework. Because Mapsui is a component it should not dependend on a specific one. But dependencies should be specified by constructor.

Tests/Mapsui.Tests/Mapsui.Tests.csproj
inforithmics: removed now. I tried something and then I forgot to remove it afterwards.

Tests/Mapsui.Tests/Projections/ProjectionTests.cs
inforithmics: Is no removed and a Parameter in the Constructor.

Mapsui.Extensions/Mapsui.Extensions.csproj
inforithmics: cleaned up

Mapsui/Providers/ProjectingProvider.cs
pauldendulk: I have been looking in MAUI a bit more and they are indeed using Microsoft.Extensions.DependencyInjection.Abstractions. Perhaps we can add helper methods to register Mapsui core services.

Mapsui.Extensions/Projections/DotSpatialProjection.cs
```diff
@@ -0,0 +1,233 @@
...
+    private static readonly ConcurrentDictionary<int, ProjectionInfo> Projections = new();
+    private static readonly ConcurrentDictionary<(int From, int To), GeometryTransform> GeometryTransformations = new();
+    private static readonly ConcurrentDictionary<string, string> CrsFromEsriLookup = new();
+
+    public static void Init()
+    {
+        ProjectionDefaults.Projection = new DotSpatialProjection();
+    }
+
+    public int? GetIdFromCrs(string? crs)
```
pauldendulk, line 30: Looks like this method can be private now.

pauldendulk (approved):
Thanks for picking up the comments. This going back and forth is making things better. One more remark you could look at.

PR #1529, opened 2022-02-28
Added Geometry Simplify Provider

inforithmics:
This can be used to Simplify the Geometries and increase the Drawing speed (less complexity).


inforithmics:
This is a provider that I wrote for the V2 TiliningLayer Pull request and can be used without a tiling Layer

inforithmics:
The Sample number 5 is intentional because I have The Sample number 4 used in the TilingProvider pull request.

Samples/Mapsui.Samples.Common.Desktop/GeometrySimplifySample.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            var map = new Map();
+
+            var countrySource = new ShapeFile(GetAppDir() + "\\GeoData\\World\\countries.shp", true);
+            countrySource.CRS = "EPSG:4326";
+            var projectedCountrySource = new ProjectingProvider(countrySource) {
+                CRS = "EPSG:3857",
+            };
+
+            var simplifyCountrySource = new GeometrySimplifyProvider(projectedCountrySource, distanceTolerance: 200000);
```
pauldendulk, line 33: ah yes, put the projected data into the simplifier, so in this case it will be simplifying sphericalmercator coordinates instead of wgs84 which makes more sense.

pauldendulk (approved):
Nice to use an IProvider for this so it can be stacked on other providers. Thanks a lot.

PR #1530, opened 2022-02-28
Checkin for Sample Regression Tests

inforithmics:


inforithmics:
Main Changes:
1. Added ISampleTest Interface for making the Sample testable. (Stopping animations for example).
2. Moved SqlitePersistentCache to the Mapsui.Tiling Project and added cacheTimeOut to the Constructor.
3. All samples are run, the Samples that have consistent Created Images the Images are compared for the others they are only run.

inforithmics:
Maybe the SqlitePersistCache could be moved to mapsui.Extensions Project too.

pauldendulk:
> 2. Moved SqlitePersistentCache to the Mapsui.Tiling Project and added cacheTimeOut to the Constructor.

So Mapsui.Tiling will have a dependency on sqlite. Not great, but caching is important and may have to be standard functionality of the tiling layer we ship.

> 3. All samples are run, the Samples that have consistent Created Images the Images are compared for the others they are only run.

What is the benefit of only running the samples? Would exceptions be detected by the test?

Mapsui.Tiling/Cache/Tile.cs
```diff
@@ -1,9 +1,12 @@
-﻿namespace Mapsui.Samples.Common.Desktop.Cache;
+﻿using System;
+
+namespace Mapsui.Samples.Common.Desktop.Cache;
 
 public class Tile
```
pauldendulk, line 5: This tile class gets more prominent. Should it be in BruTile? Or could an existing BruTile class be extended so it could be used?

Mapsui.Tiling/Mapsui.Tiling.csproj
```diff
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="BruTile" Version="4.0.0" />
+    <PackageReference Include="sqlite-net-pcl" Version="1.6.292"/>
```
pauldendulk, line 9: Hmmm, I understand why we need it be I don't want to release Mapsui core with a dependency on sqlite.

Mapsui.Tiling/OpenStreetMap.cs
```diff
@@ -11,6 +12,8 @@ namespace Mapsui.Tiling
 {
     public static class OpenStreetMap
     {
+        public static IPersistentCache<byte[]>? DefaultCache = null;
```
pauldendulk, line 15: Perhaps we should have caching on by default but then all tile layers depend on sqlite. I am also thinking about moving Mapsui.Tiles to a separate nuget package, but almost all samples need it as background layer, so that won't work.

What about a file based cache?

Mapsui/Layers/AnimatedLayers/AnimatedPointLayer.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace Mapsui.Layers.AnimatedLayers;
 
-public class AnimatedPointLayer : BaseLayer
+public class AnimatedPointLayer : BaseLayer, ILayerDataSource
 {
```
pauldendulk, line 8: Why was this needed?

pauldendulk:
> Maybe the SqlitePersistCache could be moved to mapsui.Extensions Project too.

Okay, Lets go for Mapsui.Extensions. It will probably get a huge number of dependencies quickly but perhaps it should be seen as a more experimental nuget. We could parts in to separate nugets later on. Perhaps we should have just no restrained at all wrt to adding dependencies to Mapsui.Extensions. ... It is bound to become a mess though.

inforithmics:
I think splitting things up is better. When something doesn't introduce more dependencies it could still lif in mapsui.
Because I don't like Frankenstein nugets, because they become dificcult to manage the dependencies.

So additional dependencies could be added each into its own mapsui nuget.

Mapsui.Caching ...
Mapsui.Projection ...
Mapsui.GisProvider ...  

So the additional number of dependencies dos not grow, only for things that are needed.
Mapsui.Arcgis I think set the precedent here.

Maybe every Mapsui.dll should be in its own Mapsui ... nuget package.

inforithmics:
> > 2. Moved SqlitePersistentCache to the Mapsui.Tiling Project and added cacheTimeOut to the Constructor.
> 
> So Mapsui.Tiling will have a dependency on sqlite. Not great, but caching is important and may have to be standard functionality of the tiling layer we ship.
> 
> > 3. All samples are run, the Samples that have consistent Created Images the Images are compared for the others they are only run.
> 
> What is the benefit of only running the samples? Would exceptions be detected by the test?

Yes exception would fail the test. 
These are samples, that have to unreliable Image Genereation they sometimes fail every 3rd Time Or don't work yet like the WfsSample, because the Wfs call is not cached and some Test rendering Problem.

Mapsui/Layers/AnimatedLayers/AnimatedPointLayer.cs
inforithmics: This is for getting the DataSource and stopping\Controlling the Animation for the Test

inforithmics:
I think for now I make a Mapsui.Extensions project and nuget, which could be split up if it gets to large.

pauldendulk (approved):
This is going to make a big difference. Errors introduced by refactoring will be found much more often. It is a very effective way to make a big improvement in the coverage with a fairly simple solution. This is the coverage of the non-UI projects as determined by dotcover:

image

This is what it was before:

image


### About rendering versus regression tests
In the skia tests there are now the **rendering tests** besides the **regression tests**. 

The existing **rendering tests** are proper unit tests in the sense that they use only local data but are regression tests in the sense that they only compare the full state (and can not say anything about which isolated part failed).

Do the **regression tests** also depend on webservices which make them less reliable (but still very useful), but since they use the cache in many cases some of them could be considered to be equal to the rendering tests. Is that correct? Should they perhaps be moved? Or is that hard because they work very different because of the RegressionMapControl?

inforithmics:
> This is going to make a big difference. Errors introduced by refactoring will be found much more often. It is a very effective way to make a big improvement in the coverage with a fairly simple solution. This is the coverage of the non-UI projects as determined by dotcover:
> 
> image
> 
> This is what it was before:
> 
> image
> 
> ### About rendering versus regression tests
> In the skia tests there are now the **rendering tests** besides the **regression tests**.
> 
> The existing **rendering tests** are proper unit tests in the sense that they use only local data but are regression tests in the sense that they only compare the full state (and can not say anything about which isolated part failed).
> 
> Do the **regression tests** also depend on webservices which make them less reliable (but still very useful), but since they use the cache in many cases some of them could be considered to be equal to the rendering tests. Is that correct? Should they perhaps be moved? Or is that hard because they work very different because of the RegressionMapControl?

The Regression Tests should not depend on WebServices, for that I introduced The SqliteCache, so during the First run of the Regression Test all Tiles are loaded into the Cache (with no expiration time) and afterwards they don't depend on the WebServices.
The Wfs Test I have not included in the regression Test, because for that I have to implement cacheing into the Wfs Call. (Tiling the Wfs Calls and caching the Tiled Wfs Calls).

Tests/Mapsui.Rendering.Skia.Tests/RegressionMapControl.cs
```diff
@@ -0,0 +1,112 @@
...
+using Mapsui.UI;
+using Mapsui.Utilities;
+
+#pragma warning disable CS8600 // Converting null literal or possible null value to non-nullable type.
+#pragma warning disable IDISP008 // Don't assign member with injected and created disposables
+#pragma warning disable CS0067 // The event is never used
+
+namespace Mapsui.Rendering.Skia.Tests;
+
+public class RegressionMapControl : IMapControl
```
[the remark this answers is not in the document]
inforithmics: This is not depend on a specific platform, maybe The regression tests could be run on all platforms but this is for another time.

pauldendulk:
Tiling the WFS call would be fancy. No priority for that, but good to keep it in the back of our mind.

When I run the tests without wifi there are six failing tests, so some still depend on external services. They are related to WFS, WMTS, WMS, TMS. I suspect it is mostly related to fetching meta data.

pauldendulk:
@inforithmics The shapefile with labels sample does not work anymore. I have not investigated it, but if you have a hint I would appreciate it.

inforithmics:
I'll have a look at it.

PR #1535, opened 2022-03-05
eto : use movecursor when moving

rafntor:


inforithmics:
I'll looked at the pull request serms ok for me. I have one question though In Win forms the Cursor was saved in a local/field currentcursor and this was reatored instead the default one. The reason was when the user/programmer had changed it outside of the compnent. That the custom cursor was kept.

rafntor:
Good idea, does it look more complete now?

inforithmics (approved):
Looks good to me.

PR #1536, opened 2022-03-05
eto: use overrides instead of eventhandlers

rafntor:
I think this is preferable (?)


inforithmics:
I too think this is the way to go.

inforithmics (approved):

pauldendulk:
Yes, that's better, and could be changed in many places.

@inforithmics Thanks for taking care of the eto reviews.

PR #1545, opened 2022-03-12
Fix projection of raster and rect

pauldendulk:


pauldendulk:
This was a bug of functionality that is not tested anywhere in our tests or samples.

PR #1547, opened 2022-03-15
Fixed ShapeFile with Label

inforithmics:
Set Default Style For Rasterizing Style Layer or else it won't display because Mapsui has default VectorStyleLayer and this does not know how to display RasterFeatures

This is fixed
https://github.com/Mapsui/Mapsui/pull/1530#issuecomment-1065838694


pauldendulk (approved):
Awesome!

PR #1549, opened 2022-03-17
Enabling Android, iOS, Avalonia and Eto in build scripts

pauldendulk:


pauldendulk:
@inforithmics The build here fails on Mapui.UI.iOS while it succeeds on my machine. The changes in the PR seem unrelated. I do not want to spend time on Mapsui.UI.iOS/Android. Could you take a look at it?

pauldendulk:
Merging anyway because I want to move on with the release of other components.

PR #1552, opened 2022-03-17
Update Skia.Sharp 2.88.0-preview.232 for Maui Preview 14 Compatibility

inforithmics:


inforithmics:
Upgrading Skia Sharp caused to Render the Images slightly different, But the Bitmap Atlas sample is some more changes. I think it has to do with Clipping or scaling.

pauldendulk:
@inforithmics Thanks a lot!! I am just now starting to dig into Maui and how it compiles and am very confused by these compile errors because I have little background so far. I will continue with this and update the tests. I will accept the problem with Bitmap Atlas at this point.

pauldendulk (approved):

pauldendulk:
@inforithmics Is my interpretation correct that the build broke because the new version of the MAUI preview was released and the build servers were automatically updated? And could we fix a MAUI version? Or perhaps we should not, and just go up quickly to a new version when it is there?

inforithmics:
The problem with MAUI is that when a preview gets released the sdk-manifests get updated to the new Preview 14. And `dotnet workload install maui` s to the Maui Preview. Unfortunatly they introduced breaking changes like renames and Interfaces changes. That's why an upgrade to skia sharp was necessary. I'll make a pull request later to update the Test images than the tests should work and the appveyor build should become green. The different rendering of Atlas Bitmap Image I have to investigate.

pauldendulk:
>  I'll make a pull request later to update the Test images than the tests should work and the appveyor build should become green.

Fantastic. Could you describe how you go about updating the cache, so I would be able to do this myself?

inforithmics:
The regression test images are in Generated folder with regression in the
name. They need to ve moved to the regression folder in the test project.
The Cache files are in the cache folder simply move the contents to the
cache folder. And the test images are in the generated folder and need to
be copies to the image folder in the project.


Paul den Dulk ***@***.***> schrieb am Fr., 18. März 2022,
7:46 AM:

PR #1555, opened 2022-03-19
Regression images to separate output folder

pauldendulk:
Why: Generally easier to understand what belong where and makes it easier to copy generated to original folder with cmd.


Scripts/test-image-copier.cmd
```diff
@@ -2,4 +2,5 @@ REM Run this cmd from the sln folder to copy all generated images over the origi
 REM scripts\test-image-copier.cmd
 REM Only do this if you have checked that the generated images are correct.
 REM todo: run unit tests C:\Program Files (x86)\NUnit.org\nunit-console\nunit3-console.exe 
-xcopy .\Tests\Mapsui.Rendering.Skia.Tests\bin\Debug\netcoreapp3.1\Resources\Images\Generated .\Tests\Mapsui.Rendering.Skia.Tests\Resources\Images\Original  /Y
\ No newline at end of file
+xcopy .\Tests\Mapsui.Rendering.Skia.Tests\bin\Debug\netcoreapp3.1\Resources\Images\GeneratedTest .\Tests\Mapsui.Rendering.Skia.Tests\Resources\Images\OriginalTest  /Y
+xcopy .\Tests\Mapsui.Rendering.Skia.Tests\bin\Debug\netcoreapp3.1\Resources\Images\GeneratedRegression .\Tests\Mapsui.Rendering.Skia.Tests\Resources\Images\OriginalRegression  /Y
```
pauldendulk, line 6: The test-image-copier.cmd now also copies the regression images back to GeneratedRegression. Before this PR there as a separate 'Original' and 'Regression' folders but only one 'Generated' folder were both were writte to. This makes copying back hard. The main reason for this PR was to make this possible, but it also makes more sense to have one on one folders.

Tests/Mapsui.Rendering.Skia.Tests/File.cs
```diff
@@ -33,29 +44,35 @@ public static void WriteToGeneratedFolder(string fileName, MemoryStream? stream)
...
+                    throw new Exception($"Assembly.GetExecutingAssembly().CodeBase was null") ;
 
                 var uri = new UriBuilder(codeBase);
                 var path = Uri.UnescapeDataString(uri.Path);
-                return Path.GetDirectoryName(path);
+                return Path.GetDirectoryName(path)!;
             }
         }
 
+        public static Stream ReadFromImagesFolder(string fileName)
```
pauldendulk, line 61: There were some svg utility tests that also used the Images/Original folder but 'Original' only makes sense if you generate an image and compare to that original.  So added a regular ReadFromImagesFolder.

Tests/Mapsui.Rendering.Skia.Tests/Mapsui.Rendering.Skia.Tests.csproj
```diff
@@ -94,460 +52,22 @@
...
   <ItemGroup>
-    <Folder Include="Resources\Images\Regression\" />
+    <Folder Include="Resources\Images\OriginalTest\">
+    </Folder>
+    <Folder Include="Resources\Images\OriginalRegression\">
+    </Folder>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="Resources\Images\OriginalTest\*.*">
```
pauldendulk, line 65: Could remove all the explicit copy settings on individual files by copying the entire folder.

Tests/Mapsui.Tests/Mapsui.Tests.csproj
```diff
@@ -58,13 +58,13 @@
 		<Content Include="Resources\example.tif">
 			<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
 		</Content>
-		<Content Include="Resources\Images\Original\Pin.svg">
+		<Content Include="Resources\Images\Pin.svg">
```
pauldendulk, line 61: Moved these files to Images.

Tests/Mapsui.Rendering.Skia.Tests/File.cs
```diff
@@ -33,29 +44,35 @@ public static void WriteToGeneratedFolder(string fileName, MemoryStream? stream)
             stream?.WriteTo(fileStream);
         }
 
-        public static string? AssemblyDirectory
+        public static string AssemblyDirectory
```
pauldendulk, line 47: @inforithmics I am still working out how to go about nullability. I think this is an example we we can check early and just throw if it is null, because there is no alternative way to go about it, and also I doubt if we ever encounter a scenario where it actually is null. Perhaps I should have added a throw before the return as well.

PR #1556, opened 2022-03-19
Update to BruTile 5.0.0

pauldendulk:


Mapsui.Tiling/Layers/TileLayer.cs
```diff
@@ -144,7 +144,7 @@ private void TileFetchDispatcherOnDataChanged(object sender, DataChangedEventArg
 
         private RasterFeature? ToFeature(TileInfo tileInfo)
         {
-            var tileData = _tileSource.GetTile(tileInfo);
+            var tileData = _tileSource.GetTileAsync(tileInfo).Result;
```
pauldendulk, line 147: .Result should be avoided but I updated without any other changes in Mapsui.

Issue #1546, opened 2022-03-15
[Regression] Pin callout is not shown when user clicks on the pin


g-sharat:
### Description
Pin callout is not shown when user clicks on the pin in my attached project and also in the sample project

### Steps to Reproduce

1. Download the attach project and run it with android
2. Click on the pin

### Expected Behavior
Should open the callout somewhere near the pin

### Actual Behavior
No callout shows up

### Basic Information

- Version with issue:
- Last known good version: N/A
- Platform Target Frameworks: 
  - iOS:  Not checked in iOS
  - Android: Android API 30, mdpi 5.4 inch emulator
  - UWP:  N/A 

### Screenshots

![Screenshot_1647352237](https://user-images.githubusercontent.com/70692056/158392829-feaf06b4-9055-4657-9360-3be1db2c7c1e.png)
![Screenshot_1647352249](https://user-images.githubusercontent.com/70692056/158392837-36f92f8f-2237-41ab-85d7-3f29ef7ecf2b.png)

### Reproduction Link

[MapsText.zip](https://github.com/Mapsui/Mapsui/files/8253892/MapsText.zip)

### Workaround

None found

### Extra notes

This issue happens in sample project provided in the github as well, if you want to test in there then open ``Mapsui.Samples.Forms.Android`` as start up project and open ``AddManyPins`` and put break point in ``OnPinClicked`` in ``MapPage.Xaml.cs`` and observe that it fires when clicked on pin but doesn't show anything. But in my attached  project by adding ``pin.ShowCallout();`` ``OnAppearing()`` shows the content but ui is not proper(as shown in 2nd screenshot) and doesn't reopen once clicked again.
Any fixes or suggestions are appreciated.

janusw:
Confirmed. With current master, I see the following errors with the "Add Pin Sample":

```
Error: Unexpected error in skia renderer, System.InvalidCastException: Specified cast is not valid.
  at Mapsui.Rendering.Skia.CalloutStyleRenderer.Draw (SkiaSharp.SKCanvas canvas, Mapsui.IReadOnlyViewport viewport, Mapsui.Layers.ILayer layer, Mapsui.IFeature feature, Mapsui.Styles.IStyle style, Mapsui.Rendering.ISymbolCache symbolCache, System.Int64 iteration) [0x00001] in /Users/weija001/github/Mapsui/Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs:15 
[...]

PR #1559, opened 2022-03-20
Fix Pin callout for #1546

pauldendulk:


Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
```diff
@@ -12,12 +12,17 @@ public class CalloutStyleRenderer : ISkiaStyleRenderer
     {
         public bool Draw(SKCanvas canvas, IReadOnlyViewport viewport, ILayer layer, IFeature feature, Styles.IStyle style, ISymbolCache symbolCache, long iteration)
         {
-            var pointFeature = (PointFeature)feature;
+            var centroid = feature.Extent?.Centroid;
```
pauldendulk, line 15: Using the Feature.Extent.Centroid so that it works for any kind of feature. At some point we may need to think about better placing for polygons. The centroid of the extend might not be the best location.

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
```diff
@@ -12,12 +12,17 @@ public class CalloutStyleRenderer : ISkiaStyleRenderer
...
+
+            if (centroid is null) 
+                return false;
+
             var calloutStyle = (CalloutStyle)style;
 
+            // Todo: Use opacity
             var opacity = (float)(layer.Opacity * style.Opacity);
```
pauldendulk, line 23: At the moment the opacity is not use here.

PR #1569, opened 2022-03-21
Fixed WFS PropertyName using POST requests on master branch

JBou:
This is a follow up PR on #1483 and fixes WFS PropertyName using POST requests on the master branch.


pauldendulk (approved):

PR #1582, opened 2022-03-30
Some work on samples

pauldendulk:


Mapsui.Rendering.Skia/BitmapInfo.cs
```diff
@@ -96,6 +96,8 @@ public float Width
                         return Svg?.Picture?.CullRect.Width ?? 0;
                     case BitmapType.Sprite:
                         return Sprite?.Width ?? 0;
+                    case BitmapType.Picture:
```
pauldendulk, line 99: This was apparently a bug. A Picture had width/height zero, which messed up positioning.

Samples/Mapsui.Samples.Common/Maps/Special/WritableLayerSample.cs
```diff
@@ -1,51 +1,48 @@
...
         }
 
         private Map CreateMap()
         {
             var map = new Map();
 
             map.Layers.Add(OpenStreetMap.CreateTileLayer());
+
             var writableLayer = new WritableLayer();
+            var pinId = typeof(Map).LoadSvgId("Resources.Images.Pin.svg");
```
pauldendulk, line 29: Using svg now. I think there should be helper methods that load specific symbols.

Samples/Mapsui.Samples.Maui/Readme.txt
```diff
@@ -7,11 +7,7 @@ Following things need to be installed.
 2. ios Workload
    In Administrative Console run folowwing program.
    dotnet workload install maui
-3. Install Maui-check
```
pauldendulk, line 10: Also not related, removed some more Maui-check

appveyor.yml
```diff
@@ -24,8 +24,6 @@ install:
 - ps: ./dotnet-install.ps1 -Version 6.0.201 -InstallDir "dotnetcli"
```
pauldendulk, line 24: In the image above (I can not add a comment to an image) you can see the green pin is now above the location it is revering to, that is how it is supposed to work.

PR #1583, opened 2022-04-02
Introduced IAsyncProvider for Async Providers

inforithmics:
I made a IAsyncProvider interface with a AsyncProviderBase.

In the Layer.cs it asks if it is an IAsyncProvider if yes it runs it asynchronous if not it takes the synchronous implementation.

With this Synchronous Providers like MemoryProvider stay synchronous and the ILayer Interface does not need to change.


inforithmics:
When this is the right direction I will change the implementation of the async providers

Mapsui/Providers/AsyncProviderBase.cs
```diff
@@ -0,0 +1,26 @@
+using System.Collections.Generic;
+using Mapsui.Layers;
+
+namespace Mapsui.Providers;
+
+public abstract class AsyncProviderBase<T> : IAsyncProvider<T> where T : IFeature
+{
+    public string? CRS { get; set; }
+    
+    public IEnumerable<T> GetFeatures(FetchInfo fetchInfo)
```
pauldendulk, line 10: I assume the plan is to remove this sync version when the async version is implemented?

Mapsui/Providers/AsyncProviderBase.cs
```diff
@@ -0,0 +1,26 @@
...
+        if (enumerator.MoveNextAsync().Result)
+#pragma warning restore VSTHRD002
+        {
+            yield return enumerator.Current;
+        }
+
+        enumerator.DisposeAsync().GetAwaiter().GetResult();
+    }
+
+    public abstract MRect? GetExtent();
```
pauldendulk, line 23: Since the Extent is in most cases also retrieved from a remote web resource or database it also needs to be async. Unfortunately this will make the upcoming changes more complicated because in the current implementation the IDataSource.GetExtent is called from the Layer.Extent. This is actually a flaw in our current implementation. The Layer.Extent should return a cached version just as it returns cached features. Perhaps for now it should return the IDataSource.GetExtent().Result for now, and we can think about a solution in a following PR.

image

pauldendulk (approved):
Yes, Lets do this. See my remarks.

pauldendulk:
> With this Synchronous Providers like MemoryProvider stay synchronous and the ILayer Interface does not need to change.

Oh, I had missed this. Guess this means the plan is to support both sync and non-sync interfaces. For the short term this is acceptable. I think I would like to have it separated in the long run.

inforithmics:
> > With this Synchronous Providers like MemoryProvider stay synchronous and the ILayer Interface does not need to change.
> 
> Oh, I had missed this. Guess this means the plan is to support both sync and non-sync interfaces. For the short term this is acceptable. I think I would like to have it separated in the long run.

Do you mean with seperated an async version and a synchronous version? So I think I could introduce an iproviderbase interface and the memorylayer would only accept a synchronous iprovider interface.

pauldendulk:
> Do you mean with separated an async version and a synchronous version? So I think I could introduce an iproviderbase interface and the memorylayer would only accept a synchronous iprovider interface.

I am sorry but I don't have a full plan yet. Some thoughts:
- I think the IAsyncProvider and IProvider should be separate interfaces, and implementations should implement either one or the other.
- But perhaps the sync IProvider is not necessary at all. Removing it will cause some problems for existing classes.
  - Perhaps the MemoryProvider could have one of the standard collection classes as data source, and could calculate the extent from that source itself.
  - For the regression tests we want a layer that does sync requests so there is no delay and the first iteration draws the data. Perhaps we should add a layer for that specifically for testing. It could be part of Mapsui.Tests.csproj. It has a IAsyncProvider data source but calls it directly, not from the fetcher.
  - There are probably other layers and providers that will have a problems if we remove the sync IProvider. No plan yet.
- The FeatureFetchDispatcher should only accept the IAsyncProvider. That is it intended for. The sync IProvider should always return data that is available in the cache and there should be no need to fetch it async.
- We have some transforming providers like the ProjectingProvider and the RasterizingProvider. I think they should be IAsyncProviders but this may complicate interaction with an IProvider.
- Perhaps the first MR on this topic should have separate IAysncProviders and IProviders and we should think about how to remove the IProviders later.

Mapsui/Providers/AsyncProviderBase.cs
inforithmics: I looked through the Code and everywhere the Extent is used synchronous, So I didn't change that.

inforithmics:
I Introduced an IAsyncSample Interface for Samples that are asynchronous. This interface is used in the Regression Tests Too. I made an ExtensionMethod SetupAsync that calls the AsyncVersion if available or else the synchronous one.

inforithmics:
Still WIP because the Async Sample has some Trheading Problems

inforithmics:
I think the WMSProvider with the url Constructor should be created by a WmsProviderFactory, So that the Creation is Async and afterwards the operation can be synchronous.

Mapsui/Providers/AsyncProviderBase.cs
pauldendulk: You mean in the implementations of the providers there is no data fetching in the GetExtent?

Mapsui/Providers/IAsyncProvider.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Collections.Generic;
+using Mapsui.Layers;
+
+namespace Mapsui.Providers
+{
+    /// <summary>
+    /// Interface for data providers
+    /// </summary>
+    public interface IAsyncProvider<out T> : IProvider<T> where T : IFeature
```
pauldendulk, line 15: What fails if the IAsyncProvider is not derived from IProvider?

Mapsui/Providers/AsyncProviderBase.cs
```diff
@@ -0,0 +1,26 @@
...
+
+public abstract class AsyncProviderBase<T> : IAsyncProvider<T> where T : IFeature
+{
+    public string? CRS { get; set; }
+    
+    public IEnumerable<T> GetFeatures(FetchInfo fetchInfo)
+    {
+        var enumerator = GetFeaturesAsync(fetchInfo).GetAsyncEnumerator();
+#pragma warning disable VSTHRD002
+        if (enumerator.MoveNextAsync().Result)
```
pauldendulk, line 14: When is GetFeatures called on an IAsyncProvider? What needs to be done to remove it?

pauldendulk (commented):
No full review, just two questions.

Mapsui/Providers/AsyncProviderBase.cs
inforithmics: Yes everything is Iinitialized in the Constructor or before.

Mapsui/Providers/AsyncProviderBase.cs
inforithmics: I'll investigate if this is still needed, because I made a IBaseProvider Interface and made Extension Methods that call the Sync or Async Version.

inforithmics:
So I cleaned it up IAsyncProvider is providing only GetFeaturesAsync

Mapsui/Providers/AsyncProviderBase.cs
inforithmics: Conclusion it isn't needed. Fixed the remaining places.
The only thing is that encapsulating Providers need to implement both interfaces, to be able to be used in Sync Cases.

One Sample of this case is the ProjectionSample.

Here the MemoryLayer calls directly the ProjectionProvider which Calls the MemoryProvider. So for example IterateLayers  
goes always through Projection for every call. 

The question could now be if there is a better solution, But I think this could be discussed another time.

At the Moment I see two options.

1. Keep as is.
2. Only implement IAsyncProvider in the Wrapping Providers and use Layer instead of MemoryLayer in the sample.
3. Improve MemoryLayer somehow to fetch data only once (But this might be a behavorial change)

pauldendulk (approved):
I see you investigated my question to support both sync and async (now still done in the ProjectingProvider). I understand the problem as you describe it here: 
https://github.com/Mapsui/Mapsui/pull/1583#discussion_r851033712
I think the solution should be to adjust the MemoryProvider. The ILayer.GetFeatures of the ILayer is sync and should stay sync because it is fetching directly from memory, and the IProvider.GetFeaturesAsync should be async because it is fetching from a remote source. So calling IProvider.GetFeaturesAsync from the MemoryLayer.GetFeatures is just wrong. It is something that needs to be fixed in a later request. 

One issue here is that we use the direct fetching of the MemoryLayer in the tests. Perhaps we should add a TestLayer (similar to the MemoryLayer) that is used specifically for testing and only exists in our test library.

pauldendulk:
@inforithmics Thanks again for all the work you have done. 🤩 btw, I would like to have video chat sessions as well. Just a few short conversations can speed up some things a lot.

PR #1585, opened 2022-04-07
(WIP) Set up of dynamic memory provider

pauldendulk:
The intent is to have an event in a dynamic memory provider that triggers a refresh of the data, but that does not work in the current setup. Workarounds become complicated. I need to think about another design.

### DIFF
--- a/Mapsui.UI.Wpf/MapControl.cs
+++ b/Mapsui.UI.Wpf/MapControl.cs
@@ -192,6 +192,7 @@ namespace Mapsui.UI.Wpf
         {
             var mousePosition = e.GetPosition(this).ToMapsui();
 
+
             if (_previousMousePosition != null)
             {
                 if (IsInBoxZoomMode())
@@ -204,6 +205,10 @@ namespace Mapsui.UI.Wpf
                 {
                     HandleFeatureInfo(e);
                     OnInfo(InvokeInfo(mousePosition, _downMousePosition, e.ClickCount));
+                    _mouseDown = false;
+                    _previousMousePosition = new MPoint();
+                    ReleaseMouseCapture();
+                    return;
                 }
             }
 
--- a/Mapsui/Layers/IDynamic.cs
+++ b/Mapsui/Layers/IDynamic.cs
@@ -12,7 +12,7 @@ public interface IDynamic
     /// Event called when the data within the layer has changed allowing
     /// listeners to react to this.
     /// </summary>
-    event DataChangedEventHandler DataChanged;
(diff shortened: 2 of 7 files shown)


Mapsui.UI.Wpf/MapControl.cs
```diff
@@ -204,6 +205,10 @@ private void MapControlMouseLeftButtonUp(object sender, MouseButtonEventArgs e)
                 {
                     HandleFeatureInfo(e);
                     OnInfo(InvokeInfo(mousePosition, _downMousePosition, e.ClickCount));
+                    _mouseDown = false;
```
pauldendulk, line 208: These lines were added to prevent the call to RefreshData a few lines below. This is because I want to test a refresh trigger from the dynamic memory provider.

Mapsui/Layers/IDynamic.cs
```diff
@@ -12,7 +12,7 @@ public interface IDynamic
     /// Event called when the data within the layer has changed allowing
     /// listeners to react to this.
     /// </summary>
-    event DataChangedEventHandler DataChanged;
+    event DataChangedEventHandler? DataChanged;
```
pauldendulk, line 15: I think such an event handler should always be nullable.

Mapsui/Layers/Layer.cs
```diff
@@ -70,6 +70,16 @@ public int FetchingPostponedInMilliseconds
...
                     _fetchDispatcher.DataSource = _dataSource;
+                    
+                    if (_dataSource is IDynamic dynamic)
+                        dynamic.DataChanged += (s, e) => {
+                            // Todo: Make this work.
+                            // The Problem: DataHasChanged signals the Map that the data has
+                            // changed within the layer and triggers a RefreshGraphics. But that
+                            // is not what is needed here. The data has changed in the provider
+                            // and a RefreshData call is needed.
+                            DataHasChanged();
```
pauldendulk, line 81: The DataHasChanged() call does not have the effect I am looking for. It tells the Map that the data within the layer has changed, but it hasn't, the data in the DataSource has changed, so the Layer should fetch data from the source first. We could think of of way to signal this as well, but why is this so difficult. From my perspective the situation is simple, I want to trigger RefreshData. There should be a comprehensible way to do that.

pauldendulk (commented):
This is a failed attempt which needs a different kind of solution.

pauldendulk:
I will close this for now. This does not have my priority at the moment. Perhaps reopen later.

PR #1586, opened 2022-04-09
Use Fill.Color for colored svg

pauldendulk:
I am working on improving the some of the samples that I think users are most interested in. I think this is mostly simple pins and feature info on those pins. One other thing that I think is relevant is selecting colors for the pins that are placed. This PR allow the user of SymbolStyle.Brush.Color to determine the color of an SVG pin. 

A problem with this PR is that it changes the default behavior of adding a pin because the SymbolStyle.Brush.Color is default set to White. So If you do not remove this color the svg will be drawn in white. In most cases this is not what people want. Options:
- Explain that this is a breaking change and users need to set the Brush.Color to null (current solutions)
- Make the SymbolStyle.Brush default null. This is also a breaking change.
- Introduce a new kind of style for this. This may be confusing for starting users. Which style to use?
- Add a new field SymbolStyle.ImageColor
- Generate or modify svg images. Perhaps we can set the color in the svg itself.

### DIFF
--- a/Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
+++ b/Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
@@ -91,7 +91,7 @@ namespace Mapsui.Rendering.Skia
                         (float)destX, (float)destY,
                         rotation,
                         (float)offsetX, (float)offsetY,
-                        opacity: opacity, scale: (float)symbolStyle.SymbolScale, color: symbolStyle.Fill?.Color);
+                        opacity: opacity, scale: (float)symbolStyle.SymbolScale, color: symbolStyle.BlendModeColor);
                     break;
                 case BitmapType.Svg:
                     // Todo: Perhaps remove BitmapType.Svg and SvgRenderer?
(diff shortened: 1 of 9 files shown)


pauldendulk:
Decided to go with a new field ColorBlendMode, so there are no breaking changes. Not sure if it is a great name. Just based on the named in skia. Perhaps we need to add a BlendMode type as well at some point. In general I don't think it is bad to stay close to skia. 

I was also thinking about experimenting with a SkiaStyleStyle where you could use the skia SKPaint so there is no translation whatsoever, and users could be pointed to the skia documentation to know learn how it works. May also be an advantage for performance.

Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
```diff
@@ -91,9 +91,11 @@ public static bool DrawImage(SKCanvas canvas, IReadOnlyViewport viewport, ILayer
                         (float)destX, (float)destY,
                         rotation,
                         (float)offsetX, (float)offsetY,
-                        opacity: opacity, scale: (float)symbolStyle.SymbolScale);
+                        opacity: opacity, scale: (float)symbolStyle.SymbolScale, blendModeColor: symbolStyle.BlendModeColor);
                     break;
                 case BitmapType.Svg:
+                    // Todo: Perhaps remove BitmapType.Svg and SvgRenderer?
```
pauldendulk, line 97: @inforithmics Currently we do not seem to use the BitmapType.Svg anywhere in our tests. If we load an svg we treat it as BitmapType.Picture. Can I remove the BitmapType.Svg code?

PR #1587, opened 2022-04-10
Cleanup on renderers

pauldendulk:


Mapsui.Rendering.Skia/SkiaStyles/LineStringRenderer.cs
```diff
@@ -7,7 +7,7 @@ namespace Mapsui.Rendering.Skia
 {
     public static class LineStringRenderer
     {
-        public static void Draw(SKCanvas canvas, IReadOnlyViewport viewport, VectorStyle vectorStyle, IFeature feature,
```
pauldendulk, line 10: feature was not used

Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
```diff
@@ -1,61 +1,74 @@
...
     public class SymbolStyleRenderer : ISkiaStyleRenderer
     {
         public bool Draw(SKCanvas canvas, IReadOnlyViewport viewport, ILayer layer, IFeature feature, IStyle style, ISymbolCache symbolCache, long iteration)
         {
+            var symbolStyle = (SymbolStyle)style;
             switch (feature)
             {
                 case PointFeature pointFeature:
-                    DrawPointFeature(canvas, viewport, layer, pointFeature, (SymbolStyle)style, symbolCache, iteration);
+                    DrawXY(canvas, viewport, layer, pointFeature.Point.X, pointFeature.Point.Y, symbolStyle, symbolCache);
```
pauldendulk, line 21: DrawXY allows it to be called with both MPoint and Point without having to create a PointFeature (or a GeometryFeature as done below) which is IDisposable. Just like the async methods the IDisposable classes should be kept away from the core methods if possible.

Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
```diff
@@ -1,61 +1,74 @@
...
+                            foreach (var point in GetPoints(collection))
+                                DrawXY(canvas, viewport, layer, point.X, point.Y, symbolStyle, symbolCache);
                             break;
                         case Point point:
-                            Draw(canvas, viewport, layer, new PointFeature(point.X, point.Y), style, symbolCache, iteration);
+                            DrawXY(canvas, viewport, layer, point.X, point.Y, symbolStyle, symbolCache);
                             break;
                     }
                     break;
-                case GeometryCollection geometryFeatureCollection:
```
pauldendulk, line 33: A GeometryCollection is not derived from IFeature so this case can never happen. I would expect the compiler to complain about this.

Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
```diff
@@ -43,16 +41,12 @@ public bool Draw(SKCanvas canvas, IReadOnlyViewport viewport, ILayer layer, IFea
...
                                 break;
                             case LineString lineString:
-                                LineStringRenderer.Draw(canvas, viewport, vectorStyle, feature, lineString, opacity);
+                                LineStringRenderer.Draw(canvas, viewport, vectorStyle, lineString, opacity);
                                 break;
                             default:
                                 throw new ArgumentException("Unknown geometry of Feature");
                         }
                         break;
-                    case (GeometryCollection geometryFeatureCollection):
```
pauldendulk, line 52: again: A GeometryCollection is not derived from IFeature so this case can never happen. I would expect the compiler to complain about this.

Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
```diff
@@ -43,16 +41,12 @@ public bool Draw(SKCanvas canvas, IReadOnlyViewport viewport, ILayer layer, IFea
...
-                        for (var i = 0; i < geometryFeatureCollection.NumGeometries; i++)
-                            Draw(canvas, viewport, layer, new GeometryFeature(geometryFeatureCollection.GetGeometryN(i)), style, symbolCache, iteration);
-                        break;
                 }
             }
             catch (Exception ex)
```
pauldendulk, line 52: You can add a try catch in every individual method, just to be sure, but it also adds clutter. I propose that the MapRenderer itself is responsible for catching and logging the error and that all rendering have no catches. They could have a try finally if things need to be disposed.

PR #1588, opened 2022-04-13
Upgrade to MAUI RC1

pauldendulk:


inforithmics:
Skia Sharp is now updated so everything should be now Rc1 compatible.

Maybe we should wait for the Release of 6.0.300 so that the prevew versions could be removed.

https://github.com/dotnet/installer Here there are already 6.0.300 Installers available so this shouldn't take to long.

PR #1600, opened 2022-05-10
Remove DataSource from MemoryLayer

pauldendulk:
This change has been suggested elsewhere. MemoryLayer called the DataSource directly from the GetFeatures method. The GetFeatures methods is called in the render loop and should only return cached data. The current solution compiles and the tests are green but it needs to get a closer look.


inforithmics:
I had a quick lock at it and I like the general architecture, makes much more sense now. The MemoryLayer has now no DataSource only the Items. And the Layer can be used as the MemoryLayer with a DataSource. 

The build failure could be that a new update for skiashapr is needed for maui.rc3 which is not yet available. maybe tommorrow.

pauldendulk:
> The build failure could be that a new update for skiashapr is needed for maui.rc3 which is not yet available. maybe tommorrow.

The windows github action build is now fixed but updating a package. The appveyor build is still failing on 'nuget restore mapsui.sln', while we also do that in the windows github action, but there is succeeds🤔.  Perhaps nuget itself should be updated?

pauldendulk:
> > The build failure could be that a new update for skiashapr is needed for maui.rc3 which is not yet available. maybe tommorrow.
> 
> The windows github action build is now fixed but updating a package. The appveyor build is still failing on 'nuget restore mapsui.sln', while we also do that in the windows github action, but there is succeeds🤔. Perhaps nuget itself should be updated?

Perhaps this is a good moment to figure out what we still need from appveyor, add it to github actions and remove appveyor. What I can think of right now is that we generate the nugets of every build (I think we already publish these from actions as well, but I need to update the documentation, and perhaps we need to change something in the version number) and generating the docs using docfx.

inforithmics:
appveyor build runs the unit tests and compiles the test projects this needs to be enhanced.

pauldendulk (commented):
Added comment comments to my own PR. Some points could be picked up in a later PR, like how to work with coordinate projection. I will changes something in the MemoryLayer.Extent logic.

Mapsui/Layers/MemoryLayer.cs
```diff
@@ -50,6 +51,20 @@ public override void RefreshData(FetchInfo fetchInfo)
             // DataHasChanged should be called.
         }
 
-        public override MRect? Extent => DataSource?.GetExtent();
+        public override MRect? Extent
+        {
+            get
+            {
+                MRect? result = null;
```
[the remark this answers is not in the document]
pauldendulk: I did spend some time looking into this and decided to leave it as it was. So the current implementation is not great for performance. 

There are two alternatives:
- Leave setting the extent up to the user entirely. This is clear, but works different in other layers and is a big breaking change.
- Update the extent for the most commons scenarios but not ub all scenarios. This will cause come confusion.

pauldendulk:
I removed the WIP and propose to merge it as it is.

A following MR will replace the IProvider.GetFeatures with GetFeaturesAsync everywhere.

Tests/Mapsui.Tests.Common/Maps/ProjectionSample.cs
```diff
@@ -61,19 +60,9 @@ public static Map CreateMap()
...
-
-            var dataSource = new ProjectingProvider(memoryProvider)
-            {
-                CRS = "EPSG:3857"
-            };
-
             return new MemoryLayer
             {
-                DataSource = dataSource,
+                Features = features.Project("EPSG:4326", "EPSG:3857"),
```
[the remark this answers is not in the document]
inforithmics: In the old memory layer the CRS was in the DataSource, now with the new Memory Layer the CRS of the data is not so clear. Maybe a wrapper should be made where the CRS field is set. So that when a projection is needed it still can be done.
This is only relevant in a scenario when the same memory layer is attached to different Maps with differen crs. (I don't do this. but it could be done and before this would have worked)

Mapsui.UI.Forms/MapView.cs
```diff
@@ -45,9 +45,9 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
         private const string CalloutLayerName = "Callouts";
         private const string PinLayerName = "Pins";
         private const string DrawableLayerName = "Drawables";
-        private readonly MemoryLayer _mapCalloutLayer;
```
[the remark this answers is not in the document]
inforithmics: Maybe this is a question for @charlenni

Mapsui.UI.WinUI/Mapsui.UI.WinUI.csproj
```diff
@@ -18,7 +18,7 @@
 
   <ItemGroup>
 	<PackageReference Include="Microsoft.WindowsAppSDK" Version="1.0.3" GeneratePathProperty="true" />
-	<PackageReference Include="Microsoft.Graphics.Win2D" Version="1.0.1" />
```
inforithmics, line 21: Good catch is needed for Maui Rc3 (WinUI ...)

Mapsui/Extensions/FeatureExtensions.cs
```diff
@@ -46,5 +47,19 @@ public static string ToDisplayText(this IEnumerable<KeyValuePair<string, IEnumer
             }
             return result.ToString();
         }
+
+        public static IEnumerable<IFeature> Project(this IEnumerable<IFeature> features, string? fromCRS,
```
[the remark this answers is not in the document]
inforithmics: Yes this is a nice solution for Projecting with out the need of Layers .. Providers

Mapsui/Layers/MemoryLayer.cs
```diff
@@ -1,45 +1,42 @@
...
+        public MemoryLayer() : this(nameof(MemoryLayer)) { }
 
         /// <summary>
         /// Create layer with name
         /// </summary>
         /// <param name="layerName">Name to use for layer</param>
         public MemoryLayer(string layerName) : base(layerName) { }
 
-        public IProvider<IFeature>? DataSource { get; set; }
+        public IEnumerable<IFeature> Features { get; set; } = new List<IFeature>();
```
inforithmics, line 26: Was the Code in the Rasterizing Tile Layer Improved to read  this Features too? Or the RasterizingLayer for the Unit tests.

Mapsui/Layers/MemoryLayer.cs
```diff
@@ -1,45 +1,42 @@
...
+            if (rect == null) { return new List<IFeature>(); }
 
-            var biggerBox = box.Grow(
+            var biggerRect = rect.Grow(
                     SymbolStyle.DefaultWidth * 2 * resolution,
                     SymbolStyle.DefaultHeight * 2 * resolution);
-            var fetchInfo = new FetchInfo(biggerBox, resolution, CRS);
 
-            return DataSource?.GetFeatures(fetchInfo) ?? Array.Empty<IFeature>();
+            return Features.Where(f => f.Extent?.Intersects(biggerRect) == true);
```
inforithmics, line 39: good solution for fetching the data

Mapsui/Layers/MemoryLayer.cs
inforithmics: Ah I see the GetFeatures is used so this would still work.

Mapsui/Layers/MemoryLayer.cs
```diff
@@ -50,6 +47,6 @@ public override void RefreshData(FetchInfo fetchInfo)
             // DataHasChanged should be called.
         }
 
-        public override MRect? Extent => DataSource?.GetExtent();
+        public override MRect? Extent => Features.GetExtent();
```
inforithmics, line 50: Could this maybe somhow be cached, Because I don't think The Features would change. Or is the Intention to make them Changeable. Like in the Pull Request with the Dynamic Memory Source for Memory Layer?

Mapsui/Providers/ProjectingProvider.cs
```diff
@@ -15,7 +15,7 @@ public class ProjectingProvider : IAsyncProvider<IFeature>, IProvider<IFeature>
         public ProjectingProvider(IProviderBase provider, IProjection? projection = null)
         {
             _provider = provider;
-            _projection = projection ?? new Projection();
+            _projection = projection ?? ProjectionDefaults.Projection;
```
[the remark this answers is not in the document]
inforithmics: I agree

Mapsui/Providers/ProjectingProvider.cs
```diff
@@ -30,41 +30,19 @@ public async IAsyncEnumerable<IFeature> GetFeaturesAsync(FetchInfo fetchInfo)
...
-            }
-            
-            if (!CrsHelper.IsCrsProvided(_provider.CRS, CRS))
-                throw new NotSupportedException($"CRS is not provided. From CRS: {_provider.CRS}. To CRS {CRS}");
-
-            var copiedFeatures = features.Copy().ToList();
-            _projection.Project(_provider.CRS, CRS, copiedFeatures);
-            foreach (var it in copiedFeatures)
-                yield return it;
+            return features.Project(_provider.CRS, CRS, _projection);
```
inforithmics, line 45: Good that only one place makes the projection.

Samples/Mapsui.Samples.Common.Desktop/ThemeStyleSample.cs
```diff
@@ -90,7 +90,7 @@ public static ILayer CreateCityHoverPoints()
 
             return new MemoryLayer
             {
-                DataSource = new MemoryProvider<IFeature>(features),
+                Features = features,
```
inforithmics, line 93: All samples with MemoryLayer need now to be fixed. I don't comment the other fixes because they are obvious.

Samples/Mapsui.Samples.Common/Maps/Symbols/AtlasSample.cs
```diff
@@ -42,18 +42,12 @@ private static ILayer CreateAtlasLayer(MRect? envelope)
...
             {
                 Name = AtlasLayerName,
-                DataSource = CreateMemoryProviderWithDiverseSymbols(envelope, 1000),
+                Features = CreateAtlasFeatures(RandomPointGenerator.GenerateRandomPoints(envelope, 1000)),
                 Style = null,
                 IsMapInfoLayer = true
             };
         }
 
-        public static MemoryProvider<IFeature> CreateMemoryProviderWithDiverseSymbols(MRect? envelope, int count = 100)
```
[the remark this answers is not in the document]
inforithmics: Make it as simple as possible but not simpler.

Tests/Mapsui.Tests.Common/TestTools/TestLayer.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Mapsui.Tests.Common.TestTools
+{
+    /// <summary>
+    /// This layer calls the DataSource directly from the GetFeatures method. This should be avoided
+    /// in a real application because the GetFeatures methods is called in the rendering loop. For 
+    /// testing this can be useful when wnat to generate an image from the data source without 
+    /// having to wait for the asynchronous data fetch call.
+    /// </summary>
+    public class TestLayer : BaseLayer, ILayerDataSource<IProvider<IFeature>>
```
inforithmics, line 19: Good Addtion for Testing, makes things more reliable.

inforithmics (approved):

Mapsui/Layers/MemoryLayer.cs
pauldendulk: I tried to find some solution for this but could not find one that was good enough. The extent can only be cached if the features can not be mutated or there is an event when they are mutated, but this can not be enforced. The Features field could be reassigned or even coordinates of individual features can be set. The most honest solution would be to make it fully the responsibility of the developer, but that would be too much of a breaking change.

inforithmics: This is okay for me. Maybe in a later improvement The Features could become an observableCollection. And listing to changes of the Collection would recalculate the extent.

inforithmics: https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.observablecollection-1?view=net-6.0

Samples/Mapsui.Samples.Common.Desktop/ThemeStyleSample.cs
pauldendulk: I went through all samples and they work as in master. There a few samples that don't work but that is also in master. Those are:
- WMTS (Could be the data source)
- WMTS Michelin (Could be the data source)
- WMS (Error: The current SynchronizationContext may not be used as a TaskScheduler. This can be fixed by removing the TaskScheduler, which is what I intend to do.)
- Geo Tiff (I thought this one was fixed not too long ago).
- Keep Within Extent (I broke this with my work on animations).
- Atlas sample (The sample, but the test works properly).

Mapsui/Layers/MemoryLayer.cs
pauldendulk: I see the ObservableCollection as something that is useful for a limited number of items in specific context where you know roughly the number of items you have to deal with. Even in such a context I have struggled with performance problems with ObservableCollection. It triggers many updates and it is how to track what is going on. For our purposes we need more control.

Samples/Mapsui.Samples.Common.Desktop/ThemeStyleSample.cs
inforithmics: When I added the Geo Tiff sample it worked, I'll have a look at it.

inforithmics: The atlas sample can be fixed by using the png from the test (I forgot to copy the fixed png to the atlas sample)

Mapsui/Layers/MemoryLayer.cs
pauldendulk: An alternative could be the Channel
https://docs.microsoft.com/en-us/dotnet/api/system.threading.channels.channel-1?view=net-6.0
It has more options to regulate the flow of notifications.

Samples/Mapsui.Samples.Common.Desktop/ThemeStyleSample.cs
inforithmics: I had a quick look at it and the Problem is that the RasterFeature has no Default Style attached and therefre it isn't displayed.

In The GeoTiffProvider follwoing has to be changed:

from:
`_feature.Styles.Add(new VectorStyle());`
to:
`_feature.Styles.Add(new RasterStyle());`

Then it works because the Vectorstyle does not know how to render a RasterImage so a RasterStyle should be added.

The same Problem occours in WMTS/WMSProvider:

So the question is should in the Constructor of the RasterFeature a RasterStyle be attached or should it be done in the Providers when a RatsterFeature is created?

inforithmics: I looked at the WMTSSample Brutile doesn't deliver no Tiles so that's why it doesn't work maybe the webservice access was changed.

inforithmics: I found out why WMTS and WMTSMichelin did not work. The reason was changes for the Unit Testing to work.

fail:
`tileSource.PersistentCache = Michelin.DefaultCache;`

work:

 ```
 if (Michelin.DefaultCache != null)
{
    tileSource.PersistentCache = Michelin.DefaultCache;
}
```

The reason i set the PersistentCache if it is null and then it crashes Brutile because it is not checked for null.

The Same here:

not work:
`nature2000TileSource.PersistentCache = DefaultCache;`

works:
```
if (DefaultCache != null)
{
    nature2000TileSource.PersistentCache = DefaultCache;
}
```

Added the changes in this pull request: https://github.com/Mapsui/Mapsui/pull/1601

Mapsui/Layers/MemoryLayer.cs
inforithmics: Intresting I didn't know the Channel Type. Regarding the Observable Collection the Performance Trick is to Trigger Correct changes. So that when you add 2 Elements only an Add 2 Elements Event is raised instead of a whole ObservableCollection Refresh. And when it's done this way it can hold thousends of elements and display it in a Grid for example. But Maybe in a Map this is different.

inforithmics: And in MapView ObservableRangeCollection is used for drawables and pins and callouts.

PR #1603, opened 2022-05-18
Replace all remaining IProvider.GetFeatures with GetFeatureAsync

pauldendulk:
And renamed IAsyncProvider back to IProvider so wrt beta.3 only the method name has changed.


inforithmics:
I agree with you on the 5 points. 
Especially IAsyncEnumerable<IFeature> replacing it with Task<IEnumerable<IFeature>>

PR #1604, opened 2022-05-18
Replace T in IProvider with IFeature

pauldendulk:
Only one cast was needed in the entire code base so the generic IProvider did not provide enough added value.

### DIFF
--- a/Mapsui/Layers/AnimatedLayers/AnimatedPointLayer.cs
+++ b/Mapsui/Layers/AnimatedLayers/AnimatedPointLayer.cs
@@ -6,13 +6,13 @@ using Mapsui.Providers;
 
 namespace Mapsui.Layers.AnimatedLayers;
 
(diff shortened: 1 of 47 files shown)


Mapsui/Layers/AnimatedLayers/AnimatedPointLayer.cs
```diff
@@ -28,7 +28,7 @@ public async Task UpdateDataAsync()
         if (_fetchInfo is null) return;
         if (_dataSource is null) return;
         var features = await _dataSource.GetFeaturesAsync(_fetchInfo).ToListAsync();
-        _animatedFeatures.AddFeatures(features);
+        _animatedFeatures.AddFeatures(features.Cast<PointFeature>());
```
pauldendulk, line 31: This is the only Cast that was needed.

pauldendulk (commented):
There we many (perhaps one hundred) templated provider but only one place where a cast could be avoided.

inforithmics:
Good work.

PR #1605, opened 2022-05-18
Move RefreshData to IAsyncDataFetcher

pauldendulk:
The purpose of RefreshData is to fetch new data from the source of the layer (This source could be an IProvider or ITileSource), so it does not apply to all layers and should be in a more specific interface. This allowed to remove some implementations that made no sense.

### DIFF
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -623,16 +623,6 @@ namespace Mapsui.UI.Wpf
             Refresh();
         }
 
-        /// <summary>
-        /// Clear cache and repaint map
-        /// </summary>
-        public void Clear()
-        {
-            // not sure if we need this method
-            _map?.ClearCache();
-            RefreshGraphics();
-        }
-
         private void CommonDispose(bool disposing)
         {
             if (disposing)
--- a/Mapsui/Layers/MemoryLayer.cs
+++ b/Mapsui/Layers/MemoryLayer.cs
@@ -39,14 +39,6 @@ namespace Mapsui.Layers
             return Features.Where(f => f.Extent?.Intersects(biggerRect) == true);
         }
 
-        public override void RefreshData(FetchInfo fetchInfo)
-        {
-            // RefreshData needs no implementation for the MemoryLayer. Calling OnDataChanged here
-            // would trigger an extra needless render iteration.
-            // If a user changed the data in the provider and needs to update the graphics
-            // DataHasChanged should be called.
-        }
-
         public override MRect? Extent => Features.GetExtent();
     }
 }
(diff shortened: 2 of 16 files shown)


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -623,16 +623,6 @@ private void SetViewportSize()
             Refresh();
         }
 
-        /// <summary>
-        /// Clear cache and repaint map
-        /// </summary>
-        public void Clear()
```
pauldendulk, line 629: Removed this method because I don't know what the purpose is. If you remove the cache and refresh the graphics the layers that have a cache will show no data, but as soon as you start panning (or trigger RefreshData some other way) there will be features visible again.

Mapsui/Layers/MemoryLayer.cs
```diff
@@ -39,14 +39,6 @@ public override IEnumerable<IFeature> GetFeatures(MRect? rect, double resolution
             return Features.Where(f => f.Extent?.Intersects(biggerRect) == true);
         }
 
-        public override void RefreshData(FetchInfo fetchInfo)
```
pauldendulk, line 42: This method can now be removed.

Mapsui/Layers/WritableLayer.cs
```diff
@@ -42,11 +42,6 @@ public override IEnumerable<IFeature> GetFeatures(MRect? box, double resolution)
 
         public override MRect? Extent => GetExtent();
 
-        public override void RefreshData(FetchInfo fetchInfo)
```
pauldendulk, line 45: This method can now be removed.

Samples/Mapsui.Samples.Wpf.Editing/Layers/VertexOnlyLayer.cs
```diff
@@ -34,10 +34,5 @@ public override IEnumerable<IFeature> GetFeatures(MRect box, double resolution)
                     }
             }
         }
-
-        public override void RefreshData(FetchInfo fetchInfo)
```
pauldendulk, line 38: This method can now be removed.

Tests/Mapsui.Tests.Common/TestTools/TestLayer.cs
```diff
@@ -42,10 +42,5 @@ public override IEnumerable<IFeature> GetFeatures(MRect box, double resolution)
         public override MRect? Extent => DataSource?.GetExtent();
 
         IProvider? ILayerDataSource<IProvider>.DataSource => throw new NotImplementedException();
-
-        public override void RefreshData(FetchInfo fetchInfo)
```
pauldendulk, line 46: This method can now be removed.

PR #1617, opened 2022-05-26
Add MapControl with Map to Maui sample

pauldendulk:


Mapsui.UI.Forms/MapControl.cs
```diff
@@ -39,7 +39,13 @@ namespace Mapsui.UI.Forms
     /// </summary>
     public partial class MapControl : ContentView, IMapControl, IDisposable
     {
+#if __MAUI__
+        // GPU does not work currently on MAUI
```
pauldendulk, line 43: Moved setting the default this to a compiler directive so that app developer does not have to take this into account.

Samples/Mapsui.Samples.Maui/MauiProgram.cs
```diff
@@ -7,6 +9,7 @@ public static MauiApp CreateMauiApp()
             var builder = MauiApp.CreateBuilder();
             builder
                 .UseMauiApp<App>()
+                .UseSkiaSharp()
```
pauldendulk, line 12: Took me a while before I figures this out. The app developer will also have to do this. I do not know what happens in this method but perhaps we could call this logic in our MapControl constructor somehow.

PR #1621, opened 2022-05-29
Use Central Package Management

inforithmics:
This Pull request Centralises all Nuget Versions in the Directory.Packages.props File:

see here:

https://devblogs.microsoft.com/nuget/introducing-central-package-management/

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -27,17 +27,17 @@
 
   <ItemGroup>
     <!--Specify Compiler for Mac Builds-->
-    <PackageReference Include="Microsoft.Net.Compilers.Toolset" Version="4.2.0">
+    <PackageReference Include="Microsoft.Net.Compilers.Toolset">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
     <!--Disposable Analyzer-->
-    <PackageReference Include="IDisposableAnalyzers" Version="4.0.2">
+    <PackageReference Include="IDisposableAnalyzers">
         <PrivateAssets>all</PrivateAssets>
         <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
     <!--Threading Analyzer mainly Async Code-->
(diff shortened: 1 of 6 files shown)


inforithmics:
This pull request had the nice side effect that instead of several Versions of a nuget package only one is used (The one specified in the Directory.Packages.props). So this reduced the Downloaded nuget packages from 3.5GB to around 2GB.

I fixed also some Package Inconsistency that came to light with this Pullrequest.

When I had to upgrade Versions I tried to upgrade to the minimal next higher Versions so that it would compile.

For Eto (Sample) I had to make following workaround.

```
<ItemGroup>
    <!-- Explicitly reference Linux native dlls to get them included -->
    <PackageReference Include="SkiaSharp.NativeAssets.Linux" />
    <PackageReference Include="Eto.Forms" />
  </ItemGroup>

  <ItemGroup Condition="'$(BuildPlatform)'=='Wpf'">
    <PackageReference Include="Eto.Platform.Wpf" />
  </ItemGroup>
```
The References of Eto.Platform.Wpf and Eto.Forms are not needet, But with the Central Package Managment they need to specified or else it won't compile. Currently I only specified the Additional References for WPF for the other Build Platforms it has to be done like for Wpf with the Correct Eto.Platform.... Nuget package.

Maybe @rafntor Could help to specify the other platforms beside Wpf.

inforithmics:
With this Checkin the Versions need only to be changed in two places. Directory.Packages.props and nuspec Files.

rafntor:
The build-platform is auto-detected based on some rules in ['Sdk.targets'](https://github.com/picoe/Eto/blob/226778b2609870b424d1db7ace05cfb2ae1cfe12/src/Eto/sdk/Sdk.targets#L10).

I think it will be fine if you include these ;
"Eto.Platform.Wpf" - "Eto.Platform.Gtk" - "Eto.Platform.Mac64"
That should allow 'dotnet build' to execute and produce an executable on each platform if all works out.
Does it work to add also these to 'Directory.Packages.props' instead of Eto.Sample ?

rafntor:
I made a PR .. please have a look ;-)

pauldendulk (approved):
Excellent! This is a huge improvement. Dealing with package versions was a major hassle, this will make it so much easier.

PR #1624, opened 2022-06-01
Fix last few warnings in Mapsui.Tiling

pauldendulk:


.editorconfig
```diff
@@ -11,6 +11,26 @@ root = true
 #use soft tabs (spaces) for indentation
 indent_style = space
 indent_size = 4
+csharp_indent_labels = one_less_than_current
```
pauldendulk, line 14: Apparently defaulting to var was in the IDE, not in our .editorconfig

pauldendulk:
There are actually not that many warnings left in the project (about 24). Most are wrt nuget packages, so lets first merge your other PR and then fix what remains. We are close.

PR #1627, opened 2022-06-01
Add SkiaSharp.Views.Forms.WPF Directory.Package.props

pauldendulk:
I know nothing about Directory.Package.props but got a compile error on SkiaSharp.Views.Forms.WPF and since it was missing from Directory.Package.props I added it there.


pauldendulk:
This was the error:
```console
NU1010	The PackageReference items SkiaSharp.Views.Forms.WPF do not have corresponding PackageVersion
```

inforithmics:
This is the correct solution when this error occours

inforithmics (approved):

PR #1630, opened 2022-06-06
Fix Central Package Management Warnings in Visual Studio 2022 Preview for Uno

inforithmics:
Ignore NU1801 Central Package Management Warnings in Visual Studio 2022 Preview


inforithmics:
Main changes.

1) Fixed Uno Projects they compile and work now again.
2) Fixed tons of DuplicatePackage References
3) Fixed missing Version number in SkiaSharp.Views.WPF
4) I updated global.json to the dotnet Version of 6.0.400-preview.22301.10 (So Visual Studio 2022 Preview is needed for working with Mapsui until Visual Studio 2022 17.3 is released)
5) Fixed Mapsui so that it works with the newest Visual Studio 2022 Preview

inforithmics:
8) Enabled Mapsui.Uno and Mapsui.WinUI

inforithmics:
Maybe https://github.com/Mapsui/Mapsui/pull/1644 can be merged first, because it does not introduce an Preview Sdk.

Directory.Build.props
```diff
@@ -18,6 +18,7 @@
 	  <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
 	  <LangVersion>10.0</LangVersion>
 	  <Nullable>enable</Nullable>
+    <NoWarn>$(NoWarn);NU1008;NU1701</NoWarn>
```
pauldendulk, line 21: 👍Putting the NoWarn in the Directory.Build.props can make the other files less cluttered with NoWarns.

Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Skia.Wpf/Mapsui.Samples.Uno.Skia.Wpf.csproj
```diff
@@ -5,16 +5,10 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" />
-    <PackageReference Include="Uno.UI.Skia.Wpf" />
+    <PackageReference Include="Uno.UI.Skia.Wpf" NoWarn="NU1701" />
```
pauldendulk, line 8: Can this 'NoWarn="NU1701' can now be removed because it is already in the Directory.Build.props? There are 3 other cases of 'NoWarn="NU1701'.

Samples/Mapsui.Samples.Uno/Directory.Build.props
```diff
@@ -0,0 +1,6 @@
+<Project ToolsVersion="15.0">
```
pauldendulk, line 1: I did not know you could have several Directory.Build.props at different levels in the hierarchy.

Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Mobile/Mapsui.Samples.Uno.Mobile.csproj
```diff
@@ -25,19 +25,46 @@
...
-  <ItemGroup Condition="'$(TargetFramework)'=='net6.0-ios' or '$(TargetFramework)'=='net6.0-maccatalyst'">
-    <PackageReference Include="Uno.Extensions.Logging.OSLog" />
-  </ItemGroup>
-  <PropertyGroup Condition="'$(TargetFramework)'=='net6.0-maccatalyst'">
-    <!-- Required for unknown crash as of .NET 6 Mobile Preview 13 -->
-    <MtouchExtraArgs>--registrar:static</MtouchExtraArgs>
-    <!-- Full globalization is required for Uno -->
-    <InvariantGlobalization>false</InvariantGlobalization>
-  </PropertyGroup>
+ <Choose>
```
pauldendulk, line 28: Impressive, advanced csproj-ing. The fact that we need this kind of logic is worrying though, It means this is harder to maintain. Perhaps we can ask the Uno team how to go about there problems that are solved here.

pauldendulk (approved):
Great! I will merge this but please take a look at the remarks.

Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Mobile/Mapsui.Samples.Uno.Mobile.csproj
inforithmics: This is from the default uno template I simply updated it to stay compliant

jeromelaban: Most of these options are indeed needed for production apps, as of .NET 6.0.300 at least. I agree that it makes it harder to maintain, but that's what a "single project" means... :)

pauldendulk: @jeromelaban Thanks for your feedback. @inforithmics did all of the work on Uno so far and I am still trying to understand it. You mention 'single project', but apart from Mapsui.Samples.Uno.Mobile.csproj we have quite a few other projects in our solution. As a maintainer I would like to reduce that number. Do al still have an added value?

image

jeromelaban: Yes they do. The .NET team has "single-project" in mind because the targets they have works with what .NET provides, but for the other targets (Gtk, Wpf, Framebuffer, Wasm, UWP), those cannot be merged into one. The only upcoming changes is the Wpf and Wpf.Host that will be mergeable into one project (but still not as a single project).

pauldendulk:
@inforithmics I would love to release Uno but I still get an error if I build from the commend line:

> C:\Users\pauld\.nuget\packages\msbuild.sdk.extras\3.0.44\Build\Workarounds.targets(27,5): error : If you are building p
> rojects that require targets from full MSBuild or MSBuildFrameworkToolsPath, you need to use desktop msbuild ('msbuild.
> exe') instead of 'dotnet build' or 'dotnet msbuild' [c:\code\github\Mapsui\Mapsui.UI.Uno\Mapsui.UI.Uno.csproj]

And from Visual Studio:

> Severity	Code	Description	Project	File	Line	Suppression State
Error	NU1008	Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items: Uno.UI.RemoteControl;Xamarin.Google.Android.Material;Microsoft.VisualStudio.Threading.Analyzers;Microsoft.Net.Compilers.Toolset;IDisposableAnalyzers;SkiaSharp.Views.Uno;Uno.UI;Uno.UniversalImageLoader;Uno.UI.Adapter.Microsoft.Extensions.Logging;Microsoft.Extensions.Logging.Console;Microsoft.Extensions.Logging.	Mapsui.Samples.Uno.Mobile	C:\code\github\Mapsui\Samples\Mapsui.Samples.Uno\Mapsui.Samples.Uno.Mobile\Mapsui.Samples.Uno.Mobile.csproj	1

jeromelaban:
@pauldendulk for the first issue, this is not a behavior we can fix. On Windows, to build the current set of targets (Xamarin and UWP) you'll need to use `msbuild` as `dotnet build` is not supported by Microsoft. This changes with the `net6.0` targets (what's created by the `dotnet new unoapp-net6` and `dotnet new unoapp-winui-net6` project templates), where everything can target `dotnet build`.

For the second issue, this is is a nuget preview feature that we have not tested and I would guess does not work with "legacy" projects like Xamarin and UWP. If you have a branch/repro we can look at we'll see if there's a workaround.

inforithmics:
@pauldendulk  do you use buildpack.bat for building the nugets? If yes then I can fix it with the workarounds I use for passing the build.

pauldendulk:
Thanks a lot @jeromelaban, I slowly start to understand a bit more.

@inforithmics I see you use msbuild in the dotnet.yml, I guess that is related to the workaround. But I was also thinking if this could be solved by replacing the UWP target in the csproj with a WinUI one. We replaced support for regular UWP in Mapsui with WinUI, so it makes sense to do the same for Uno UWP and Uno WinUI. ... but perhaps I do not fully understand the Uno situation. Or is Uno WinUI not ready yet.

inforithmics:
1) Currently Mapsui.Uno is based on the Uno (UWP Implementation because until recently there was no SkiaSharp.Uno.WinUI implementation).

nuget package:
https://www.nuget.org/packages/SkiaSharp.Views.Uno.WinUI/2.88.1-preview.71

There are two Uno implementations Uno.UI (UWP based) and Uno.WinUI (WinUI) based. 

Currently Uno.Ui is downloaded roughly twice as often as Uno.WinUI but I think this will change. 
I have added the Mapsui.Samples.Uno.UWP for debugging purposes because at the beginning not much worked. But I think this can safely now removed. (Because the other targets work now).

inforithmics:
But I think I could easily add Mapsui.Uno.WinUI package and project and one samples project for it.

jeromelaban:
> There are two Uno implementations Uno.UI (UWP based) and Uno.WinUI (WinUI) based.

This is correct. Both implementations have different packages like the rest of the winui/uwp ecosystem split.

> Currently Uno.Ui is downloaded roughly twice as often as Uno.WinUI but I think this will change.

Indeed. We recently switched our default template to use Uno.WinUI as WinAppSDK is getting lots more stable and featureful, but both are still going to be used for a while.

> I think for supporting Uno.WinUI there needs to be a Mapsui.Uno.WinUI nuget package. But I think this can be done only with adding more project files without adding code files. (Only guess I didn't investigate).

That's the way to go indeed, so that it stays in line with what developers expect. You're going to mainly reference the same files in the 2 projects. They (`Mapsui.Uno` and `Mapsui.Uno.WinUI`) can't really be merged as visual studio can't handle such a split (you would not be able to build both without changing some properties).

inforithmics:
I'll make another pull request to fix this. But I'll wait until the Mac Fixes pull request gets first merged. Not to introduce merge confilicts again.

inforithmics:
I think I know what the Problem is the restore has to be done with .nuget/nuget restore Mapsui.sln and then it compiles and works. (Is done when executing the nuget.restore.cmd

I'll think I can call in the prebuild project event nuget.restore.cmd for the projects that don't work and then it should work on commandline and in Ide. (The main problem is that depending on call not the same nuget version is called). and in the .nuget folder nuget 6.2.1 is used which always works.

inforithmics:
In this pull request it should work now

https://github.com/Mapsui/Mapsui/pull/1646

inforithmics:
> For the second issue, this is is a nuget preview feature that we have not tested and I would guess does not work with "legacy" projects like Xamarin and UWP. If you have a branch/repro we can look at we'll see if there's a workaround.

with nuget 6.2.1 central package management is out of preview unfortunately in msbuild a nuget version is included with the central package management still in preview.

pauldendulk:
@inforithmics @jeromelaban Alright, I guess the only logical thing to do in the current situation is to support both Mapsui.Uno and Mapsui.Uno.WinUI. That means another nuget package to publish. I try to reduce the number of things we need to support all the time, but rarely succeed 🙄, but it just makes sense.

PR #1644, opened 2022-06-16
Mapsui.Mac.Legacy.sln and Mapsui.Mac.slnf (Rider/Fixed Visual Studio for Mac)

inforithmics:


inforithmics:
This works now in Visual Studio for Mac 2022 Preview with following setting made in the Console.

launchctl setenv MSBuildEnableWorkloadResolver false

I have to investaget why the windows build failed

inforithmics:
I think this has to be first merged.
https://github.com/Mapsui/Mapsui/pull/1630

Or else I have to do certain fixes again which would cause merge confilicts

inforithmics:
I renamed the Mapsui.Mac.Legacy to Mapsui.Mac.sln (should work with the Current Visual Studio Mac Version (With workaround launchctl setenv MSBuildEnableWorkloadResolver false)

And created a Mapsui.Mac.slnf (That filters all Windows Projects from Mapsui.Mac.sln (But for this a Fixed Visual Studio for Mac is necessary).

pauldendulk:
> I think this has to be first merged. #1630
> 
> Or else I have to do certain fixes again which would cause merge confilicts

I merged that one first and then fixed a merge conflict in the browser, hope that is correct https://github.com/Mapsui/Mapsui/pull/1644/commits/ef36178df0b8b3424914916c69716a4b167d5550

pauldendulk:
This PR is okay with me, but I can hardly judge it. @janusw could you review this? Perhaps you could check out the branch to see if it fixes you problem.

inforithmics:
What I saw was that some Forms projects where disabled I’ll fix them for GitHub Actions.
Because I found a workaround.

inforithmics:
The merge conflict fix is fine.

janusw (commented):
Thanks for looking into this, @inforithmics. I just tried the branch, with the following results:

- Mapsui.Mac.sln works well for me (restoring and building) with VS4Mac 17.3 build 1028.
- With Mapsui.Mac.slnf I get the following warnings/errors when restoring. Both come form Mapsui.UI.Uno, which is not included in Mapsui.Mac.sln on master.
```
/Applications/Visual Studio (Preview).app/Contents/MonoBundle/MSBuild/Current/bin/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.DefaultItems.targets(257,5): warning NETSDK1023: A PackageReference for 'Microsoft.NETCore.UniversalWindowsPlatform' was included in your project. This package is implicitly referenced by the .NET SDK and you do not typically need to reference it from your project. For more information, see https://aka.ms/sdkimplicitrefs [/Users/janus/github/Mapsui/Mapsui.UI.Uno/Mapsui.UI.Uno.csproj]

/Applications/Visual Studio (Preview).app/Contents/MonoBundle/MSBuild/Current/bin/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.TargetFrameworkInference.targets(127,5): error NETSDK1045: The current .NET SDK does not support targeting .NET Core 6.0.  Either target .NET Core 5.0 or lower, or use a version of the .NET SDK that supports .NET Core 6.0. [/Users/janus/github/Mapsui/Mapsui.UI.Uno/Mapsui.UI.Uno.csproj]
```

IMHO having the same name for the .sln and .slnf can be a bit confusing. Maybe Mapsui.Mac.Legacy.sln is a better name after all (since it makes the distinction clearer)?

inforithmics:
I'll rename it to Mapsui.Mac.sln to Mapsui.Mac.Legacy.sln for clarity

inforithmics:
I'll remove the Uno Projects from the Mapsui.Mac.slnf (Because I didn't test it or build on Mac)

inforithmics:
I fixed now the unit tests and all projects build, so in my opinion this can be merged.

.github/workflows/dotnet.yml
```diff
@@ -48,8 +48,10 @@ jobs:
          include-prerelease: true    
     - name: install maui macos android ios maccatalyst wasm-tools
       run: dotnet workload install maui macos android ios maccatalyst wasm-tools
-    - name: Restore dependencies
-      run: dotnet restore Mapsui.Mac.sln   
+    - name: Restore dependencies Legacy
+      run: nuget restore Mapsui.Mac.Legacy.sln
```
janusw, line 52: Is this step actually necessary? Since restoring Mapsui.Mac.slnf works at the command line, this should already cover everything, right?

inforithmics: This is a workaround on the Mac because the restore of the slnf does not restore the forms projects. So this is a workaround until it gets fixed in the Mac Tooling. That’s one of the reasons why there is a mapsui.Mac.legacy solution

PR #1646, opened 2022-06-21
Fix uno

inforithmics:
This Fixes the Uno.Mobile Build


pauldendulk:
@inforithmics Great, I can now run Wpf.Host, UWP, Wasm and Gtk.

If we can get buildpack.bat to work we can release.

pauldendulk:
Looks like simply replacing 'dotnet build' with 'msbuild' solves the problem.

PR #1651, opened 2022-06-22
Introduce ISample.CreateMapAsync

pauldendulk:
Why: Because for MVVM we want to set the Map in the ViewModel
so we want to create the Map without a dependency on the MapControl


pauldendulk:
I did not try to migrate the tests to ISample, did not look at them at all, just because I did not have time for that today.

I think that in all other cases where IMapControlSample is used there is a reason. There are quite a few. There are probably ways to remove that need, but in some cases it will be hard. One thing I would like to do it move the Navigator and Viewport to the Map.

inforithmics:
I looket over the changes and I espacially liked the ISample interface.

PR #1652, opened 2022-06-23
MVVM support for Maui

pauldendulk:


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -444,34 +444,61 @@ public void CallHomeIfNeeded()
             }
         }
 
+#if __MAUI__
```
pauldendulk, line 447: This is now in the Shared MapControl. The original idea was that #if #else logic would be limited to the namespaces and this kind of code should be in the platform dependent MapControl. So, that is a todo.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -444,34 +444,61 @@ public void CallHomeIfNeeded()
...
+        public static readonly BindableProperty MapProperty = BindableProperty.Create(nameof(Map),
+            typeof(Map), typeof(MapControl), default(Map), defaultBindingMode: BindingMode.TwoWay, propertyChanged: MapPropertyChanged);
+
+        static void MapPropertyChanged(BindableObject bindable, object oldValue, object newValue)
+        {
+            var mapControl = (MapControl)bindable;
+            mapControl.SetMap(newValue as Map);
+        }
+
+        public Map? Map
```
pauldendulk, line 458: There is now still a problem with the underlying _map field that is used if the property is not bindable. I think the _map should be made platform dependent as well.

Samples/Mapsui.Samples.Maui/View/MainPage.cs
```diff
@@ -12,10 +13,11 @@ public class MainPage : ContentPage
...
     MapControl mapControl = new MapControl();
 
-    public MainPage()
-	{
+    public MainPage(MainViewModel mainViewModel)
+    {
         sampleCollectionView.SelectionChanged += CollectionView_SelectionChanged;
-        mapControl.Map?.Layers.Add(OpenStreetMap.CreateTileLayer());
+        BindingContext = mainViewModel;
+        mapControl.SetBinding(MapControl.MapProperty, new Binding(nameof(MainViewModel.Map)));
```
pauldendulk, line 20: Binding works with this setup. Next I would like to use ISample.GetMapAsync to select between different samples in the ViewModel.

inforithmics:
Strange the Unit Tests pass on Github WinBuild but faill on Appveyor. Maybe trigger a checkin again (Maybe HeisenBug).

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -15,6 +15,7 @@
 #pragma warning disable IDISP008 // Don't assign member with injected and created disposables
 
 #if __MAUI__
```
pauldendulk, line 17: Added functionality specific for maui in the shared file with a __MAUI__. This can make the code less readable. It is preferred to add such code in the platform specific files. In this case this would mean that all the other platform specific files would need the regular Map property. This seemed a bit too much at this point, but perhaps this would be better.

Mapsui.UI.Shared/MapControl.cs
pauldendulk: This is was I did. The _map field not only exists if __MAUI__ is false.

Samples/Mapsui.Samples.Maui/View/MainPage.cs
pauldendulk: This is now implemented. Selecting of sample is done in the ViewModel.

Samples/Mapsui.Samples.Maui/ViewModel/MainViewModel.cs
```diff
@@ -0,0 +1,68 @@
...
+        ISampleBase selectedSample;
+
+        [ObservableProperty]
+        Map? map;
+
+        public ObservableCollection<ISampleBase> Samples { get; set; } = new();
+        public ObservableCollection<string> Categories { get; } = new();
+
+        // MapControl is needed in the samples. Mapsui's design should change so this is not needed anymore.
+        public MapControl MapControl { get; set; }
```
pauldendulk, line 40: This MapControl field should not be necessary.

Mapsui.UI.Shared/MapControl.cs
inforithmics: I think this is okay because the other platforms have mvvm constructs too (winui, wpf,xamarin.forms, uno, ...)

Samples/Mapsui.Samples.Maui/View/MainPage.cs
inforithmics: Good idea to make the Map Property Bindable.

inforithmics (approved):
Nice improvement for MAUI, maybe in a later step I can update the other platforms with MVVM

PR #1657, opened 2022-07-04
(WIP) Comment Out Central Package Management

inforithmics:
The Central Package Management does not work on MAC (Visual Studio) and Uno (needs workarounds) therefore I disable it for now until the Tooling is there.

In the Next Round of Tooling updates this could be enabled again easily (With Regex Remove Version Attributes from PackageReferences) and enabling the commented out code in Directory.Packages.props .

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -28,17 +28,17 @@
 
   <ItemGroup>
     <!--Specify Compiler for Mac Builds-->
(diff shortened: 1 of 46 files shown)


inforithmics:
I'll think I will close this pull request and write a script to set the Versions from the Directory.Packages.props so that the Script can be run when the projects need versions.

inforithmics:
I'll make another pull request with a script

Issue #1653, opened 2022-06-24
Mapsui.Forms: problems on Android with 4.0.0-beta.4/5


janusw:
**Describe the bug**

In my MapView-based application, I'm seeing problems and instabilities on Android with 4.0.0-beta.4 and beta.5, while beta.3 works fine.

At first everything seems to work, but after a while the problems begin: Sometimes the Callouts are not being shown any more, or other parts of the map (Pins or Drawables) are not being rendered. Sometimes the map freezes completely and no further updates happen at all. This happens only on Android, iOS seems to be ok.

I had a look at the Mapsui log, and the only error I found was this:

```
Error: The CancellationTokenSource has been disposed. - System.ObjectDisposedException: The CancellationTokenSource has been disposed.
  at System.Threading.CancellationTokenSource.ThrowObjectDisposedException () [0x00000] in /Users/builder/jenkins/workspace/archive-mono/2020-02/android/release/external/corert/src/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs:538 
  at System.Threading.CancellationTokenSource.ThrowIfDisposed () [0x00008] in /Users/builder/jenkins/workspace/archive-mono/2020-02/android/release/external/corert/src/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs:533 
  at System.Threading.CancellationTokenSource.Cancel (System.Boolean throwOnFirstException) [0x00000] in /Users/builder/jenkins/workspace/archive-mono/2020-02/android/release/external/corert/src/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs:320 
  at System.Threading.CancellationTokenSource.Cancel () [0x00000] in /Users/builder/jenkins/workspace/archive-mono/2020-02/android/release/external/corert/src/System.Private.CoreLib/src/System/Threading/CancellationTokenSource.cs:291 
  at Mapsui.Fetcher.FetchWorker.FetchAsync (System.Threading.CancellationTokenSource cancellationTokenSource) [0x00082] in <41f3bb80db3a454e9cc79a38d1817734>:0 
```

This must have started somewhere between beta.3 and beta.4.

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy S21
 - OS: Android 12

pauldendulk:
Thanks for the wonderfully clear report and testing on specific versions. The stacktrace is informative. Looking in the history this PR could have caused the problem:
[...]

PR #1663, opened 2022-07-13
use asynclock instead of semaphore slim

inforithmics:
Semaphore Slim could deadlock

https://github.com/Mapsui/Mapsui/issues/1653#issuecomment-1166556328

so I use an AsyncLock Implementation (Unfortunately there is currently no Framework AsyncLock Implementation so I took a small one). To reduce dependencies


inforithmics:
Alternatives for the nuget package would be
1) nuget package, (more dependencies)
2) git submodule.  (more complicated project setup for contributing).
3) import the source directly (More code has to be maintained).

pauldendulk (approved):
Indeed unfortunate the NeoSmart.AsyncLock needed to be added. Also I am not sure why the lock was needed in the first place (although I might have written it myself). There is no modification of the features within this lock. Is it because of what is happening within the _dataArrived.Invoke method? Or just the put the requests in sequence so to limit the number requests to the server at the same time. Anyway, my plan is to rewrite the fetcher logic entirely, so this is okay for now.

inforithmics:
The reason was that before there was a classic lock(object) and this can deadlock in AsyncCode that's why I changed it to Semaphore slim and because this could also deadlock to the asynclock.

PR #1664, opened 2022-07-13
Log All exceptions to the Mapsui Logger add a default Trace Logger

inforithmics:
Log all exceptions on Mapsui Logger and add a default Trace Output as default logger


inforithmics:
For debugging this log all exceptions that occour:

pauldendulk (approved):
This is an important improvement. To have this errors visible can save developers a lot of time. The focus is usually on the happy flow.

One remark, all exceptions are now logged as errors. In some cases they could be warnings. For instance if an internet interruption causes an exception I usually log that as warning. Opinions on what to log when vary, but I think many developers would agree. Here some discussion about this: https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels

PR #1667, opened 2022-07-16
Another round of Warning fixes

inforithmics:


Mapsui/Widgets/IWidget.cs
```diff
@@ -18,7 +18,7 @@ public interface IWidget
         /// <param name="navigator">Navigator of MapControl</param>
         /// <param name="position">Screen position</param>
         /// <returns>True, if the Widget had handled the touch event</returns>
-        bool HandleWidgetTouched(INavigator navigator, MPoint position);
+        bool HandleWidgetTouched(INavigator? navigator, MPoint position);
```
pauldendulk, line 21: Nullabilty should be avoided if possible. In general it is better to check for nullabilty early and then pass a non-nullable argument down the line. In this case the call to HandleWidgetTouched should check for it and return false immediately.

NuSpec/Mapsui.Maui.nuspec
```diff
@@ -26,19 +26,11 @@
         <dependency id="SkiaSharp.Views.Maui.Controls" version="[2.88.1-preview.63,3.0.0)"/>
         <dependency id="SkiaSharp.Views.Maui.Controls.Compatibility" version="[2.88.1-preview.63,3.0.0)"/>
       </group>
-	  <group targetFramework="net6.0-windows10.0.19041.0">
```
pauldendulk, line 29: This does not seem to belong to this PR but I assume this can just be removed.

Samples/Mapsui.Samples.Common/Maps/Special/AsyncFetchSample.cs
```diff
@@ -51,7 +52,7 @@ private static IEnumerable<IFeature> GetCitiesFromEmbeddedResource()
         {
             var path = "Mapsui.Samples.Common.EmbeddedResources.congo.json";
             var assembly = typeof(PointsSample).GetTypeInfo().Assembly;
-            using var stream = assembly.GetManifestResourceStream(path);
+            using var stream = assembly.GetManifestResourceStream(path) ?? throw new InvalidOperationException(@"Mapsui.Samples.Common.EmbeddedResources.congo.json not found");
```
pauldendulk, line 55: Perhaps using 'path' here would be better:
```$"{path} not found"```

File: Samples/Mapsui.Samples.Common/Maps/Geometries/PointsSample.cs [hunk 4]
```diff
@@ -46,7 +47,7 @@ private static IEnumerable<IFeature> GetCitiesFromEmbeddedResource()
         {
             var path = "Mapsui.Samples.Common.EmbeddedResources.congo.json";
             var assembly = typeof(PointsSample).GetTypeInfo().Assembly;
-            using var stream = assembly.GetManifestResourceStream(path);
+            using var stream = assembly.GetManifestResourceStream(path) ?? throw new InvalidOperationException(@"Mapsui.Samples.Common.EmbeddedResources.congo.json not found");
```
pauldendulk, line 50: Perhaps using 'path' here would be better:
$"{path} not found"

pauldendulk (changes requested):
The request for change is to check if Navigator is null before the call to IWidget.HandleWidgetTouched.

Mapsui/Widgets/IWidget.cs
inforithmics: I made this nullable because on some code Pathes the navigator is not used, So returning instead of calling it would change the logic.

NuSpec/Mapsui.Maui.nuspec
inforithmics: This is for removing tons of warnings in the other targets, this is remaining code of a workaround for earlier MAUI Versions.

Samples/Mapsui.Samples.Common/Maps/Special/AsyncFetchSample.cs
inforithmics: Good catch copy paste error

Mapsui/Widgets/IWidget.cs
inforithmics: Examples where the Navigator is not used.

1) ScaleBarWidget.HandleWidgetTouched.              returned false
2) PerformanceWidget.HandleWidgetTouched.       WidgetTouched called without Navigator
3) ButtonWidget.HandleWidgetTouched.                WidgetTouched called without Navigator
4) HyperLink.HandleWidgetTouched.                      Touched called without Navigator

Example where the Navigator is used.
1) ZoomInOutWidget.HandleWidgetTouched        Here the navigator is used for zoom in and out.

inforithmics: I change it anyway, because when the Navigator is null the mapcontrol is not properly initialized.

inforithmics: Done

pauldendulk (approved):
🚀

PR #1668, opened 2022-07-16
Script for disabling and enabling central package management

inforithmics:
Here are 2 Powershell scripts to disable or enable central Package mangement (When debugging a Problem central package management can be disabled (like for adding Target for Uno.WinUI making it run) than enable Central Package Management again to bugfix it.


inforithmics:
I normalized the csproj files to utf8 with bom and </Project>\r\n at the end

pauldendulk (approved):

PR #1673, opened 2022-07-18
Make ObservableCollectionProvider ThreadSafe

inforithmics:
By adding a Shadow Collection.

Trying to fix this


pauldendulk:
@inforithmics Could you explain a bit more about the general idea behind using a shadow collection? I see you can now avoid a lock, that is great.

charlenni:
I didn‘t looked in it, but I would think, that the idea is to centralize the lock by hiding it in the class. So all access to the Collection is inside the class and you don‘t need a lock.

inforithmics:
@charlenni  is right.

1) The main idea is that ObservableCollection is not threadsafe so every change is applied to the ShadowCollection which is thread safe. So the operation of the ObservableCollectionProvider are done on the ShadowCollection which is thread safe. So no looks are needed.

2) Because the ObservableCollection is not threadsafe all changes normally should happen on the UIThread, so the change events occour only at one Thread and no lock is needed. when accessing the ObservableCollection like in the Reset case.

3) If the ObservableColllection is updated by several Threads than a ConcurrentObservableCollection needs to be used so that it works. For that a ConcurrentObservableCollectionProvider is needed, easy to implement this doesn't need a Shadow Collection because the CucurrentObservableCollectionProvider is thread safe.

4) Further improvment could be that instead a ShadowCollection an SRTree is used so that the Fetch can use the SRTree as an index. But I think the SRTree is not threadsafe and a lock would be needed. (But because it is a private collection the Provider can guard the access to the Index, which it cannot do on an ObservableCollection which is used outside the Provider.

inforithmics:
I updated the pull reqeust with the latest master (More logging around Task.Run (wasn't done before so this could catch more exceptions.)

janusw:
> I updated the pull reqeust with the latest master (More logging around Task.Run (wasn't done before so this could catch more exceptions.)

Thanks. I tried again with revision c14a6fc56a5403ca3f802fe807608a50e3ce5697, but did not see any new error messages.

pauldendulk (approved):

PR #1674, opened 2022-07-18
Catch Exceptions in Task.Run for better logging and stability

inforithmics:
While looking at

I found some more places where exceptions where swallowed and not logged.

### DIFF
--- a/Mapsui.UI.Avalonia/MapControl.cs
+++ b/Mapsui.UI.Avalonia/MapControl.cs
@@ -12,6 +12,7 @@ using Avalonia.Platform;
 using Avalonia.Rendering.SceneGraph;
 using Avalonia.Skia;
 using Avalonia.Threading;
+using Mapsui.Extensions;
 using Mapsui.Layers;
 using Mapsui.UI.Avalonia.Extensions;
 using Mapsui.Utilities;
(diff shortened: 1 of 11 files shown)


Mapsui.UI.Avalonia/MapControl.cs
```diff
@@ -215,7 +216,7 @@ private void MapControlSizeChanged()
 
         private void RunOnUIThread(Action action)
         {
-            Task.Run(() => Dispatcher.UIThread.InvokeAsync(action));
+            Catch.TaskRun(() => Dispatcher.UIThread.InvokeAsync(action));
```
pauldendulk, line 219: Why did we use this construct in the first place. A thread is started to only send the call to another thread (the UI thread).

Mapsui/Extensions/CatchExceptions.cs
```diff
@@ -17,5 +17,35 @@ public static async void Exceptions(Func<Task> func)
                 Logger.Log(LogLevel.Error, e.Message, e);
             }
         }
+
+        public static void TaskRun(Action func)
```
pauldendulk, line 21: General remark not aimed at this PR. This class is not an extension method, so should not be in the Extensions folder but somewhere else. Utilities perhaps.

pauldendulk (approved):
Important improvement. In another project we recently agreed that all Task.Run calls should have a try catch for logging, and made it a rule in our guidelines. This is something a reviewer of the PR should find.

PR #1675, opened 2022-07-19
Some more Error Catching

inforithmics:
Some more Error Catching


inforithmics:
Main changes add more errorhandling

Mapsui/Viewport.cs
```diff
@@ -251,6 +251,7 @@ public void Transform(MPoint positionScreen, MPoint previousPositionScreen, doub
             var newX = _centerX + previous.X - current.X;
             var newY = _centerY + previous.Y - current.Y;
 
+            // TODO: deltaResolution is double and it is compared to an int maybe this is always true.
```
pauldendulk, line 254: The method can be called without the deltaResolution parameter. In that cause it will have the default value 1. I assume that if you would call it like this from a unit test the deltaResolution == 1 would be true. We could test that.

pauldendulk (approved):

Mapsui/Viewport.cs
thomasstockermc: I tested it 1.0 == 1 is true so the equality is smart enough to detect 1 (integer) is equals to 1.0 (double)

inforithmics: I tested it 1 (integer) is equal to 1.0 (double) so the == is smarter than I thought so I can remove this comment

charlenni: I would say, that the double is converted to the integer and then tested. With this, 1 == 1.1 should be true too.

inforithmics:
I added a Unit test for delta Resolution. (To make meaningful change, because somehow the build failed).

PR #1678, opened 2022-07-22
Some logging improvements

pauldendulk:
These would have helped me to find a problem when using Mapsui.

### DIFF
--- a/Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
+++ b/Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
@@ -44,8 +44,10 @@ namespace Mapsui.Rendering.Skia
                             case LineString lineString:
                                 LineStringRenderer.Draw(canvas, viewport, vectorStyle, lineString, opacity);
                                 break;
+                            case null:
+                                throw new ArgumentException($"Geometry is null, Layer: {layer.Name}");
                             default:
-                                throw new ArgumentException("Unknown geometry of Feature");
+                                throw new ArgumentException($"Unknown geometry type: {geometryFeature.Geometry?.GetType()}, Layer: {layer.Name}");
                         }
                         break;
                 }
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -90,7 +90,7 @@ namespace Mapsui.UI.Wpf
             _performance?.Add(_stopwatch.Elapsed.TotalMilliseconds);
 
             // Log drawing time
-            Logger.Log(LogLevel.Information, $"Time for drawing control [ms]: {_stopwatch.Elapsed.TotalMilliseconds}");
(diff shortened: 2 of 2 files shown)


Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
```diff
@@ -44,8 +44,10 @@ public bool Draw(SKCanvas canvas, IReadOnlyViewport viewport, ILayer layer, IFea
                             case LineString lineString:
                                 LineStringRenderer.Draw(canvas, viewport, vectorStyle, lineString, opacity);
                                 break;
+                            case null:
+                                throw new ArgumentException($"Geometry is null, Layer: {layer.Name}");
```
pauldendulk, line 48: It helps to distinguish null values, which I think is the more common situation, from unknown geometry types. Also adding the LayerName in the message would have helpt me to diagnose the problem sooner.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -90,7 +90,7 @@ private void CommonDrawControl(object canvas)
             _performance?.Add(_stopwatch.Elapsed.TotalMilliseconds);
 
             // Log drawing time
-            Logger.Log(LogLevel.Information, $"Time for drawing control [ms]: {_stopwatch.Elapsed.TotalMilliseconds}");
+            Logger.Log(LogLevel.Debug, $"Time for drawing control [ms]: {_stopwatch.Elapsed.TotalMilliseconds}");
```
pauldendulk, line 93: I forward the Mapsui logs to my own logs. The drawing time logged at information levels spams the logs with lines that are not that interesting. Even logging this at debug level may be too much. Perhaps the drawing information should be stored somewhere so that the users can collect that information when they want.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -374,15 +374,15 @@ private void MapDataChanged(object? sender, DataChangedEventArgs? e)
                     }
                     else if (e.Cancelled)
                     {
-                        Logger.Log(LogLevel.Warning, "Fetching data was cancelled", e.Error);
+                        Logger.Log(LogLevel.Warning, "Fetching data was cancelled.");
                     }
                     else if (e.Error is WebException)
                     {
-                        Logger.Log(LogLevel.Warning, "A WebException occurred. Do you have internet?", e.Error);
+                        Logger.Log(LogLevel.Warning, $"A WebException occurred. Do you have internet? Exception: {e.Error?.Message}", e.Error);
```
pauldendulk, line 381: Although the more specific information is in the exception the users needs to log the entire exception itself in order to see it, and this is something often not done when a warning comes in. In my case I forwarded the message to the Maui logs and I did not add the exception in the case of a warning. So, adding the exception message to the log message so that it can not be missed.

inforithmics (approved):
This a reasonable changes, to help debug problems.

PR #1683, opened 2022-08-01
Rasterizing Tiling Bugfixes

inforithmics:
In this pullrequest i enhance the ShapefiletilingSample so that it displays the same data as the ShipeFileSample. Doing that I already discovered a concurrency bugs and other rendering bugs, I'll fix them as they occour


inforithmics:
Main changes.

1) Updated sqlite so that it works on Windows Arm64.
2) Fixed Label Transformation Bug the Coordinates where transformed but not the center point.
3) Fixed Tiling of Points where part of the Symbol rendered in another Tile (so for example only half circle was rendered instead of a full circle (featureSearchGrow parameter default 25%).
4) Added FilterProvider and sample for it (I used it for debugging the half circle Rendering in the Tiles.
5) Now the RasterizingTileLayer should be able to Render everything that the RasterizingLayer can render.

inforithmics:
I discovered that the Symbol Rendering is not Multithreading save so. I fixed some Multithreading Bugs but in the end I render only one Tile at a time with the map Renderer.

pauldendulk (changes requested):
It looks like featureSearchGrow has a similar purpose as MemoryLayer.SymbolSize (or RasterizingLayer.SymbolSize or StackedLabelProvider.SymbolSize). SymbolSize was more like a workaround than a real solution, but we should try to come up with single solution for everything. 

Some things about SymbolSize:
- It should increase the fetch extent to the provider as well (apart from the features currently in the layer).
- It it used on a couple of layers, but I think now it should be on all layers. This PR uses featureSearchGrow for tile layers. When using tile layer the problem becomes clearly apparent on the tile boundaries, but the same problem occurs for all layers on the edge of the map, only there you are not aware something is missing, but it still is.
- SymbolSize is now used with point symbols in mind, but I think it is also relevant for polygons and lines if they have a line width bigger than zero.
- It increases the search by an amount in screen size and not in geographic coordinate size, and I think this is what we need. Do you see scenarios where this could be different?
- It is hard to know what the size of the symbols is for a layer. You could try to calculate it based on the symbols, but that is not easy and symbol size could change from one feature to another. Guess the solution is to make it configurable.
- It could be that symbol size extents in just one direction, but I we should ignore that on use the max in all directions.
- Perhaps SymbolSize should be called MaxSymbolSize
- Perhaps it should be a field on BaseLayer.

I briefly glanced at featureSearchGrow so I may not fully understand it, please help understand it better.

charlenni:
It is more about screen pixel sizes than about geographical sizes. It is important which size in screenpixels the symbol has. If you want to use geographical sizes, you have to convert symbols to geographical sizes and check, if some of them are inside of the visible area.

It could be getting more difficult, if you have a polyline. The outlining rectangle could have common parts, but the polyline is total outside.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -24,12 +26,12 @@ public class MapRenderer : IRenderer
 
         public ISymbolCache SymbolCache => _symbolCache;
 
-        public IDictionary<Type, IWidgetRenderer> WidgetRenders { get; } = new Dictionary<Type, IWidgetRenderer>();
+        public IDictionary<Type, IWidgetRenderer> WidgetRenders { get; } = new ConcurrentDictionary<Type, IWidgetRenderer>();
```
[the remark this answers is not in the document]
inforithmics: I'll revert this changes, before this changes the Rendering worked Multi threaded but because not all Parts are multithreading capable in the MapRenderer I only Render now 1 Tile at a Time. So this changes aren't anymore necessary.

inforithmics:
> It looks like featureSearchGrow has a similar purpose as MemoryLayer.SymbolSize (or RasterizingLayer.SymbolSize or StackedLabelProvider.SymbolSize). SymbolSize was more like a workaround than a real solution, but we should try to come up with single solution for everything.

I didn't know SymbolSize, yes this could be a Solution for the Symbols. But there is a second Problem and this is Labels. I don't know the maximum Label Size, maybe I have to Calcualte the Maximum LabelSize and use than this Size.

> 
> Some things about SymbolSize:
> 
> * It should increase the fetch extent to the provider as well (apart from the features currently in the layer).
> * It it used on a couple of layers, but I think now it should be on all layers. This PR uses featureSearchGrow for tile layers. When using tile layer the problem becomes clearly apparent on the tile boundaries, but the same problem occurs for all layers on the edge of the map, only there you are not aware something is missing, but it still is.
> * SymbolSize is now used with point symbols in mind, but I think it is also relevant for polygons and lines if they have a line width bigger than zero.
> * It increases the search by an amount in screen size and not in geographic coordinate size, and I think this is what we need. Do you see scenarios where this could be different?
> * It is hard to know what the size of the symbols is for a layer. You could try to calculate it based on the symbols, but that is not easy and symbol size could change from one feature to another. Guess the solution is to make it configurable.
> * It could be that symbol size extents in just one direction, but I we should ignore that on use the max in all directions.
> * Perhaps SymbolSize should be called MaxSymbolSize
> * Perhaps it should be a field on BaseLayer.
> 
> I briefly glanced at featureSearchGrow so I may not fully understand it, please help understand it better.

An Image says more than 1000 words.

Tile Rendering without featureSearchGrow
image

Tile Rendering with featureSearchGrow = 0.25
image

The Problem with Symbols and Labels at the Border of a Tile is that the Center of the Label or the Symbol is in the other Tile. But it renders into the current Tile. So to workaround this I load features outside of the Tile 0.25 means 25% bigger on all sides.

so a tile of (1,1,2,2) would load the Data of the Tile (0.75,0.75,2.25,2.25) So the Symbols and Lables are Rendered into the current Tile and the User does not see the Tile boundries anymore like in the first picture.
The red line should display the bigger Feature size.

But I think with the SymbolSize, and the Maxium Label Size I could calculate the Value automatically and removing it from the Interface, But I have to investigate that.

inforithmics:
> It is more about screen pixel sizes than about geographical sizes. It is important which size in screenpixels the symbol has. If you want to use geographical sizes, you have to convert symbols to geographical sizes and check, if some of them are inside of the visible area.
> 
> It could be getting more difficult, if you have a polyline. The outlining rectangle could have common parts, but the polyline is total outside.

Yes you are right, The OnScreen Symbol size has to be transformed to the geographic size which depends on the zoom level

charlenni:
You couldn't use values like 25 %, because there could be much longer labels.

I do it in the vector map with a two step algorithm:
1. Get all symbols (symbols and labels) and calc their size
2. Get alle features from the tile and all surrounding tiles and check, which features are overlapping with the tile you want to draw.

inforithmics:
I calculate now the Additional Size to load the features with following methodology.

1. I calcluate the maximum Feature size of the tiles sourounding the current tile and cache the result, so I don't Need to load all features.
2. The I use half of the Feature Size to add to the search of the features.
3. I Implemented it on the StyleRenderers of Symbol and Label because the other renderers don't need to load more data.

inforithmics:
@pauldendulk I calcuate now the Symbol size in the StyleRenderer where it makes sense. (Added Interface for it, currently I use it in the RasterizingTiling. The question is where should the Functionality go into the IRender Interface or another interface? Or in the Layer (But than it needs a Map Renderer to Calculate the Size).

inforithmics:
Fixed a Bug iterating the Features, and fixed a too large Feature Search Area

inforithmics:
https://github.com/Mapsui/Mapsui/pull/1722 was merged into this to not fix the same bugs again in this pull request

Mapsui.Rendering.Skia/SymbolCache.cs
```diff
@@ -1,12 +1,13 @@
...
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using Mapsui.Styles;
 
 namespace Mapsui.Rendering.Skia
 {
     public class SymbolCache : ISymbolCache
     {
-        private readonly IDictionary<int, BitmapInfo> _cache = new Dictionary<int, BitmapInfo>();
+        private readonly IDictionary<int, BitmapInfo> _cache = new ConcurrentDictionary<int, BitmapInfo>();
```
pauldendulk, line 10: See my earlier remark, please no thread save things in the core.

Mapsui.Tiling/Provider/RasterizingTileProvider.cs
```diff
@@ -62,31 +66,147 @@ public class RasterizingTileProvider : ITileSource
         var result = PersistentCache.Find(index);
         if (result == null)
         {
-            var renderer = GetRenderer();
-            var (viewPort, renderLayer) = await CreateRenderLayerAsync(tileInfo);
+            MemoryStream? stream = null;
+
+            try 
+            { 
+                using (await _renderLock.LockAsync())
```
pauldendulk, line 73: Why is this necessary. What recourses are accessed by which threads? I would be great if we could avoid this.

Mapsui/Providers/FilteringProvider.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Mapsui.Layers;
+
+namespace Mapsui.Providers
+{
+    public class FilteringProvider : IProvider
```
pauldendulk, line 9: What is the purpose of the FilteringProvider?

Mapsui/Providers/ProjectingProvider.cs
```diff
@@ -31,7 +31,8 @@ public async Task<IEnumerable<IFeature>> GetFeaturesAsync(FetchInfo fetchInfo)
                 return Enumerable.Empty<IFeature>();
 
             var features = await _provider.GetFeaturesAsync(fetchInfo);
-            return features.Project(_provider.CRS, CRS, _projection);
+            var result = features.Project(_provider.CRS, CRS, _projection);
```
pauldendulk, line 34: Why is the result first assigned to a local variable before it is returned?

Mapsui.Rendering.Skia/SymbolCache.cs
inforithmics: Checkin that changed that is now reverted. Because Tiling Rasterizing is now single threaded this isn't needed anymore.

Mapsui.Rendering.Skia/SkiaStyles/LabelStyleRenderer.cs
```diff
@@ -7,16 +7,17 @@
...
 using System.Text;
 
 namespace Mapsui.Rendering.Skia
 {
-    public class LabelStyleRenderer : ISkiaStyleRenderer
+    public class LabelStyleRenderer : ISkiaStyleRenderer, IFeatureSize
     {
         private static readonly IDictionary<string, BitmapInfo> LabelCache =
-            new Dictionary<string, BitmapInfo>();
+            new ConcurrentDictionary<string, BitmapInfo>();
```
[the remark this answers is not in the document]
inforithmics: Checkin that changed that is now reverted. Because Tiling Rasterizing is now single threaded this isn't needed anymore.

Mapsui/Providers/ProjectingProvider.cs
inforithmics: I changed it for debugging, but because it works now I revert it.

inforithmics: was for debuggin is now reverted

Mapsui/Providers/FilteringProvider.cs
inforithmics: I use the FilteringProvider to display only cities with more than 5 million. I made it general with a func for filtering around other critieria.

Mapsui.Tiling/Provider/RasterizingTileProvider.cs
inforithmics: Because in Tiling more than one Tile can be requested at the moment. And Rendering is not Threadsafe even when Rendering with different MapRenderers (due to global caches objects, for example symbols and Labels, I only render one Tile at a time, before there where Concurrency exceptions because of Used Dictionary objects and similar things. 
Another method would be to specifiy in the Layer Interface if it is thread safe and than render Threadsafe layers with out lock.
and the others with a lock.

Mapsui.Rendering.Skia/SkiaStyles/LabelStyleRenderer.cs
```diff
@@ -383,5 +384,34 @@ private static Line[] SplitLines(string? text, SKPaint paint, float maxWidth, st
                 return result.ToArray();
             }).ToArray();
         }
+
+        public double FeatureSize(IFeature feature, IStyle style, ISymbolCache symbolCache)
```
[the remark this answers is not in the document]
inforithmics: I'll make this method static

inforithmics:
It seems Github Action changes broke Mapsui.Samples.UI.Forms.Mac compilation

will fix it here
https://github.com/Mapsui/Mapsui/pull/1722
and merge into this

Mapsui.Tiling/Provider/RasterizingTileProvider.cs
pauldendulk: Okay, it is in a provider, not in a the renderer, so not that bad. The caches really complicate things. So now that we have a lock in the provider do we not need the thread save caches? Perhaps we still do because the map renderer itself is not in this lock. What about injecting the caches and using a different cache for the provider? Perhaps make it optional to inject a thread save cache? .... just throwing in some thoughts, no worked out ideas.

inforithmics: Could be a solution. I'll have to try it out.

inforithmics: I removed the lock, the reason was that labelstylerenderer had one statock global skpaint. I made it an instance of the Labelstylerenderer now it can render in parallel. There is a static label cache because this is accessed by all labelstylerenderers, i made it a concurrentdictionary. or I could make it a dictionary then it doesn't need to be concurrent.

inforithmics: Another possibility would be to create an ILabelCache and give this instance into the renderer

inforithmics:
In this PullRequest I fixed now the LabelStyleRenderer (removing the static SkPaint and static LabelCache).

removing the static Skpaint fixes the parallel rendering of Lables (for example in the TilingProvider).

removing the Static LabelCache has as sideeffect that bitmap data is now freed when the MapRenderer isn't used anymore, before that this was a possible memory leak and made problems in parallel rendering in the LabelStyleRenderer.

Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
```diff
@@ -255,5 +255,48 @@ private static void DrawTriangle(SKCanvas canvas, float x, float y, float sideLe
             if ((fillColor != null) && fillColor.Color.Alpha != 0) canvas.DrawPath(path, fillColor);
             if ((lineColor != null) && lineColor.Color.Alpha != 0) canvas.DrawPath(path, lineColor);
         }
+
+        public double FeatureSize(IFeature feature, IStyle style, ISymbolCache symbolCache)
+        {
+            var symbolStyle = (SymbolStyle)style;
```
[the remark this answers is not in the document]
inforithmics: I added Unit Tests for LabelStyle and SymbolStyle size

PR #1694, opened 2022-08-20
update skia sharp to 2.88.2

inforithmics:
Update skia sharp to release 2.88.1


inforithmics:
The reason why it doesn't build is fixed here https://github.com/mono/SkiaSharp/pull/2231

inforithmics:
Updated to the newest stable skiasharp version there are some bugs in uno.winui but they are already fixed and will be in the next version probably 2.88.2

pauldendulk (approved):

Issue #484, opened 2018-07-26
Started using SKGLView once it is fixed


pauldendulk:
The Mapsui iOS MapControl was reverted to use SKCanvasView because SKGLView can crash when rendering certain polygon outlines.

pauldendulk:
This is the Mapsui issues discussing it:
https://github.com/Mapsui/Mapsui/issues/457

This is the SkiaSharp issue:
https://github.com/mono/SkiaSharp/issues/571

pauldendulk:
The original issue was fixed in the version of skiasharp we are currently using so we could go back to SKGLView. Let's do that after release v2.

charlenni:
The Forms implementation already uses the SKGLView.

pauldendulk:
> The Forms implementation already uses the SKGLView.

Yes, I saw that. An indication it is stable. Still I prefer to not change the iOS implementation in this stage.

pauldendulk:
Release in 2.0.0-beta.35

setallunder:
In a Xamarin.iOS (Xamarin.Native) project SKGLView leads to the OOM crash (check the [SkiaSharp memory issue](https://github.com/mono/SkiaSharp/issues/1683)). It takes like a dozen navigations to a view controller with a fullscreen map to crash (iPhone 6). Changing canvas back to the SKCanvasView fixes the issue

inforithmics:
I think the UseGPU Property should be available in base MapControl so that the other platforms can specify if GPU is used or not.

pauldendulk:
With `MapControl.UseGPU = true` this is now a choice the user can make. It can not be used on all platforms yet, but that is not something to track in this issue.

PR #1696, opened 2022-08-20
useGpu roperty on Mapsui.ios

inforithmics:
Fixes this: https://github.com/Mapsui/Mapsui/issues/484 (With setting usepgu to false)


inforithmics:
Can somehow not run the Mapsui.ios sample

inforithmics:
Main change is that the rendering can be changed for iOS from gpu to cpu due to some gpu bugs on certain platforms

pauldendulk (approved):

PR #1704, opened 2022-09-03
(WIP) Remove unnecessary lock from Layer

inforithmics:
For trying to fix
Removing or improving locks


inforithmics:
This is a test pull request if it fixes the freezing of mapsui

janusw:
Sorry for the late feedback here. I just tried the nupkg from rev. 32ddf20c319b2ae28c9adb50919e5295d02269cf, but unfortunately it does not seem to fix the freezing issues from #1653.

inforithmics:
@janusw thanks for the feedback then I think I will close this pull request, because it is risky and does not improve the situation.
On Main Mapsui I did some Fixes in the LabelRenderer (was not thread safe). Maybe this helps

janusw:
> On Main Mapsui I did some Fixes in the LabelRenderer (was not thread safe). Maybe this helps.

I tried the current master (1eef6fcbe896bf2b511c233938638f2b6d2c7a58), but the freezing issues still seem to be present.

PR #1705, opened 2022-09-04
Introduced ObservableMemoryLayer for trying to solve freezes

inforithmics:
Trying to solve the Freezing of Mapsui, by making an ObservableMemoryLayer trying to fix https://github.com/Mapsui/Mapsui/issues/1653

This is a pull request that tries to use Again MemoryLayer as the Layer for Pin,Callout,Drawable


inforithmics:
Because I cannot Test it @janusw has to Test this.

inforithmics:
This is a test pull request if it fixes the freezing of mapsui

janusw:
I just tried the nupkg from rev. d87f5133f9f69fb8e5b55d5c3796c61c0c17af13, but it does not seem to work. The callouts, pins and drawables are not rendered at all.

inforithmics:
@janusw I fixed it with this checkin
https://github.com/Mapsui/Mapsui/pull/1705/commits/6fb8c8d6a29875770d06366ef46e415314c0b788
Updated it from main.

janusw:
> @janusw I fixed it with this checkin [6fb8c8d](https://github.com/Mapsui/Mapsui/commit/6fb8c8d6a29875770d06366ef46e415314c0b788)

Thanks for your continued efforts here! A first quick test of this looks very promising. I was not able to observe any freezing issues so far. 👍

But let me test it a bit more thoroughly ...

janusw (approved):
So, I did some more testing, running a couple of longish sessions with complex rendering scenarios: Large amount of pins, several complex drawables (mostly long polylines) plus some callouts.

The upshot is: Everything is golden. Mapsui works like a charm again on Android (first time since half a year). Thanks a lot for this fix, @inforithmics! Please feel free to merge it.

Oh, and could you maybe explain how and why this fixes the freezing issues?

inforithmics:
I think the reason why this pull request fixes the freezes is that in pullrequest

https://github.com/Mapsui/Mapsui/pull/1600

MemoryLayer was replaced with Layer the reason was that the MemoryLayer was simplified so it couldn't be used here.
The Usage scenario stays the same I introduced an ObservableMemoryLayer, because (ObservableCollection is not thread safe I copy the contents to the MemoryLayer on changes on the ObservableCollection, So you can still use an ObservableCollection and still be Multithreading save. (Because there are two threads at least a rendering thread and a ui thread).

I used the same technic here https://github.com/Mapsui/Mapsui/pull/1673 to make the ObservableCollection Provider threadsafe.

I think the true reason is that Layer has some locking issues this should be adressed with using Channels. But for the scenario of pins and callout the Layer has many things that are not needed and simplifying the implementation with ObservableMemoryLayer solves the Problem too, and it is more close to the Original MemoryLayer.

janusw:
Thanks for the explanation.

Btw, wouldn't it be a good idea to tag a new beta version soon? It's been more than four months since 4.0.0-beta.5, and beta.3 (the last tag that was working for me) is almost 8 months old. @pauldendulk, what do you think?

PR #1722, opened 2022-09-24
Fixing ios/uno/forms build (GitHub Action changes)

inforithmics:
An update occured on the Mac Github Image and iOS doesn't compile anymore so I try fixing this.

PR #1723, opened 2022-09-24
Avalonia.MapControl: fix zoom with touchpad on MacOS & Linux

yapus:
Contrary to integer +1/-1 "Y" deltas sent by ordinary mouse devices, on MacOS & Linux, touchpads emulate the scroll wheel by dragging two fingers. These synthetic mouse-wheel events are sent in batches of 10-20 events with Y-deltas as fractional decimal numbers (0.01-0.1). So zooming the map by scrolling does not work as expected on Touchpad.

This commit introduces a double accumulator variable `_mouseWheelPos` that is used to sum individual Y-deltas as they are received and trigger zoom at integer steps only.

Related info and solution from: https://github.com/mrkite/minutor/issues/224

### DIFF
--- a/Mapsui.UI.Avalonia/MapControl.cs
+++ b/Mapsui.UI.Avalonia/MapControl.cs
@@ -28,6 +28,7 @@ namespace Mapsui.UI.Avalonia
         private bool _mouseDown;
         private MPoint? _previousMousePosition;
         private double _toResolution = double.NaN;
+        private double _mouseWheelPos = 0.0;
 
         public event EventHandler<FeatureInfoEventArgs>? FeatureInfo;
         public MouseWheelAnimation MouseWheelAnimation { get; } = new() { Duration = 0 };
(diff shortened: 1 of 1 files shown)


yapus:
HI @janusw @tmijieux @bertt 
please check/merge this - i'm really looking forward for this to be part of next nuget package release (4.0.0-beta6?)

inforithmics:
The Build failures are unrelated and are fixed here https://github.com/Mapsui/Mapsui/pull/1722

inforithmics:
The build failures are unrelated and are fixed here.

https://github.com/Mapsui/Mapsui/pull/1722

I tested it and on the other platforms the Y value changes only 1 plus or -1 so my comments don't need to be adressed (so merging this now)

inforithmics (approved):

yapus:
any idea when the next beta release would be published?

yapus:
@inforithmics Hi, was this been reverted for some reason? Can't see these fixes in the latest 4.0.0-beta11

yapus:
Should i fix it again and send new PR? The zoom not working with touchpad is really-really annoying....

inforithmics:
Yes would be nice to send a new PR? The mouse handling moved to the navigator and your code got optimized away. 

https://github.com/Mapsui/Mapsui/commit/ce0f5edee10fffcf424d776318853d9e097c863f

yapus:
I see, so what you think would be better solution here - put that logic back to the MapControl, or change `mouseWheelDelta` type to double and handle that logic in Navigator?

inforithmics:
In my opinion I would handle this in MapControl. Because I think this is an Avalonia specific Problem.

yapus:
Ok, see new pull request [pull/1984](https://github.com/Mapsui/Mapsui/pull/1984) please

PR #1725, opened 2022-10-01
(WIP) Target Blazor

inforithmics:


inforithmics:
The first sample is displaying the moving arrows. So the Mapsui.UI.Blazor is working.
Things to Do.

Let the User the sample to select which he wants to be displayed. Fixing Bugs (Like mouse navigation etc.)

inforithmics:
Cleaned up the pull request here with less changes

https://github.com/Mapsui/Mapsui/pull/1747

PR #1727, opened 2022-10-10
Remove the 'Time for drawing control' log line

pauldendulk:
When starting on this change I intended to store the list of the last X times somewhere so that others would at least have to opportunity to draw it somewhere. The 'Performance' class seemed the proper place to add this list but it turns out this class already stores the last x (default is 20) drawing times (thanks @charlenni ). Users can request this list with MapControl.Performance.DrawingTimes. It may also be an option for users to create their own version of the performance widget and draw the last x drawing times there.

This PR simply removes the log line.


inforithmics (approved):
For me it is fine when the Drawing Time is not logged (could potentially affect drawing performance).

PR #1732, opened 2022-10-12
Improve mbtile deployment for samples

pauldendulk:
Two of our samples make use of an mbtiles file which needs to be deployed for it to work. The previous setup did this deployment for each platform which caused more platform specific code and made it possible to forget it. With the new setup the mbtile files are deployed in the static constructor of the samples themselves.

Some notes. 
- I tested this with Maui on Windows and Xamarin.Forms on Android. Nothing else.
- I use this folder: Environment.SpecialFolder.LocalApplicationData. I don't know where that is exactly on each platform and I am not sure it exists on all platforms, or if you need access rights.


inforithmics:
Good idea. Some more MBTilesLocation Code has to be removed.

inforithmics (approved):
It works quite good I tested eto,ios,android,winui,avalonia,wpf

The only thing that didn't work was Uno.Wasm but I think this didn't work before.

inforithmics:
I fixed the Eto Sample so that it can start on windows (on other platforms I didn't test)

inforithmics:
Found this:
https://platform.uno/docs/articles/features/file-management.html

inforithmics:
The build failure will be fixed by this pullrequest
https://github.com/Mapsui/Mapsui/pull/1735

I checked wasm didn't work before with mbtiles. I guess that the reading of the embedded resources does not work and that is why it isn't working so no regression.

inforithmics:
I did some more research. And it seems that for wasm special nugets need to be used.

https://platform.uno/blog/uno-platform-sqlite-for-webassembly-support-on-net-5-and-net-6/

Tried it and it didn't work, so waiting for .NET 7

PR #1733, opened 2022-10-12
Install .Net Sdk 6.0.401 on document generation

inforithmics:
Install .Net Sdk 6.0.401 for fixing document generation


pauldendulk (approved):

PR #1734, opened 2022-10-12
Make CacheTypeFace an instance Field (Then I don't need ConcurrentDic…

inforithmics:
Make CacheTypeFace an instance Field (Then I don't need ConcurrentDictionary anymore)


inforithmics:
Needs more work to fix this so closing this pull request

PR #1735, opened 2022-10-14
remove and ignore dotnetcli and dotnet-install.ps1

inforithmics:
The Generate Size checkin has added the dotnet cli and the dotnet-install.ps1 With this checkin I removed them and added it to the ignore list


inforithmics:
Main changes here
1) removed the dotnetcli from the git repository and ignored it
2) improved the appveyor-doc.yml to first checkout and then installing the dotnetcli (before the checkout removed the dotnetcli and thats why the Generate site failed).

pauldendulk:
oh, the entire dotnetcli was pushed to git when it tried to push the website. 😲. Thanks for fixing it.

pauldendulk (approved):

PR #1736, opened 2022-10-15
Put every mapsui dll into its own nuget package and use csproj for nuget package creation

inforithmics:
Every mapsui dll is now in its own nuget package.
i added Mapsui.Rendering.Skia as additional dependency to the (Mapsui.Maui, Mapsui.Forms nuget packages ...)


inforithmics:
Main changes:
1. Update VersionUpdater to update the mapsui Versions in Directory.Packages.props
2. Added Skript to change project References in Samples to nuget packages ([SamplesMapsuiNugetReferences.ps1](https://github.com/Mapsui/Mapsui/pull/1736/files#diff-a28fbc08d2462b11174dd119c618cc37ebc9441722e6421491b5f0d24595bcd3))

inforithmics:
3. In the Github action i change the projectreference with the nuget references and compile it then so it is tested if the nuget packages work or have the needed dependencies included.

inforithmics:
4. The AsyncLock package I integrated in mapsui because there where nuget reference problems in the generated nugets (was only 1 class and mit license).

inforithmics:
Should I add Mapsui.Tiling as an additional dependency on Mapsui.Forms, Mapsui.Maui (it is not directly referenced so when someone uses it and upgrades he has to add it (like osm maps).

pauldendulk:
> Should I add Mapsui.Tiling as an additional dependency on Mapsui.Forms, Mapsui.Maui (it is not directly referenced so when someone uses it and upgrades he has to add it (like osm maps).
> 
> Mapsui.Tiling is now added as dependency

Yes, I think that makes sense. We should make the most common use case as easy as possible.

Another option would be to add a separate project that combines other projects, but I think that is a bit too much.

inforithmics:
The appveyor build seems to fail do to networking errors because installing .net sdk 6.0.401 takes roughly one hour

inforithmics:
With the script where I change the Project references with the created nuget packages. I discovered that the Mapsui.iOS nuget package was wrongly written (didn't work for net6.0-ios) this is now fixed.

What I discovered Too nuget package Mapsui.iOS project name Mapsui.UI.iOS should the project name and the nuget package name not be the same.

If yes then I would like to rename all ui Projects to a name without ui (the folders too).

inforithmics:
> > Should I add Mapsui.Tiling as an additional dependency on Mapsui.Forms, Mapsui.Maui (it is not directly referenced so when someone uses it and upgrades he has to add it (like osm maps).
> > Mapsui.Tiling is now added as dependency
> 
> Yes, I think that makes sense. We should make the most common use case as easy as possible.
> 
> Another option would be to add a separate project that combines other projects, but I think that is a bit too much.

I thought too, because every time I use Mapsui in the Frontend I use Tiling anyway. In the backend it is another case. But then I can now Select the packages I need instead of everything

inforithmics:
I replaced <licenseurl>.... with <license type="expression">MIT</license>
1. licenseurl is deprecated
2. easier discoverability of the license because with this MIT is displayed as license in nuget.org

pauldendulk:
I would like to do a release before merging this.

One question. For simple scenarios it is possible to generate the nuget from the csproj alone without the nuspec. NuGet dependencies are retrieved from the csproj. Do you think this would be possible for us as well at some point?

inforithmics:
> I would like to do a release before merging this.
> 
> One question. For simple scenarios it is possible to generate the nuget from the csproj alone without the nuspec. NuGet dependencies are retrieved from the csproj. Do you think this would be possible for us as well at some point?

Yes It would be possible, Because for this Scenario to work only one Nuget Can be created per project so this is the first step in that direction.

"Roadmap":

1) One Dll one nuget
2) Eventually rename the csproj to the same name as the nuget package.
   ex. Mapsui.UI.WPF => Mapsui.WPF (Only the UI projects are different) 
3) Set the Nuget Properties in the csproj
4) use the pack command to generate a nuget from the csproj
5) Things to investigate (Currently we use a Version range of our dependencies   for example         <dependency id="Svg.Skia" version="[0.5.12,0.6)"/> Version 0.5.12 < 0.6
There is now a possibility to specify in the PackageReference a VersionRange maybe this gets automatically translated to the generated nuget package. Else a specific reference is generated in this case 0.5.12.

inforithmics:
I reenabled now the Script that changes Project References to nuget PackageReferences and builds the samples with the generated nugets as a Test that the nuget packages are sound.

inforithmics:
I changed now the the nuget package generation to generate them from the csproj file.
1) in the buildpack.cmd

inforithmics:
github actions now use the csproj too for generating the nuget packages

inforithmics:
So the generated nuget packges seem to work. 

I changed the build order.

1) Tests 
2) Release nuget packtes
3) Change in the Samples and Tests Project References to the generated nuget packages
4) Build the Samples with the Generated nuget packages (Found some errors during the Process like wrong references and fixed them).

inforithmics:
Removed nuspec files and use now Version Ranges in the Versions, so that the resulting nuget packages have version ranges too.

Mapsui.ArcGIS/Mapsui.ArcGIS.csproj
```diff
@@ -3,6 +3,10 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
     <TargetFrameworks Condition="'$(SolutionName)' != 'Mapsui.Mac.Legacy'">$(TargetFrameworks);net6.0</TargetFrameworks>
+    <Description>
+      Mapsui components to access an ArcGIS dynamic layer and image layer.
+    </Description>
+    <PackageTags>$(PackageTags) ArcGIS</PackageTags>
```
pauldendulk, line 9: How does $(PackageTags) work? What to you need to do to initialize this? Or is this simple the field set in Directory.Build.props?

pauldendulk (approved):

Mapsui.ArcGIS/Mapsui.ArcGIS.csproj
inforithmics: PackageTags is taken from Directory.Build.props or from the Project when this Property is overwritten in the project

PR #1737, opened 2022-10-16
(Appveyor Build fix) download 6.0.402 sdk to see if this works better

inforithmics:


inforithmics:
Use 6.0.402 sdk (latest stable) seems to work with appveyor 6.0.401 seems to have problems.

PR #1741, opened 2022-10-19
CentralPackageTransitivePinningEnabled

inforithmics:
Reenable CentralPackageTransitivePinningEnabled because enabling this increases the build speed of github und appvoyor builds because less nuget packages are downloaded. Only one Version of every nuget package instead of several


inforithmics:
With this checkin caching works again on github actions

pauldendulk (approved):

PR #1742, opened 2022-10-26
Remove Center from viewport

pauldendulk:
I am looking into simplifying the Viewport. The goals is to have a simple Viewport state, that is just state and other classes that manipulate it. The state in the current Viewport class holds both a Center MReadonlyPoint and CenterX and CenterY doubles. The idea of the MPoint was for ease of use, but it also confusing to have both.

The MReadonlyPoint was mainly added for this Center field. The idea was to prevent users to set the X and Y because it would not trigger an event. Since the Center field was removed the entire MReadonlyPoint had little purpose and was also removed.


inforithmics:
For fixing the build https://github.com/Mapsui/Mapsui/pull/1737 can be used

PR #1743, opened 2022-10-26
Further preparation for viewport changes

pauldendulk:
Introduced a simple ViewportState, which is just the state, hardly any logic (name can change). Not using it yet. Also some other changes to prepare for its use.


Mapsui/UI/ViewportLimiterKeepWithin.cs
```diff
@@ -40,7 +40,7 @@ public class ViewportLimiterKeepWithin : IViewportLimiter
 
         public void Limit(Viewport viewport, IReadOnlyList<double> mapResolutions, MRect? mapEnvelope)
         {
-            viewport.Resolution = LimitResolution(viewport.Resolution, viewport.Width, viewport.Height, mapResolutions, mapEnvelope);
+            viewport.SetResolution(LimitResolution(viewport.Resolution, viewport.Width, viewport.Height, mapResolutions, mapEnvelope));
```
pauldendulk, line 43: Removed some use of the properties, this will work different in a coming release.

Mapsui/Viewport.cs
```diff
@@ -253,29 +257,29 @@ public void Transform(MPoint positionScreen, MPoint previousPositionScreen, doub
 
             if (deltaResolution != 1)
             {
-                Resolution = Resolution / deltaResolution;
+                _resolution = Resolution / deltaResolution;
```
pauldendulk, line 260: Using the local fields instead of the properties internally.

Mapsui/ViewportState.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) The Mapsui authors.
+// The Mapsui authors licensed this file under the MIT license.
+// See the LICENSE file in the project root for full license information.
+
+using Mapsui.Extensions;
+using Mapsui.Utilities;
+
+namespace Mapsui
+{
+    public record class ViewportState
```
pauldendulk, line 10: This is the new ViewportState. Six fields and two fields derived from those.

inforithmics:
For fixing the build https://github.com/Mapsui/Mapsui/pull/1737 can be used

pauldendulk:
The build fail is because you can not push to the github pages from a PR. This should not be possible because of security reasons. Not every PR pushes to github pages, only when something is changed in the docs, and in this case it was probably triggered by merging the doc fix PR into this branch. So this problem does not show up every time. We could try to fix this by only pushing to github pages when we are master, but perhaps it would be better to try to find out how we could do this with github actions, and not using appveyor.

inforithmics:
I think this should be used for gh pages generation

https://github.com/marketplace/actions/deploy-to-github-pages

PR #1744, opened 2022-10-27
Render cache

inforithmics:
The main idea behind this is that IRenderCache inherits from ISymbolCache and also implements a new ILabelCache interface (caching from LabelRenderer moved into this (removing static ConcurrentDictionary on the way).

Then I will introduce a IVectorCache interface where SkiaPaint is cached and SkiaPath. SkiaPath is then drawn by .Draw(path, ref matrix, paint). I will leave the old VectoryStyleRenderer as is and make a new VectoryStyleCachedRenderer and measuring with Benchmark .Net which is faster.

I saw that in https://github.com/charlenni/VectorTileLayers a similar approach is used.

There the SkCanvas is transformed then the path is drawn maybe this is faster I have to measure it.


inforithmics:
|                         Method |      Mean |     Error |    StdDev |       Min |       Max |    Median |     Gen0 |     Gen1 |  Allocated |
|------------------------------- |----------:|----------:|----------:|----------:|----------:|----------:|---------:|---------:|-----------:|
| RenderDefaultWithoutCacheAsync | 11.015 ms | 0.2192 ms | 0.2850 ms | 10.534 ms | 11.686 ms | 11.013 ms | 546.8750 | 265.6250 | 2290.06 KB |
|             RenderDefaultAsync |  4.428 ms | 0.0747 ms | 0.0699 ms |  4.279 ms |  4.552 ms |  4.434 ms |  39.0625 |        - |  166.94 KB |
|           RenderTilingPngAsync |  1.432 ms | 0.0050 ms | 0.0047 ms |  1.424 ms |  1.441 ms |  1.432 ms |   1.9531 |        - |   11.13 KB |
|          RenderTilingWebPAsync |  1.435 ms | 0.0036 ms | 0.0032 ms |  1.427 ms |  1.438 ms |  1.435 ms |   1.9531 |        - |   11.22 KB |
|           RenderTilingSkpAsync |  2.945 ms | 0.0588 ms | 0.0722 ms |  2.753 ms |  3.072 ms |  2.944 ms |        - |        - |   11.41 KB |

inforithmics:
Added now an lru cache with 10000 entries per default

inforithmics:
So in my opinion this can be merged but maybe @charlenni  can have a look at it. 
What I'm not so sure is for what the skCanvas.ClipRect clipping is used. Because the bounds of the visible area are already in the viewport.
So I made in the caching case a viewport.ToSkRect function that creates the cliprect from the Viewport. Maybe this is wrong but I think @charlenni could know it.

charlenni:
A quick answer about clipping rect. If I remember it correct, then the problem long time ago was, that dash lines were drawn very slowly. You could test this with the sample with the dashed rects going over the whole world. If you now zoom in, then it is getting slower and slower. This is, because Skia calculates the dashes from the beginning of the line up to where you see the line to get the right dash to start. So you draw a very, very big rect. So the idea was to calculate a smaller rect instead, so that Skia has less to draw. Don‘t know, if this is still needed. Was in CPU times.

Regarding the caching: This is a good idea. As your measurements show, the most time is wasted by converting the LineString again and again into a SKPath. I‘m not sure, if you need a cache for SKPaint, because they say, that SKPaint is a light weight structure.

What I didn‘t find up to now is, if it is possible to optimize it some more. Skia converts the SKPath to vertices, that could be stored in a buffer of the GPU. Last time I asked, they said, that Skia caches this too. But I didn‘t found it in the source code.

Why I use a matrix drawing the path? Vector data in the format I use is tile oriented and each tile has 4096x4096 coordinates which then build the 512x512 tile. So it is easy to create a matrix so that the left upper corner is 0/0 of the canvas. Then all the coordinates of the tile must not converted.

Mapsui.Rendering.Skia/Cache/RenderingCache.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
```
charlenni, line 1: File should be RenderCache.cs and not RenderingCache.cs

Mapsui.Rendering.Skia/Extensions/ViewportExtensions.cs
```diff
@@ -17,5 +19,13 @@ public static SKMatrix ToSKMatrix(this IViewport viewport)
             matrix = SKMatrix.Concat(matrix, SKMatrix.CreateTranslation((float)-viewport.CenterX, (float)-viewport.CenterY));
             return matrix;
         }
+
+         /// <summary> Converts the Extent of the Viewport to a SKRect </summary>
+        /// <param name="viewport">viewport</param>
+        /// <returns>SkRect</returns>
+        public static SKRect ToSkRect(this IReadOnlyViewport viewport)
```
charlenni, line 26: We call the convert functions always `ToSkia...()`. Perhaps here would be better to use `ToSkiaRect()`

charlenni (commented):
For me it seems a good improvement for Mapsui. To converting all the geometries again and again to the Skia counterpart costs many time and has no advantage. So I like this PR. Perhaps with this it is obsolete to have an own preparing cycle in front of drawing. 

What I don´t know is, if the LruCache is thread save. But if it is only used by the rendere thread (is there an independent thread for rendering?), then it shouldn´t be a problem.

inforithmics:
> For me it seems a good improvement for Mapsui. To converting all the geometries again and again to the Skia counterpart costs many time and has no advantage. So I like this PR. Perhaps with this it is obsolete to have an own preparing cycle in front of drawing.
> 
> What I don´t know is, if the LruCache is thread save. But if it is only used by the rendere thread (is there an independent thread for rendering?), then it shouldn´t be a problem.

Because the Rendering is Single Threaded the Caches don't need to be Multithreaded. If more than one Thread is used for example in TilingRenderingProvider, than more than one Cache is used one Cache Per Thread.
Because the other Caches are not thread safe either.

Mapsui.Rendering.Skia/Cache/RenderingCache.cs
inforithmics: Fixed

Mapsui.Rendering.Skia/Extensions/ViewportExtensions.cs
inforithmics: Fixed

IldarKhayrutdinov:
🚀  this PR will give good UI responsiveness when large/many objects

charlenni:
@pauldendulk Perhaps you have some time to look into this. It is a good PR while we not have separate render and draw processes.

pauldendulk:
> @pauldendulk Perhaps you have some time to look into this. It is a good PR while we not have separate render and draw processes.

Thanks for the notification. I expect this to have some impact and some discussion to come out of it, so I ignored it for now to be able to finalize the release of beta.6. I will have time for this next week and hope to have another release in a few weeks time.

pauldendulk:
Thanks for this PR. Because of you earlier PR it is now possible to measure performance, these optimizations are the result, that's great. Below are some general remarks. I will try to add some specific remarks later on.

### LRUCache
I like the concept of LRUCache. The [MemoryCache](https://github.com/BruTile/BruTile/blob/master/BruTile/Cache/MemoryCache.cs) in BruTile which is used in Mapsui serves a similar purpose. I am sure the MemoryCache could be improved if I study LRUCache design.

### Performance improvements
The most important thing about this PR is that it improves performance. This is what the users will notice

### IFeature.RenderedGeometry
There was a previous implementation of feature caching. It had pointer pairs to a skia path and the skia paint. The field still exists but the values are not cached anymore. It somehow got broken in the changes made in the past. 

### Caching complications
I worked with caching within BruTile and Mapsui and it can cause issues that are hard to debug, and some arbitrary choices had to be made. Some things to take into account.
- The cache should not be to big, because the apps memory will get to big.
- The cache size should al least be bigger than the features that fits on the screen in one render cycle. If it is smaller items will be added and pushed out again by items later in the cycle.
- The number of items in a single iteration can vary a lot, and the size of the cached items can vary a lot. You could have 16 tiles or 1000 points. You need a caching strategy that works well for both.

Idea: We could perhaps add a cleanup mechanism more specific to our use case. We could add an iteration field that is updated to the last iteration it was used. We could choose to cache a number of iterations. 

### A lot of caches
There are a lot of cache classes involved now
- LabelCache
- RenderCache using LRUCache
- VectorCache
- SymbolCache
- UrlPersistentCache

It becomes hard to keep track of it. Perhaps this could be simplified. 

### Caching on the outside
The caches are injected deep inside the core. This makes these core methods harder to grasp (I think they can still be considered pure functions if you consider the cache as an input argument). Perhaps we can rewrite it so that the renderers return a collection of 'drawables' that are added to the cache (or not) on the outside. An added bonus could be that every render method does not have to implement caching itself. Custom renderers do not have to take caching into account and would still be cached.

### Two step rendering
Although this setup makes the need for two step rendering less necessary creating the drawable is still in the render loop. A loop that does only rendering will still have an added value. I think such a setup could be combined with other things mentioned above (like caching on the outside).

### The implementation of *value equals* is hard to maintain
When implementing caching with IFeature.RenderedGeometry I implemented value equals for styles (I think it was suggested by resharper). Afterwards I was unhappy with that choice. It involves a lot of code. It is easy to introduce a bug because of a property that is overlooked. It is hard to maintain because when a property is added users should not forget to include it in the equals check. When we allow custom rendering the custom implementation would also need to implement this. I wonder if there is an alternative, perhaps just use reference equals, or perhaps we could use some id somehow.

Mapsui.Avalonia.slnf
```diff
@@ -0,0 +1,26 @@
+{
+  "solution": {
+    "path": "Mapsui.sln",
+    "projects": [
+      "Benchmarks\\Mapsui.Rendering.Benchmarks\\Mapsui.Rendering.Benchmarks.csproj",
```
pauldendulk, line 5: When do you use this individual filters? I prefer to have the minimal number of filters we need for our workflow so that there is less to maintain.

Mapsui.Rendering.Skia/Cache/LabelCache.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Mapsui.Rendering.Skia.Cache
+{
+    public class LabelCache : ILabelCache
+    {
+        private readonly Dictionary<Font, object> _cacheTypeface = new();
+        
+        private readonly Dictionary<(string? Text, Font Font, Brush? BackColor, Color ForeColor, float Opacity), IBitmapInfo> _labelCache = new();
+        
+        public T GetOrCreateTypeface<T>(Font font, Func<Font, T> createTypeFace)
```
pauldendulk, line 16: Perhaps it is possible to threat a label like a style in which a feature is rendered, to avoid any specific code for labels.

Mapsui.Rendering.Skia/Cache/RenderCache.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+using Mapsui.Rendering.Skia.Extensions;
+using Mapsui.Styles;
+
+namespace Mapsui.Rendering.Skia.Cache
+{
+    public class RenderCache : IRenderCache
```
pauldendulk, line 7: This the parent of all caches. Good that have that in one place. Makes it easier to understand.

pauldendulk (approved):

inforithmics:
@pauldendulk thanks for the review. 

**Caching**

I think there are 2 Caches Memory and Persistent But they could implement a general Caching interface like


 public TR GetOrCreate<TP1, TR, T>(TP1 parameter1, Func<TP1, TR> pureFunc>
 public TR GetOrCreate<TP1, TP2, TR, T>(TP1 parameter1,  TP2 parameter2, Func<TP1, TP2, TR> pureFunc>
....

So the Cache has severel GetOrCreate Functions, and the parameters for createing the cached value and the purefunction are paremters of the call like GetOrCreateTypeFace (But general without speicifc implementations.

With this 
LabelCache
RenderCache using LRUCache
VectorCache
SymbolCache
Can be unified to a Memory Cache and which would impelment an LRUCache.
UrlPersistentCache

Could implement the same interface lets call it IMCache.

**Two Step Rendering**

I thought about this too. And a not so invasive implementation could be to Change the SKCanvas to an ICanvas.
Where only Certain Draw operations without transformations are allowed, or Draw Operation with the Transformation.

There are two Implementations of ICanvas one directly rendering and one Putting the items to a ConcurrentBag. Which the Real renderer uses to draw the SkObjects that are created up to this point.

The advantage of this design is that the creation of SKObjects can be paralleized. Because an SkCanvas can only used by one thread at a time to draw items.

**The implementation of value equals is hard to maintain**

Maybe a sourcegenerator can be used for that either writing one for mapsui only
or using something like this
https://github.com/diegofrata/Generator.Equals
(This introduces no addtional nuget package dependencies because its only a compile time dependency).

**IFeature.RenderedGeometry**

Intresting didn't know that have to investigate this.

pauldendulk:
> @pauldendulk thanks for the review.
> **The implementation of value equals is hard to maintain**
> 
> Maybe a sourcegenerator can be used for that either writing one for mapsui only or using something like this https://github.com/diegofrata/Generator.Equals (This introduces no addtional nuget package dependencies because its only a compile time dependency).

Did not investigate this, but this might be a very good idea!


> **IFeature.RenderedGeometry**
> 
> Intresting didn't know that have to investigate this.

There is no real need for that, and the current implementation is broken, but it is good to know there was some experience with caching in the past. I think for two step rendering it would be better to fully separate the feature and the drawables.

I will have to further look into your other suggestions.

charlenni:
I don't think so. I would say, the Feature and the Drawable are always connected. The approach of @inforithmics is good, but only a workaround. At the end, the renderer for this special type of Feature is the only, who knows how to convert a Feature into a Drawable. I think, there are normally no two Features, that have the same Drawable. With this in mind, a central cache isn't very good. The RenderedGeometry is what the render process should fill (convert Feature to a SKPath), and the drawing process takes the RenderedGeometry and draws it on the screen (SKCanvas.DrawPath). Thomas does this without any knowledge, which data has to be rendered. And the live time of this caching is also a problem. If you destroy the Feature, the entry in the cache still exists.

inforithmics:
Rendered Geometry is a good idea to use. But the Key needs to be changed to (IStyle, IViewPort). Because the IViewport influences the rendering of the Geometry. (Like Scaling, Limiting, ...)

pauldendulk:
With RenderedGeometry it became important to dispose features because they had pointers to disposable skia objects. When features needs to be disposed all owners of the features need to be disposed. I would like to avoid that.

pauldendulk:
> But the Key needs to be changed to (IStyle, IViewPort). 

This is something I had not considered yet. This could make things more complex. It would also mean that while panning and zooming (when we want optimal performance) the clipped edge of the polygon will be visible, or it will be shown just in a single iteration.

PR #1747, opened 2022-11-01
Mapsui.Ui.Blazor

inforithmics:
Second Try with Blazor support


inforithmics:
The Shapefiles Samples don't work yet on Blazor (Need to make the deployment the same as mbtiles)

inforithmics:
Fixes: 
1. Fixed Pathes in Desktop samples (were Windows Convention)
2. Fixed BrotliCompression is not supported on every Platform (Wasm for example in this case save uncompressed)

inforithmics:
Most Desktop samples work now (Bugfixes for other platforms too). 
So because next Version is a Beta Version then this can be merged in my opninion.

janusw:
I just tried this branch. I was able to open Mapsui.Blazor.slnf in VS4Mac, build Mapsui.Samples.Blazor and run it in Firefox (on MacOS). I can confirm that most samples seem to work pretty well, which is awesome.
But I also noticed some problems/limitations:
* The navigation buttons are not visible, therefore zooming is not possible (panning via mouse works fine, though).
* When selecting the samples "3 WMTS" or "8 TMS openbasiskaart" from the "Data" category, I get the message: "An unhandled error has occurred."
* When selecting the samples "4 WMS called tiled", "5 WMTS Michelin" or "6 WMS" from the "Data" category, I only see a white screen (no map rendering) but no error message.

IMHO the fact that some specific samples do not work is not a big problem (and should not necessary block the merging of this PR), but having the navigation buttons working would be important, I guess.

Thanks for all your work here, @inforithmics! This is great progress for Mapsui!

janusw:
> I just tried this branch. I was able to open Mapsui.Blazor.slnf in VS4Mac, build Mapsui.Samples.Blazor and run it in Firefox (on MacOS).

When running the samples in Google Chrome, "5 WMTS Michelin" seems to work, but the other samples mentioned above behave the same as with Firefox.

janusw:
In Chrome also two-finger scrolling seems to work, but it's a bit tricky to use.

In the "OpenStreetMap" sample in the "Demo" category, buttons for zooming in and out are visible, but they don't seem to work (neither in Firefox, nor in Chrome).

pauldendulk (approved):
I will merge this but did not review it. Since it is mostly Blazor I assume it will not break the existing functionality. Please look at the comment of @janusw.

pauldendulk:
@inforithmics I will merge it if the conflicts are resolved.

inforithmics:
@janusw thanks for the review will look at the specific samples.

inforithmics:
I looked at the samples, 
1 Sample uses Webrequest (does not work (will change it to httpclient call)
And the httpclient calls that do not work seem to have missing credentials.

inforithmics:
The Zooming Buttons come from the MapView (which is currently only availabe in Forms and Maui)

Here in this pull request I try to make MapView available in all Targets
https://github.com/inforithmics/Mapsui/tree/MapView

But it isn't working yet. When there is something usefull I'll make a pull request. I found workarounds so that on most platform the databinding does work with some compiler trickery without ugly #if #endif code.

inforithmics:
But I discovered that the zooming and rotation widgets are not displayed have to investigate why this is the case.

inforithmics:
fixed the tms wmts michelin and arcgis samples in blazor. 

The wmts and wms tiled samples seem not to work in blazor.

the wms sample seems generally not work I have to investigate here.

pauldendulk:
**About Webrequest.** I worked on BruTile in december and it was still using Webrequest in a number of places. I did not want to fix that at that point but maybe I should.

PR #1748, opened 2022-11-02
fixed wrong skiasharp versions in nuspec files

inforithmics:
During the merging I saw that in 3 nuspec files wrong skia sharp versions were specified.
Fixing this now


inforithmics:
When I opend the solution in Rider it complained about a vulnerabilty
see here
https://www.nuget.org/packages/System.Text.RegularExpressions/4.3.0
I upgraded it to 
https://www.nuget.org/packages/System.Text.RegularExpressions/4.3.1

pauldendulk (approved):

PR #1749, opened 2022-11-06
Use Net7

inforithmics:
Target Net7 (Improves things for Blazor)


inforithmics:
This pull request needs that Github Action and Appveyor are updated to Visual Studio 2022 17.4 (.Net 7 support).

inforithmics:
Added .net 7 targets to maui and uno because I discoverd today that maui is not binary compatible from .net6 to .net7 so it needs to be compiled for the specific framework.
and uno has wasm which is not binary compatible too between .net6 and .net7
I updated uno to the latest stable nuget version 4.6.19

inforithmics:
This pull reqeust is for blazor too because it sets the used .net sdk to 7.0 for allowing blazor with target 7.0 to be built

pauldendulk (approved):
Thanks a lot

PR #1751, opened 2022-11-10
Fix warnings and WFS Sample in Wasm and ArcGis Sample

inforithmics:
Fixing Warnings and replacing WebRequest Call in WFSProvider with HttpClient because in WASM 
WebRequest is not implmented and does not work (is already deprecated)


inforithmics:
Currently I'm fixing the WFS Sample the coordinates seem to be off.

inforithmics:
If Fixed the Arcgis image example and fixed some small bugs in the ArcGis

inforithmics:
This pull request is for the blazor implementation, because the Webrequest calls are replaced with httpclient calls so that the samples work in Blazor

pauldendulk (approved):

PR #1757, opened 2022-11-13
Shape file deployer 

inforithmics:
Created Deployers for Shapefiles and Geotiff and MapTiles so that the Desktop samples can be used in Mobile and Blazor


inforithmics:
Main Changes:
1. Added the Desktop samples to all Sample Projects because they are now mobile friendly
2. Added to the Regression Sample tests the Mapsui.Tests.Common Samples

pauldendulk (approved):
Wonderful!. This makes things easier to maintain as well. A next step is to remove the Desktop project (which was intended for those samples which were not cross platform).

PR #1763, opened 2022-12-03
Improve the doc generation on github actions

inforithmics:
only run on changes in docfx and don't install .net 7 (is already insalled)


pauldendulk (approved):
Thanks for the fixes

PR #1764, opened 2022-12-04
Allow BoundingBox to be null when calling ExecuteIntersectionQueryAsync to allow search with OGCFilter across the whole map

JBou:
The null checks are currently not consistent across the WFS 1.1.0 and WFS 1.0.0 requests.
The following code was possible in v3, but with v4 it's only possible when using WFS version 1.0.0:

**I made the bounding box nullable to allow the following queries:**

```csharp
var ogcFilter = new OGCFilterCollection();
ogcFilter.AddFilter(new PropertyIsEqualToFilter_FE1_1_0("PPOL_CCAT_CODICE", "123"));
ogcFilter.AddFilter(new PropertyIsEqualToFilter_FE1_1_0("PPOL_CODICE", "55/1"));
ogcFilter.Junctor = OGCFilterCollection.JunctorEnum.And;

_wfsProvider.OgcFilter = ogcFilter;

var features = _wfsProvider.ExecuteIntersectionQuery(null);
_wfsProvider.OgcFilter = null;

return features;
```


inforithmics:
The changes look reasonable. But somehow the wfssample ski rendering test fails probably because the picture changed.

inforithmics:
It worked locally the reason was that the wfssample.sqlite (wfs cache hadn't all calls cached) so I updated it and it should work now.

inforithmics (approved):
I'll merge this when the build is green, the changes are reasonable and good. Good work

JBou:
Thanks for having a look! I appreciate the work you and the team is making :)

PR #1766, opened 2022-12-05
Prerender CalloutStyle only if it is visible

charlenni:
In the moment the `CalloutStyleRenderer` creates a bitmap even if the callout style isn't enabled. With this, the `CalloutStyleRenderer` render the callout and register a bitmap only, if the style is enabled.


inforithmics (approved):
is reasonable

PR #1768, opened 2022-12-14
Fix StyleCollection initialization

pauldendulk:
This fixes a bug I ran into in a project I worked on. If the StyleCollection is within a ThemeStyle it was ignored because of the default values. A sample and unit test were added.


pauldendulk:
The sample uses a ThemeStyle. If the feature in the ThemeStyle method is selected a StyleCollection is returned.

image

PR #1774, opened 2022-12-21
Fix check when StyleCollection directly on Layer

pauldendulk:
There was no check on the StyleCollection when it was directly on the layer. The 'ToLayer' method converted it to list of it's children, checking was done later. This showed up in a previous PR where it turned out the default StyleCollection was not Enabled.


Mapsui/Layers/BaseLayer.cs
```diff
@@ -21,18 +21,6 @@ public abstract class BaseLayer : ILayer
         private object? _tag;
         private MRect? _extent;
 
-        /// <summary>
-        /// Get a layer's styles
-        /// </summary>
-        /// <param name="layer">Layer, for which styles should be returned</param>
-        /// <returns>Enumerable with styles belonging to layer</returns>
-        public static IEnumerable<IStyle?> GetLayerStyles(ILayer? layer)
```
pauldendulk, line 29: Remove this method that is not tested and not used in our code. Users should use the GetStylesToApply IStyle extension instead.

Mapsui/Rendering/VisibleFeatureIterator.cs
```diff
@@ -28,59 +30,48 @@ public static class VisibleFeatureIterator
...
-                        if (styleForFeature == null) continue;
-                        style = styleForFeature;
-                    }
-
-                    if (ShouldNotBeApplied(style, viewport)) continue;
-
-                    if (style is StyleCollection styles) // The ThemeStyle can again return a StyleCollection
-                    {
-                        foreach (var s in styles)
+                        var stylesFromThemeStyle = themeStyle.GetStyle(feature).GetStylesToApply(viewport.Resolution);
```
pauldendulk, line 40: Now using the GetStylesToApply for the style directly on the layer and the style returned by the ThemeStyle. This fixes some inconsistencies.

Mapsui/Rendering/VisibleFeatureIterator.cs
```diff
@@ -28,59 +30,48 @@ public static class VisibleFeatureIterator
...
             }
 
             foreach (var feature in features)
             {
-                var featureStyles = feature.Styles ?? Enumerable.Empty<IStyle>(); // null check
+                var featureStyles = feature.Styles ?? Enumerable.Empty<IStyle>();
                 foreach (var featureStyle in featureStyles)
                 {
-                    if (ShouldNotBeApplied(featureStyle, viewport)) continue;
+                    if (featureStyle is IThemeStyle themeStyle)
```
pauldendulk, line 58: ThemeStyle and StyleCollection never worked on the IFeature.Style. I don't think it should because it makes no sense, ThemeStyle is not necessary because you already added a style for this specific feature, StyleCollection is not necessary because it already is an enumerable.

Tests/Mapsui.Tests/Rendering/VisibleFeatureIteratorTests.cs
```diff
@@ -11,21 +11,96 @@ namespace Mapsui.Tests.Rendering
...
+        public static IEnumerable<TestCaseData> FeatureStylesThatShouldBeAppliedOrNot
+        {
+            get
+            {
+                yield return new TestCaseData(new StyleCollection { Styles = { new VectorStyle() } }, false, "A StyleCollection on a feature should not be applied");
+                yield return new TestCaseData(new ThemeStyle(f => new VectorStyle()), false, "A ThemeStyle on a feature should not be applied");
+            }
+        }
+
+        [Test, TestCaseSource(nameof(StylesThatShouldBeAppliedOrNot)), TestCaseSource(nameof(LayerStylesThatShouldBeAppliedOrNot))]
```
pauldendulk, line 40: It was not possible to use the TestCase tag here because an instance of an object is created. The TestCaseSource construction makes it a bit more abstract, but I think this is quite powerful. I could test a lot of scenarios with just a few tests.

PR #1776, opened 2022-12-23
Fix rotation bug and do some rewriting

pauldendulk:


Mapsui.UI.Android/MapControl.cs
```diff
@@ -255,23 +255,10 @@ public void MapView_Touch(object? sender, TouchEventArgs args)
...
-                                    {
-                                        if (Math.Abs(_innerRotation) <= Math.Abs(ReSnapRotationDegrees))
-                                            rotationDelta = -Viewport.Rotation;
-                                        else
-                                            rotationDelta = _innerRotation - Viewport.Rotation;
-                                    }
+                                    _virtualRotation += angle - previousAngle;
+
+                                    rotationDelta = RotationCalculations.CalculateRotationDeltaWithSnapping(
+                                        _virtualRotation, _viewport.Rotation, _unSnapRotationDegrees, _reSnapRotationDegrees);
```
pauldendulk, line 261: The rotation code is now duplicated in 5 places. I moved part of the duplicated code into pure functions. So, less duplication. There could be more shared code.

Mapsui.UI.Forms/MapControl.cs
```diff
@@ -706,27 +706,15 @@ private bool OnTouchMove(List<MPoint> touchPoints)
...
 
-                        if (!(Map?.RotationLock ?? false))
+                        if (!(Map?.RotationLock ?? true))
                         {
-                            _innerRotation += angle - prevAngle;
-                            _innerRotation %= 360;
+                            var deltaRotation = angle - prevAngle;
+                            _virtualRotation += deltaRotation;
 
-                            if (_innerRotation > 180)
```
pauldendulk, line 714: It should not be necessary anymore to worry about normalization. This is done inside the function.

pauldendulk:
There are currently five places where the rotation code is duplicated. I don't think it is tested everywhere.
- Forms (XamarinForms and Maui) - tested
- Android - tested
- iOS - not tested but looks like it is correct.
- WPF - not tested but looks like it is not correct.
- UNO - not tested but looks like it is not correct.

PR #1781, opened 2022-12-31
Reorganize samples

pauldendulk:
Done:
- Remove the Desktop category
- Move some more samples to the Data category
- Add a Performance category

Todo:
The Mapsui.Samples.Common.Desktop should be removed entirely, and sample files should  move to other folders. Since this will cause conflicts I will put this PR on hold for now.


inforithmics:
Good idea. I think Git should be able to resolve the conflicts. Because it detects moves of Files.

pauldendulk:
> Good idea. I think Git should be able to resolve the conflicts. Because it detects moves of Files.

Thanks, I continued, removed the desktop project entirely and did a lot of cleanup of the Mapsui.Samples.Common project. Moved all the GeoData to one folder. Moved some other code to subfolders. Fixed all the warnings. Ran code cleanup. Finally could set TreatWarningsAsErrors to true in Mapsui.Samples.Common.csproj.

PR #1791, opened 2023-01-15
Improving wms test

inforithmics:
I improved the WMS Test so that a Visible Picture is tested. So to discover Regressions faster.

On the way to this I fixed some warnings and added some more log messages (Because during debugging it worked but not on running).
I to fixed that to many changed events where fired when the property didn't really change anything.


Benchmarks/Mapsui.Rendering.Benchmarks/RenderToBitmapPerformance.cs
```diff
@@ -124,7 +124,7 @@ public async Task RenderDefaultAsync()
             await map.WaitForLoadingAsync();
             using var bitmap = mapRenderer.RenderToBitmapStream(map.Viewport, map.Map!.Layers, Color.White);
 #if DEBUG
-            File.WriteAllBytes(@$"{OutputFolder()}\Test.png", bitmap.ToArray());
+            File.WriteAllBytes(@$"{OutputFolder()}\Test.png", bitmap!.ToArray());
```
pauldendulk, line 127: On most projects I work on we have turned nullalbe checks on, but I see a lot of possible null values are now hidden with the !. In this case you know the value will not be null, but I would rather rewrite the code a bit so that we can let the nullable check do the checking for us. In order to get there we need to check early and in most cases throw an exception. In most cases this will mean a problem is found early. In some cases it was not a real problem and the user will now run into an exception. This can lead to some frustration but I think in the long run it will be better for our users. 

In this case I am going to try to make RenderToBitmapStream instead of return null.

Mapsui/Layers/BaseLayer.cs
```diff
@@ -62,6 +63,7 @@ protected BaseLayer(string name)
             get => _tag;
             set
             {
+                if (_tag == value) return;
```
pauldendulk, line 66: All those checks on changes in property values are a good improvement.

Mapsui/Layers/ImageLayer.cs
```diff
@@ -42,6 +42,7 @@ private class FeatureSets
         private readonly Timer _startFetchTimer;
         private IProvider? _dataSource;
         private readonly int _numberOfFeaturesReturned;
+        private bool _busyExtent;
```
pauldendulk, line 45: I see you need to deal with the extent. This is a general problem. Also in Layer.cs the extent is fetched directly from the Provider which could be a web or database request. The caller of ILayer.Extent does not expect this and it is against the Microsoft guidelines. The easy solution is to make it an async method. I think it should be field instead and initializing it should be an async process that the users of ILayer do not need to wait for.

This is code from Layer.cs:
```csharp
        public override MRect? Extent
        {
            get
            {
                lock (_syncRoot)
                {
                    return DataSource?.GetExtent();
                }
            }
        }
```

pauldendulk (approved):

inforithmics:
I think too that the Extent should be set in an Asynchronous Init Call.

PR #1794, opened 2023-01-18
Throw if RenderToBitmapStream and GetSnapshot fail

pauldendulk:
I think we should throw more and allow less null. This may have consequences for our users:
- Sometimes they may have an implicit bug which now becomes apparent and they become happier.
- Sometimes they have a glitch that they don't care about and now they have to deal with the exception and they become less happy, but we tell them it is all for the greater good and we hope to get away with it.
- Sometimes we will make a mistake to throw and have to rethink our design. In some cases we will have to revert it in others we will have to rewrite other parts as well. The users will have more reason to be unhappy, but still I think trying to allow less null is for the greater good.

This PR does not allow null RenderToBitmapStream and GetSnapshot and throws instead. There are many other places where we could throw.

### DIFF
--- a/Benchmarks/Mapsui.Rendering.Benchmarks/RenderToBitmapPerformance.cs
+++ b/Benchmarks/Mapsui.Rendering.Benchmarks/RenderToBitmapPerformance.cs
@@ -124,7 +124,7 @@ namespace Mapsui.Rendering.Benchmarks
             await map.WaitForLoadingAsync();
             using var bitmap = mapRenderer.RenderToBitmapStream(map.Viewport, map.Map!.Layers, Color.White);
 #if DEBUG
-            File.WriteAllBytes(@$"{OutputFolder()}\Test.png", bitmap!.ToArray());
+            File.WriteAllBytes(@$"{OutputFolder()}\Test.png", bitmap.ToArray());
 #endif
         }
 
@@ -134,7 +134,7 @@ namespace Mapsui.Rendering.Benchmarks
             await pngMap.WaitForLoadingAsync();
             using var bitmap = mapRenderer.RenderToBitmapStream(pngMap.Viewport, pngMap.Map!.Layers, Color.White);
 #if DEBUG
-            File.WriteAllBytes(@$"{OutputFolder()}\Testpng.png", bitmap!.ToArray());
+            File.WriteAllBytes(@$"{OutputFolder()}\Testpng.png", bitmap.ToArray());
 #endif
         }
 
@@ -144,7 +144,7 @@ namespace Mapsui.Rendering.Benchmarks
             await webpMap.WaitForLoadingAsync();
             using var bitmap = mapRenderer.RenderToBitmapStream(webpMap.Viewport, webpMap.Map!.Layers, Color.White);
 #if DEBUG
-            File.WriteAllBytes(@$"{OutputFolder()}\Testwebp.png", bitmap!.ToArray());
+            File.WriteAllBytes(@$"{OutputFolder()}\Testwebp.png", bitmap.ToArray());
 #endif
         }
         
@@ -154,7 +154,7 @@ namespace Mapsui.Rendering.Benchmarks
             await skpMap.WaitForLoadingAsync();
             using var bitmap = mapRenderer.RenderToBitmapStream(skpMap.Viewport, skpMap.Map!.Layers, Color.White);
 #if DEBUG
-            File.WriteAllBytes(@$"{OutputFolder()}\Testskp.png", bitmap!.ToArray());
+            File.WriteAllBytes(@$"{OutputFolder()}\Testskp.png", bitmap.ToArray());
 #endif
         }              
 
--- a/Mapsui.Rendering.Skia/MapRenderer.cs
+++ b/Mapsui.Rendering.Skia/MapRenderer.cs
@@ -73,11 +73,10 @@ namespace Mapsui.Rendering.Skia
             Render(canvas, viewport, widgets, 1);
         }
 
-        public MemoryStream? RenderToBitmapStream(IReadOnlyViewport? viewport, IEnumerable<ILayer> layers, 
+        public MemoryStream RenderToBitmapStream(IReadOnlyViewport viewport, IEnumerable<ILayer> layers,
             Color? background = null, float pixelDensity = 1, IEnumerable<IWidget>? widgets = null, RenderFormat renderFormat = RenderFormat.Png)
         {
-            if (viewport == null)
-                return null;
+            if (viewport == null) throw new ArgumentNullException(nameof(viewport));
 
             try
             {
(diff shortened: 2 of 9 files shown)


Benchmarks/Mapsui.Rendering.Benchmarks/RenderToBitmapPerformance.cs
```diff
@@ -124,7 +124,7 @@ public async Task RenderDefaultAsync()
             await map.WaitForLoadingAsync();
             using var bitmap = mapRenderer.RenderToBitmapStream(map.Viewport, map.Map!.Layers, Color.White);
 #if DEBUG
-            File.WriteAllBytes(@$"{OutputFolder()}\Test.png", bitmap!.ToArray());
+            File.WriteAllBytes(@$"{OutputFolder()}\Test.png", bitmap.ToArray());
```
pauldendulk, line 127: Now the ! can be removed.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -73,11 +73,10 @@ public MapRenderer()
             Render(canvas, viewport, widgets, 1);
         }
 
-        public MemoryStream? RenderToBitmapStream(IReadOnlyViewport? viewport, IEnumerable<ILayer> layers, 
+        public MemoryStream RenderToBitmapStream(IReadOnlyViewport viewport, IEnumerable<ILayer> layers,
             Color? background = null, float pixelDensity = 1, IEnumerable<IWidget>? widgets = null, RenderFormat renderFormat = RenderFormat.Png)
         {
-            if (viewport == null)
-                return null;
+            if (viewport == null) throw new ArgumentNullException(nameof(viewport));
```
pauldendulk, line 79: Although viewport is not nullable users could still pass in a null value, so, there is reason to check for null. But if we apply this argument consistently we should add null checks to all of our public methods, which I don't want to do. So I think this check should just be seen as an extra check because it could have been null before.

inforithmics:
I generally like your approach. To validate the input parameters and fail fast, because it is much easier to debug, when the call fails immediatly when the parameters are wrong instead of later, with an nullreferenceexception.

The wmssample skiasharp rendering test is somehow unreliable.

Working Debug Output: (Debug)

LogLevel: Debug Message: Viewport Extent Changed: BL: (X=4383.601354761078,Y=357546.62554659054)  TR: (X=286028.2781357203,Y=568780.13313231)
LogLevel: Debug Message: Viewport Extent Changed: BL: (X=-140410.33839047962,Y=-104300.75379285972)  TR: (X=141234.33839047962,Y=106932.75379285972)
LogLevel: Debug Message: Viewport Extent Changed: BL: (X=-136438.73703571854,Y=-104300.75379285972)  TR: (X=145205.9397452407,Y=106932.75379285972)
LogLevel: Debug Message: Viewport Extent Changed: BL: (X=-136438.73703571854,Y=199305.01919615606)  TR: (X=145205.9397452407,Y=410538.5267818755)
LogLevel: Debug Message: Refresh Data: Resolution: 352.05584597619907 Change Type: Discrete Extent: BL: (X=-136438.73703571854,Y=199305.01919615606)  TR: (X=145205.9397452407,Y=410538.5267818755) 
LogLevel: Debug Message: Start image fetch at 06:39:27.6314102
LogLevel: Debug Message: Cached Load Url https://geodata.nationaalgeoregister.nl/windkaart/ows?SERVICE=WMS&REQUEST=GetMap&BBOX=-158970.3111781953,176773.44505367934,167737.51388771745,433070.10092435224&WIDTH=928&Height=728&Layers=windsnelheden100m&FORMAT=image/png&CRS=EPSG:28992&VERSION=1.3.0&TRANSPARENT=true&Styles=
LogLevel: Debug Message: Finished image fetch at 06:39:27.6608196

Not Working Unit Test Output: (Debug)

LogLevel: Debug Message: Viewport Extent Changed: BL: (X=12,Y=1016)  TR: (X=812,Y=1616)
LogLevel: Debug Message: Viewport Extent Changed: BL: (X=3983.6013547610783,Y=1016)  TR: (X=4783.601354761078,Y=1616)
LogLevel: Debug Message: Viewport Extent Changed: BL: (X=3983.6013547610783,Y=304621.7729890158)  TR: (X=4783.601354761078,Y=305221.7729890158)
LogLevel: Debug Message: Refresh Data: Resolution: 1 Change Type: Discrete Extent: BL: (X=3983.6013547610783,Y=304621.7729890158)  TR: (X=4783.601354761078,Y=305221.7729890158) 
LogLevel: Debug Message: Start image fetch at 06:41:24.4772173
LogLevel: Debug Message: Cached Load Url https://geodata.nationaalgeoregister.nl/windkaart/ows?SERVICE=WMS&REQUEST=GetMap&BBOX=3919.6013547610783,304557.7729890158,4847.601354761078,305285.7729890158&WIDTH=928&Height=728&Layers=windsnelheden100m&FORMAT=image/png&CRS=EPSG:28992&VERSION=1.3.0&TRANSPARENT=true&Styles=
LogLevel: Debug Message: Finished image fetch at 06:41:24.4789687

As you can see the Extents are differnt, I think this is a Timing issue. So I think I will make a pull request to allow the DataSource to be set asynchronous.

SetDataSourceAsync

Benchmarks/Mapsui.Rendering.Benchmarks/RenderToBitmapPerformance.cs
inforithmics: Better be explicit. because if it is null something else has happened (For example on my machine it sometimes doesn't render a Bitmap because the viewport.Extent is width=0 and height=0, I think it is a timing issue

Mapsui.Rendering.Skia/MapRenderer.cs
inforithmics: Yes public interfaces should be always checked that the user adheres to the contract.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -86,73 +85,64 @@ public MapRenderer()
...
+                            using var pictureRecorder = new SKPictureRecorder();
+                            using var skCanvas = pictureRecorder.BeginRecording(new SKRect(0, 0, Convert.ToSingle(width), Convert.ToSingle(height)));
+                            RenderTo(viewport, layers, background, pixelDensity, widgets, skCanvas);
+                            using var skPicture = pictureRecorder.EndRecording();
+                            skPicture?.Serialize(memoryStream);
+                            break;
+                        }
                     case RenderFormat.Png:
-                    {
-                        using var surface = SKSurface.Create(imageInfo);
```
inforithmics, line 107: intrestingly the surface can be null when the imageinfo has width=0 and height=0 (Failing unit test on my machine).

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -86,73 +85,64 @@ public MapRenderer()
 
                 var imageInfo = new SKImageInfo((int)Math.Round(width * pixelDensity), (int)Math.Round(height * pixelDensity),
```
inforithmics, line 86: maybe throw if width or height is 0. because this is not a real image with 0 width or 0 height

inforithmics (approved):

pauldendulk:
@inforithmics Thanks for the analysis of the WMS, not easy to find.

PR #1801, opened 2023-01-22
Remove the Async setting of the Extent because the it is a sync metho…

inforithmics:
…d meaning it doesn't take to long

### DIFF
--- a/Mapsui.UI.Blazor/Extensions/MapsuiJsInterop.cs
+++ b/Mapsui.UI.Blazor/Extensions/MapsuiJsInterop.cs
@@ -16,6 +16,11 @@ public class MapsuiJsInterop : IAsyncDisposable
 
     public MapsuiJsInterop(IJSRuntime jsRuntime)
     {
+        if (jsRuntime == null)
+        {
+            throw new ArgumentNullException(nameof(jsRuntime));
+        }
+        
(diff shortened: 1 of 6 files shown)


inforithmics:
This pull request removes the async setting of the Extent in the ImageLayer. Because the GetExtent method is synchronous this Method doesn't take to long, if it would be an http call the method had to be asynchrounous or this wouldn't work in Blazor or uno.wasm.

https://github.com/Mapsui/Mapsui/blob/master/Mapsui/Providers/Wms/WmsProvider.cs

The WmsProvider is Create with CreateAsync Call (there the extent gets set).

inforithmics:
The alternative would be to make GetExtent and asynchrouns call => Task<MRect?> GetExtentAsync. I could that too but it would make the change bigger, because every sample and provider would be affected.

inforithmics:
I'm investigating why the RasterizingLayer sample crashes

inforithmics:
The problem was that the RegressionMapControl (used for the Regression Unit Tests the PixelDensity 0 was which caused the Test to fail).

pauldendulk (approved):
Without studying the details, I would think the ILayer should be synchronous an the IProvider asynchronous. So, the ILayer.Extent should never directly call the IProvider.GetExtent()

PR #1802, opened 2023-01-23
Build docs on GHA

pauldendulk:


pauldendulk:
Merge because the 'Run workflow' button of workflow_dispatch will only show when the code is on the main branch.

PR #1806, opened 2023-01-28
add tiling dependency to ui projcts

inforithmics:
Add Tiling Reference to Ui Projects so that the nuget packages have tiling dependency too. 
So that the nuget packages have the same dependency as with the nuspec files


pauldendulk (approved):
Great that you noticed it was missing.

PR #1809, opened 2023-01-29
Added Geo json Samples (existing cities.geojson)

inforithmics:


pauldendulk (approved):
This will be a very useful sample to many.

PR #1814, opened 2023-02-01
Fix error caused by auto merge in previous PR

pauldendulk:


pauldendulk:
This fixes a flaw introduced by the auto merge yesterday at the end of the day. I was quite confused. The errors seemed to be unrelated. I thought it was caused by the workload and VS updates I did during the day.

After this (and mainly the previous) PR is will be possible to release nugets from github actions. This will also replace the nuspec with csproj for generating nupkg, thanks to an earlier PR by Thomas.

@inforithmics It is now possible to build everything with dotnet except for UnoUwp. In [this action](https://github.com/Mapsui/Mapsui/blob/master/.github/workflows/dotnet-release-nugets.yml) dotnet pack is run on a slnf without UnoUwp and UnoUwp is built in a separate step. Could this approach also be used in this [general build action](https://github.com/Mapsui/Mapsui/blob/master/.github/workflows/dotnet.yml) where we have many separate steps, or does it miss something?

PR #1815, opened 2023-02-02
add ios targets to uno

inforithmics:


Directory.Packages.props
```diff
@@ -152,22 +152,22 @@
...
     <PackageVersion Include="Uno.UniversalImageLoader" Version="[1.9.36,2.0.0)" />
-    <PackageVersion Include="Uno.UI" Version="[4.6.19,5.0.0)" />
-    <PackageVersion Include="Uno.UI.Adapter.Microsoft.Extensions.Logging" Version="[4.6.19,5.0.0)" />
-    <PackageVersion Include="Uno.UI.RemoteControl" Version="[4.6.19,5.0.0)" />
-    <PackageVersion Include="Uno.UI.Skia.Gtk" Version="[4.6.19,5.0.0)" />
-    <PackageVersion Include="Uno.UI.Skia.Linux.FrameBuffer" Version="[4.6.19,5.0.0)" />
-    <PackageVersion Include="Uno.UI.Skia.Tizen" Version="[4.6.19,5.0.0)" />
-    <PackageVersion Include="Uno.UI.Skia.Wpf" Version="[4.6.19,5.0.0)" />
-    <PackageVersion Include="Uno.UI.WebAssembly" Version="[4.6.19,5.0.0)" />
+    <PackageVersion Include="Uno.UI" Version="[4.7.30,5.0.0)" />
```
nickrandolph, line 155: Is there a reason to force update to 4.7? For libraries I would try to avoid this unless there are breaking changes that prevent it from working.

inforithmics: Was only a trial to see if the mac build would pass with this changes. But I reverted it because other things failed

inforithmics:
Fixed uno ios and android samples to run on macos. 
1. I had to remove a workaround that wasn't anymore necessary (nuget restore in prebuild event).
2. I had to increase xcode to 14.2 (but legacy still needs 14.0) so it is switch during the build from 14.0 for legacy to 14.2 to newer components.
3. I renenabled the Mapsui.ios.sample compilation it works now (after usting 14.2 xcode)

pauldendulk (approved):
Thanks a lot. I would not have the time/means/knowledge to fix this.

Perhaps the dotnet.yml can be simplified at some point. We had to fix all kinds of issues over the years. Some workarounds might not be necessary anymore and some complications can perhaps be fixed in the csproj or by using an slnf.

In the dotnet-release-nugets.yml I use a single pack on the NoUnoUwp slnf and then only msbuild step for UnoUwp: https://github.com/Mapsui/Mapsui/blob/c49ad15022468119e883c035008d23a9b0a61a67/.github/workflows/dotnet-release-nugets.yml#L49

PR #1826, opened 2023-02-08
Cleanup after changes to release logic

pauldendulk:


Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 ﻿<Project DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <Version>4.0.0-beta.6</Version>
+    <Version>4.0.0-beta.7</Version>
```
pauldendulk, line 4: Updated the version number but it is only for our local development since the git tag is used in the release and builds.

Mapsui.Avalonia.slnf
```diff
@@ -19,8 +19,7 @@
       "Tests\\Mapsui.Nts.Tests\\Mapsui.Nts.Tests.csproj",
       "Tests\\Mapsui.Rendering.Skia.Tests\\Mapsui.Rendering.Skia.Tests.csproj",
       "Tests\\Mapsui.Tests.Common\\Mapsui.Tests.Common.csproj",
-      "Tests\\Mapsui.Tests\\Mapsui.Tests.csproj",
-      "Tools\\VersionUpdater\\VersionUpdater.csproj"
+      "Tests\\Mapsui.Tests\\Mapsui.Tests.csproj"
```
pauldendulk, line 21: I updated all the sln's and slnf's by editing them in VS code. I only tested Mapsui.sln. I could have broken something.

Scripts/Mac/buildpack.sh
```diff
@@ -1,11 +1,12 @@
+# REM We have a new way of releasing therefor I remove the nuspec and VersionUpdater references.
```
pauldendulk, line 1: Perhaps buildpack.sh can be deleted.

Directory.Packages.props
```diff
@@ -33,23 +33,23 @@
...
-    <PackageVersion Include="Mapsui.Maui" Version="4.0.0-beta.6" />
-    <PackageVersion Include="Mapsui.Nts" Version="4.0.0-beta.6" />
-    <PackageVersion Include="Mapsui" Version="4.0.0-beta.6" />
-    <PackageVersion Include="Mapsui.Rendering.Skia" Version="4.0.0-beta.6" />
-    <PackageVersion Include="Mapsui.Tiling" Version="4.0.0-beta.6" />
-    <PackageVersion Include="Mapsui.Uno" Version="4.0.0-beta.6" />
-    <PackageVersion Include="Mapsui.Uno.WinUI" Version="4.0.0-beta.6" />
-    <PackageVersion Include="Mapsui.WinUI" Version="4.0.0-beta.6" />
-    <PackageVersion Include="Mapsui.Wpf" Version="4.0.0-beta.6" />
+    <PackageVersion Include="Mapsui.Android" Version="4.0.0-beta.7" />
```
pauldendulk, line 36: @inforithmics We reference our own packages. I think you mentioned this was for testing. We do not use our own packages anywhere though. So, it is just for testing if they have been published correctly?

pauldendulk:
There is more cleanup to do but I think it is better to wait for 5.0.0-beta.1 when we go full .NET 6.

PR #1827, opened 2023-02-08
Use list of key values in popup and use pin style

pauldendulk:
This popup is a basic solution for many practical purposes. For me it seems like the most common scenario, something you start out with, but we had no samples for this yet. If users want something really pretty they need to create a platform dependent popup. 

It looks like this now:
image

Some future improvements:
- An outline for svg can be important if you have many symbols overlapping. It can help to make them transparent. It is always to use your own svg which can have all kinds of shapes and colors in it.
- The current CalloutStyle has title which is now based on the feature. To allow this we create a style for every individual feature. It would be better that have a single style on the layer and use the features data only when rendering.


pauldendulk:
@charlenni I think you refer to "If users want something really pretty they need to create a platform dependent popup."? Sorry, was that was not carefully formulated. What I was trying to say is that you can do anything that you want if you implement a platform dependent popup. I thought that was important because Mapsui has no limitation in that regard. It offers some options but you can also do anything else if you would want that. Implementing a platform specific popup is also not very hard (I think we should derive from AbsoluteLayout in MAUI to make it easier). This sample was intended to show Mapsui itself also has some options that are easy to get started with. There is also the 'Info | Custom Callout' sample, and perhaps the Callouts in the Forms samples could also be under the 'Info' category. Perhaps there should be a page in the documentation that explains the general mechanism and lists the options.

btw, the 'Info | Custom Callout' sample uses bitmaps and then shows them as image. The svg option is similar but uses svg instead of bitmap?

charlenni:
I talked about the SVG option, which should already working.

pauldendulk:
I would like to add a sample for that. Do we have a sample or test for this in our current code? Or do you have some pointers about how this should work? How should the the current custom callout sample be changed? 

https://github.com/Mapsui/Mapsui/blob/3ebdd31b80f64694f65ebd31a57e1881694727f3/Samples/Mapsui.Samples.Common/Maps/Callouts/CustomCalloutSample.cs#L70

PR #1828, opened 2023-02-08
Add single image sample

pauldendulk:
A simple scenario that I encountered several times in recent use cases.


pauldendulk:
This is what it looks like:

image

PR #1829, opened 2023-02-08
Upgrade Avalonia to 11.0.0-rc1.1

pauldendulk:
Was hoping for a quick upgrade but there are some changes that I don't know how to deal with yet.


Mapsui.UI.Avalonia/MapControl.cs
```diff
@@ -284,17 +287,19 @@ public bool Equals(ICustomDrawOperation? other)
...
-            var canvas = (context as ISkiaDrawingContextImpl)?.SkCanvas;
-            if (canvas == null)
-                context.DrawText(Brushes.Black, new Point(), _noSkia.PlatformImpl);
-            else
-            {
-                canvas.Save();
-                _mapControl.CommonDrawControl(canvas);
-                canvas.Restore();
-            }
+            //!!!
```
pauldendulk, line 293: No idea yet how direct skia rendering should work in 11.0.0-preview5. Did no research either.

Samples/Mapsui.Samples.Avalonia/Mapsui.Samples.Avalonia.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <OutputType>WinExe</OutputType>
-    <TargetFramework>net6.0-windows10.0.19041.0</TargetFramework>
+		<OutputType>WinExe</OutputType>
+		<TargetFramework>net6.0</TargetFramework>
```
pauldendulk, line 5: Perhaps I should generate a 11.0.0-preview5 project from scratch. Not sure how to install those templates.

Samples/Mapsui.Samples.Avalonia/Program.cs
```diff
@@ -21,7 +21,7 @@ public static AppBuilder BuildAvaloniaApp()
                        .UsePlatformDetect()
                        .With(new Win32PlatformOptions
                        {
-                           EnableMultitouch = true,
+                           // Not in v11. Is there an alternative?: EnableMultitouch = true,
```
pauldendulk, line 24: Not blocking if missing in the first version.

Mapsui.UI.Avalonia/MapControl.cs
IldarKhayrutdinov: maybe needs to wait for RC?
https://github.com/AvaloniaUI/Avalonia/discussions/10193#discussioncomment-4868225

IldarKhayrutdinov: I tried to update to preview in my project, there are really a serious API changes

Samples/Mapsui.Samples.Avalonia/Program.cs
maxkatz6: No, it's always enabled.

Samples/Mapsui.Samples.Avalonia/Mapsui.Samples.Avalonia.csproj
maxkatz6: You need to add Avalonia.Themes.Fluent or Avalonia.Themes.Simple nuget package (depending on which theme is used in the sample). Other than that, for sample purposes, it looks good.

maxkatz6:
Also, `<FluentTheme />` should be restored. Mode property was removed, it's now Application.RequestedThemeVariant. Not sure if Mapsui needs any special support for dark/light theme. If so, you can ping me for help.

Mapsui.UI.Avalonia/MapControl.cs
maxkatz6: Please check this page - https://github.com/AvaloniaUI/Avalonia/wiki/Breaking-Changes - it should contain all important breaking changes. 

"ISkiaDrawingContextImpl" isn't there, as it never was a proper public API. More like a hack to provide access to the Skia. 
Now there is a proper public API.

maxkatz6: About Skia feature answered here https://github.com/AvaloniaUI/Avalonia/issues/10304

pauldendulk:
@rgustv @maxkatz6 @IldarKhayrutdinov Thanks a lot for you input. We should have this PR fixed and it may not be that hard. But perhaps it is better to wait with the merge (not sure how long) and keep master on 0.10.18.

inforithmics:
I fixed the Avalonia Rendering now.
1) Use SimpleTheme as Theme
```
<Application.Styles>
        <SimpleTheme />
    </Application.Styles>
```
2) AllowEglInitialization = false or else only a Transparent Window is displayed (Windows Dev Kit 2023)
`AllowEglInitialization = false`
3) Get the SkiaCanvas correctly
```
            var leaseFeature = context.GetFeature<ISkiaSharpApiLeaseFeature>();
            if (leaseFeature == null)
                return;
            using var lease = leaseFeature.Lease();
            var canvas = lease.SkCanvas;
            canvas.Save();
            _mapControl.CommonDrawControl(canvas);
            canvas.Restore();
```

thanks for the Tips @maxkatz6

inforithmics:
@pauldendulk maybe we should wait to merge this until Avalonia 11 is stable, but I don't have a strong opinion about it. So because Mapsui is itself in Beta we can have a preview reference itself.

pauldendulk:
> @pauldendulk maybe we should wait to merge this until Avalonia 11 is stable, but I don't have a strong opinion about it. So because Mapsui is itself in Beta we can have a preview dependency itself.

Lets wait, at least don't merge right now. Reasons:
- I don't have a good impression of how stable 11 is and how long it will take before they will release.
- I assume many Avalonia developer are still using 0.12.18. They would not have an option to use Mapsui.

pauldendulk:
@inforithmics and @maxkatz6 Thanks a lot for fixing this!

Mapsui.UI.Avalonia/MapControl.cs
```diff
@@ -49,22 +51,30 @@ private void Initialize()
...
+        base.OnPropertyChanged(change);
+
         switch (change.Property.Name)
         {
             case nameof(Bounds):
                 // size changed
                 MapControlSizeChanged();
                 break;
-        }
+        }
```
robloo, line 77: Note that Avalonia now has the SizeChanged event: https://github.com/AvaloniaUI/Avalonia/blob/cbe0bb861a767b7bf7b268bbaa1d89dcf8523aea/src/Avalonia.Controls/Control.cs#L203-L210

It is no longer necessary to watch for Bounds changes but you are free to still do so.

inforithmics:
Things done:
Upgraded to Preview 6

inforithmics:
Things done:
I fixed the dark mode in the sample (it was white text on white background) now it changes according to theme.

inforithmics:
Things done:
Updated to Preview7

pauldendulk:
We have created a separate package called Mapsui.Avalonia.V0. So, the sequel to Mapsui.Avalonia-rc.2 will be in Mapsui.Avalonia.V0-rc.3. While Mapsui.Avalonia-rc.3 will contain this V11 version. Thanks everybody who contributed to this branch.

slater1:
very good news, I will start trying it out!

slater1:
Working like a charm on Windows Desktop, Android, and iOS Simulator (real iOS currently broken by .NET runtime)

pauldendulk:
> On macOS can't seem to zoom in, only out

Is that mouse wheel zoom?

slater1:
Correct, mouse wheel up and down only zoomed out, at least in my app. Haven't tried the example.

robloo:
macOS sometimes does Y coordinate flipping. Avalonia should abstract that away though. I also had a look at what ScottPlot does and they don't seem to account for anything special.

https://github.com/ScottPlot/ScottPlot/blob/ceb1f6531969cba588f17097fbaaa1d63a94d32a/src/ScottPlot5/ScottPlot5%20Controls/ScottPlot.Avalonia/AvaPlot.axaml.cs#L171-L180

slater1:
Ran the Mapsui.Samples.Avalonia on macOS, zoom works fine. False alarm.

PR #1832, opened 2023-02-11
UpdateExtent and Call ViewPort Changes on Transform

inforithmics:
Maybe this fixes this.
https://github.com/Mapsui/Mapsui/issues/1830

Issue #1830, opened 2023-02-09
[4.0.0-beta.7 Regression] viewport changes not always detected


janusw:
**Describe the bug**

I recently tried updating my Xamarin.Forms application from Mapsui.Forms 4.0.0-beta.6 to 4.0.0-beta.7, and found a new problem that did not exist previously ...

I'm using a `MapView` and subscribe to `Viewport.ViewportChanged`, in order to trigger specific actions. But with beta.7 those actions do not seem to be triggered on every viewport change (only sometimes). With beta.6 (and earlier) this worked flawlessly.

I also checked the Mapsui logger, which sometimes shows messages like:

```
Viewport Extent Changed: BL: (X=...,Y=...)  TR: (X=...,Y=...)
```

But, again, these message do not appear on every viewport change, only sometimes.

I see the same behavior both on iOS and Android.

**Expected behavior**

`MapView.Viewport.ViewportChanged` should be triggered on every viewport change.

**Smartphone (please complete the following information):**
 - Device: iPhone 8, Samsung Galaxy S21
 - OS: iOS 16.3, Android 13

janusw:
A quick look through the git history leads me to suspect that PR #1791 (by @inforithmics) might be the culprit. In particular the following commits seem related:
* 5225885c79bb7047f3abeddd8cc197af1ff3a90f
* 4bcf4b6965d326406de2da42c88e549deebdc378
* fbc87f4e0660fadb7db362e00c9d4a94e04a3a5b

janusw:
I verified that Mapsui.Forms version 4.0.0-build.4666 (built via AppVeyor from 956875e15e0e5d2c8453f1ae199f31f8c8af39da) works well. This is the version just before the merge of #1791.

janusw:
Unfortunately the AppVeyor build failed on the merge commit of #1791, so I tried the last commit inside the PR (1011f4cde5ab299feb21420028055c9c9a7e8065), where AppVeyor succeeded and built package version 4.0.0-build.4672. This version indeed shows the same problem as 4.0.0-beta.7, which I take as evidence that #1791 actually caused this problem (as suspected).

pauldendulk:
Thanks for you excellent analysis. I am eager to work on this but I think won't have time the coming days. I would like to rewrite this part of the code to make problems like these likely.

inforithmics:
[...]

PR #1834, opened 2023-02-11
Version to 4.0.0-beta.8

pauldendulk:


janusw:
I would appreciate very much if beta.8 could wait for a proper fix of #1830!

pauldendulk:
> I would appreciate very much if beta.8 could wait for a proper fix of #1830!

Will wait for that!

janusw (approved):
#1830 appears to be fixed by #1832, so beta.8 is good to go from my side! 👍

Issue #1835, opened 2023-02-11
GHA: packages built on Windows don't have proper version


janusw:
**Describe the bug**

In the GitHub Actions CI build, the packages built on Mac ("nupkg.mac") have a proper version (determined via `git describe --tags`), but the packages build on Windows don't ("nupkg.win"). They seem to have the version of the last tag/release.

**To Reproduce**

As an example:
1. Go to https://github.com/Mapsui/Mapsui/actions/runs/4150324257
2. Navigate to Artifacts.
3. Download both.
4. Unzip.
5. Inspect the nupkg file names ...

Mac has: Mapsui.4.0.0-beta.7-89-g640c84e1c.nupkg
Win has: Mapsui.4.0.0-beta.7.nupkg

**Expected behavior**

Both package variants should have the same version (the one from Mac).

**Additional context**

It seems the problem is that the Win build uses the `buildpack` script to create the packages:

```
    - name: Build nuget packages
      run: Scripts\buildpack $(git describe --tags)
```

It receives the version as an argument, but apparently that version is not being used anywhere in the script.

pauldendulk:
This is something that I broke when removing the version updater. Here is a fix https://github.com/Mapsui/Mapsui/pull/1836. I tested it by running `scripts\buildpack.bat 4.0.0-beta.9` from the command line and to check if the packages are in Artifacts folder with the 4.0.0-beta.9 version (they were).

pauldendulk:
Over the years the dotnet.yml has grown more complex. I think it can be simplified now that the tools have improved. The dotnet-release-nugets.yml builds a solution with only UnoUwp using msbuild as an exception. Then again the dotnet.yml also does extra checks. Some of which I may not understand yet. For example, just yesterday I had the chance to look at the SamplesMapsuiNugetReferences.ps1, a nifty solution.

janusw:
Fixed via #1836. Thanks, @pauldendulk!

PR #1836, opened 2023-02-11
Fix version in winbuild for #1835

pauldendulk:
The versionupdater was removed without alternative. In this PR the /p:Version parameter is used.

### DIFF
--- a/Scripts/buildpack.bat
+++ b/Scripts/buildpack.bat
@@ -8,38 +8,38 @@ rmdir Release /s /q
 REM create Artifacts if not exists
 if not exist "Artifacts" mkdir Artifacts
 
-dotnet pack /p:RestorePackages=true /p:Configuration=Release Mapsui/Mapsui.csproj --output Artifacts
+dotnet pack /p:RestorePackages=true /p:Configuration=Release /p:Version=%Version% Mapsui/Mapsui.csproj --output Artifacts
 
-dotnet pack /p:RestorePackages=true /p:Configuration=Release Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj --output Artifacts
+dotnet pack /p:RestorePackages=true /p:Configuration=Release /p:Version=%Version% Mapsui.Rendering.Skia/Mapsui.Rendering.Skia.csproj --output Artifacts
 
-dotnet pack /p:RestorePackages=true /p:Configuration=Release Mapsui.Tiling/Mapsui.Tiling.csproj --output Artifacts
+dotnet pack /p:RestorePackages=true /p:Configuration=Release /p:Version=%Version% Mapsui.Tiling/Mapsui.Tiling.csproj --output Artifacts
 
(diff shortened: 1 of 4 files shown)


inforithmics:
The 2. restore dependencies should be changed to this and then it should run through.

- name: Restore dependencies
      run: .nuget\nuget restore Mapsui.sln
      continue-on-error: true

inforithmics:
Or to comment out following code
```
    - name: Cleanup
      run: git clean -fx -d -e Artifacts
    # Change Project References to nuget package references in samples    
    - name: nuget ProjectReferences to PackageReferences
      shell: pwsh
      run: |
        ./Scripts/SamplesMapsuiNugetReferences.ps1        
    - name: Restore dependencies
      run: .nuget\nuget restore Mapsui.sln      
```
But then the created nuget packages wouldn't be used to restore the samples (I used this to test the created nuget samples and found a few bugs along the way).

inforithmics:
I'll fix the project references to nuget packages script

Scripts/buildpack.bat
```diff
@@ -8,38 +8,38 @@ rmdir Release /s /q
...
 
-dotnet pack /p:RestorePackages=true /p:Configuration=Release Mapsui.UI.Eto/Mapsui.UI.Eto.csproj --output Artifacts
+dotnet pack /p:RestorePackages=true /p:Configuration=Release /p:Version=%Version% Mapsui.UI.Eto/Mapsui.UI.Eto.csproj --output Artifacts
 
-dotnet pack /p:RestorePackages=true /p:Configuration=Release Mapsui.UI.Maui/Mapsui.UI.Maui.csproj --output Artifacts
+dotnet pack /p:RestorePackages=true /p:Configuration=Release /p:Version=%Version% Mapsui.UI.Maui/Mapsui.UI.Maui.csproj --output Artifacts
 
-msbuild /p:RestorePackages=true /p:Configuration=Release Mapsui.UI.Uno/Mapsui.UI.Uno.csproj /t:Pack 
+msbuild /p:RestorePackages=true /p:Configuration=Release /p:Version=%Version% Mapsui.UI.Uno/Mapsui.UI.Uno.csproj /t:Pack 
 xcopy Mapsui.UI.Uno\bin\Release\*.nupkg Artifacts /Y
```
janusw, line 40: Stupid question here: Why is there an extra copy to `Artifacts` instead of outputting to that folder directy (as it is done for the other projects above)?

janusw (approved):
The GHA run proves that it works, so ok with me. Thanks for the fix!

Scripts/buildpack.bat
pauldendulk: The buildpack.bat script is very old. Older than the first github actions. The release script we had then pushed to nuget.org from the Artifacts folder. So, historic reasons.

pauldendulk:
@inforithmics Thanks for fixing my build 😁

pauldendulk:
> But then the created nuget packages wouldn't be used to restore the samples (I used this to test the created nuget samples and found a few bugs along the way).

Lets keep it, although it is less likely there will be bugs because we do not use nuspecs anymore.

PR #1841, opened 2023-02-15
Use ViewportState

pauldendulk:
The Viewport now holds all state in the ViewportState. The existing properties now se the properties of the ViewportState. They can probably be remove altogether at some point.

The ZoomOnCenterAnimation now sets the ViewportState as one statement instead of setting CenterX, CenterY and Resolution separately. 

https://github.com/Mapsui/Mapsui/blob/7fd495cd3aef71e12a8b54f108fb452c2db057fa/Mapsui/ViewportAnimations/ZoomOnCenterAnimation.cs#L28-L36

In the old code a single mousewheel click caused around 192 notifications
Screenshot 2023-02-15 113639

With the new code about one third (there is some variation between events)
image

### DIFF
--- a/Mapsui/Fetcher/Postponer.cs
+++ b/Mapsui/Fetcher/Postponer.cs
@@ -1,0 +1,54 @@
+﻿using System;
+using System.Threading;
+
+namespace Mapsui.Fetcher;
+
+/// <summary>
+/// Makes sure a method is always called 'MillisecondsToDelay' after the previous call.
+/// </summary>
+public class Postponer : IDisposable
+{
+    private readonly Timer _waitTimer;
+    private Action? _action;
+    private int _millisecondsToWait;
+
+    public Postponer(int millisecondsToWait = 500)
+    {
+        _millisecondsToWait = millisecondsToWait;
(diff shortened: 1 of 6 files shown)


Mapsui/Extensions/ViewportExtensions.cs
```diff
@@ -10,49 +8,6 @@ public static class ViewportExtensions
     public static bool HasSize(this IReadOnlyViewport viewport) =>
         !viewport.Width.IsNanOrInfOrZero() && !viewport.Height.IsNanOrInfOrZero();
 
-    /// <summary>
-    /// IsRotated is true, when viewport displays map rotated
-    /// </summary>
-    public static bool IsRotated(this IReadOnlyViewport viewport) =>
```
pauldendulk, line 16: IsRotated and GetExtent are moved to VierportStateExtensions

pauldendulk:
The difference in behavior is quite noticeable when running the 'Data Formats' | 'GeoJson' sample when mouse wheeling in Mapsui.Samples.Wpf.

PR #1843, opened 2023-02-15
NetTopologySuite.IO.GeoJSON4STJ to 3.0.0

pauldendulk:
@inforithmics The GeoJson sample works. Can we go to 3.0.0? Should we perhaps only increase the upper limit to allow a lower version?


IldarKhayrutdinov:
It seems that there are no breaking changes: https://github.com/NetTopologySuite/NetTopologySuite.IO.GeoJSON/compare/stj-v2.1.0...stj-v3.0.0

inforithmics:
It's okay. I wouldn't have a Dependency Range over 2 major versions.

inforithmics (approved):

PR #1845, opened 2023-02-16
Remove viewport getters

pauldendulk:
The viewport property getters are wrappers to the ViewportState fields. In this PR they are replaced by direct access to the ViewportState fields. This is a conceptually simple thing but impacts a lot of files.


pauldendulk:
I will look for an alternative solution. More direct use of the ViewportState.

PR #1851, opened 2023-02-22
Keep extent and resolution in sync

pauldendulk:


pauldendulk:
There PR is part of the work on Viewport improvements. The main purpose was to make more use of immutable classes (records), specifically ViewportState and MSection. Now changing the state becomes a single operation and once a record is passed into a method it's state can not be changed by other thread  which eliminates a category of bugs related to threading.

Done:
- Add MSection record (combination of extent and resolution).
- Make more use of ViewportState 
- Removed RenderResolutionMultiplier. It was not used in samples or tests. RenderResolutionMultiplier seemed to be replaced by PixelDensity.
- Removed Overscan. It was not used in samples or tests. It was not used in samples or tests.  It seemed to be implemented incorrect. It was replaced by featureSearchGrowth

Todo:
- Remove IViewportState
- Replace more (IReadOnly)Viewport with ViewportState or MSection.
- Improve naming of ViewportState and MSection?

PR #1856, opened 2023-02-25
Add net standard2.1 to Xamarin.Forms  and Fix System.Memory iOS problems

inforithmics:
Add net standard2.1 to Xamarin.Forms and Xamarin.iOS target to avoid System.Memory problems
Because System.Memory is included in net standard 2.1 and Xamarin.iOS

The main change is that again there where some nuget packages inconsistencies which I fixed.
So the result is that there are less nuget package references (for the samples - for the library it stayed the same).


inforithmics:
Things done:
1. Added Mapsui.Maui.slnf
2. Removed some more unnecessary nuget packages. (Are added automatically).
3. add .Net Standard 2.1 Target for better Xamarin.iOS Compatibility. (System.Memory nuget package).

inforithmics:
Things done:
1) Removed microsoft.netcore.platforms from the nuget packages (this is done automaticlly by the framework)
2) Updated xam.plugin.geolocator (only used in samples). The old version that was referenced caused tons of nuget warnings.

inforithmics:
Things done:
1) Fixed dotnet-release-nugets.yml dotnet pack ... -o nugets does not work anymore (use -p:PackageOutputPath="${{ github.workspace }}/nugets")
2) reintroduced the .target File for Xamarin.Forms iOS Target that removes the System.Memory nuget reference on Xamarin.Forms ios. This was lost during the migration to the csproj nuget packaging.

https://github.com/Mapsui/Mapsui/pull/1252

.github/workflows/dotnet-release-nugets.yml
```diff
@@ -43,13 +43,18 @@ jobs:
       run: |
         msbuild Mapsui.UI.Uno/Mapsui.UI.Uno.csproj /t:restore /p:Configuration=Release 
         msbuild Mapsui.UI.Uno/Mapsui.UI.Uno.csproj /t:pack /p:Configuration=Release /p:Version=${{ env.VERSION_OF_RELEASE }} /p:PackageOutputPath="${{ github.workspace }}/nugets"
+
+    - name: Pack Xamarin Forms
```
pauldendulk, line 47: How does this work. Mapsui.UI.Forms is part of the slnf so it is packed in the existing step. Since this is executed later would the nuget not get overridden?

Directory.Build.props
```diff
@@ -39,13 +39,6 @@
 	  <Compile Include="$(MSBuildThisFileDirectory)/InitProperty.cs" />
   </ItemGroup>
 
-  <!-- Workaround for when setting CentralPackageTransitivePinningEnabled to true it includes in .net Standard
```
pauldendulk, line 42: Nice this workaround can now be removed.

pauldendulk (approved):
Thanks a lot for taking care of the nasty stuff. Going to all .NET 6 in v5 will make things a lot easier. In the meantime we have to get this functioning properly, so this is important.

.github/workflows/dotnet-release-nugets.yml
inforithmics: The nuget package isn't produced with dotnet pack. Because it needs msbuild.

pauldendulk: ohw, it did build with dotnet previously, but it does not anymore because xamarinios1.0 was added. And xamarinios1.0 was added because System.Memory needs to be removed specifically for iOS using the Mapsui.Forms.targets.

PR #1857, opened 2023-02-26
Publish Blazor App to Samples sub url of mapsui

inforithmics:


inforithmics:
I puplished the Blazor samples to this url to test out that the publishing works.

https://inforithmics.github.io/samples/

inforithmics:
When everything works it should be deplyed to https://mapsui.com/samples

pauldendulk (approved):
This is absolutely wonderful. An important improvement for Mapsui, for all platforms. Not a lot of code was needed. I don't use Powershell a lot but it seems a good choice here.

Future features:
- It should be possible to navigate to a specific category/sample through the url so we can point to a specific sample in a discussion with a url.
- Show the sample code in the sample itself (this would be helpful for all other samples).

pauldendulk:
oh, the build fails on master with `Container cannot be copied onto existing leaf item.`
https://github.com/Mapsui/Mapsui/actions/runs/5110804086/jobs/9187100960#step:5:119

PR #1869, opened 2023-03-04
Improve performance samples and use source name in RasterizingTileLayer

pauldendulk:


pauldendulk:
Done:
- Mostly samples except for one thing in the RasterizingTileLayer where the source of the RasterizingTileLayer.Name is set to the source. This is like in the RasterizingLayer.
- Changed the naming of the ISample.Name and the files itself to reflect the technology used in the samples which I think is the important part for the users. This is what they need to start looking into.
- Added a sample for the RasterizingTileLayer and the LineString with many vertices. It is much faster than the RasterizingLayer.

image

PR #1873, opened 2023-03-08
WIP Limiter using ViewportState

pauldendulk:


pauldendulk:
### Done
- Big breaking change: Move Viewport and Navigator to Map, for better MVVM support. Also I think this can simplify a lot of things we do in the samples, tests and benchmarks.
- Merged LimitedViewport in Viewport. Deleted LimitedViewport, IViewport, IReadonlyViewport.
- Made more not nullable. It is either not possible or an exception is thrown it is assigned null.
  - Made MapControl.Map not nullable
  - Made Map.Navigator not nullable
  - Made Map.Viewport not nullable
- Limiting is now always done in one of the core methods. It is hard to avoid this.
- Fix Limit to extent (Keep Confine to Extent)
- Fix PanLock and Zoom lock (perhaps needs more work)
- Fix the regiression test for the design changed. I did not check if the benchmark tests still work.

### Todo
- Replace all IMapControlSample with ISample
- Where we have a long chane of properties we should see for ways to reduce that. Move logic to a module with a more specific purpose that only need the right most field. In practice: Move more out of classes like MapControl, Map and Navigator. Also where we have validation checks (like HasSize) we could do them insize the called method).

### Considering:
- Merge Viewport to Navigator
- Rename ViewportState to Viewport

PR #1877, opened 2023-03-11
remove .nuget\nuget.exe

inforithmics:
Removes the local nuget.exe (was a workaround because the nuget update task didn't update to the latest nuget)


pauldendulk (approved):
I love cleanup

PR #1882, opened 2023-03-12
Use ISample instead of IMapControlSample where possible

pauldendulk:
This PR is meant to replace all use of IMapControlSample with ISample. This is possible because everything needed for the samples is now accessibly from the Map. @inforithmics Do you see possible problems with this approach?


inforithmics (approved):
Good Improvement, I like simplifications.

pauldendulk:
Things I run into now:
- SymbolAnimationSample. Don't think this has to be real problem. It just has to change a bit.
- PixelDensity in the RasterizingLayer. Typically something that belongs to the platform. But then it is also weird a layer needs to know about this. In an app I would make this an injected singleton service.
- SkiaCustomStyleRenderer. For MVVM support this does not need to be on the Map but it is a problem for our samples.
- OnClick. For the sample Info could be used instead
- UseDoubleTap. Usually configured once in the setup I presume, so could stay on the MapControl for MVVM.

PR #1885, opened 2023-03-14
Fixed Vector Style Point TileRendering Fix (didn't increase the search area so it cut of rendered points)

inforithmics:


inforithmics:
Things done:

1) Fixed Tile Rendering of Poits (VectorStyle)
2) Added Index to GeoJsonProvider to faster query the Geometry Features

inforithmics:
Improved the algorithm to increase the search area.

pauldendulk:
I would like to have a better understanding of the existing problem. What could I do to reproduce it?

inforithmics:
Adding this sample to the current Mapsui Master.

https://github.com/Mapsui/Mapsui/pull/1885/files#diff-76898329af5ab8d790e04018a0260a168c79971625c97bd0160606fab1e220b5

And then the Circles are Cut at the bounds of the Tiles.

Mapsui.Nts/Providers/GeoJsonProvider.cs
```diff
@@ -79,17 +81,46 @@ private bool IsGeoJsonContent()
...
+                        foreach (var feature in featureCollection)
+                        {
+                            var boundingBox = BoundingBox(feature);
+                            if (boundingBox != null)
+                            {
+                                var geometryFeature = new GeometryFeature();
+                                geometryFeature.Geometry = feature.Geometry;
+                                FillFields(geometryFeature, feature.Attributes);
+
+                                _index.Insert(boundingBox, geometryFeature);
```
inforithmics, line 109: insert the geometry into the index, Because searching in the index is log n searching on a list is n. So with a lot of items these is a Performance improvement. Because the loaded Geometries aren't changed STRTree is used (doesn't support modifcation and is compact).

Mapsui.Nts/Providers/GeoJsonProvider.cs
```diff
@@ -79,17 +81,46 @@ private bool IsGeoJsonContent()
...
+
+                                // build extent
+                                var mRect = boundingBox.ToMRect();
```
inforithmics, line 112: calculating bounds on same iteration

Mapsui.Nts/Providers/GeoJsonProvider.cs
```diff
@@ -131,19 +145,8 @@ public Task<IEnumerable<IFeature>> GetFeaturesAsync(FetchInfo fetchInfo)
...
-            {
-                var geometryFeature = new GeometryFeature();
-                geometryFeature.Geometry = feature.Geometry;
-                FillFields(geometryFeature, feature.Attributes);
-                list.Add(geometryFeature);
-            }
-        }
-
-        return Task.FromResult((IEnumerable<IFeature>)list);
+        IEnumerable<IFeature> result = FeatureCollection.Query(fetchExtent);
```
inforithmics, line 148: query is now done over the index

Mapsui.Rendering.Skia/SkiaStyles/LabelStyleRenderer.cs
```diff
@@ -381,8 +381,12 @@ private static Line[] SplitLines(string? text, SKPaint paint, float maxWidth, st
         }).ToArray();
     }
 
-    double IFeatureSize.FeatureSize(IFeature feature, IStyle style, IRenderCache renderingCache)
+    bool IFeatureSize.NeedsFeature => true;
+
+    double IFeatureSize.FeatureSize(IStyle style, IRenderCache renderingCache, IFeature? feature)
     {
+        if (feature == null) throw new ArgumentNullException(nameof(feature));
```
inforithmics, line 388: The Needs Feature says I need Feature instances to detect the Feature size so not providing one is an error.
Because on a Label the Text influences the Feature size

Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
```diff
@@ -287,7 +289,8 @@ public static double FeatureSize(SymbolStyle symbolStyle, ISymbolCache symbolCac
             case SymbolType.Ellipse:
             case SymbolType.Rectangle:
             case SymbolType.Triangle:
-                symbolSize = new Size(SymbolStyle.DefaultWidth, SymbolStyle.DefaultHeight);
+                var vectorSize = VectorStyleRenderer.FeatureSize(symbolStyle);
```
inforithmics, line 292: Ask now the VectoryStyle for  the Featuresize because the base class of a symbolStyle is a Vector style. And here I render vector (symbols).

Mapsui.Rendering.Skia/SkiaStyles/VectorStyleRenderer.cs
```diff
@@ -59,4 +59,36 @@ public bool Draw(SKCanvas canvas, ViewportState viewport, ILayer layer, IFeature
...
+    {
+        if (style is VectorStyle vectorStyle)
+        {
+            return FeatureSize(vectorStyle);
+        }
+
+        return 0;
+    }
+
+    public static double FeatureSize(VectorStyle vectorStyle)
```
inforithmics, line 75: provide implementation of the Feature Size of the Vector style. This calculates the Feature Size of Points and  thick lines, Borders.

Mapsui.Tiling/Provider/RasterizingTileProvider.cs
```diff
@@ -27,7 +27,7 @@ public class RasterizingTileProvider : ITileSource
     private Attribution? _attribution;
     private readonly IProvider? _dataSource;
     private readonly RenderFormat _renderFormat;
-    private IDictionary<TileIndex, double> _searchSizeCache = new ConcurrentDictionary<TileIndex, double>();
+    private readonly IDictionary<TileIndex, double> _searchSizeCache = new ConcurrentDictionary<TileIndex, double>();
```
inforithmics, line 30: can be made readonly

Mapsui.Tiling/Provider/RasterizingTileProvider.cs
```diff
@@ -140,22 +140,32 @@ private async Task<double> GetAdditionalSearchSizeAsync(TileInfo tileInfo, IRend
...
         if (!_searchSizeCache.TryGetValue(tileInfo.Index, out var result))
         {
-            result = 0;
-            var features = await GetFeaturesAsync(tileInfo);
-            var layers = new List<ILayer> { new RenderLayer(_layer, features) };
+            double tempSize = 0;
 
-            void MeasureFeature(IStyle style, IFeature feature)
+            var layerStyles = _layer.Style.GetStylesToApply(section.Resolution);
+            foreach (var style in layerStyles)
```
inforithmics, line 146: optimization to only load features when the FeatureSize needs to be calculated over every Feature.

Mapsui/Styles/IFeatureSize.cs
```diff
@@ -5,10 +5,17 @@ namespace Mapsui.Styles;
...
+    bool NeedsFeature { get; }
+
     /// <summary> The Feature Size is in points of the screen </summary>
-    /// <param name="feature">feature to detect size</param>
     /// <param name="style">symbol style to detect size</param>
     /// <param name="symbolCache">symbol Cache</param>
+    /// <param name="feature">feature to detect size</param>
     /// <returns>size in points</returns>
-    double FeatureSize(IFeature feature, IStyle style, IRenderCache symbolCache);
+    double FeatureSize(IStyle style, IRenderCache symbolCache, IFeature? feature = null);
```
inforithmics, line 20: move the feature to the end because it is now optional, was only used in LabelStyleRenderer.FeatureSize

Samples/Mapsui.Samples.Common/Maps/Performance/RasterizingTileLayerGeoJsonSample.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System.Diagnostics.CodeAnalysis;
```
inforithmics, line 1: sample that shows the cutting off of Rendered Points

Tests/Mapsui.Rendering.Skia.Tests/SymbolStyleFeatureSizeTests.cs
```diff
@@ -20,7 +20,7 @@ public void DefaultSizeFeatureSize()
         var symbolCache = new SymbolCache();
         var size = SymbolStyleRenderer.FeatureSize(symbolStyle, symbolCache);
 
-        Assert.AreEqual(size, Math.Max(SymbolStyle.DefaultHeight, SymbolStyle.DefaultWidth));
+        Assert.AreEqual(size, Math.Max(SymbolStyle.DefaultHeight, SymbolStyle.DefaultWidth) + 1);
```
inforithmics, line 23: line size is now added to the symbol size. Without it the Sorounding Circle Line is Cut off.

pauldendulk (approved):
Thanks for the fix. This is how the sample looks in master:

image

Some thoughts:
- This is my understanding of how it works. We are drawing tiles. Adjacent tiles might contain features that flow over into the current tile. In order to resolve this we want to draw those features while rendering the current tile. In order to know which features on the outside to draw the size is calculated of all the features in the eight adjacent tiles. 
- For labels the size depends on the feature because of the rendered text.
- In this PR NeedsFeature is false for SymbolStyle or VectorStyle. How would it work with a ThemeStyle that does not feature information at this point? Would the size not be calculated because the ThemeStyle has no IFeatureSize interface?
- Perhaps it should still be possible to specify a maximum size on a layer that can be used as a default if the style renderer does not implement IFeatureSize.
- Something to think about is if we want to implement this for all renderers? Is there another solution? Could we calculate from the outside?
- This problem is now solved for tiles but it is actually a general problem. On the edges of a (non tiled) map there could also be symbols that is (or should be) partly visible even if the center is outside the extent. So this is important.
- We iterate through adjacent tiles now, but it is not impossible that a label or symbol would cover three tiles (unlikely perhaps but it will definitely occur for some of our users at some point, perhaps as part of an unusual use case).
- There is now quite some code in the RasterTileLayer related to measuring. Perhaps this could be moved outside of the RasterTileLayer. Perhaps to some extension method. Also because this code might be necessary for other layers as well.
- An alternative solution is to not measure all the features in the surrounding tiles but to render them all. The reason not to do this is performance I guess. How much does it matter? What is the cost of drawing something that is completely outside the canvas? (I have used libraries where this was quite high, but I have also used libraries where the cost of measuring was quite costly). All in all, this could be seen as a performance optimization of rendering that happens in the background. 
- Are there alternative solutions to this problem? Here is one thought. If a feature is drawn and the pixels of the outer rectangle of the tile do change color after it is drawn then it might be cut off, so that same feature should be drawn in the adjacent tile. This seems like an interesting idea. What is problematic is that the information in the current tile is important to the other tile.

inforithmics:
1. Yes the FeatureSize Calculation should be moved outside of the RasterTileLayer, because it is a General problem, during Tiling it becomes visible.
2. I think I have to look at the ThemeStyle or other Styles where an overlapp in Rendering could occour.

pauldendulk:
In the past we had a solution for this using an 'overscan' parameter. This functionality broke at some point and I removed the parameter in one the PRs a month before. That overscan parameter just fetched a bigger extent and all those features were rendered. In the current there is still such an overscan, which is the width a tile for this tile layer solution. The difference is that for the features in the overscan area it is calculated if they need to be drawn. It is unfortunate that the size of this overscan is a arbitrary choice. Perhaps it should be configurable. For fixed size symbols the oversize area could be as small as the max symbol size. This would prevent a lot of unnecessary calculations.

PR #1886, opened 2023-03-14
Url is case sensitive, needs to be /Mapsui

skipru:
Lower case link generates a 404 not found error.


inforithmics:
Good catch

inforithmics (approved):

PR #1897, opened 2023-03-22
Use ViewportState instead of Viewport in animations

pauldendulk:


pauldendulk:
One advantage of this approach is that the limiting is now done in the viewport after the animation calculation and is not the responsibility of all the animation implementations. So, it is not possible to circumvent the limiting.

PR #1899, opened 2023-03-23
Cleanup and moving namespaces

pauldendulk:
- Move Animations to Mapsui.Animations
- Move Info related classes to Mapsui namespace
- Only IMapControl is left in UI folder of the Mapsui project. I would like to move it to Mapsui.UI.Shared but that is not possible because the samples depend on it. I tried to removed that in an earlier PR but did not succeed yet.

In general I think less namespaces should be needed by our users. We could move more classes to the root of the Mapsui project, but folders can be helpful to keep the overview. We could indicate that a folder should not have its own namespace in the cproj. In some cases it does make sense to have things in a separate namespace, but that should only be the case if you are using some special feature, not for the basic stuff. We could go through our samples to check which namespaces are needed and decide if we really want it that way.


pauldendulk:
Looks like it is not possible with the current dotnet tools to ignore the folder as namespace provider. https://github.com/dotnet/project-system/issues/5372. I used Resharper for this in the past but I now use the dotnet and VS tools exclusively. I don't see a good solution at the moment. I don't want to move everything in Mapsui.Layers and Mapsui.Styles to the root of the Mapsui project. Perhaps it is possible to exclude those folders to the tools.

Issue #1902, opened 2023-03-25
Dragging a polygon's vertex


pauldendulk:
### Discussed in https://github.com/Mapsui/Mapsui/discussions/1901

<div type='discussions-op-text'>

<sup>Originally posted by **DmitryEvstakhov** March 24, 2023</sup>
Hi guys!
I use example Mapsui.Examples.Mpf.Editing. When I dragging point from a polygon I have some problems:
1. Dragging vertex is not smooth. I mean that the vertex from the VertexOnlyLayer drags smoothly, but the rendered geometry of polygon is not smooth (jerky).
2.  Sometimes the geometry doesn't redraw at all. You have moved one vertex of polygon, but the polygon rendered geometry remains the same. But when I move the map the geometry will be rendered correctly.

Initial geometry:
![01](https://user-images.githubusercontent.com/59865880/227527846-9e796f0c-30d2-4e6a-b112-93ef6761f61d.png)

When I dragged vertex:
![02](https://user-images.githubusercontent.com/59865880/227528006-1d097422-de3c-4872-a56e-c493a19ab00e.png)

When I move the map:
![03](https://user-images.githubusercontent.com/59865880/227528324-9f4a9265-4411-46d2-b238-ef1c21b97e12.png)

Why is this happening and how to solve this problem?
</div>

pauldendulk:
Video by @DmitryEvstakhov. Thanks.
https://user-images.githubusercontent.com/59865880/227552763-b25cfccf-2b4a-477d-b6be-eb210a2b56f4.mp4

pauldendulk:
I can reproduce it with todays master. It seems the vertex in the geometry that is being dragged around jumps back to a previous state while panning. Sometimes to a quite recent position but later it can go back to an older position. Somehow this older state is preserved. It is not just slow. 

When you stop dragging the vertex stays at an older position.

When you drag the map it immediately goes to the correct position.

pauldendulk:
When I remove the renderCache parameter of the PolygonRenderer.Draw in the VectorStyleRenderer.Draw method at line 42 the problem is solved. So, something with caching.

```csharp
case Polygon polygon:
    PolygonRenderer.Draw(canvas, viewport, vectorStyle, feature, polygon, opacity, renderCache);
    break;
```

inforithmics:
From the described behavior, I think the HashCode does not Change when a Point is changed, causing the Hash to return the old rendered Polygon.

inforithmics:
[...]

PR #1908, opened 2023-03-25
Fix editing and caching

inforithmics:
1) Updated NTS to newest Version, some Bugfixes when Manipulating and Editing Geometries.
fixing for example https://github.com/NetTopologySuite/NetTopologySuite/pull/604

2) Writable Layers are not cached, because when I want to cache mutable objects I would have to clone them and put them into the cache (Which would destroy the Performance wins). But when they are mutated the cache state becomes invalid causing this problems, They are equal but different in Geometry.

Fixes:
https://github.com/Mapsui/Mapsui/issues/1902

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -65,7 +65,7 @@
     <PackageVersion Include="Microsoft.Windows.CsWinRT" Version="[2.0.0,3.0.0)" />
     <PackageVersion Include="Microsoft.WindowsAppSDK" Version="[1.2.230217.4,2.0.0)" />
     <PackageVersion Include="NeoSmart.AsyncLock" Version="[3.2.1,4.0.0)" />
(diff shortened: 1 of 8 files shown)


pauldendulk:
Fix for the current problem. Although it is possible for the user to edit the feature in other layer types as well. So, a better fix may be needed in the future.

I did not look at the related code but some things that come to mind:
- I think a HashCode in dotnet does not have to unique (but unique enough to help with performance when using hashing). 
- Not having a unique enough HashCode should affect performance but should not cause a bug. Usually the HashCode is used in an initial step but value compare is done in a later step.
- Perhaps we don't do a value compare but reference compare.
- If we use a hash based on values and a comparison based on reference this might not make sense. Perhaps the hash should be based on the reference in that case, or no hash at all.

Sorry, if this makes no sense at all. Just doing some guessing.

PR #1912, opened 2023-03-27
Clean up Navigator (aftercare of merging in old Viewport)

pauldendulk:


pauldendulk:
@inforithmics In this MR I had a problem with the WFS sample. I assumed that this was because of a problem with the service itself because it seemed to me unlikely to be related to my changes (but it is possible), so I renamed the reference image to skip the test. I see some issues related to WFS, so I thought it was important to you that I made this change. 

My focus is to release beta 9.

PR #1920, opened 2023-03-31
WMS provider

gpsaliola:
I would like to point out that there are some WMS services which in the BBOX parameter invert minx with miny and maxx with maxy.
I read on the net that this inversion can be understood from the GeographicBoundingBox parameter present in the xml file obtained with GetCapabilities.
Looking at the sources in the Mapsui\Providers folder it doesn't seem that this is taken into consideration.

An example of one such server is:
http://sgi2.isprambiente.it/arcgis/services/raster/igm25k_lazio_wgs/ImageServer/WMSServer?service=wms&request=getCapabilities&version=1.3.0

An example of GetMap:
https://sgi2.isprambiente.it/arcgis/services/raster/igm25k_lazio_wgs/ImageServer/WMSServer?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&BBOX=41.91,12.57,41.92,12.58&WIDTH=256&HEIGHT=256&FORMAT=image/png&STYLES=&TRANSPARENT=TRUE&CRS=EPSG:4326&LAYERS=igm25k_lazio_wgs

As you can see in GetMap BBOX the longitude and latitude are inverted.
Read this also: https://gis.stackexchange.com/questions/310950/public-site-wms-bounding-box-seems-wrong
Thank you!


pauldendulk:
This PR merges master to develop/3.0. This is probably not what you intended. You need to create your own fork. Make the changes there. And in the PR point  your branch with changes to the mapsui master or develop/3.0.

pauldendulk:
I think you meant to post this as an issue but created a pull request instead. You point is valid and we should deal with it. Please post it as an issue.

gpsaliola:
Ok, Thank you.
Sorry, I didn't quite understand what pull requests were.

PR #1922, opened 2023-04-01
Implement Mapy map (http headers)

gpsaliola:
I'm trying to implement XYZ map "Mapy" (https://api.mapy.cz/v1/maptiles/basic/{z}/{x}/{y}).
This web service needs to include the following in the http headers:
key="Referer", value="https://en.mapy.cz/".
Without this, the map cannot be displayed.
Is there a way to pass these parameters to the HttpTileSource class?
Thank you.


pauldendulk:
@gpsaliola I am closing this merge request. It seems to be more like a question and the merge request itself merges the mapsui master into the mapsui develop/3.0, so something has gone wrong.

Regarding the question. The HttpTileSource takes a tileFetcher as argument. This is a function that gets the url as input and returns the byte array of the tile. By injecting this function you override the way the tile is fetched. You can use your own HttpClient and set it's headers the way you like.

gpsaliola:
Ok, thank you.

Issue #1919, opened 2023-03-30
WFSProvider does not honor proxy settings


wilhelmberg:
## Describe the bug

`WFSProvider` does not use these proxy properties:
* `Credentials`
* `ProxyUrl`

Even when those properties are set, requests to a WFS server are direct requests not using the specified proxy settings.

## To Reproduce

To reproduce a proxy is needed.
I use [Fiddler Classic](https://www.telerik.com/download/fiddler) for local development (Windows only). 
There's also [Fiddler Everywhere](https://www.telerik.com/download/fiddler-everywhere) (Win, Linux, Mac) and others, but I have no experience with those.

Fiddler Classic can be setup to act as a system proxy and all applications that listen to that proxy change start using it immediately.
But it isn't a catch-all, application that connect directly are still able to establish outbound connections.

That's why I test like this:
* Fiddler OFF, but define proxy settings in app
  * If requests go through, this means the app is not trying to use the proxy -> more fixing needed
  * If requests fail -> app is trying to use proxy, good
* After all requests fail, Fiddler ON and check if the expected requests also show up in the log

----

I tried several of the available `WFSProvider` constructors and also the factory methods, but none of them worked.

One problem, the `ProxyUrl` property can only be set after a `WFSProvider` instance has been initialized, but the WFS Capabilities are already being requested during said initialization (more findings further down under `Additional context`):

```csharp
WFSProvider provider = await WFSProvider.CreateAsync(
	_wfsUri
	, _nsPrefix
	, _layerName
	, WFSProvider.WFSVersionEnum.WFS_1_1_0
	, persistentCache: DefaultCache
);

provider.ProxyUrl = "http://localhost:8888";  <==== after GetCapabilities
```

_Note:_
I haven't tried the constructor which allows to bypass the intial `GetCapabilities` and `DescribeFeatureType` as I don't know the necessary parameters upfront.

## Expected behavior
Requests to a WFS server are routed via the specified proxy.

## Desktop (please complete the following information):
 - OS: Win11

## Additional context

The solution suggested in other issues reading proxies
```
BruTile.Web.HttpClientBuilder.HttpClientHandler = clientHandler;
```
does not work with  `WFSProvider`, because it brings its own [HttpClientUtil](https://github.com/Mapsui/Mapsui/blob/f2cc703fbacdd3d55e1b0df9259ed697d5b9d539/Mapsui.Nts/Providers/Wfs/Utilities/HttpClientUtil.cs).

[...]

PR #1924, opened 2023-04-01
WFS Proxy and WfsSample fix

inforithmics:
Add proxy and credientials parameter to WfsProvider.CreateAsync so that first call could already run over the proxy


inforithmics:
Fixes this: https://github.com/Mapsui/Mapsui/issues/1919

inforithmics:
Things done:
1) The proxy isn't reset anymore and is always used.
2) fixed the wfssample, I fixed it by always creating a new httpclient instance per http call. If not a new httpclient is created calls could overlap and strange errors could occour.

pauldendulk (approved):

Issue #1938, opened 2023-04-03
KeepWithinExtentSample seems not to work.


inforithmics:
**Mapsui Version**
Current Master

**Mapsui Platform**
Xamarin.Forms

**Device**
Windows (WPF)

**Describe the bug**
The Keep within Extent Sample should keep the zoom and panning inside the extent. But with zoom and panning I can move outside the extent.

**To Reproduce**
1.  Get Master
2. Open Mapsui.Forms.slnf
3. Navigation => Keep Within Extent
4. I can zoom and pan outside of Madagaskar.

**Expected behavior**
Zoom and Pan should be within Madagaskar.

**Screenshots**
Initial zoom is within madagaskar.
![image](https://user-images.githubusercontent.com/2290863/229554726-1cad3c44-32dc-4749-a1d7-ba3899bba7ee.png)

Afterwards I can move outside of madagascar
![image](https://user-images.githubusercontent.com/2290863/229554993-fc7d4837-7132-40cc-8b54-7d90cbd3fe00.png)

**Additional context**

pauldendulk:
Fixed with https://github.com/Mapsui/Mapsui/pull/1942

PR #1942, opened 2023-04-04
Don't override user set resolutions, zoomextremes panextent

inforithmics:
Fixes this https://github.com/Mapsui/Mapsui/issues/1938

When resolutions, zoomextremes and panextent are set externally they have priority over default (zoomextrems,panextent, resoultions) which are set during layerchanged events.

### DIFF
--- a/Mapsui/Map.cs
+++ b/Mapsui/Map.cs
@@ -236,13 +236,10 @@ public class Map : INotifyPropertyChanged, IDisposable
 
     private void LayersChanged()
     {
-        // Todo: It is possible for the user to assign the three fields below override
-        // this initialization. However, whenever a layer changes this will override the
-        // the users settings again. This is a bug that needs to be fixed with a better
-        // design.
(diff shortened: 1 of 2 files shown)


pauldendulk (approved):
Wow! I considered many different options (for instance an Action that could be overridden, or an internal boolean that tracked if the user set it) but yesterday in the train back home I decided I should go for a solution that is almost identical to this. I made a draft (with different naming though) where I even made the default fields internal like you did. 

image

pauldendulk:
This solution also takes into account later changes in the layers. It was easy to make the mistake to only check on the first layer changed.

A critique on this solution could be that the properties behave a bit unpredictable. If you `set` the property to null it will the `get` will not return null but the default value.

The alternative would be to treat the internal and external value is separate properties throughout the code. In that case you have the do the `DefaultResolutions ?? Resolutions` logic in many places, and could be forgotten at some point, so that is also a drawback. So, this is find.

PR #1944, opened 2023-04-07
Use Override solutions for Resolutions and Pan/Zoom bounds

pauldendulk:
@inforithmics I implemented the 'override property' solution you suggested in your previous MR. I also renamed PanExtent to PanBounds and ZoomExtremes to ZoomBounds, which seems a bit better. Also renamed the ZoomToPanBounds method accourdingly.

### DIFF
--- a/Mapsui/Navigator.cs
+++ b/Mapsui/Navigator.cs
@@ -16,9 +16,7 @@ public class Navigator
     private Viewport _viewport = new(0, 0, 1, 0, 0, 0);
     private IEnumerable<AnimationEntry<Viewport>> _animations = Enumerable.Empty<AnimationEntry<Viewport>>();
     private IReadOnlyList<double>? _resolutions;
-    private MMinMax? _zoomExtremes;
-    private MRect? _panExtent;
-
+    
     /// <summary>
     /// Called when a data refresh is needed. This directly after a non-animated viewport change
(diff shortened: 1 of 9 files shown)


inforithmics:
Looks good to me. Is the better solution because there are no surprises in the API

inforithmics (approved):

Mapsui/Navigator.cs
```diff
@@ -123,13 +122,13 @@ public void ZoomToBox(MRect? box, MBoxFit boxFit = MBoxFit.Fit, long duration =
     public void ZoomToPanExtent(MBoxFit boxFit = MBoxFit.Fill, long duration = -1, Easing? easing = default)
     {
         if (!Viewport.HasSize()) return;
-        if (PanExtent is null)
+        if (PanBounds is null)
         {
             Logger.Log(LogLevel.Warning, "ZoomToPanExtent was called but PanExtent was null");
```
charlenni, line 129: Shouldn’t the log text be updated too?

pauldendulk: Good catch. I squeezed the fix into my current PR with this commit https://github.com/Mapsui/Mapsui/pull/1948/commits/20c15c324da1e14777f0b55cd46df0ccaa9825cb, using `nameof` as I should have done in the first place.

PR #1947, opened 2023-04-07
Move locks to navigator

pauldendulk:
They were not used in the Limiters at all, but are in the Navigator. This makes it also possible to remove the BaseViewportLimiter.

Note to the future: We should go through all use of the Lock outside the Navigator and check if that is still necessary. If no, then remove it, if yes think about away to make it not necessary.


pauldendulk:
> Note to the future: We should go through all use of the Lock outside the Navigator and check if that is still necessary. If no, then remove it, if yes think about away to make it not necessary.

Had a look at it and removed one PanLock check, but the remaining checks are related to touch/pinch/manipulation and should be tested on a device, and I did not want to do that right now.

PR #1951, opened 2023-04-11
Fix scale issue for Uno running on Android API 33

pauldendulk:


FrozDark:
A little notice that WinUI itself doesn't support DisplayInformation.GetForCurrentView()

https://github.com/microsoft/microsoft-ui-xaml/issues/4228

pauldendulk:
@FrozDark ah, I did not know. xamlRoot was what we used, but it also returned 2.8125. So, this means this fix should really be regarded as a workaround, and it gets more important to get to the root of this.

Mapsui.UI.WinUI/MapControl.cs
```diff
@@ -257,7 +258,7 @@ public void OpenBrowser(string url)
     private float GetPixelDensity()
     {
 #if __WINUI__
-        return (float)(XamlRoot?.RasterizationScale ?? 1f);
+        return (float)DisplayInformation.GetForCurrentView().ResolutionScale / 100.0f;
```
inforithmics, line 261: Shouldn't it be this so that the #if __WINUI__ can be removed?
` return (float)DisplayInformation.GetForCurrentView().RawPixelsPerViewPixel;`

inforithmics: So I think this is a Mapsui.Uno.WinUI Bug and not an Uno Bug.

pauldendulk: No, that is one of the first things I tried. `(float)DisplayInformation.GetForCurrentView().RawPixelsPerViewPixel` returns the same value as `(XamlRoot?.RasterizationScale`, 2.8125. And both are different from what is returned on API 30, 2.75. Which is different again from the scaling of the skia canvas which is 2.0.   

I now think it is a bug in SkiaSharp or Uno. 2.75 is the correct scaling for that device it should work properly.

inforithmics: I would suggest to change the code to

```
 #if __UNO__
         return (float)DisplayInformation.GetForCurrentView().ResolutionScale / 100.0f;
 #elif __WINUI__
         return (float)(XamlRoot?.RasterizationScale ?? 1f);
#else
        return (float)DisplayInformation.GetForCurrentView().RawPixelsPerViewPixel;
#endif
```

or else it doesn't work for winUI because DisplayInfrmation.GetForCurrentView does not work for winui

inforithmics (changes requested):

PR #1959, opened 2023-04-17
Fix clipping while rotated

### LINKED ISSUE #1958: Clipping edges visible in map when rotated
**Mapsui Version**
todays master

**Mapsui Platform**
All

**Device**
All

**Describe the bug**
The clipping of the MRect is not not correct when rotated. The clip edged can be seen in the viewport and should be invisible.

**To Reproduce**
1. Go to the Mapsui.Samples.Avalonia 
2. Select Navigator | Keep Center in Map
3. Drag the rotation slider bottom left.
4. The top and bottom clipping edges are visible in the map. The rotation angle of the clipping is correct.

**Expected behavior**
No visible clipping

**Screenshots**
image

pauldendulk:
Fixes #1958.

MRect WorldToScreen did not take into account rotation. Since the MRect does not have individual coordinates for the four corners but just has min/max x/y values it can not be rotated. In case of rotation the encompassing rectangle of the rotated rectangle is returned.

With this fix the clipping edges are not visible. So the clip rect is not too small. Yet the clip rect could also be too big. In order to test this I shrunk the clip rect a bit after the transformation and then it showed the clip edges inside the viewport as expected.

image

Notes:
- Perhaps WorldToScreen(MRect) should not be an extension method, or perhaps it should have a different name. What is does is different to the other WorldToScreen methods.

### DIFF
--- a/Mapsui/Extensions/ViewportExtensions.cs
+++ b/Mapsui/Extensions/ViewportExtensions.cs
@@ -12,12 +12,22 @@ public static class ViewportExtensions
     public static bool HasSize(this Viewport viewport) =>
         viewport.Width > 0 && viewport.Height > 0;
 
-    /// <summary> World To Screen Translation of a Rect </summary>
-    /// <param name="viewport">view Port</param>
-    /// <param name="rect">rect</param>
+    /// <summary>Transforms the MRect from world coordinates to screen coordinates. Note, that
+    /// an MRect always represents and unrotated box. If the Viewport is rotated this will result
+    /// in an unrotated box that encompasses the rotated transformation.</summary>
+    /// <param name="viewport">Viewport</param>
+    /// <param name="rect">The MRect to transform</param>
     /// <returns>Transformed rect</returns>
     public static MRect WorldToScreen(this Viewport viewport, MRect rect)
     {
+        if (!viewport.IsRotated()) // Checking on IsRotated for performance reasons
+        {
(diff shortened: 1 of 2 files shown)


pauldendulk:
@inforithmics I should have asked you for a review on this one. 

I replaced some of the WorldToScreen code with the existing extension method. It seems to work correct, so just a bit more code reuse.

The WorldToScreen method for MRect bothers me, because it is not really transforming the MRect itself (because that is not possible). I now think it would be a better solution to add a WorldToScreen for MQuad instead and and to a rect.ToMQuad(), then WorldToScreen and then request the encompassing MRect (there is a method for it, but I don't know it from the top of my head).

Mapsui/Extensions/ViewportExtensions.cs
```diff
@@ -10,15 +12,40 @@ public static class ViewportExtensions
...
+            viewport.WorldToScreen(rect.TopRight),
+            viewport.WorldToScreen(rect.TopLeft)
+        };
+
+        var minx = screenPoints.Select(p => p.X).Min();
+        var miny = screenPoints.Select(p => p.Y).Min();
+        var maxx = screenPoints.Select(p => p.X).Max();
+        var maxy = screenPoints.Select(p => p.Y).Max();
+
+        return new MRect(minx, miny, maxx, maxy).Grow(-50);
```
pauldendulk, line 48: The Grow(-50) is test code 😬

PR #1962, opened 2023-04-17
added shapefile projection sample

inforithmics:
Added Shapefile Projection sample. for reproducing dotspatial not working on platforms like uwp or ios (where dynamic assembly loading is disabled).

Somehow the RasterizingTileProvider and Projection don't play so nice together, because some data seems not to be rendered.


inforithmics:
Things done
1) The Rasterizing Tile Bug is now fixed.

PR #1964, opened 2023-04-18
added linux github action

inforithmics:


inforithmics:
Before I remove the WIP I have to run the linux github action

inforithmics:
Things done:

Added Linux Github Action with mapsui.Tests.
update nunit and nunit test adapter.

.github/workflows/dotnet.yml
```diff
@@ -10,9 +10,82 @@ on:
...
+      run: dotnet pack --no-restore --configuration Release Mapsui.Extensions/Mapsui.Extensions.csproj -o Artifacts -p:PackageVersion=$(git describe --tags)
+    - name: Build Mapsui.UI.Android
+      run: dotnet pack --no-restore --configuration Release Mapsui.UI.Android/Mapsui.UI.Android.csproj -o Artifacts -p:PackageVersion=$(git describe --tags)    
+    - name: Build Mapsui.UI.Avalonia
+      run: dotnet pack --no-restore --configuration Release Mapsui.UI.Avalonia/Mapsui.UI.Avalonia.csproj -o Artifacts -p:PackageVersion=$(git describe --tags)    
+    - name: Upload packages
+      uses: actions/upload-artifact@v3
+      with:
+        name: nupkg.mac
+        path: Artifacts/*.nupkg
```
pauldendulk, line 84: Would it not be sufficient to do a single pack on the Mapsui.Linux.slnf, like it is done in the dotnet-release-nugets.yaml? Less code is better and perhaps it is also faster that way (the build already takes a long time). What is the reason to build the separate csprojs?

pauldendulk (approved):
Approved but please have a look at the comment.

.github/workflows/dotnet.yml
inforithmics: The reason is (I copied it from the mac build (closest to linux that was there)).
I don't know if the pack on mapsui.linux.slnf already works. I'll try that on a different pull request, beacuse i saw that the nupkg.mac should be nupkg.linux

PR #1965, opened 2023-04-18
Index memory provider

inforithmics:
Added an IndexedMemoryProvider, so that when a lot of polygons are used the getfeatures can use an index. 

An alternative solution would be to add the index to the memoryprovider. But as far as I remember I have seen code that adds items to the MemoryProvider after initialization. And this would work with the index because the new items are not indexed.

Should help and improve this
https://github.com/Mapsui/Mapsui/discussions/1867


pauldendulk (approved):
Great. Good to keep it separated from the ordinary MemoryProvider.

PR #1970, opened 2023-04-19
Move MapView to shared project

inforithmics:
This moves the mapview to a shared project, so that other targets could implement a MapView. Is although the recommanded way of sharingt code between forms and maui. (Because eventually forms will dissappear).


pauldendulk:
I would like to wait with this until after some rework has been done on MapControl/MapView. I would like to share more functionality between MapControl and MapView but in order to do that the MapView needs some rewriting first. There are also some problems in the current MapView implementation that I would like to fix first. Like the layer ordering, but this needs an improvement to the core. Also, some of the things that are completely off the shelf in MapView right now will need to be added a bit more explicit. In general I would like to have the functionality of MapView but the users will be a bit more aware of how it works, I might need to do a bit more to get it to work.

pauldendulk:
But of course I support the idea that we should share this kind of code for all all platforms.

inforithmics:
When the mapview is a shared project, parts that can be shared can be moved now to mapsui, or mapsui.nts ... and so the features  are available to all platforms. Because I need the mapview in wpf, this will problably the platform where i will fix the mapview shared project to be ported. But I think this should be another pull request.

pauldendulk:
@inforithmics ah, I see, this is the first step for any kind of refactoring we would want to do.

Why do you want to use MapView in WPF? I would like know because we need some understanding of the use cases for refactoring. MapView makes some things easier. It has Location off the shelf, I think that is important, but it could be more generic layer. It makes working with LatLon easier. Default zoom buttons.

pauldendulk (approved):

charlenni:
Perhaps I could answer this.

I made the MapView once upon a time to make things easier for a first contact with Mapsui, especially for Xamarin.Forms users. They had already a wrapper over map, which exists for the platform they use: on Android for Google Maps, on iOS for Apple Maps. But both weren't usable without a license. 

The idea of MapView is to have an easy way of creating pins and other drawables while using the coordinate system you are used to when you have used Google Maps before. You have ready-made buttons on the control and also the displaying of your location isn't very difficult. All of this could be made with MapControl too, but you said, that you want to have this separated and not included in Mapsui.

Long story short: MapView is a convenience wrapper around MapControl. It provides things, that aren't possible in an easy way with the existing MapControl. The questions seem now, do you need an extra MapView for all platforms, or could you include the benefits from MapView into MapControl (easy placement and handling of pins, predefined buttons, Google coordinate system)?

inforithmics:
The things I would need from Mapview are Pin/Location Layer, the Google Coordinate System, and Predfined Buttons.

And what another advantage of MapView is that a lot of things are Bindable (Xaml, But this proved the biggest hurdle too because on every Platform the Binding is done different).

PR #1971, opened 2023-04-20
WIP MapView.Samples Project

inforithmics:


pauldendulk (commented):
I did not study this carefully but still I think it is better to comment now than to wait until I will have proper time for this. As mentioned earlier I would like to rewrite the MapControl and MapView, but my goal would be to make it simpler. Perhaps even support less off the shelf but support many scenarios by composing smaller components. Perhaps using a builder pattern to make specific scenarios easier. There is now a lot of code in MapView and I want less. I want more code that is completely platform independent. This introduces interfaces for the classes currently in MapView, but this is not the direction I think it should be going. Rather we should look at what these classes are doing and try to think of ways this could be solved with more platform independent code, or no code, or a builder that puts some simpler components together.

inforithmics:
I think I'll close this pull request. Following smaller changes I discovered in this pull requst.

1). MyLocationLayer move from Xamarin/Maui Animation to Mapsui Animation.
2) Move following classes into core Mapsui.

Mapview Classes.
* Distance
* Position
* MapSpan
* PinType
General Classes:
* ObservableRangeCollection
* ObservableCollectionProvider

PR #1972, opened 2023-04-21
Move ObservableRangeCollection and ObservableCollectionProvider

inforithmics:
This two MapView classes have no MapView dependencies and should have lived anyway in the corresponding Mapsui Assemblies.


pauldendulk (approved):
This moves ObservableRangeCollection and ObservableCollectionProvider to more logical locations. However, this will also make them available for wider use, and that is something I would like to avoid. I have often seen that the ObservableCollection becomes problematic when there is a lot of data, and also when data changes a lot. This what happens a lot in Mapsui  where a simple mouse wheel movement can bring you to an entire different set of features. 

Since this is a just a step to further allow refactoring of the MapView I think this should be merged, but eventually I would rather not have any reference to an ObservableCollection in our code.

Issue #1894, opened 2023-03-18
Jerky AnimatedPointFeature animations in Avalonia


pingzing:
Hi! I'm trying out Mapsui with Avalonia, and wanted to implement a small local bus tracker. As a tiny proof of concept, I've implemented a basic `AnimatedPoint` to stand in for a bus. However, I've noticed that when the point's position changes, sometimes it will briefly flicker to its final position before animating into place.

Is this a known bug? The documentation for animation seems sparse, and from looking around the repo, I get the impression that animation is still kind of in the works.

Here's a short gif demonstrating:

![mapsui-flicker](https://user-images.githubusercontent.com/5133649/226103341-a417113c-da8d-4260-9c98-91ef15765599.gif)

Version info:
.NET: 6.0
Avalonia: 0.10.18
Mapsui.Avalonia: 4.0.0-beta.8 
Target platform: Windows 10, build 19045.2728

And some minimal reproduction sample code:

```csharp
public partial class MainWindow : Window
{    
    private MapControl _map;

    public MainWindow()
    {
        InitializeComponent();

        _map = new MapControl();
        _map.Map!.Layers.Add(Mapsui.Tiling.OpenStreetMap.CreateTileLayer("trippit-kiosk-avalonia"));
        _map.Map.Layers.Add(new AnimatedPointLayer(new BusPointProvider())
        {
            Name = "Buses",
            Style = new LabelStyle
            {
                BackColor = new Brush(Color.Black),
                ForeColor = Color.White,
                Text = "Bus",
            }
        });
        Content = _map;

        _map.Map.CRS = "EPSG:3857";
    }


    private class BusPointProvider : MemoryProvider, IDynamic
    {
        public event DataChangedEventHandler? DataChanged;
        
        private readonly PeriodicTimer _timer = new PeriodicTimer(TimeSpan.FromSeconds(1));

        public BusPointProvider()
        {            
            Task.Run(() => RunTimerAsync());
        }

        private (double Lon, double Lat) _prevCoords = (24.945831, 60.192059);
        private async Task RunTimerAsync()
        {
            while(true)
            {
                await _timer.WaitForNextTickAsync();
                
                _prevCoords = (_prevCoords.Lon + 0.00005, _prevCoords.Lat + 0.00005);                                

                OnDataChanged();
            }
        }

        void IDynamic.DataHasChanged()
        {
            OnDataChanged();
        }

        private void OnDataChanged()
        {
[...]

PR #1974, opened 2023-04-21
AnimatedBusSample

inforithmics:
reproduces this. https://github.com/Mapsui/Mapsui/issues/1894

I think the reason is that the Animation duration is 1 second and the next animation entry is 1 second too. So it seems that an existing animation is finished rendering the end position, but from the current position causing this jerky behavior.


inforithmics:
Things done:
Fixed the bumpy behavior by waiting for the old animation to finish, before starting a new animation.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -161,12 +161,12 @@ private void Render(SKCanvas canvas, Viewport viewport, IEnumerable<ILayer> laye
     private void RenderFeature(SKCanvas canvas, Viewport viewport, ILayer layer, IStyle style, IFeature feature, float layerOpacity, long iteration)
     {
         // Check, if we have a special renderer for this style
-        if (StyleRenderers.ContainsKey(style.GetType()))
+        if (StyleRenderers.TryGetValue(style.GetType(), out var renderer))
```
inforithmics, line 164: only access the dictionary once and only call GetType once

Mapsui.Rendering.Skia/SkiaStyles/LineStringRenderer.cs
```diff
@@ -20,7 +20,7 @@ public static class LineStringRenderer
 
         SKPaint paint;
         SKPath path;
-        if (vectorCache == null || layer is IWritableLayer)
+        if (vectorCache == null || layer is IModifyFeatureLayer)
```
inforithmics, line 23: Renamed WritableLayer to ModifyFeature Layer

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -120,6 +123,12 @@ private void InvalidateTimerCallback(object? state)
             return;
         }
 
+        if (_invalidated)
```
inforithmics, line 126: when the invalidate does not trigger a drawing fast enough it could trigger two invalidates in short order making the animation bumpy.

Mapsui/Layers/AnimatedLayers/AnimatedFeatures.cs
```diff
@@ -38,10 +38,15 @@ public AnimatedFeatures()
     public int AnimationDuration { get; set; }
     public EasingFunction Function { get; set; }
 
-    public void AddFeatures(IEnumerable<PointFeature> features)
+    public async Task AddFeaturesAsync(IEnumerable<PointFeature> features)
     {
-        var previousCache = _cache;
+        while (_animating)
```
inforithmics, line 43: wait the current animation to finish

Samples/Mapsui.Samples.Common/Maps/Animations/BusPointProvider.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
```
inforithmics, line 1: Bumpy Bus Sample

charlenni:
I'm not sure, but I would think, that you shouldn't wait for the old animation, but start the new animation from the point, where the last animation is at this moment. E.g. if you fly from New York to Paris and start a new animation to London, if you're over Dublin, then I expect to change to the new animation, in this case from Dublin to London.

inforithmics:
@charlenni good idea I'll try to stop the current animation instead of waiting.

pauldendulk:
> I'm not sure, but I would think, that you shouldn't wait for the old animation, but start the new animation from the point, where the last animation is at this moment. E.g. if you fly from New York to Paris and start a new animation to London, if you're over Dublin, then I expect to change to the new animation, in this case from Dublin to London.

Yes, that is how it is supposed to work.

inforithmics:
The animation is now stopped the following way. I change the start animation time so that the animation is finished during one task.delay wait. The reason I stop the animation this way is because when the animation is running and I update the animation a jump occurs. I guess that the rendering and the animation update happen on two different threads. This is the only way I found to reliable render it without a glitch. 

Things I tried but still caused glitches.

1) Use a lock so that the update and adding of animation.
2) Animation progress on the animation feature.
3) clear the animation before updating. The animation does not work anymore because it needs the old value for animating.

pauldendulk (approved):
Thanks for the fix! 

This is general idea of how it should have worked, but perhaps this needs more fundamental changes:
- There is some initial position
- Then a new position comes in. At that moment animation is started from the initial position to the new position over some time period. During that animation there is a difference between the actual position and the animation position. The renderer only shows the animated position.
- If a new position comes in during the animation the previous animation should be replaced by a new animation which should have the current animated position as starting point.
- The AnimationSource would have the actual position. The AnimationLayer should only return the animated position. 
- The animation entry has a Final tick which puts the animation exactly on the destination position. This Final method should not be called if a running animation is replaced with  a new animation. In that situation that final position should never be shown.

inforithmics:
Somehow without my fix it calls the final with the new final destination so it jumps. The only reliable way I found is to finish the current animation. And then starting the new animation.

pauldendulk:
I still see bumps in the animations now and then. I am looking into it right now. 

btw, copying these user generated samples into our own source is a good way to improve our samples and allows us to easily test functionality and fix it.

PR #1980, opened 2023-05-05
fix artifical clipping on rotated map

### LINKED ISSUE #1975: [Regression] Artifical clipping on rotated map
**Mapsui Version**
4.0.0-beta.11

**Mapsui Platform**
Xamarin.Forms

**Device**
Android 13

**Describe the bug**
I'm seeing a problem with rotated maps with v4.0.0-beta.11 on Android. If I draw a circle on a rotated map, it is somehow clipped artifically (smaller than the actual viewport). See screenshot below (map rotated by 90°).

This was working with beta.10 (the problem appeared with beta.11). It works on iOS, but fails on Android. Non-rotated maps work fine. I suspect it was broken by #1959.

**Expected behavior**
The circle should be drawn over the full viewport and not be clipped artificially.

**Screenshots**
Screenshot_20230424_115218

janusw:
This is supposed to fix #1975 (a problem introduced by #1959).


inforithmics:
For making the build green
You have to cherry pick following commit
https://github.com/Mapsui/Mapsui/pull/1974/commits/222942b0f29d9b17bd5b9abfb32ea3a8241c6f3f

janusw:
> For making the build green You have to cherry pick following commit [222942b](https://github.com/Mapsui/Mapsui/commit/222942b0f29d9b17bd5b9abfb32ea3a8241c6f3f)

Thanks for the hint ;) I'll probably just wait until you have this on master. Btw, I don't see any difference by eye, and the commit contains zero explanation, so what's the difference? Also, why doesn't this fail on current master (from where I branched)?

inforithmics:
The updated image is 3kb bigger, I think the reason why it fails is that osm got updated in this region, making enough changes on a pixel level to fail. Normally the osm calls are in the cached so that updating the map doesn't cause this behavior.

This checkin here updates the sqlite files (cached osm calls).
https://github.com/Mapsui/Mapsui/pull/1974/commits/a9b89942755ab3a8ca7b2e159254093b70db4d06

so this shouldn't happen anymore.
Maybe I have to make a seperate pull request only for fixing the unit tests and sqlite and png files.

Probably updating the master with any new pull request with out this fix will break master.

janusw:
> The updated image is 3kb bigger, I think the reason why it fails is that osm got updated in this region, making enough changes on a pixel level to fail. Normally the osm calls are in the cached so that updating the map doesn't cause this behavior.
> 
> This checkin here updates the sqlite files (cached osm calls). [a9b8994](https://github.com/Mapsui/Mapsui/commit/a9b89942755ab3a8ca7b2e159254093b70db4d06)
> so this shouldn't happen anymore.

Thanks for the explanation!


> Maybe I have to make a seperate pull request only for fixing the unit tests and sqlite and png files.

Makes total sense, if you ask me ;)
(Better than mixing it up with another feature PR ...)

inforithmics:
pull request with updated image test file and updated cache files so that the build succeeds

https://github.com/Mapsui/Mapsui/pull/1985

pauldendulk (approved):
Thanks!

PR #1981, opened 2023-05-07
My location mapsui animation

inforithmics:
I'm trying to make the Location Animation with Mapsui Animation, it is not yet working


inforithmics:
To View the animation in action AnimatedMyLocationSample can be used (currently only for maui because of .net 6 dependency)

pauldendulk (approved):
Thanks! This improves the existing MyLocationLayer. This sample depends on the `Microsoft.Maui.Controls.Animation`. Eventually we would want this to be independent of a platform. 

My preferred solution for animations is to not use any timers, or something similar, that need to be started and stopped and that consume extra threads. Instead it should depend on the rendering loop. Within the loop the state of the animation should be updated for the rendering that immediately follows. This way there is no problem of syncing the animation updates with the render loop updates. Also it uses less threads, which is better for performance an makes the code easier to understand (threads are complicate things, they often need cross thread synchronization).

charlenni:
@pauldendulk Wasn't this the case before? If I'm remember it correct, the original implementation of animations was in this way (called before rendering).

pauldendulk:
@charlenni We have animations in a few places and they do not all follow the same logic. I did not study the code in this MR in detail but I see it uses a MAUI Animation class, so I assume it is not using the pre-render-update logic. This logic is used in the SymbolAnimations, FeatureAnimations and the ViewportAnimations, but still in slightly different ways. This could be improved.

Issue #1218, opened 2021-09-17
ScaleBarWidget can not draw because the Map.CRS is not set


pauldendulk:
This is the error seen by our user:

```
2021-09-17 12:51:17.839863+0200 caminomobileapp.iOS[32980:1360459] MAPSUI Error: ScaleBarWidget can not draw because the Map.CRS is not set -
2021-09-17 12:51:18.411406+0200 caminomobileapp.iOS[32980:1360459] MAPSUI Error: ScaleBarWidget can not draw because 
```

Perhaps the widget should assume the SphericalMercator projection in this case.

pauldendulk:
@dinonovak Could it be you do add the ScaleBarWidget somewhere in your own code? 

map.Widgets.Add(new ScaleBarWidget(....

dinonovak:
@pauldendulk yes I was adding ScaleBarWidget in my code

charlenni:
If you want to use ScaleBarWidget, then you have to initialize the map like this:

            var map = new Map
            {
                CRS = "EPSG:3857",
                Transformation = new MinimalTransformation()
            };

If you don't do it this way, the ScaleBarWidgt don't know, what type of CRS the map has and couldn't calculate any values.

charlenni:
@pauldendulk This is only included in the Forms documentation. Perhaps you want to set this values as default? Or you have to mention this on all platforms.

janusw:
> Perhaps you want to set this values as default? 

I agree that reasonable default values would be useful. 👍

janusw:
This is fixed via #1983.

PR #1983, opened 2023-05-11
Map: set default value for CRS

janusw:
* this fixes the problem that the ScaleBarWidget cannot be drawn if no CRS is set (see #1218)
* EPSG:3857 (SphericalMercator) is a reasonable default value (this was actually suggested by @charlenni)


Mapsui/Map.cs
```diff
@@ -55,9 +55,10 @@ private void Navigator_ViewportChanged(object? sender, PropertyChangedEventArgs
     public ConcurrentQueue<IWidget> Widgets { get; } = new();
 
     /// <summary>
-    /// Projection type of Map. Normally in format like "EPSG:3857"
+    /// Coordinate reference system (projection type of map).
+    /// Default: "EPSG:3857" (SphericalMercator).
     /// </summary>
-    public string? CRS { get; set; }
+    public string? CRS { get; set; } = "EPSG:3857";
```
inforithmics, line 61: I would change the logic here so that it asks first the layers if there is a crs set, And only if on the layers there is no crs set I would set it to EPSG:3857.

Something like this

   ```
 /// <summary>
    /// Projection type of Map. Normally in format like "EPSG:3857"
    /// </summary>
    public string? CRS
    {
        get
        {
            if (_crs == null && this._layers.Count > 0)
            {
                foreach (var layer in this._layers)
                {
                    if (layer is ILayerDataSource<IProvider> { DataSource: not null } dataSourceLayer)
                    {
                        if (!string.IsNullOrEmpty(dataSourceLayer.DataSource?.CRS))
                        {
                            _crs = dataSourceLayer.DataSource?.CRS;
                        }
                    }
                }
            }

            return _crs ?? "EPSG:3857";
        }
        set { _crs = value; }
    }
```

pauldendulk: But the DataSource.CRS is not necessarily the same as the Map.CRS. The Layer could project from DataSource.CRS to Map.CRS.

pauldendulk:
> * this fixes the problem that the ScaleBarWidget cannot be drawn if no CRS is set (see [ScaleBarWidget can not draw because the Map.CRS is not set #1218](https://github.com/Mapsui/Mapsui/issues/1218))
> * EPSG:3857 (SphericalMercator) is a reasonable default value (this was actually suggested by @charlenni)

An alternative would be to let the ScaleBarWidget itself assume a default CRS. Fixed like that only the ScaleBarWidget itself would be impacted. Perhaps better.

For the longer term I am thinking of builders for specific scenarios. A builder with an OSM background would also set the CRSes.

pauldendulk (approved):
There my be other options but this one is easy to understand.

janusw:
> > * this fixes the problem that the ScaleBarWidget cannot be drawn if no CRS is set (see [ScaleBarWidget can not draw because the Map.CRS is not set #1218](https://github.com/Mapsui/Mapsui/issues/1218))
> > * EPSG:3857 (SphericalMercator) is a reasonable default value (this was actually suggested by @charlenni)
> 
> An alternative would be to let the ScaleBarWidget itself assume a default CRS. Fixed like that only the ScaleBarWidget itself would be impacted. Perhaps better.

Thanks for the comments here (and for merging). If someone wants to follow up on this, some open questions from my side:
* What else is the Map.CRS used for (appart from the ScaleBarWidget)?
* How to deal with differing CRSs in the map/layers (and possibly widgets)? Should this be allowed or do we need checks to avoid it? Probably the latter, right?
* Should one use the CRSs from the layers to set the CRS of the map (as suggested by @inforithmics )? And/or the other way around?

inforithmics:
I can life with the current default: Because I always use this default. But setting the crs and the layer to different crs might cause the scalebarwidget to display wrong values. 
Because the Transformation to the Map.Crs has to be done anyway explicitly it doesn't matter there. But this was/is a Main difference between MapView and MapControl and because this default Mapcontrol assumes now the same crs as the MapView.

pauldendulk:
> 
> - What else is the Map.CRS used for (apart from the ScaleBarWidget)?

How it should work, and how it works in most other map components is that the Map.CRS determines the projection in which the map is displayed, and different DataSource can be added which are automatically transformed to the Map.CRS. This way the user would have to choose only the Map.CRS and could add all kinds of DataSources without having to worry about their projection. Although sometimes projection from one CRS to another may not be possible. In that case an error should be reported for that DataSource. There are various ways in which a source could be projected to the Map.CRS, for instance:
- Projection of the locally available vector data using the projection component of the application. In the case of Mapsui this could be NTS geometries or the MPoint, projected with DotSpatial.Projections. In that case the DataSource.CRS needs to be set.
- A WMS often supports projection server side, in that case the Map.CRS should be passed into the request for the WMS image.
- A WMTS sometimes supports a number of CRSs. So the Map.CRS should be used to request the compatible tile set.
- Other data sources sometimes also support projection, like some database.

> - How to deal with differing CRSs in the map/layers (and possibly widgets)? Should this be allowed or do we need checks to avoid it? Probably the latter, right?

The DataSource has a CRS. These could all be different. In the current implementation the Layer has the responsibility to project the data to the Map.CRS. If this is not the case an error should be shown.

> - Should one use the CRSs from the layers to set the CRS of the map (as suggested by @inforithmics )? And/or the other way around?

You choose the Map.CRS, the layers should be compatible. 

In the current implementation there is no automatic transformation. It had partial automatic transformation in V3 but it was not centralized. I removed this, but I would like to add this again in the future, but there should be one place in our code for all projection (so it should not be in the individual layers).

PR #1984, opened 2023-05-11
Avalonia.MapControl: fix zoom with touchpad on MacOS & Linux

yapus:
fix regression in commit ce0f5edee10fffcf424d776318853d9e097c863f


yapus:
@inforithmics Could you please merge this ?

inforithmics:
I looked at the old pull request and the logic is the same as in the old version so this is okay.
But out of curiosity I looked at other platforms and saw that Eto and Blazor have floating point numbers for mousewheel too. 
than I looked at the Navigator Mousewheel Implementation and saw that it could handle double values (it simply defined the mousewheel delta as int.

So another solution as you suggested would be to simply change the mousewheel delta to double and it should work without changes. 

What this would change is that in avalonia more mousewheel events would be triggered,

@pauldendulk  what's your opinion on this.

pauldendulk:
I need to understand the problem or I'll break it again. 
- So Avalonia uses the touchpad to trigger the mousewheel event. Why is this not a problem of Windows? I see in the debugger that on Windows the value of Delta.Y is always -1 or 1. What kind of values does it have in Linux? Could be smaller than 1?
- What would happen if we did: `var mouseWheelDelta = Math.Sign(e.Delta.Y);` instead? Are there more touchpad events on Linux, would it go to fast like this? 
- For real mouse wheel events we step to discrete resolutions, but for touchpad we could do a more granular zoom, similar to pinch zoom. Is that what you were suggesting @inforithmics?.

inforithmics:
> I need to understand the problem or I'll break it again.
> 
> * So Avalonia uses the touchpad to trigger the mousewheel event. Why is this not a problem of Windows? I see in the debugger that on Windows the value of Delta.Y is always -1 or 1. What kind of values does it have in Linux? Could be smaller than 1?
as I understand it the Y values are very small for example 0.1 (but are more often triggered) so the sum gets eventuellty to 1 and the a mapsui event is triggered. Maybe it is better to trigger every smal value and change the mousewheel delte paramter to double.

> * What would happen if we did: `var mouseWheelDelta = Math.Sign(e.Delta.Y);` instead? Are there more touchpad events on Linux, would it go to fast like this?
It would be 0 i think because the value is for example 0.1

> * For real mouse wheel events we step to discrete resolutions, but for touchpad we could do a more granular zoom, similar to pinch zoom. Is that what you were suggesting @inforithmics?.
Yes that's what i meant and probably what is happening in Linux. Touche => mousewheel small values.

pauldendulk (approved):
Thanks for the fix. 

The alternative would be to use doubles for mouse wheel zoom. In theory this is possible but there are some complications. It would cause more updates, that is fine for RefreshGraphics, but RefreshData may be called too often. Also there is some logic for accumulation of mouse wheel events (this allows faster zoom in/out), I am not sure how this would be affected. So, this is the more conservative approach. I will add some comments in the code to prevent this regression from happening again.

Issue #1995, opened 2023-05-22
[Regression] Pins are not rendered with rc.1


janusw:
**Mapsui Version**
4.0.0-rc.1

**Mapsui Platform**
Xamarin.Forms

**Device**
Android, iOS, Mac

**Describe the bug**
With rc.1, the Pins on a MapView are not being rendered any more. With beta.11 they are still visible. I'm using `PinType.Icon`, but changing this to `PinType.Pin` does not seem to make a difference. Happens on all platforms that I tried.

I would expect that the problem should be visible in the samples, but I haven't checked that yet. Looking at the recent git history, #1973 seems like it might be a possible culprit (it's the only PR since beta.11 that touches the Pin class).

pauldendulk:
I can reproduce this in the MapView 'Add Pin Sample'. The callout is visible but not the pin.

janusw:
> I can reproduce this in the MapView 'Add Pin Sample'. The callout is visible but not the pin.

Thanks for the confirmation. I have verified that 4.0.0-beta.11-37-g4788199a1 still works.

janusw:
> I have verified that 4.0.0-beta.11-37-g4788199a1 still works.

OTOH, 4.0.0-beta.11-47-gce2544005 fails (which is the merge commit of #1973), so I'd conclude that #1973 indeed caused this issue.

@inforithmics, do you understand how the problem comes about?

inforithmics:
I‘ll investigate it

janusw:
Fixed via #1997.

PR #1997, opened 2023-05-23
fixed wrong defaults after BindableObject => IPropertyChanges (Pins are not visible by default)

inforithmics:
fixes this: https://github.com/Mapsui/Mapsui/issues/1995

The Change from BindableObject, I changed the defaults for the Properties. And for example the default value of the Visible Property was false, that's why the pins didn't render. I reviewed all the Properties and they should have now again the same default values.

I look into the other BindableObject, if I have forgotten some defaults there too.


inforithmics:
I reviewed the changes in the BindableObject => IPropertyChanges commit and it seems I only forgot the set the defaults in the Pins.

janusw (approved):
> The Change from BindableObject, I changed the defaults for the Properties. And for example the default value of the Visible Property was false, that's why the pins didn't render. I reviewed all the Properties and they should have now again the same default values.

Ah, it's that simple :)

I can confirm that this makes the pins show up again. Thanks for fixing!

PR #1999, opened 2023-05-26
Upgrade Microsoft.Net.Compilers.Toolset to 4.6.0

pauldendulk:
After upgrading my Visual Studio yesterday to the latest I got this error:
```console
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS9057	The analyzer assembly 'C:\Program Files\dotnet\sdk\7.0.302\Sdks\Microsoft.NET.Sdk\codestyle\cs\Microsoft.CodeAnalysis.CodeStyle.dll' references version '4.6.0.0' of the compiler, which is newer than the currently running version '4.5.0.0'.	Mapsui (netstandard2.0)		
```
The upgrade fixed it.

### DIFF
--- a/.github/workflows/dotnet-docs.yml
+++ b/.github/workflows/dotnet-docs.yml
@@ -18,7 +18,14 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v3
       with:
-        fetch-depth: 0      
+        fetch-depth: 0 
+    # .Net 7 update     
+    - name: Setup .NET 7 SDK
+      uses: actions/setup-dotnet@v3
+      with:
+         dotnet-version: |
+           7.0.x
+         include-prerelease: false
     - name: Run DocFx
       # Use a .cmd so this script can be run locally for testing
       run: docfx\build-site.cmd      
--- a/.github/workflows/dotnet-release-nugets.yml
+++ b/.github/workflows/dotnet-release-nugets.yml
@@ -23,6 +23,14 @@ jobs:
 
     - name: Setup NuGet
       uses: NuGet/setup-nuget@v1.0.5
+
+    # .Net 7 update     
+    - name: Setup .NET 7 SDK
+      uses: actions/setup-dotnet@v3
+      with:
+         dotnet-version: |
+           7.0.x
+         include-prerelease: false
       
     - name: Setup msbuild
       uses: microsoft/setup-msbuild@v1.1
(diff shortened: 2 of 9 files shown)


inforithmics:
The error message means something in the windows app sdk isn't anmyore compatbile i'll upgrade the analyzers and the windows app sdk to see if it runs

inforithmics:
intrestingly locally this works. So it seems that to merge this upgrade we have to wait until the Image gets upgraded to Visual Studio 17.6

inforithmics:
<RunApiCompat>false</RunApiCompat> was the fix for the build.

Should I make a second pull request with the changes for winui (removing workarounds, and fixing dark mode).

pauldendulk:
Thanks a lot for helping out. This is the kind of issue I struggle with. I am very eager to upgrade everything to .NET 6. Very unfortunate that simply upgrading VS breaks the build. That should not happen.

Fine that just work with me in the same MR.

The release nugets actions still fails for this branch on `dotnet workload install maui macos android ios maccatalyst wasm-tools wasm-tools-net6`, but that command is identical to the windows build:
https://github.com/Mapsui/Mapsui/actions/runs/5097832115/jobs/9164604923

inforithmics:
I forget to add the .net 7 install to the release yml. will do it now.

pauldendulk:
> I forget to add the .net 7 install to the release yml. will do it now.

Thanks, that fixed it. Now there is an error later in the script 
`error MSB4018: The "Microsoft.DotNet.ApiCompat.Task.ValidatePackageTask" task failed unexpectedly` https://github.com/Mapsui/Mapsui/actions/runs/5103128241/jobs/9173132346#step:10:479

inforithmics:
I set 
`    <RunApiCompat>false</RunApiCompat>`
no in 
Directory.Build.props
so that in every project the runapicompat task is disabled. (We don't use it currently anyway). The task checks for breaking changes between Versions. It enforces the Correct Version numbering for Releases.

pauldendulk:
> I set 
` <RunApiCompat>false</RunApiCompat>` 
now in Directory.Build.props so that in every project the runapicompat task is disabled. (We don't use it currently anyway). The task checks for breaking changes between Versions. It enforces the Correct Version numbering for Releases.

I know nothing about this. Read a bit now here: https://learn.microsoft.com/en-us/dotnet/fundamentals/package-validation/overview

> So can you try again?

Fixed it!

PR #2002, opened 2023-05-28
Improvements in feature animations

pauldendulk:


pauldendulk:
- Turned the AnimatedFeature into a AnimatedPointFeature which is now an actual feature derived from PointFeature. I think this class is now easy to grasp. It holds, the current, start and end position, and calculates the new current position at a given moment in time.
- Removed the AnimatedFeatures, and moved its functionality to the AnimatedPointLayer. In v3 there were Feature classes. These were removed. Now we have just collections of features. It was not clear what the purpose of this class was. An alternative setup would be to use a MemoryLayer and introduce a AnimatedPointProvider which wraps a provider of the source data. Similar to the ProjectingProvider.
- Now using the Easing function of Mapsui.Animations (instead of a custom animation).

pauldendulk:
Some possible future improvements:
- All values related to the animation could be encapsulated within AnimatedPointFeature. There should be a way to change the default on a more global level though.
- There should be a PollingProvider which could encapsulate an ordinary provider so that someone implementing an animated layer could write even simpler code.
- The current solution expects a user to implement a provider. This should not be necessary. The user could have whatever event which comes up with a new position, this could be code not related to Mapsui. From there it should be easy to update an AnimatedPointFeature in for instance a MemoryProvider. 
- Perhaps all layers should have the ability to update the AnimatedPointFeature. Perhaps the AnimatedPointLayer.UpdateAnimations method should be in the BaseLayer. Iterating over all features all the time may impact performance. Perhaps there should be a `BaseLayer.HasAnimatedFeatures = false;`

Issue #1918, opened 2023-03-30
WFS Provider


asim27:
**Describe the bug**
I am not able to create a wfsprovider from latest version [4.0.0-beta.8](https://github.com/Mapsui/Mapsui/releases/tag/4.0.0-beta.8)
WfsProvider method with URI parameter raises System.NullReferenceException: 'Object reference not set to an instance of an object.'
on calling InitAsync() method
**To Reproduce**
1. Create a WFSProvidere with 
![image](https://user-images.githubusercontent.com/24899310/228777758-6096c088-4a9c-4546-80f8-fb3ce0156c6e.png)
2. Call .InitAsync() method on the WFSProvider instance.
3. Handle the exception
4. See the error

**Expected behavior**
System.NullReferenceException: 'Object reference not set to an instance of an object.'

**Screenshots**
![image](https://user-images.githubusercontent.com/24899310/228778308-13889023-fdd7-4fe1-b5b4-5eec3d758e22.png)

**Desktop (please complete the following information):**
 - Windows 11
 -  [4.0.0-beta.8](https://github.com/Mapsui/Mapsui/releases/tag/4.0.0-beta.8)

**Smartphone (please complete the following information):**
 - Windows Application

inforithmics:
You should use WfsProvider.CreateAsync ... and WmsProvider.CreateAsync ... because Constructors cannot be asynchronous this methods initialize the wfs and wms provider asynchronous because a http call is involved.

gpsaliola:
Ok, now i've tried this code:

```
        protected async override void OnAppearing()
        {
            base.OnAppearing();

            mapView.Map = await CreateMapAsync();

            Mapsui.Widgets.ScaleBar.ScaleBarWidget scale = new Mapsui.Widgets.ScaleBar.ScaleBarWidget(mapView.Map)
            {
                TextAlignment = Alignment.Center,
                HorizontalAlignment = HorizontalAlignment.Center,
                VerticalAlignment = VerticalAlignment.Top
            };
            mapView.Map.Widgets.Add(scale);
        }

        public static async Task<Mapsui.Map> CreateMapAsync()
        {
            var map = new Mapsui.Map { CRS = "EPSG:3857" };
            map.Layers.Add(await CreateLayerAsync());
            return map;
        }

        public static async Task<ILayer> CreateLayerAsync()
        {
            return new ImageLayer("mainmap") { DataSource = await CreateWmsProviderAsync() };
        }

[...]

PR #2008, opened 2023-05-30
Projection wms sample

inforithmics:
Reproduction of this https://github.com/Mapsui/Mapsui/issues/1918 (No fix yet)


inforithmics:
Things done:
Fixed Projection Provider (Resolution was not projected).
Fixed WmsProvider Extent (Axis switch)
Fixed WmsProvider general axis switch not only 4326
Added Projection Wms Sample

Mapsui/Providers/ProjectingProvider.cs
```diff
@@ -45,7 +45,9 @@ private bool GetFetchInfo(ref FetchInfo fetchInfo)
 
         // throws exception when CRS or _provider.CRS is null (so I don't have to check it here)
         _projection.Project(CRS!, _provider.CRS!, extentToProject);
-        fetchInfo = new FetchInfo(new MSection(extentToProject, fetchInfo.Resolution), CRS, fetchInfo.ChangeType);
+        var scale = fetchInfo.Extent.Width / extentToProject.Width;
+        var projectedResolution = fetchInfo.Resolution / scale;
```
pauldendulk, line 49: Never realized that this was a problem. How did the problem show?

Mapsui/Layers/ImageLayer.cs
```diff
@@ -14,16 +14,22 @@
...
 using Mapsui.Providers;
+using Mapsui.Styles;
 
 namespace Mapsui.Layers;
 
 public class ImageLayer : BaseLayer, IAsyncDataFetcher, ILayerDataSource<IProvider>, IDisposable, ILayer
 {
+    public ImageLayer()
+    {
+        Style = new RasterStyle();
```
pauldendulk, line 25: This will save a lot of time for many people.

Mapsui/Providers/Wms/WmsProvider.cs
```diff
@@ -41,6 +41,7 @@ public class WmsProvider : IProvider, IProjectingProvider
     private readonly Client? _wmsClient;
     private Func<string, Task<Stream>>? _getStreamAsync;
     private readonly IUrlPersistentCache? _persistentCache;
+    private static readonly CrsAxisOrderRegistry _axisOrder = new();
```
pauldendulk, line 44: Great that you now can reuse this really old class, one generated by @Fobermaier

pauldendulk (approved):
Great PR. Fixes many things. Love to see the CrsAxisOrderRegistry used again.

Mapsui/Providers/ProjectingProvider.cs
inforithmics: It showed that the image size was width= 0 and height = 0 because there is such a big scale difference. And because of that the wms server said invalid image size.

Mapsui/Providers/Wms/WmsProvider.cs
inforithmics: The same class is used in brutile, the question is should I reuse this class in brutile, but then I would have to move the wms provider to Mapsui.Tiling. So that I have access to this class.

pauldendulk: I think this is the right choice, we don't want to add a BruTile dependency here also I was considering to split of the CrsAxisOrderRegistry to a separate repository, this was simply because I wanted to get rid of the Windows dependency, but still it might be separate thing in the future. Perhaps a nuget package with just csharp code (never used that but I read this is possible) so that Mapsui has a dependency but not our users.

Issue #2014, opened 2023-06-03
Callouts do not work in Blazor


pauldendulk:
**Mapsui Version**
4.0.0-rc.2

**Mapsui Platform**
Blazor, not on WPF or Avalonia.

**Device**
Windows, in the Chrome browser

**Describe the bug**
The Info | Single Callout sample should show a callout when clicking on a symbol but it does not. The logger mentions a DllNotFoundException with as message libHarfBuzzSharp.

![image](https://github.com/Mapsui/Mapsui/assets/963462/a5996711-eb6f-474f-ab63-b95b0ae62038)

**To Reproduce**
1. Run Samples.Mapsui.Blazor which is part of the Mapsui repository, or go to https://mapsui.com/samples/
2. Select category 'Info'
3. Select sample 'Single Callout'
4. Click on a symbol.
5. Don't see a callout

**Expected behavior**
Show a callout.

pauldendulk:
Current status: Fixed for our sample, but not for our users.

mattleibow:
I think this will fix it: https://github.com/mono/SkiaSharp/pull/2497

CI is still working, but a `2.88.4-preview.77` version should appear in the feed https://aka.ms/skiasharp-eap/index.json

pauldendulk:
@mattleibow  I upgraded to SkiaSharp 2.88.4-preview.77 but did not fix the problem. Upgrading to 2.88.4-preview.82 that not fix it either. Here is the PR with the upgrade https://github.com/Mapsui/Mapsui/pull/2086/files

This line needs to be enabled:
```<NativeFileReference Include="libHarfBuzzSharp.a" />```

This is the message of the exception:
```Unexpected error in skia rendererException in BuildFontRuns() with original length of 53 now 53, style run count 1, font run count 0, direction overrides: False```

inforithmics:
Made a pull request for this

https://github.com/Mapsui/Mapsui/pull/2103

pauldendulk:
Assume this is fixed with the MR. Please test on the next release.

PR #2015, opened 2023-06-03
Implement workaround for missing callout in Blazor with workaround #2014

pauldendulk:
- Add a simple test to show a callout
- Fix the missing callout in Blazor by adding a workaround suggested here: 
   - https://github.com/mono/SkiaSharp/issues/1725#issuecomment-978843629

The workaround fixes the problem for our sample but our users will have to apply the same workaround. 

How to apply the workaround.
- Add this to the csproj of your main project 
```xml
<ItemGroup>
  <NativeFileReference Include="libHarfBuzzSharp.a" />
</ItemGroup>
```
- Now you need to copy that file to your project. For this you need to download the nuget package https://www.nuget.org/packages/HarfBuzzSharp.NativeAssets.WebAssembly/
- Unzip the nuget. 
- Copy the file from this folder `build\netstandard1.0\libHarfBuzzSharp.a\2.0.23` to you project folder.


pauldendulk:
A real fix would deploy libHarfBuzzSharp automatically. I would except that adding HarfBuzzSharp.NativeAssets.WebAssembly to the csproj should solve the problem, but it did not.

pauldendulk:
@inforithmics I would expect HarfBuzzSharp.NativeAssets.WebAssembly to the samples csproj should fix the problem. What do you think? Did I miss something? If not if will submit a bug or question in SkiaSharp.

inforithmics:
https://github.com/mono/SkiaSharp/issues/1725

inforithmics:
here in this pull request I added a targets file.
https://github.com/Mapsui/Mapsui/pull/1856

and included it in the nuget package. So maybe we could make a blazor.targets file.

```
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
	</ItemGroup>
		<NativeFileReference Include="libHarfBuzzSharp.a" />
	</ItemGroup>
</Project>

```
and add it to the nuget.

```
<!--target to add Harfbuzzsharp.NativeAssets.Webassembly -->
  <ItemGroup>
    <None Include="Mapsui.Forms.targets" PackagePath="build/net6.0/Blazor.targets" Pack="true" />
    <None Include="Mapsui.Forms.targets" PackagePath="build/net7.0/Blazor.targets" Pack="true" />
  </ItemGroup>

```


https://github.com/Mapsui/Mapsui/pull/1856

inforithmics (approved):
Nice that a specific sample is added

Testet it and it works. The targets approach I have to test if it works

pauldendulk:
> https://github.com/mono/SkiaSharp/issues/1725
> And I applied the workaround here in the csproj
> https://github.com/Mapsui/Mapsui/blob/master/Samples/Mapsui.Samples.Uno/Mapsui.Samples.Uno.Wasm/Mapsui.Samples.Uno.Wasm.csproj
> in the uno.winui.wasm too.

I found my workaround on that same thread. The one use used for Uno did not work for Blazor (just tested again, it still does not work). I had to use the one mentioned earlier in the thread. 

Let's add the targets file like you proposed because now everyone of our users has to find that same obscure thread to fix their problem.

PR #2018, opened 2023-06-03
Shared mapsui samples mapview

inforithmics:
Move MapView Samples to Mapsui.Samples.MapView shared project


pauldendulk (approved):
Makes sense

PR #2019, opened 2023-06-03
fixed warnings

inforithmics:


Samples/Mapsui.Samples.Maui/ViewModel/MainViewModel.cs
```diff
@@ -39,7 +39,7 @@ public MainViewModel()
     ISampleBase selectedSample;
 
     [ObservableProperty]
-    Map map;
+    Map? map;
```
pauldendulk, line 42: I am still struggling with the nullability of the map. I made MapControl.Map non nullable in order to make a lot of code simpler. You would think we could just decide that. The Map is just not allowed to be null, if it is assigned null we throw an exception. But if you want to bind the Map to a Map field in a ViewModel in MAUI the MapControl.Map is still assigned a null value at some moment during the assignment, even if you ViewModel.Map is not null either. Perhaps this is a bug, or at last something in MAUI that could be improved. 

Now here you made the ViewModel.Map nullable. I would think that as a consequence the MapControl.Map is nullable. I don't have a solution for this yet.

pauldendulk (approved):
There is a conflict now

PR #2020, opened 2023-06-03
MapControl: fix missing TouchEnded events

### LINKED ISSUE #2017: MapControl.TouchEnded is not always triggered
**Mapsui Version**
4.0.0-rc.2

**Mapsui Platform**
Xamarin.Forms

**Device**
Android & iOS

**Background**
I want to implement a custom behavior for single-finger dragging in a `MapView`. To achieve this, I enable `MapView.PanLock` (and possibly `MapView.ZoomLock`) and implement handlers for the events `MapControl.TouchStarted`, `MapControl.TouchMove` and `MapControl.TouchEnded`.

**Describe the bug**
`TouchStarted` and `TouchMoved` are basically triggered as expected, but I noticed that sometimes the `TouchEnded` event seems to be missing at the end of a single-finger dragging gesture.
At first I was completely puzzled about when the problem occurs, but after playing around for a while I noticed:
* When keeping the finger steady for a moment, before lifting it off the screen, `TouchEnded` is triggered quite reliably.
* When performing a fast dragging gesture and lifting the finger right during the movement (without stopping), `TouchEnded` is omitted.

janusw:
This is supposed to fix #2017. For discussion see the issue.

### DIFF
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -323,11 +323,6 @@ public partial class MapControl : ContentView, IMapControl, IDisposable
 
                 _flingTracker.RemoveId(e.Id);
 
-                if (_touches.Count == 1)
-                {
-                    e.Handled = OnTouchStart(_touches.Select(t => t.Value.Location).ToList());
-                }
-
                 if (!e.Handled)
                     e.Handled = OnTouchEnd(_touches.Select(t => t.Value.Location).ToList(), releasedTouch.Location);
             }


janusw:
Unfortunately it seems like this does not fix the problem :/

pauldendulk:
> Unfortunately it seems like this does not fix the problem :/

TouchEnded is still not called or is something else failing?

janusw:
> > Unfortunately it seems like this does not fix the problem :/
> 
> TouchEnded is still not called or is something else failing?

TouchEnded is simply not being called, neither with nor without the change from this PR.

My initial assumption here was that a `TouchStarted` is being triggered instead of a `TouchEnded`, but I checked this now, and it's actually not the case. For further discussion see the issue.

I'm closing this experimental PR as NOTWORKING ;)

PR #2021, opened 2023-06-03
Blazor sample with code snippets

inforithmics:
Add a tab in the Blazor Samples so that the source code can be seen.


inforithmics:
The main idea is that with docfx I generate the html pages with the c# code. And in the deployed Blazor I display in an iframe the code html. with a relative path.

pauldendulk (approved):
Wow, this is great!

The code snippets in html generated by docfx, but I guess adding the .md and .html files is still a manual thing that we need to do when adding a new sample?

inforithmics:
I made a powershell script
https://github.com/Mapsui/Mapsui/blob/master/Scripts/BuildSamplesDocumentation.ps1
that adds the .md files from there the .md files generate the html files of the code. in mapsui.com/codesamples/animatedbussample.html for example

pauldendulk:
To get it prettier we might need to replace docfx with some other tool. A tool that takes a C# file as input and convert is into a html code snippet. Guess there are libraries for that, mostly javascript I fear.

pauldendulk:
oh, there is an html `<code>` tag. I guess everyone knew that already. So, simply pasting the file inside a `<code>` tag would already work. And then there are some css libraries that can add syntax highlighting. I scanning this blog: https://www.codiga.io/blog/display-code-snippets-in-html/

PR #2022, opened 2023-06-04
WIP adding libharfbuzzsharp.a to fix blazor

inforithmics:
trying to fix this https://github.com/Mapsui/Mapsui/issues/2014


inforithmics:
When the build is done I need to test the nuget package if it solve the problem

inforithmics:
I tried the resulting nuget package it seems not to work. Maybe I need to specify in Maspui.UI.Blazor every skiasharp Harfbuzz packages explicitly

inforithmics:
seems the Blazor.targets file was invalid. I'll try the resulting nuget again.

pauldendulk:
In the workaround I use in the Blazor sample the native file libHarfBuzzSharp.a is itself copied into the project folder. So, I think it has to be in the nuget. The workaround for Uno refers to the native file in the nuget, so no separate file needed.

inforithmics:
I found out that there could be a Mapsui.Blazor.props File (for properties and nuget packages gets evaluated earlier than the Mapsui.Blazor.targets File) i will try that if I suceed if not, I will explicitly include the  libHarfBuzzSharp.a

inforithmics:
It would work with Mapsui.Blazor.props but because it is a blazor component it has already an Mapsui.Blazor.props component so a cannot change that. I made a test with Mapsui.Blazor.NativeAssets with only Mapsui.Blazor.NativeAssets.props and it would work but than you would have to Reference Mapsui.Blazor and Mapsui.Blazor.NativeAssets. Not so good. So for now I have made another pull request which updates the blazor getting started. Hopfefully someday this bug will be fixed.

mattleibow:
I think this will fix it.

pauldendulk:
@mattleibow  I upgraded to SkiaSharp 2.88.4-preview.77 but did not fix the problem. Upgrading to 2.88.4-preview.82 that not fix it either. Here is the PR with the upgrade https://github.com/Mapsui/Mapsui/pull/2086/files

This line needs to be enabled:
```<NativeFileReference Include="libHarfBuzzSharp.a" />```

This is the message of the exception:
```Unexpected error in skia rendererException in BuildFontRuns() with original length of 53 now 53, style run count 1, font run count 0, direction overrides: False```

PR #2026, opened 2023-06-05
MapControl: add missing Swipe events

janusw:
* see discussion in issue #2017
* OnSwiped was not called at all
* also improve the velocity condition for Fling events (so that it depends on the absolute velocity only)
* FlingTracker.CalcVelocity: add basic documentation, use Vector2 as return type


janusw:
> Don't understand why the build is failing with the error below.

It fails because I haven't updated the Wpf parts (I'm working on MacOS using Mapsui.Mac.sln). Will fix.

pauldendulk:
Closing this because the alternative #2032 was merged.

PR #2030, opened 2023-06-05
added wfs points sample and posnode fix

inforithmics:


inforithmics:
Sample works now axis order where 0, 1 instead of 1,0 AxisOrderRegistry seems not to handle this one correctly. I discovered that there are two AxisOrder classes in Mapsui one in mapsui.nts and one in Mapsui will unify it to use the one in Mapsui

inforithmics:
Things done:
Fixed parsing of nodes in Wfs
Unified two AxisOrder Implementations
Updated to newest AxisOrder Implementation of Brutile

pauldendulk (approved):
Thanks. Indeed AxisOrderRegistry belongs in Projections.

PR #2032, opened 2023-06-06
MapControl: remove Swipe and OnSwiped

janusw:
This PR is an alternative to #2026. It removes the `Swipe` event and `OnSwipe` method of the `MapControl` class (they are unused). See discussion in #2017.

I slightly prefer this one over #2026 (since I have no use for the swipe event). In the end, @pauldendulk  should decide which one he likes better ;)


pauldendulk (approved):
Thanks for your analysis. I prefer to remove swipe because I don't see the use case yet and I prefer to support less, and support the rest better. If users can explain why they need swipe we should think about a solution that is cross platform.

janusw:
> Thanks for your analysis. I prefer to remove swipe because I don't see the use case yet and I prefer to support less, and support the rest better.

Makes sense and is fine with me. Thanks for merging!

PR #2035, opened 2023-06-08
Remove update of _downMousePosition in OnMouseMove in Blazor MapControl

pauldendulk:
This is because the field should indicate the position where the mouse went down and not the currentMouseDown position.


pauldendulk:
Could something else fail because of this change?

inforithmics (dismissed):
Good catch, i looked at it and couldn't find a reason why it was there.

inforithmics:
removing this causes drag of the map to stop working. It jumps randomly away from the mouse

inforithmics:
_downMousePosition was used as a _previousMousePosition. I made now two variables to differentiate it, so it works now.

pauldendulk:
Thanks for fixing it.

PR #2043, opened 2023-06-09
MapView: fix HandlerTap

### LINKED ISSUE #2041: MapView: Drawable with IsClickable=false blocks MapClicked event
**Mapsui Version**
4.0.0-rc.2

**Mapsui Platform**
Xamarin.Forms

**Device**
iOS 16 (probably also Android and others)

**Describe the bug**
I have a `MapView` with several `Drawables`. The `IsClickable` property is true for some and false for others. The clickable Drawables receive `Clicked` events as expected. The unclickable ones do no receive such events (also expected), but they seem to block the `MapClicked` event, which is actually not what I expected.

**Expected behavior**
If the Drawable at the clicked position is not clickable, i.e. does not handle the click event, I'd expect that the event is handed down to the map itself and the MapView emits that MapClicked event, as if there was no drawable at the position.

If there is another (clickable) Drawable at the same position, it should be able to handle the click event (which also does not seem to work).

janusw:
* for the case that we have a Drawable that is not clickable
* this is an attempt to fix #2041, but I still have to verify that it works

### DIFF
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -714,12 +714,18 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
                 }
             }
 
-            // Check, if we hit a drawable
-            // Is there a drawable at this position
+            // Check if we hit a drawable/pin/callout etc
             var mapInfo = GetMapInfo(e.ScreenPosition);
 
-            if (mapInfo?.Feature == null)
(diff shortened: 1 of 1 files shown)


pauldendulk:
@janusw You could add a 'WIP' in front of the title to indicate you still need to verify if this works.

janusw:
> @janusw You could add a 'WIP' in front of the title to indicate you still need to verify if this works.

Ah, good point. Thanks for the hint.

By now I have verified that this fixes #2041 for me (at least the part with the missing MapClicked event). There might still be an issue with overlapping Drawables, I'm not fully sure. In any case, this PR is not expected to help with this. A separate fix will be necessary, so this PR is ready for merging from my side.

pauldendulk (approved):

PR #2046, opened 2023-06-09
fixing mousewheel scrolling

inforithmics:


pauldendulk (approved):
👍

pauldendulk:
@inforithmics The mouse wheel scroll fix has no effect in my development environment. Do you know why this is? I ask because I need to suppress touch as well, which has no effect so far, but it could be because of my development environment too.

inforithmics:
I would suggest to install newest .NET 7 SDK and call dotnet workload update on the commandline. I discovered that Scrolling with mouse does not work in firefox (but the scroll page suppression does).

pauldendulk:
@inforithmics Thanks, got it work. Not sure what it was in the end. Also updated Visual Studio.

PR #2048, opened 2023-06-09
Nullable warning fixes mac (after Targeting .net 7 Maui/Android/iOs)

inforithmics:


pauldendulk (approved):

Mapsui.UI.iOS/MapControl.cs
```diff
@@ -265,9 +266,16 @@ public override void LayoutMarginsDidChange()
         SetViewportSize();
     }
 
-    public void OpenBrowser(string url)
+    public async void OpenBrowser(string url)
```
IldarKhayrutdinov, line 269:
```suggestion
    public async Task OpenBrowser(string url)
```

maybe `async Task`?

inforithmics: Open Browser implements an interface here, The Interface should be changed to OpenBrowserAsync

IldarKhayrutdinov: oh yeah, the name from interface. I just mean that better to avoid using `async void`

Issue #2000, opened 2023-05-26
MAUI Flyout Locks Map


nm4568:
**Mapsui Version**
Mapsui.Maui@4.0.0-rc.1

**Mapsui Platform**
MAUI

**Device**
iOS

**Describe the bug**
Swiping to open the flyout menu will eventually fully lock the map.  

**To Reproduce**

1. When swiping from the left to open/close the flyout menu, the map will eventually become locked.  The first step in becoming locked will manifest in the map thinking that it already has a Touch Point so triggering a TouchMove event with 1 finger will act as if 2 fingers were being used.  
2. Continue to open/close the Flyout and the map will eventually become locked. 

**Expected behavior**
The map should function as normal when the Flyout is closed.

**Additional context**
I suspect what is happening is that the TouchStarted event is being fired but the TouchEnded event is not.

pauldendulk:
Thank you for the clear bug report. We should investigate this before the release.

pauldendulk:
I can reproduce it. It takes quite a few retries. Once it is locked I can unlock it again by dragging the locked map and letting go. Is this also what you see?

pauldendulk:
I investigated what happend and as turns out there is no OnTouch event coming in at all in the scenario at all. So, it looks like it is a bug in MAUI or SkiaSharp. Since trying again solves the problem (or do you see something else @nm4568?) I don't see it as a big issue.

I don't thing there is a workaround possible on our side, since we do not get an event at all.

nm4568:
I have no problem reproducing it on my end.  In fact, here is a video that shows the problem.

https://github.com/Mapsui/Mapsui/assets/3853981/c67a7d9d-6680-4e9d-bbfc-85a443b37dc4

Steps in Video 
1) Shows the map functioning normally (1 finger)
2) Flout menu is opened
3) Flout menu is closed
4) Shows the map zooming and tilting as if I am using more than 1 finger but in reality, only 1 finger is used.

As I mentioned before, if the Flyout Menu is opened and closed again, the entire map locks up and doesn't respond to a drag at all.  I didn't feel this would provide any value showing this in the video.

Also, here is a link to the repo that I created the video from.

https://github.com/nm4568/MapsuiMauiFlyoutLocksMap

Other details: 

Device - iPhone Pro Max 14 (physical device)
.NET Workloads - 

Installed Workload Id      Manifest Version       Installation Source
---------------------------------------------------------------------
[...]

PR #2053, opened 2023-06-10
Fixing to many _touches after flyout

inforithmics:
fixed freezing after flyout there where to many touches, so it froze. Removing now to many touches. with the older touches first to remove

fixed this https://github.com/Mapsui/Mapsui/issues/2000


inforithmics:
https://github.com/Mapsui/Mapsui/issues/2000

pauldendulk (commented):
How does this work. What I expect is that after opening a flyout there is one touch remaining (because Released is missing). If you start panning again with one finger _touches.Count will be two and the mode will be zooming and no finger will be removed. What am I missing.

What would work is you start pinching after the flyout. Two mode will be zooming, but there will be three touches and the third one will be removed.

inforithmics:
The Problem was that there were for example 3 touchpoints and the control frooze (because it exited early because too many points where there). I couldn't really debug it, because i couldn't set breakpoints. But what i discovered with logging is that to many _touches where there for the state of the mouse moving.

charlenni:
What @inforithmics solution does is, that the crash after the second opening of the flyout isn't happening.

As said: Wouldn't it be better to use the Unfocused event (see [here](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.visualelement.unfocused?view=net-maui-6.0)) to clear _touches array, so that all fingers are gone?

pauldendulk:
> Wouldn't it be better to use the Unfocused event (see [here](https://learn.microsoft.com/en-us/dotnet/api/microsoft.maui.controls.visualelement.unfocused?view=net-maui-6.0)) to clear _touches array, so that all fingers are gone?

Yes, something like that. On Android the Cancelled event is missing. On iOS we need an alternative to do a _touches.Clear()

inforithmics:
I'll try my luck with the unfocused event.

inforithmics:
Unfocused isn't thrown, The only way i discovered is to detect if a flyout is shown and if yes clear the touches. Added A clearTouches public function so that users could call it too.

Mapsui.UI.Forms/MapControl.cs
```diff
@@ -652,6 +692,9 @@ private bool OnTouchMove(List<MPoint> touchPoints)
         {
             case TouchMode.Dragging:
                 {
+                    // remove old points avoids freeze after flyout
+                    LimitPoints(1, touchPoints);
```
pauldendulk, line 696: I don't see how this could have any effect. We set the Dragging mode based on the number of touchPoints

```csharp
if (touchPoints.Count >= 2)
{
    (_previousTouch, _previousRadius, _previousAngle) = GetPinchValues(touchPoints);
    _mode = TouchMode.Zooming;
    _virtualRotation = Map.Navigator.Viewport.Rotation;
}
else
{
    _mode = TouchMode.Dragging;
    _previousTouch = touchPoints.First();
}
```

For Zooming it could have effect when in case a Release was missing, but it will also change behavior when zooming with three fingers. Perhaps the LimitPoints method can be removed now there is a new workaround.

inforithmics: Yes the LimitPoints can be removed. And because there is now a ClearTouchState function, for external clearance when another issue with touchpoints occours. The Programmer can call this function.

pauldendulk (approved):
Fantastic you managed to find a workaround for this nasty issue.

PR #2055, opened 2023-06-11
Improve edit sample (With Buttons and Layer selection with Widget instead of real Buttons)

inforithmics:

PR #2057, opened 2023-06-11
Improved widgets

inforithmics:
This pull request is based on this pull request:
https://github.com/Mapsui/Mapsui/pull/2055

It introduces a IWidgetExtended interface with that I can implement an EditingWidget with handles all editing.
I introduce a MousePositionWidget too. That displays the geographic Position of the mouse.

With pull request I only need to call HandleMoving/HandleTouching/HandleTouched at the right Mouse/Touch Event positions and editing works for a platform, I implemented it in Balzor and Wpf. If the change is acceptable I will implement it for the other platforms.


inforithmics:
The editing Widget is a "virtual" widget it doesn't display anything but catches the IExtendedWidget Events and handles the Mouse interaction in this way.

the MouseCoordinateWidget is a "semi-virtual" widget because it displays the Mouse Position but catches the mousemoves over the whole map.

pauldendulk:
The IWidget already has an HandleWidgetTouched with one parameter less. It is a bit weird that this is separate. What about adding the three new methods to the IWidget interface. Perhaps with a default implementation that does nothing in Widget.

We should remove HandleWidgetTouched but not this close to the release because it would be a breaking change for users implementing a custom widget.

inforithmics:
Yes I agree this should go into the next Version of Mapsui, because of the breaking change. And the next Version of Mapsui is .Net 6 IWidget could implement default interface Implementations, so adding Members to an interface does not break the inheritors.

https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/interface-implementation/default-interface-methods-versions

pauldendulk:
> this should go into the next Version

Do you mean updating the IWidget interface, or this entire PR?

inforithmics:
I meant updating the IWidget Interface

I Implemted Edting for Avalonia, WinUI, Uno, Uno.WinUI, Maui, Forms

pauldendulk (approved):
Fabulous. 

Something on our to do list is more cleanup the MapControls/MapViews. Which was also something that came to mind when working on touch. We should move more to shared components. The platform dependent events should translate early to the platform independent logic.

inforithmics:
With the new Widget Interface it should be possible to create a ZoomBoxWidget so that all platforms get a Zoom Box almost for free.

pauldendulk:
> With the new Widget Interface it should be possible to create a ZoomBoxWidget so that all platforms get a Zoom Box almost for free.

That is great!

> I think with the current Changes we could introduce Drawables in Mapsui.Nts and therefore implement MapView on all platforms. Because the LocationLayer Is already everywhere available.

Instead of adding a MapView for all platform I would prefer to make the MapControl easier to use for the most common scenarios. We may need big changes here, not guided by where we happen to be right now.

charlenni:
@inforithmics But why do you need a Global? Could you not create a Widget which uses the whole MapControl. With this, you would get all events, because Widget are the topmost parts.

charlenni:
Regarding the idea behind the MapView: I made it, because I like to have a replacement for the existing MapView of Forms, which used Google or Apple map tools. It mimics this object. The advantage of it is, that it handles easier than MapControl. Users can use the well known coordinates for their Features.

Over the time more and more parts move from MapView to shared part of MapControl. There are no special things in MapView. It uses normal parts from MapControl.

So, to get ride of MapView, you have to move this comfort functions (use last/Lon, pin layer, polygon and polyline layer and so on) to MapControl.

charlenni:
@inforithmics The ZoomBoxWidget is old and should be removed. For this is the ButtonWidget (not sure about the name), which is much more flexible.

pauldendulk:
> Regarding the idea behind the MapView: I made it, because I like to have a replacement for the existing MapView of Forms, which used Google or Apple map tools. It mimics this object. The advantage of it is, that it handles easier than MapControl. Users can use the well known coordinates for their Features.
> 
> Over the time more and more parts move from MapView to shared part of MapControl. There are no special things in MapView. It uses normal parts from MapControl.
> 
> So, to get ride of MapView, you have to move this comfort functions (use last/Lon, pin layer, polygon and polyline layer and so on) to MapControl.

Yes, that is roughly what I had in mind.  The MapControl will not be entirely the same because it has to support more scenarios. One important feature is that is a a simple solution for automatic projections so that users can easily add lat/lon coordinates. 

I was thinking of adding MapBuilders for the most common scenarios. The MapBuilders should make it easy to add individual components, like the widgets, or a pin/line/polygon layers, all with a single short statement. There could also Builders that compose a MapControl with a single line. For instance:
`MapBuilder.CreateOsmLayerWithWidgets`.

inforithmics:
@charlenni You are right the Widget could virtually use the whole area and then get the events for the virtual area.
When a Widget wants to display some Information too, I could introduce some Child Widget (where the Information is displayed) and the Capture Widget that gets the whole area. For that no changes would be necessary only two Widgets would be added to the Widgets Collection. (One Displaying the Information and the other Capturing the Mouse Events and relaying it to the Display Widget).

PR #2058, opened 2023-06-12
Add color constants

pauldendulk:
We already had a few constants and all web colors as string. Asked ChatGPT to generate all the constants.


pauldendulk:
We started with our own colors in the Silverlight days but I assume that these days we could use System.Drawing.Color in all platforms. Perhaps not yet in .netstandard2.0 yet. In Mapsui 5.0 we should perhaps make the switch to System.Drawing.Color.

PR #2059, opened 2023-06-12
MapView: provide read-only access to callouts

janusw:
This PR provides read-only access to the callouts in a MapView.

Unlike for the pins and drawables, we don't want to modify the list of callouts here, but only provide the list of callouts that are currently visible. Normally the callouts are accessed and modified through the corresponding pin, but up to now it was not easily possible to obtain the full list of visible callouts.

Moreover there is some minor cleanup in MapView (second and third commit).

### DIFF
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -700,9 +700,6 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
 
     private void HandlerTap(object? sender, TappedEventArgs e)
     {
-        // Close all closable Callouts
-        var pins = _pins.ToList();
-
         e.Handled = false;
 
         if (Map != null)
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -238,6 +238,11 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
     }
 
     /// <summary>
+    /// Get callouts on map (those that are currently visible).
+    /// </summary>
+    public IReadOnlyList<Callout> GetCallouts() { return _callouts.ToList(); }
+
+    /// <summary>
     /// Single or multiple callouts possible
     /// </summary>
     public bool UniqueCallout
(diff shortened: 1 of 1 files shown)


janusw:
This is really a rather small change, so hopefully still acceptable for 4.0.0?

pauldendulk (changes requested):
Thanks! 

This is intended for readonly access, but the IList also has RemoveAt and Insert methods. Although a copy is made with the ToList so the source won't change it is still better to use a IReadOnlyList.

For me it is also important the understand why you need this. In general it is good to strive for small interfaces that allow all use cases, so I would like to know what you do with it so I can think of alternatives.

janusw:
> This is intended for readonly access, but the IList also has RemoveAt and Insert methods. Although a copy is made with the ToList so the source won't change it is still better to use a IReadOnlyList.

Good point. I rebased to current master and changed that. Thanks for the review!

> For me it is also important the understand why you need this. In general it is good to strive for small interfaces that allow all use cases, so I would like to know what you do with it so I can think of alternatives.

Well, basically I do some custom handling of tap-and-drag gestures and for that I need to know where all the map objects are (including callouts). I guess my use case is pretty specific here, nothing that can be generalized in a reasonable way.

pauldendulk:
> Well, basically I do some custom handling of tap-and-drag gestures and for that I need to know where all the map objects are (including callouts). I guess my use case is pretty specific here, nothing that can be generalized in a reasonable way.

Not sure how it works for the Callouts in MapView, but the features used in for instance the 'Callouts' | 'Single Callout' sample could also be retrieved with a MapControl.GetMapInfo if ILayer.IsMapInfoLayer = true.

janusw:
> > Well, basically I do some custom handling of tap-and-drag gestures and for that I need to know where all the map objects are (including callouts). I guess my use case is pretty specific here, nothing that can be generalized in a reasonable way.
> 
> Not sure how it works for the Callouts in MapView, but the features used in for instance the 'Callouts' | 'Single Callout' sample could also be retrieved with a MapControl.GetMapInfo if ILayer.IsMapInfoLayer = true.

Yes, I'm basically working with GetMapInfo to get the relevant features, but I'd at least need a way to check which ones are clickable etc (without actually issuing a click event, cf the proposals in #2041). As long as that is not possible, I resort to checking if those features are Callouts or other things.

PR #2060, opened 2023-06-12
MapView.HandlerInfo: fix detection of clicked drawables

janusw:
* we need to look through the whole list of records
* there might be several drawables at the clicked location (possibly semi-transparent)
* the first feature in the MapInfo might not be clickable and might block the clickable ones
* this fixes the remaining part of #2041


pauldendulk (approved):
Great you could solve it in this limited way.

janusw:
Thanks for merging 👍

PR #2065, opened 2023-06-14
Add touch to Blazor

pauldendulk:
So far:
- Suppress default touch 
- Add simple dragging.
- Tested in Chrome only.

Update:
- Added pinch zoom.
- Did not implement rotation.


pauldendulk:
My plan is to release after merging this one.

PR #2073, opened 2023-06-17
Wms Projection with DotSpatial (EPSG:6706 as sample)

inforithmics:


inforithmics:
I added A Sample that uses DotSpatialProjection for EPSG:6706 (Italy) Projection.
Things done:
1) Not known Projections can be registered.
2) Transparent can now be configured in the WmsProvider (default true)
3) Fixed Bug in DotSpatialProjection did not transform MRaster

pauldendulk:
> I added A Sample that uses DotSpatialProjection for EPSG:6706 (Italy) Projection. Things done:
> 
> 1. Not known Projections can be registered.
> 2. Transparent can now be configured in the WmsProvider (default true)
> 3. Fixed Bug in DotSpatialProjection did not transform MRaster

I do not understand what this is about. 

Mapsui does not support projection of MRaster, so I would think it should not try to. Raster transformation is hard although not impossible (it was done in SharpMap at some point). So, I assume this is not what this is about, although perhaps DotSpatialProjection has functionality for this.

Also I do not understand WMS projection with DotSpatial. Usually a WMS itself supports projection. So you pass in the required CRS and get a bitmap in the expected projection, so you would not need DotSpatial for this.

I am sure you are doing something useful, I just do not understand what it is. 🤔

inforithmics:
Was wrong wording, the Extent from MRaster is now transformed or else it wouldn't work in the default projectionprovider a coordinate visitor is used to handle extentents and geometries... but in the dotspatial provider only geometries were transformed. Now it transforms other tan geometry features by coordinate visitors too.
The register was added to add not known projections dotspatial knows around 4000 but not Epsg:6706 (Italy).
Concerning the projection the sample webservice only supports epsg:6706 so mixing this map with other maps and other projections wouldn't work.

inforithmics:
This service does not allow CORS Access so it doesn't work in blazor unless Cors check is disabled in the Browser (From the security Standpoint a Bad idea).
https://wms.cartografia.agenziaentrate.gov.it/inspire/wms/ows01.php?service=WMS&request=getcapabilities

What I thought is that I could display the static Test image instead of the Map when I know that a Sample does not work in Blazor.

inforithmics:
I added a sample that tiles a WmsProvider with Projection using the RasterizingTileLayer

pauldendulk:
> This service does not allow CORS Access so it doesn't work in blazor unless Cors check is disabled in the Browser (From the security Standpoint a Bad idea). https://wms.cartografia.agenziaentrate.gov.it/inspire/wms/ows01.php?service=WMS&request=getcapabilities
> 
> What I thought is that I could display the static Test image instead of the Map when I know that a Sample does not work in Blazor.

How about adding a CORS exception just for this service? Allowing everything is the kind of thing you do during the development phase to just get things working and then forget to turn off. I think allowing specific sources is just fine from security perspective.

pauldendulk (approved):

inforithmics:
> > This service does not allow CORS Access so it doesn't work in blazor unless Cors check is disabled in the Browser (From the security Standpoint a Bad idea). https://wms.cartografia.agenziaentrate.gov.it/inspire/wms/ows01.php?service=WMS&request=getcapabilities
> > What I thought is that I could display the static Test image instead of the Map when I know that a Sample does not work in Blazor.
> 
> How about adding a CORS exception just for this service? Allowing everything is the kind of thing you do during the development phase to just get things working and then forget to turn off. I think allowing specific sources is just fine from security perspective.

I tried to implement a workaround, But because Blazor runs in the Browser the Browser settings override the .Net Code. So the only workaround is to call a Proxy Service, That loads the Data Outside of the Browser.

Here an example. But because Blazor has no WebServer instance (That's why it can be deployed on a static WebSite, I cannot Implement it currently, I'm trying to cache the initial calls so that there is now webcall making the samples work in blazor, But I discovered some Sqlite Blazor issues along the way) 
https://darnton.co.nz/2020/07/13/cors-proxy-for-client-side-blazor/

Issue #2092, opened 2023-06-23
Mouse wheel does not work in Blazor on FireFox


pauldendulk:
**Mapsui Version**
4.0.0

**Mapsui Platform**
Blazor

**Device**
FireFox on Windows

**Describe the bug**
Mouse wheel zoom has no effect and in the debugger I see the OnMouseWheel event is not called.

pauldendulk:
Fixed with #2093

PR #2093, opened 2023-06-23
Fix mouse wheel zoom on FireFox

pauldendulk:


pauldendulk:
To fix https://github.com/Mapsui/Mapsui/issues/2092

PR #2095, opened 2023-06-29
fixed naming of sample

inforithmics:
One sample was not in an ...sample.cs file so the sourcecode was not referenced


janusw (approved):
Simple enough. LGTM.

PR #2099, opened 2023-07-07
Update Svg.Skia to 1.0.0

janusw:
... in order to finally have a 'stable' version for this dependency. The changes from 0.6.0-rc1.1 to 1.0.0 are minimal AFAICS, so this update should not cause any issues.


inforithmics (approved):
svg.skia is released at the same time as avalonia, so because avalonia is now realeased so is svg.skia.

PR #2103, opened 2023-07-08
WIP Adding Mapsui.UI.Blazor.targets for fixing HarfbuzzWasm issue

inforithmics:


inforithmics:
Workaround works now. 
But when .NET 8 is supported the .targets file as to be adapted to include the right harfbuzz library

docfx/documentation/getting-started-blazor.md
```diff
@@ -43,14 +43,19 @@ In Index.razor add this to the to Page.
...
-### Step 5
+
+
+### Step 6
+Run it and you should see a map of the world.
+
+
+### Troubleshooting
+
+### Text is not displayed
```
pauldendulk, line 54: The workaround is still here but from the way it is written I assume this is just in case it still is not working. Is that correct?

pauldendulk (approved):
Good idea to check Avalonia, they have many of the same problems that we have. There could be more suggestions to our problems in their code base. I also wonder how they deal with the GPU rendering problems that we have.

Issue #2105, opened 2023-07-08
Crash - Mapsui.UI.Maui.MapControl.InvalidateTimerCallback  System.NullReferenceException


polygonsheep:
**Mapsui Version**
Mapsui.Maui 4.0.0

**Mapsui Platform**
Maui.  Not tested on other platforms.

**Device**
iOS Simulator, iPhone, iPad.  Most prominent on Simulator

**Describe the bug**
System.NullReferenceException: Object reference not set to an instance of an object.
  at at Mapsui.UI.Maui.MapControl.InvalidateTimerCallback(Object state)
  at at System.Threading.TimerQueueTimer.<>c.<.cctor>b__27_0(Object state)
  at at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
  at at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
  at at System.Threading.TimerQueueTimer.CallCallback(Boolean isThreadPool)
  at at System.Threading.TimerQueueTimer.Fire(Boolean isThreadPool)
  at at System.Threading.TimerQueue.FireNextTimers()
  at at System.Threading.TimerQueue.System.Threading.IThreadPoolWorkItem.Execute()
  at at System.Threading.ThreadPoolWorkQueue.DispatchItemWithAutoreleasePool(Object workItem, Thread currentThread)
  at at System.Threading.ThreadPoolWorkQueue.Dispatch()
  at at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
  at at System.Threading.Thread.StartCallback()

**To Reproduce**
I have an app with multiple Content Pages. One of those pages loads Mapsui and displays the map. Switching between a non-map Content page and the content page with the map it will eventually crash when loading/displaying the map page. Usually crashes within 5 - 10 page switches. Page switching about 1 or 2 seconds apart will make it crash faster, but it's crashed when not switching fast as well.

**Expected behavior**
I would expect it to handle null exceptions and not crash

**Screenshots**
N/A

**Additional context**
Doesn't happen every time, but if you continue to switch between a map page and non-map page it will crash usually within 5 - 10 switches.  Tried with both Singleton Pages and Transient Pages for the map page, same results. Here's my Map Page code.

XML
        <mapsui:MapView Grid.Row="1"
                        x:Name="MapsuiMap"/>

[...]

PR #2108, opened 2023-07-08
Fix null reference exception in timer

inforithmics:
Fixes this:
https://github.com/Mapsui/Mapsui/issues/2105


pauldendulk (commented):
There are null checks and a try-catch. Do we need both?

inforithmics:
I was extra cautious.
Null checks for fixing bugs. And try catch to catch bugs in the called functions. I guess this runs on the unthread and not caching bugs will crash the application or the thread the timer runs on.

pauldendulk (approved):

Issue #2107, opened 2023-07-08
Mapsui.Samples.Maui has problem connecting to WMS server on Android 13 platform.


FrankCAD:
**Mapsui Version**
4.0.0

**Mapsui Platform**
MAUI

**Device**
Android 13
Emulator 'Tablet M-DPI 10.1in - API 33'

**Describe the bug**
Problem connecting to WMS server.

1. Run Mapsui.Samples.Maui
2. Go to 'Data Formats'
3. Click on example '6 WMS'
4. Map does not display (should show map 'Windkaart' of The Netherlands)

Traced the problem back to line 350 in the file Mapsui/Providers/Wms/WmsProvider.cs (Task FetchOnThreadAsync) generating WebException.
See [System.Net.WebException.txt](https://github.com/Mapsui/Mapsui/files/11990734/System.Net.WebException.txt)

Similar problems with the examples: 
3 WMTS 
5 WMTS Michelin 
6 WMS Projection 
6 WMS Projection DotSpatial 
8 TMS openbasiskaart

_Running the same examples on 'Windows Machine' do not show these problems._

inforithmics:
Thanks for the Bug Report will be fixed in this pull request
https://github.com/Mapsui/Mapsui/issues/2107

pauldendulk:
Should be fixed in 4.1.0-preview.1.

PR #2109, opened 2023-07-08
make HTTPMessage Method upper case

inforithmics:
Fixees this

https://github.com/Mapsui/Mapsui/issues/2107

### DIFF
--- a/Mapsui/Providers/Wms/WmsProvider.cs
+++ b/Mapsui/Providers/Wms/WmsProvider.cs
@@ -508,11 +508,11 @@ public class WmsProvider : IProvider, IProjectingProvider
             throw new InvalidOperationException("Wms Client needs to be set");
         //We prefer get. Seek for supported 'get' method
         for (var i = 0; i < _wmsClient.GetMapRequests.Length; i++)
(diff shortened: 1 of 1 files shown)


inforithmics:
I discovered that Network requests need to be done with await `httpClient.......ConfigureAwait(false)` or else in Android in Version 11 and leter an exception is thrown because a webrequest is finished on the mainthread.

pauldendulk (commented):
I prefer to have as little clutter as possible so would only want to use the ConfigureAwait(false) when really necessary. My understanding was that is only applies in .NET Framework https://stackoverflow.com/a/42058544/85325, which might be relevant because we still support .netstandard, but even there it should only be use if there is an actual problem to solve.

inforithmics:
Platforms with an UIThread iOS/Android/Wpf ... listen to this. On Android the request is done on the UIThread and because of ConfigureAwait(false) finishes on a Background Thread. If ConfigureAwait(false) isn't set the request finishes on the UIThread (or the Thread it started) which throws an exception on Android on other platforms it slows down the UIThread Because it has mir work to do. fetching data ca be done on the Backround.

Because Asp.Net core has no UIThread/Mainthread it doesn't listen to that.

LeoCarriere (approved):

PR #2113, opened 2023-07-12
fixing build (Because of .Net Framework updates)

inforithmics:


inforithmics:
Things done:

Removed Microsoft.Graphics.Win2D from Directeroy.Packages.props because it is not needed in Mapsui.WinUI

PR #2124, opened 2023-07-28
Demo to show a possible way to add weather data as image to a map

charlenni:
Demo for uses of RasterFeature


Samples/Mapsui.Samples.Common/Maps/Demo/ImageDataSample.cs
```diff
@@ -0,0 +1,216 @@
...
+
+    private static SKPoint ConvertPolarToCartesian(double radius, double angle)
+    {
+        var x = radius * Math.Cos(angle * Math.PI / 180.0);
+        var y = radius * Math.Sin(angle * Math.PI / 180.0);
+        
+        return new SKPoint((float)x, (float)y);
+    }
+
+    private static Random _random = new Random();
```
inforithmics, line 183: for reproducability (Regression Unit Tests) it would be good to use a fixed seed

inforithmics (approved):

Samples/Mapsui.Samples.Common/Maps/Demo/ImageDataSample.cs
charlenni: I thought, that this is only a demo. But I could change this. Or I delete the PR at the end, when the TO got, what to do.

inforithmics: No need for deleting the PR. When no regression image is created it will not be tested.

inforithmics: I added the c# source code to the samples blazor project. By executing
Scripts\BuildSamplesDocumentation.ps1

inforithmics:
Merging it now, Nice sample.

Issue #2090, opened 2023-06-23
The map can only be rotated clockwise


gpsaliola:
Mapsui Version: 4.0.0
Mapsui Platform: Android/iOS

The map can only be rotated clockwise.
If I add a raster overlay this can also be rotated counterclockwise losing alignment with the map (even if RotateWithMap = true).

```
                    <mapsui:MapView x:Name="mapView"
                        Grid.Row="0" Grid.Column="0"
                        IsMyLocationButtonVisible="True"
                        IsNorthingButtonVisible="True"
                        IsZoomButtonVisible="True" 
                        RotationLock="False" />

```
This is the code used to add raster image

```
                MemoryStream memoryStream = new MemoryStream();
                memoryStream = ImageMemoryStreamFromResource("MyRaster");
                (double x, double y) = SphericalMercator.FromLonLat(42, 12);
                PointFeature feature = new PointFeature(new MPoint(x, y))
                {
                    Styles = {
                        new SymbolStyle
                        {
                            BitmapId = BitmapRegistry.Instance.Register(memoryStream),
                            SymbolScale = 1,
                            RotateWithMap = true
                        }
                    }
                };

                MemoryLayer compassLayer = new MemoryLayer()
                {
                    Name = "Compass", 
                    Features = new[] { feature },
                    Style = null 
                };
                feature.Dispose();
                mapView.Map.Layers.Add(compassLayer);

```

PR #2127, opened 2023-07-30
Touch Handling in Avalonia (Zoom and Rotation)

inforithmics:


inforithmics:
The behavior of the Clockwise rotation I discovered it too.
https://github.com/Mapsui/Mapsui/issues/2090

slater1:
Anything I can do to help this along? Really anticipating touch working on Android / iOS!

inforithmics:
You could download the artifact and Test the nuget package. And if you find issues report them here or give a thumbs up.

slater1:
I ran my Android app referencing the [latest artifact](https://github.com/Mapsui/Mapsui/actions/runs/5948164467). There's a couple of issues, see the video and comments below.

I'll try on iOS later today when wife gets home. She has the iPhone.

https://youtube.com/shorts/iVdfGTUhW7M?feature=share

Video contents in sequence:

- single-finger pan. Looks good.
- two-finger rotate, CCW and CW. Looks good.
- two finger zoom, our and in. Looks good.
- single-finger zoom and rotate (!)
- two-finger attempt to zoom and rotate (failed)

Notes:

-  First pinch and zoom works fine
- After releasing the first pinch and zoom, a single touch also pinches and zooms.
- After that single touch is released, cannot pinch and zoom with either single-finger or two-finger touch.
- Issue does not resolve until app is restarted (presumably a new MapContol would be enough)

inforithmics:
Thanks for the tests. I think somehow a touch wasn't released. That's why there was a single touch zoom. I'll investigate it.

inforithmics:
Could you provide a not working sample project. Because I tried it with Mapsui Avalonia Android Sample Project and there it works without problems.

inforithmics:
@slater1 I found an Event where I need to Clear the TouchState, maybe it works now better for you?

slater1:
Downloaded the [artifact ](https://github.com/Mapsui/Mapsui/actions/runs/5980554279) that built just a few minutes ago, and it seems to be working on Android!

I lied, didn't get to try iOS today. Will aim to get back on that this weekend.

Thanks for the work!

slater1:
I can confirm that map pinch zoom and rotate now works on iOS, at least on iPhone 14 iOS 16.4.

pauldendulk:
I need more time for this one. Touch is implemented in different ways on the various platforms. This PR shares more of the code which is a good thing but still some platforms, like Blazor, have a very different implementation. I would like to have a bigger rewrite of the touch handling but it will take some thought. A few thing I worry about in the code reused in this PR is:
- I am not sure if we should handle things like ShortTap, ShortClick, DelayTap, LongTap. I would prefer it if the platform handles it. 
- The _touches array is a stateful thing and if there is an error in keeping track of the state this gets problematic. We had a problem with this recently where some even in Maui was not triggered and because of this some touch point was not removed. Perhaps there is no way around it though. But if the platform solves this for us it would make things easier.
- Some of the shared code should perhaps we in components which could be used in the various MapControls, but would not have to be in the Shared MapControl.

The remarks about are not suggestions for improvement because I don't know myself what is the best way to improve it.

inforithmics:
@pauldendulk I think I will make another pull request with changes only in Avalonia. So this would be less risk. The touchhandling Unifcation can be done in a seperate pull request.

inforithmics:
New Pull Request (I copied the code from this pull request and added the missing functions (that where in MapControl Shared)) So this pull request is the same as this but without changes in other controls or in the Shared MapControl.
https://github.com/Mapsui/Mapsui/pull/2159

PR #2128, opened 2023-07-30
fixed counterclockwise rotation

inforithmics:
Fixes this
https://github.com/Mapsui/Mapsui/issues/2090


inforithmics:
If Rotation is initially negative it doesn't rotate.

pauldendulk (approved):
👍

PR #2129, opened 2023-07-31
Move IMapViewSample to Mapsui.Samples.Common

inforithmics:
This is an explatory pull request. Don't know yet if it leads to something useful


pauldendulk (commented):
I see a lot of code about SingleTab and I would was hoping we could leave this to the user code. Every platform has its way to distinguish between te SingleTab and DoubleTab. There are choices to make, do something different on DoubleTab, how short should the interval be for two Tabs to be a DoubleTab. This also means you need to way after the first tab to see if it is followed by another tab. I would say, let the user do that part, and design Mapsui in such a way that they could do everything that they want.

About:
```
 if (sample is IMapViewSample mapViewSample)
                    _clicker = mapViewSample.OnClick;
```
What about using OnInfo for that instead. How would that be a problem?

charlenni:
@pauldendulk Do you talk about tap (clicking on the control) or a real tab (different pages)

pauldendulk:
> @pauldendulk Do you talk about tap (clicking on the control) or a real tab (different pages)

oopf, I meant Tap. Mistyped that quite consistently. Updated now. Thanks.

charlenni:
No problem.

The question is, if all MapControls should have the same behaviors or if some of them could be more elaborated and others not. If someone like to have a special thing for a MapControl, then it could be made later.

pauldendulk:
> No problem.
> 
> The question is, if all MapControls should have the same behaviors or if some of them could be more elaborated and others not. If someone like to have a special thing for a MapControl, then it could be made later.

I am inclined to go for 'all platforms the same functionality'. And this will be easier if all our implementation to try to use things that are not platform dependent. For instance, skia rendering, like we do in our widgets. And we could work towards a common way of handling touch/mouse so that we can build things on top of that which would work for all platforms. 

But I might change my mind if we go into more detail. And I am more ready to accept platform specific things if they are in extension methods. For instance in converter to and from that platform.

inforithmics:
So I had a look at it. 
I tried to move the IMapViewSample Interface so that all platform samples could use it. (Long term plan to integrate MapView features into all Platforms).

Because IMapViewSample needs the SingleTap Event,
Because only Forms and Maui have a SingleTap Event so I implemented it on all other mapsui Platforms. Because the Handling is done right before the OnInfo Call/Event OnInfo could be used too.

inforithmics:
Should I work on this pull request further or close it in favor of a pull request where I move the Objects from MapView to Mapsui? This would involve breaking changes because moving the Objects from MapView to Mapsui (Core) would mean changing the DataTypes from Maui/Forms Types to Mapsui ones.

pauldendulk:
I think it is better to close this and do a bigger overhaul for v5. Perhaps we could release a 4v that is compatible with the release of .NET 8. 

Also, I would not want to simply merge the current MapView into the MapControl but look at the functionality in MapView and migrate each part step by step.

inforithmics:
I agree closing this

PR #2133, opened 2023-08-01
Simplify editing widget

inforithmics:
Things done. 
1) There are no virtual Widgets anymore. EditingWidget and MouseCoordinate Widget set the Envelope to get the Events for the whole Control.
2) The IWidget.HandleWidgetTouched is now handled in HandleTouched.
3) The Order of the Widgets additions define the ZOrder when they are called. The latest added Widgets are at the top. And Hide first added widgets in the EventHandling


Mapsui.Rendering.Skia/SkiaWidgets/EditingWidgetRenderer.cs
```diff
@@ -6,5 +6,7 @@ public class EditingWidgetRenderer : ISkiaWidgetRenderer
 {
     public void Draw(SKCanvas canvas, Viewport viewport, IWidget widget, float layerOpacity)
     {
+        // use all of the canvas
+        widget.Envelope = new MRect(0, 0, viewport.Width, viewport.Height);
```
pauldendulk, line 10: Does this full envelope change behavior in some of our current component?

pauldendulk (commented):
What triggered you to make these changes? It is a simplification, that is great, but was there also behavior that lead you to these changes. There don't have to be, but if there are it should be clear what changes.

Mapsui.Rendering.Skia/SkiaWidgets/EditingWidgetRenderer.cs
inforithmics: No, I removed the concept of virtual Widgets. By setting the Envelope to the whole area it catches for the whole area events. This was suggested by @charlenni. And I think it simplyfies the behavior because there is now no special casing for virtual Widgets.

Issue #2130, opened 2023-07-31
WMS Provider catch with "The SSL connection could not be established"


gpsaliola:
Mapsui Version
4.0.0

Mapsui Platform
Xamarin - Android

Device
Samsung A53

Describe the bug
WMSProvider catch with "The SSL connection could not be established" error with this url: 
https://depth.openseamap.org/geoserver/gebco2021/wms

To Reproduce

`            WmsProvider provider = await WmsProvider.CreateAsync("https://depth.openseamap.org/geoserver/gebco2021/wms", persistentCache: MyDefaultCache);
`

Expected behavior
WmsProvider.CreateAsync should not catch in error.
This server works fine with OpenLayers.

inforithmics:
It is fixed in this pull request it was a missing useragent
https://github.com/Mapsui/Mapsui/pull/2134

pauldendulk:
I assume this is fixed with the PR. If the problem persists after the next release please reopen.

Issue #2119, opened 2023-07-23
WMSProvider catch with "Could not download capabilities - Unexpected response code: Forbidden" error


gpsaliola:
Mapsui Version
4.0.0

Mapsui Platform
Xamarin - Android

Device
Samsung A53

Describe the bug
WMSProvider catch with "Could not download capabilities" (Inner exception: "Unexpected response code: Forbidden") error with this url: http://rsdi.regione.basilicata.it:80/rbgeoserver2016/maps_ctr/LC.LandCoverRaster/ows

To Reproduce
`            
WmsProvider provider = await WmsProvider.CreateAsync("http://rsdi.regione.basilicata.it:80/rbgeoserver2016/maps_ctr/LC.LandCoverRaster/ows", persistentCache: MyDefaultCache);
`

Expected behavior
WmsProvider.CreateAsync should not catch in error.
This server works fine with OpenLayers.

inforithmics:
Can you give me a codesample how this is called in openlayers?

gpsaliola:
A working OpenLayers example:

```
<!DOCTYPE html>
<html>
  <head>
    <title>Maps server</title>
	<meta content="IE=edge"> 
    <style TYPE="text/css">
    </style> 
	
	<link rel="stylesheet" href="https://openlayers.org/en/v6.5.0/css/ol.css" type="text/css">
	<script src="https://openlayers.org/en/v6.5.0/build/ol.js"></script>
	<script src="https://cdnjs.cloudflare.com/ajax/libs/proj4js/2.7.4/proj4.js"></script>

	<script type="text/javascript">
		var lat = 40.63506;
		var lon = 15.804827;
		var zoom = 12;
		
		var view;
		var map;
		
		function init() {
			view = new ol.View({
				center: ol.proj.fromLonLat([lon, lat]),
				zoom: zoom
			});

			map = new ol.Map({
				target: 'map',
				view: view,
				controls: [new ol.control.Attribution({collapsible: false, collapsed: false})],
				interactions: ol.interaction.defaults({
					doubleClickZoom: false,
					dragAndDrop: false,
					dragPan: true,
					keyboardPan: false,
					keyboardZoom: false,
					mouseWheelZoom: false,
					pointer: false,
					select: false
				})
			});

[...]

PR #2134, opened 2023-08-01
Additional wms fixes

inforithmics:
Fixes this
https://github.com/Mapsui/Mapsui/issues/2119
https://github.com/Mapsui/Mapsui/issues/2130


inforithmics:
Here in this Pull Request, I added an additional Optional Parameter, userAgent. 

When you upgrade it still compiles but it is not binary compatible because you have to recompile it.
Should I add an overload with the old parameters. so that it stays Binary compatible ?

pauldendulk:
> When you upgrade it still compiles but it is not binary compatible because you have to recompile it.
> Should I add an overload with the old parameters. so that it stays Binary compatible ?

Good enough when it is not breaking with a recompile.

pauldendulk (approved):

PR #2136, opened 2023-08-03
Fixed Merge Conflict (IndexedMemoryProvider was not used in RasterizingTileLayerWithThousandsOfPolygonsSample)

inforithmics:


inforithmics:
Things done: 
1) added missing RenderCache Parameter to setting the Size of the RenderCache.
2) in doing so i discovered that RenderCache was not Threadsafe fixed that too.

Mapsui/Features/BaseFeature.cs
```diff
@@ -28,7 +29,7 @@ public BaseFeature(BaseFeature baseFeature)
         set => _dictionary[key] = value;
     }
 
-    public IDictionary<IStyle, object> RenderedGeometry { get; set; } = new Dictionary<IStyle, object>();
+    public IDictionary<IStyle, object> RenderedGeometry { get; set; } = new ConcurrentDictionary<IStyle, object>();
```
pauldendulk, line 32: It looks like this is not used anymore. If I go through the usage I only see Clear and Dispose. I guess you only turned this into a ConcurrentDictionary because it made sense since you changed the other one?

pauldendulk (commented):
I would think the cache of the renderer should only be accessed on the render loop. How did it crash? Perhaps calling cache clean would be called on another thread, but perhaps there is another solution for that. Like signaling a cache clean is needed an let the render loop do the actually cleaning.

Also, I still had on my todo list to have try to move the caching logic out of the individual renderers. I would like these to be stateless builders, and if something is already cached that the renderer should not be called at all. Now part of the logic has to be repeated in all renderer.

inforithmics:
Closing because changes are included in this pull request

https://github.com/Mapsui/Mapsui/pull/2138

Mapsui/Features/BaseFeature.cs
inforithmics: I changed it to a concurrent dictionary because it can be potentially accessed by multiple threads. Maybe it would be better to lazy initialize this

inforithmics: This is now done in the other pull request. I think @charlenni needs this for is Vector Style Renderer but I'm not sure about that.

PR #2138, opened 2023-08-05
Add caching of symbols

inforithmics:


inforithmics:
This pull request is based on this.

Here I add caching of Circle, Rectangle and Triangle Symbols.
https://github.com/Mapsui/Mapsui/pull/2136

inforithmics:
Things done
1) Removed IModifyFeatureLayer because the Pathes are cached in the RenderedGeometry as suggested by @charlenni 
2) Symbol, Triangle, Rect and Circle are now cached.

pauldendulk (commented):
oh, here the use of the RenderedGeometry is reintroduced again, that is unfortunate. I introduced the RenderedGeometry at some point and regretted that later. Because of this caching all the features need to be disposable, and therefore all providers and layers, and everything. An external cache is a better solution. Cache is problematic and it is good to have centralized control over it. The injected VectorCache could be replaced with something that is not caching at all. That is a nice feature to have when you are debugging things that could be caused by a cache (caching causes a lot of bugs).

pauldendulk:
This thread is relevant to this PR https://github.com/Mapsui/Mapsui/pull/1744#issuecomment-1387111993.

charlenni:
As we discussed in the other thread: There must be a render and a drawing process. Converting Features in SKObjects takes much time. SKObjects then have a size and you could check, if they are visible or not. Drawing them is fast (or should be fast).

pauldendulk:
> As we discussed in the other thread: There must be a render and a drawing process. Converting Features in SKObjects takes much time. SKObjects then have a size and you could check, if they are visible or not. Drawing them is fast (or should be fast).

True, and this will also have an impact on the caching solution.

inforithmics:
Good point I will remove the RenderedGeometry, so that the main change would be to not create new geometries during rendering and thread safe caches

pauldendulk:
Because of the history of the project we came to a solution where we have a special treatment for layers with mutating features (this started with the WriteableLayer that I added, a better solution for not have introduced a special layer for mutations). Now we have to check in several for the IModifyFeatureLayer and a layer has to choose to be either IModifyFeatureLayer  or not. In an alternative solution it could perhaps be possible to mutate any feature in any layer. I have been thinking about that. 

### IFeature.Modified field
An earlier suggestion that came up was to add a modified field to IFeature. The caching mechanism could check to see if it could use the cache. A downside it that the modified field would also have to set that field to false again. Also different renderers with different caches would have a problem with this (this is not likely to happen but the fact that this could fail indicates it is not a great mechanism). Also, the render loop should not have to meddle with the state of it's input. It should be data in and image out.

### Value Equals
In other places we mentioned implementing value Equals. I think this is not a good idea because that is a a big burden to implement for every feature. Also it is a performance hit.

### Reference Equals
This is much more pragmatic solution and sufficient for many cases but it does not fix the problem with modified features. A fix would be to not let features be modified but only be replaced, but users often just want to change a feature, especially if they have state in the feature they do not want to copy to the new version.

### IFeature.Id field
My current idea is to generate an Id for every feature that could be used for caching. It could be implemented in BaseFeature, so that all features have it (it would become a requirement to derive from BaseFeature). The constructor of BaseFeature could use a static IdCounter, increment it and assign to self, so it would be unique over all features. In case a feature is modified an IFeature.IsModified() method is called which generates a new Feature.Id so that the caching mechanism would not use the old version. 

This is what I came up with today. Still does not seem a bad idea after typing this here.

Mapsui.Rendering.Skia/Extensions/LineStringExtensions.cs
```diff
@@ -12,8 +12,9 @@ internal static class LineStringExtensions
     /// <param name="lineString">List of points in Mapsui world coordinates.</param>
     /// <param name="viewport">The Viewport that is used for the conversions.</param>
     /// <param name="clipRect">Rectangle to clip to. All lines outside aren't drawn.</param>
+    /// <param name="strokeWidth">stroke Width</param>
     /// <returns></returns>
-    public static SKPath ToSkiaPath(this LineString lineString, Viewport viewport, SKRect clipRect)
+    public static SKPath ToSkiaPath(this LineString lineString, Viewport viewport, SKRect clipRect, float strokeWidth)
```
pauldendulk, line 17: So, the method does not take into account the style but it does need to know the strokeWidth (which belongs to the style) for clipping. That is unfortunate, but I don't see a better solution.

Mapsui.Rendering.Skia/SkiaStyles/LineStringRenderer.cs
```diff
@@ -1,36 +1,41 @@
...
         {
             paint = CreateSkPaint(vectorStyle.Line, opacity);
-            path = lineString.ToSkiaPath(viewport, canvas.LocalClipBounds);
+            path = lineString.ToSkiaPath(viewport, canvas.LocalClipBounds, lineWidth);
         }
         else
         {
-            paint = vectorCache.GetOrCreatePaint(vectorStyle.Line, opacity, CreateSkPaint);
+            Func<LineString, LineString>? copy = null;
+            if (layer is IModifyFeatureLayer)
```
pauldendulk, line 32: In the comment to the PR you mention " but the Geometry is cloned before it is written into the cache.". Could you explain a bit more about how it works and why it was done?

Mapsui.Rendering.Skia/SkiaStyles/PolygonRenderer.cs
```diff
@@ -16,29 +19,62 @@ internal static class PolygonRenderer
     /// </summary>
     private const float Scale = 10.0f;
 
-    [SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP001:Dispose created")]
     public static void Draw(SKCanvas canvas, Viewport viewport, ILayer layer, VectorStyle vectorStyle, IFeature feature,
-        Polygon polygon, float opacity, ISymbolCache? symbolCache = null, IVectorCache? vectorCache = null)
+        GeometryCollection collection, float opacity, ISymbolCache symbolCache, IVectorCache vectorCache)
```
pauldendulk, line 23: The Polygon renderer is expected to render a Polygon but it rendering a GeometryCollection, which could hold Polygons but also other things. Perhaps the PolygonRender could be preserved in it's original form and a new GeometryCollectionRenderer is introduced which makes use of the PolygonRenderer. 

And why is the GeometryCollection now renderer? is it because of the recursion problem?

inforithmics: In the VectoryStyle Renderer this was

```
 for (var i = 0; i < collection.NumGeometries; i++)
                                Draw(canvas, viewport, layer, new GeometryFeature(collection.GetGeometryN(i)), style, renderCache, iteration);
```

with this
                            `PolygonRenderer.Draw(canvas, viewport, layer, vectorStyle, feature, collection, opacity, renderCache, renderCache);`

And advantage is that during rendering no new GeometryFeatures get created.

Mapsui.Rendering.Skia/SkiaStyles/LineStringRenderer.cs
inforithmics: The basic idea behind this is following.

Before Layers that modify Geometries the Geometries where not cached, reducing the Performance. Now when a Layer is marked as modifying Geometries, The Geometry is cloned before the Geometry is used as a cache key. So this avoids that the cache gets corrupt because the Geometry changes. Because when the new geometry gets changed it is not the same instance as in the cache.

For caching normally unmutable objects are used, A change creates a new Geometry for example. But because the Geometry is mutable I have to use this approach.

Mapsui.Rendering.Skia/SkiaStyles/LineStringRenderer.cs
```diff
@@ -1,36 +1,41 @@
...
+using ViewportExtensions = Mapsui.Rendering.Skia.Extensions.ViewportExtensions;
 
 namespace Mapsui.Rendering.Skia;
 
 public static class LineStringRenderer
 {
     [SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP001:Dispose created")]
     public static void Draw(SKCanvas canvas, Viewport viewport, ILayer layer, VectorStyle? vectorStyle,
-        LineString lineString, float opacity, IVectorCache? vectorCache = null)
+        IFeature feature, LineString lineString, float opacity, IVectorCache? vectorCache)
```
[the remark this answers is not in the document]
inforithmics: removed now

Mapsui.Rendering.Skia/Extensions/GeometryCollectionExtensions.cs
```diff
@@ -0,0 +1,40 @@
+﻿using NetTopologySuite.Geometries;
```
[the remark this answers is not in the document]
inforithmics: Is used for GeometryCollection Rendering see comment there

inforithmics:
> ### IFeature.Modified field
Intresting idea, the problem is that when a Feature gets modified it is already in the cache and corrupts the cache. Because the caching is following it calcuates a HashCode and with that Hashcode it makes a lookup. 
and after it compares by reference and then be value.
> ### Value Equals
The first comparision is by Reference. or else the implementation is of the Value Equals is not so good.
> ### Reference Equals
Does not detect structurally equal Geometries. This is important because in a lot of Providers new Geometries are created even if there already existing Geometries.
> ### IFeature.Id field
Because when reading data from a provider, always new Features are created except in MemoryProvider or GeoJson so the numbers would increase after every refersh and the geometries wouldn't be cached. So an Cache would need to be introduced to Give duplicate Geometries the same number.

### Inmutable Geometries.
When a Geometry is changed a new Geometry would need to be created, so that existing Geometrys are not changed. Then the Geomtries wouldn't need to be cloned before putting them into the cache. I looked if the Geometry support a kind of change detection but unfortunatly no.

Mapsui.Rendering.Skia/SkiaStyles/PolygonRenderer.cs
pauldendulk: ah, the iterator could not solve this because it is iterating over features and we need to iterate over geometries for this. So this is an improvement.

pauldendulk:
Thanks for your response. It makes clear I don't have all things clear yet. So here are some more questions and remarks.

I mentioned some mechanism above, but only the IFeature.ID one is an actual proposal, the other mechanisms were more like thinking out loud. This was perhaps not clear.

### Combine *value* and *equals* compare?
Do I understand correctly that the current mechanism first does a reference equals and then a value equals? The reason for doing it this way is performance optimization because you do not want to do the value equals every time? 

### Avoid value compare
I have a problem with value compare because the burden it puts on maintaining the code. New geometries also need to implement it. Changes need to take the mechanism into account. The discussion about an effective Hashcode seems endless. If it is implemented for just NTS geometries it might be a more limited scope, and not such a big burden, but in that case I would expect this to be implemented in NTS.

### The advantage of value compare
Value compare can also detect an equals between two different instances of a geometry. But I wonder how useful that is for our application.

### How much identical geometries do we have?
I would think very few. In most datasets every geometry is unique. I try to think of counter examples, but points or equal squares in a grid include the actual coordinates so are not identical. 

Now you came up with a counter argument which is that we fetch the geometries again and again from the dataset. That made me rethink my ideas. A few thoughts:
- For what are we optimizing? My main concern is fast performance during rendering. When holding the mouse down and panning the map should be super fast. So, for performance in this scenario it does not really matter that we do another fetch.
- **Cache flooding**. You mention cache corruption, I would not use that term here. Corruption is when there is an actual bug in the logic (like a bug in a value compare). This is a problem we have to think about, but there might be solutions. I think this kind of flooding could be countered an effective cleanup mechanism. One that makes use of the specific logic that we use for our map.
- Your remark also makes me think about the data sources and fetching again and again. Perhaps we should look into that too. It is weird that we have a lot caching code to prevent another conversion of NTS to Skia, while at the same time we fetch an entire new feature, including file or web access, which is much slower than the conversion to Skia.

inforithmics:
> ### IFeature.Modified field
I thought again about it and it behaves performance wise the same as the Removed RenderedGeometry. So the Performance improves the less new Geometry objects are created. (For example like in SimplifyGeometry Provider).

pauldendulk:
> I thought again about it and it behaves performance wise the same as the Removed RenderedGeometry. So the Performance improves the less new Geometry objects are created. (For example like in SimplifyGeometry Provider).

I think you imply that value compare will work even if you fetch the feature again. I would think the advantage of this is negligible. There are many more render cycles than there are fetches. During regular rendering the caching would work properly. Only if the a fetch is done a value compare approach would be an advantage, but that is a very small advantage. The fetcher will fetch from file or web which would be much more expensive than the single conversion from NTS to Skia.

About caching corruption. If you use the IFeature.Id then that Id should be used for storing and retrieving too. That way I don't see a problem.

inforithmics:
I agree. 
1. Then I try to remove the IModifyFeatureLayer interface by creating always new Geometry Objects after manipulation of a Geometry. 
2. Then a Feature.Id can be introduced that can be used as a Cache Key

pauldendulk:
One reason I am thinking about an alternative is that I would like to keep the caching logic more in the outer loop, so that the render methods themselves can stay simple pure functions. With an IFeature.Id we do not have to know anything about the type of feature, and we could perhaps do the caching at the level of the VisibleFeatureIterator callback. So, MapRenderer.RenderFeature (This is just an idea, I have not thought about the details).

inforithmics:
Caching Improvements:
1. Introduced Feature.Id
2. Use it for caching.
3. When a feature is modified. The rendered Cache is cleared. and the Feature id is changed.

inforithmics:
I discovered something intersting. Shapefile already has a FeatureId/GeometryId (rowid).
I'm thinking about reusing this id with a Mapping Scheme to Feature Id.

rowId => -rowid -Offset = featureId.

rowid is negated to get a feature id (because feature id is incremented it is very unlikly that it gets an overlapped number especially because featureId is a long and rowid is an uint.

The -Offset is  (nth Shapefile) * uint.MaxValue

Every ShapeFile gets its own Offset Range (and nth is the number of the Shapefile used). Because again because Feature Id is a long this supports  around one Billion (English Billion) Shapefiles which should be enough.

inforithmics:
I changed the Feature Id to object so that other Keys could be used. The main reason is to create the same key for the same objects.

For example in the DbaseReader

before
`var dr = new GeometryFeature();`

after
`var dr = new GeometryFeature((Id, oid));`

id is an instance counter that is shared with baselayer and baseprovider and dbasereader
oid is the id of the feature in the shape.

So reading the same feature will create the same id.

inforithmics:
So the changes improved Rendering Performance further it is now 16.59 ms (~15% Improvement)

|                           Method |     Mean | Error |      Min |      Max |   Median |    Gen0 |   Gen1 |   Gen2 | Allocated |
|--------------------------------- |---------:|------:|---------:|---------:|---------:|--------:|-------:|-------:|----------:|
|               RenderDefaultAsync | 16.59 ms |    NA | 16.59 ms | 16.59 ms | 16.59 ms | 24.0240 | 3.0030 |      - | 210.48 KB |

pauldendulk (changes requested):
Thanks for the updates. A lot has improved. I have a problem with the object ID though. It is too obscure for my tastes. Previously I mentioned the problems with reference and value compare on the feature itself, I prefer a simple `long` instead. Now, how will compare work with the object Id? I would think it uses reference compare, which would mean there is still no caching after a new fetch (but I guess that is something you checked for). Perhaps if the object consists of structs the object is a struct and it will do an value compare? But irrespective of the answer, the fact that I don't now how this works is already a problem. It makes it harder to understand. 

I would opt for a initial merge request which only uses longs. This won't solve to problem of caching over multiple fetches, but perhaps we should discuss that problem in an issue first. Descript the problem (Perhaps the geojson provider is a good example) and then consider several ways to solve it.

pauldendulk:
So, it it will function as intended if you just use integers. I guess the requirement is that the members should implement value compare. 

```csharp

Console.WriteLine("Objects with structs");
var object1 = (1, (2, 3));
var  object2 = (1, (2, 3));
Console.WriteLine(object1 == object2);

Console.WriteLine("Objects with objects");
var object3 = (1, new Feature { Id = 1 });
var object4 = (1, new Feature { Id = 1 });
Console.WriteLine(object3 == object4);

class Feature
{
    public int Id { get; set; }
}
```
Output:
```console
Objects with structs
True
Objects with objects
False
```

pauldendulk:
oh, perhaps caching over multiple fetches will actually fail in the situations here

```csharp
        var features = Source.GetFeatures(box, resolution).Cast<GeometryFeature>().ToList();
        {
            if (feature.Geometry is Point || feature.Geometry is MultiPoint) continue; // Points with a vertex on top confuse me
            if (feature.Geometry != null)
            {
                int count = 0;
                foreach (var vertex in feature.Geometry.MainCoordinates())
                {
                    yield return new GeometryFeature((Id, feature.Id, "")) { Geometry = new Point(vertex) };
                    count++;
                }
            }
        }
```
This is how I represented that code in a test, where the two `new Feature()` calls represent two different fetches of the same feature:
```csharp
Console.WriteLine("Objects with objects");
var object3 = (1, new Feature().Id);
var object4 = (1, new Feature().Id);
Console.WriteLine(object3 == object4);

class Feature
{
    public object Id { get; set; } = 1;
}
```
Output:
```console
Objects with objects
False
```

pauldendulk:
The fact that Id is declared as object seems to be the problem. This is True:
```csharp
class Feature
{
    public int Id { get; set; } = 1;
}
```

inforithmics:
Feature needs to implement IComparable and override the == and != operators and for Dictionaries it needs to ovveride GetHashcode if this is done the 2 Feature instances are equal

inforithmics:
Console.WriteLine(object3 == object4);

Is interesting what happens if object.Equals is used

inforithmics:
I'm ll change it back to long because this is unexpected behavior.

pauldendulk:
> Console.WriteLine(object3 == object4);
> Is interesting what happens if object.Equals is used.

This makes it True again but the fact that this makes a difference makes it even harder to understand.

Mapsui.Nts/GeometryFeature.cs
```diff
@@ -14,11 +14,20 @@ public GeometryFeature()
...
+    public GeometryFeature(object id) : base(id)
+    {
+    }
+
     public GeometryFeature(GeometryFeature geometryFeature) : base(geometryFeature)
     {
         Geometry = geometryFeature.Geometry?.Copy();
     }
 
+    public GeometryFeature(GeometryFeature geometryFeature, object id) : base(geometryFeature, id)
```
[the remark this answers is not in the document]
inforithmics: This is an externally specified id normally to create the same id for the same copy for example the vertex nodes of an editing shape. The id changes when the editing shape changes. Currently it is a combination of geometry id of editing shape and point position.

Mapsui.Nts/GeometryFeature.cs
```diff
@@ -32,14 +41,6 @@ public override void Dispose()
     {
         if (_disposed) return;
         base.Dispose();
-
-        foreach (var keyValuePair in RenderedGeometry)
```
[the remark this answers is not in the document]
inforithmics: I have to investigate

Mapsui.Nts/Layers/VertexOnlyLayer.cs
```diff
@@ -1,13 +1,14 @@
...
+using System.Threading;
 using Mapsui.Layers;
 using Mapsui.Nts.Extensions;
 using Mapsui.Styles;
 using NetTopologySuite.Geometries;
 
 namespace Mapsui.Nts.Layers;
 
-public class VertexOnlyLayer : BaseLayer, IModifyFeatureLayer
+public class VertexOnlyLayer : BaseLayer
```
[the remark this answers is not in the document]
inforithmics: Good point, a point rendering style could help. Have to think about it.

Mapsui/Rendering/IVectorCache.cs
```diff
@@ -1,17 +1,27 @@
...
-        where T : class;
+    [return: NotNullIfNotNull("brush")]
+    T? GetOrCreatePaint<T>(Brush? brush, float opacity, double rotation, Func<Brush?, float, double, ISymbolCache, T> toPaint)
+        where T : class?;
 
-    TPath GetOrCreatePath<TPath, TGeometry>(Viewport viewport, TGeometry geometry, float lineWidth, Func<TGeometry, Viewport, float, TPath> toPath)
+    T GetOrCreatePath<T, TParam>(TParam param, Func<TParam, T> toSkRect);
+
+    public TPath GetOrCreatePath<TPath, TFeature, TGeometry>(
+        Viewport viewport,
```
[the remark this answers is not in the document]
inforithmics: I agree with that. I didn't want change to much and keep the uncached rendering as origanal as possible. Will try again on a follow up pull request. Because I tried it once and hit some matrix transformation problems.

Benchmarks/Mapsui.Rendering.Benchmarks/RenderToBitmapPerformance.cs
```diff
@@ -131,7 +134,6 @@ public async Task RenderDefaultAsync()
     [Benchmark]
     public async Task RenderRasterizingTilingPngAsync()
     {
-        await pngMap.WaitForLoadingAsync();
```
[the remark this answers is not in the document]
inforithmics: The wit is now done in the prepare of the data. So to measure only the rendering.

What I want to investigate is from where the 100kb of memory usage comes during one render cycle because in my opinion when the skpath is cached not much memory should be used for rendering.

Mapsui.Nts/GeometryFeature.cs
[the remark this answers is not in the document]
pauldendulk: It is an external id, but this may not be that obvious that developers and users. The fact that it has the same name as the id in the source might be confusing. By giving it another name it may be clearer that this a different kind of thing. That was my idea. If we want to give it another name we have to think about a good name. This is not something that needs to be changed in this MR.

pauldendulk:
> I'm ll change it back to long because this is unexpected behavior.

I solution with a long might not be sufficient to solve the 'caching over fetches' problem, but that problem could be addressed in another merge request.

inforithmics:
Things done: I introduced a FeatureKeyCreator for mapping differnt keys to a uint

inforithmics:
|                           Method |     Mean | Error |      Min |      Max |   Median |    Gen0 |   Gen1 |   Gen2 | Allocated |
|--------------------------------- |---------:|------:|---------:|---------:|---------:|--------:|-------:|-------:|----------:|
| RenderDefaultAsync               | 19.70 ms |    NA | 19.70 ms | 19.70 ms | 19.70 ms | 24.0240 | 3.0030 |      - | 213.92 KB |
| RenderDefaultCachedAsync         | 16.94 ms |    NA | 16.94 ms | 16.94 ms | 16.94 ms | 24.0240 | 3.0030 |      - | 213.92 KB |

So I measured the skpath caching improvement, it is 3ms per cycle, the main reason it isn't more is because in the unit tests I render to bitmap with gets serialized, and the test measures the serialization time of the bitmap too. (213.92kb is the allocated bitmap for rendering). Maybe I have to make a benchmark without bitmap serializing

inforithmics:
This are now the rendering times without Bitmap overhead (I had to repair the Cpu Benchmarks) so it is twice as fast and much less memory per image in the gpu case it would even be better but i cannot test the gpu in benchmarkdotnet

| Method                          | Mean       | Error | Min        | Max        | Median     | Gen0     | Gen1     | Allocated  |
|-------------------------------- |-----------:|------:|-----------:|-----------:|-----------:|---------:|---------:|-----------:|
| RenderDefaultWithoutCacheAsync  | 7,813.6 us |    NA | 7,813.6 us | 7,813.6 us | 7,813.6 us | 258.2583 | 255.2553 | 2123.26 KB |
| RenderDefaultAsync              | 3,808.9 us |    NA | 3,808.9 us | 3,808.9 us | 3,808.9 us |  15.0150 |        - |  143.76 KB |

Mapsui.Nts/Providers/GeometryIntersectionProvider.cs
```diff
@@ -12,37 +13,47 @@
 
 namespace Mapsui.Nts.Providers;
 
-public class GeometryIntersectionProvider : IProvider
+public class GeometryIntersectionProvider : IProvider, IProviderExtended
```
pauldendulk, line 16: IProviderExtended is not a great name, but I guess the intention is that this should become part of the interface?

pauldendulk (approved):
Great that we can use a simple long for caching. The logic to determine the next long is still not easy to follow but it is not part of the interfaces so it is easy to change if we think of something better.

Benchmarks/Mapsui.Rendering.Benchmarks/Program.cs
```diff
@@ -5,6 +5,11 @@ public class Program
 {
     public static void Main(string[] args)
     {
+#if DEBUG
```
pauldendulk, line 8: @inforithmics  Just noticed in master that this breaks the build in debug mode. I will just remove it. I am telling you because you may intended something else.

inforithmics: That's fine for me. Was some Debugging code that can be easily removed.

PR #2143, opened 2023-08-08
Fixing Build with Visual Studio 17.7

inforithmics:


inforithmics:
The failure of winBuild seems to be a temporary problem

inforithmics:
For the Windows Build it uses SDK 7.0.306 instead of the latest 7.0.400. For using 7.0.400 it needs the newest Visual Studio 17.7 which is not yet installed on the Github Windows Image. When it is updated 7.0.400 can be used.

inforithmics:
For Uno Samples following Bug needs to be fixed. Uno Works when the nuget package is referenced but not when the project is referenced.
https://github.com/dotnet/maui/issues/16680

inforithmics:
@pauldendulk I think this should be merged to make main green again

inforithmics:
Fixed Xamarin Projects.
See here for reason.
https://github.com/dotnet/roslyn/issues/69092

PR #2147, opened 2023-08-18
Add a new way for dragging: now supports to drag the whole feature on the map.

cecildelakers:
Hello! Dear maintainers, I have added a new way for dragging, which allows us to drag the whole geometry/feature from the previous position to another place on the map. Please check my codes and then merge them to the master branch, if my advice is useful to this project. Thanks!


Mapsui.Nts/Editing/EditManager.cs
```diff
@@ -159,7 +159,7 @@ public bool AddVertex(Coordinate worldPosition)
             return null;
 
         return vertices.OrderBy(v => v.Distance(mapInfo.WorldPosition.ToCoordinate()))
-            .FirstOrDefault(v => v.Distance(mapInfo.WorldPosition.ToCoordinate()) < mapInfo.Resolution * screenDistance);
+            .FirstOrDefault(v => v.Distance(mapInfo.WorldPosition.ToCoordinate()) < mapInfo.Resolution * screenDistance * 100);
```
inforithmics, line 162: Is this increased so that the nearest vertex is found when you click inside a polyogn or line? Or is there another reason for this?

Mapsui.Nts/Editing/EditManager.cs
```diff
@@ -192,19 +192,51 @@ public bool Dragging(Point? worldPosition)
     {
         if (EditMode != EditMode.Modify || _dragInfo.Feature == null || worldPosition == null || _dragInfo.StartOffsetToVertex == null) return false;
 
-        _dragInfo.Vertex.SetXY(worldPosition.ToMPoint() - _dragInfo.StartOffsetToVertex);
+        // set the boundary value at 10 to decide the way of dragging
+        // try to modify vertex itself, if the worldPosition of mouse-clicking is close to the _draginfo.Vertex
+        if (worldPosition.Distance(_dragInfo.Vertex!.ToPoint()) < 10)
```
inforithmics, line 197: I think this 10 should be scaled my the mapinfo.Resolution or else depending on the underlaying Coordinate systems it means ~10m or ~10000km

inforithmics: I tested it and it works quite well, except if I move the mouse really fast than I start moving the shape instead of the point. so that's why I think this parameter needs to be adapted to the map resolution (so it depends on the zoom level too).

Mapsui.Nts/Editing/EditManager.cs
inforithmics: Without it the moving of the shapes doesn't work.

inforithmics:
I fixed the moving of the shape and removed the magic numbers, so the dragging works now when moving fast. I have now a Vertex for moving nodes and VertexFeature for moving Shapes so, the magic numbers aren't needed anymore.

inforithmics (approved):

Mapsui.Nts/Editing/EditManager.cs
cecildelakers: Yes, at that time I didn't find a suitable way to make my idea work, so I modified the parameter. Actually I know it's not the best method.

cecildelakers:
I'm so sorry! I didn't notice your comments until now so that I couldn't offer you any help! So kind of you to correct my codes! :)

inforithmics:
No, problem can you check if my changes work for you too, if yes I can merge this.

PR #2166, opened 2023-09-11
Update roadmap.md

pauldendulk:
https://github.com/Mapsui/Mapsui/discussions/2165

### DIFF
--- a/docfx/documentation/roadmap.md
+++ b/docfx/documentation/roadmap.md
@@ -18,9 +18,8 @@
 - Everything to .NET 6
 - Remove:
   - Xamarin.Forms
-  - Xamarin.Android
-  - Xamarin.iOS
-  - Uno UWP (not Uno WinUI)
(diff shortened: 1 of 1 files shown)


inforithmics:
Uno.UI Version 5.0 will Remove UWP. So Uno.Ui and Uno.Ui.WinUi would target at least .Net 6.0

pauldendulk:
> Uno.UI Version 5.0 will Remove UWP. So Uno.Ui and Uno.WinUi would target at least .Net 6.0

So, moving tot .NET 6 does itself not require the removal of Uno.UI. Still, I want to support less, and support that better, and also move faster forward instead spending time on maintenance. If you have more stuff, than more stuff will break, that is just how it works.

inforithmics:
I agree with that, and Mapsui.Uno.WinUI is downloaded 4 times as often per day than Mapsui.Uno.

PR #2181, opened 2023-09-25
Update to BruTile 5.0.5

inforithmics:


pauldendulk (commented):
BruTile.MbTiles needs to be updated as well

inforithmics:
Good point

PR #2183, opened 2023-09-27
Fix click event on RasterizingTileLayer 

inforithmics:
- [x] Fixed the IFeatureInfo Interface (WPF) there was a compare with the == operator of two Points because == was not implemented it made a Reference comparision which was always false.
- [x] Before getting the MapInfo I look if the Layer has data Loaded. If not a RefreshData is executed.

- [x] It seems the Wms Info Sample does not work in Blazor have to Investigate why. Needs different pull request for fixing samples
- [x] Don't make MapInfo Async


inforithmics:
Before updating all images I ran the unit tests locally and they all worked.

pauldendulk:
About  https://github.com/Mapsui/Mapsui/discussions/2177

inforithmics:
Things done:
1) Obsoleted OnFeatureInfo OnInfo should be used
2) Implmented IFeatureInfo on RasterizingTileLayer
3) Use the IFeatureInfo Implementation on MapInfo

inforithmics:
Things done:
1) I Implemented the IFeatureInfo Interface on the WmsProvider so that FeatureInfo of the underlaying Wms can now be loaded

inforithmics:
Things done:
Fixed the breaking changes, obsoletet IFeatureInfo and made new ILayerFeatureInfo, that is used for MapInfo.
Use MapInfo for displaying in the two new samples fore MapInfo

Mapsui.Tiling/Provider/RasterizingTileProvider.cs
```diff
@@ -238,4 +239,40 @@ public static Viewport ToViewport(MSection section)
             section.ScreenWidth,
             section.ScreenHeight);
     }
+
+    public async Task<IDictionary<string, IEnumerable<IFeature>>> GetFeatureInfoAsync(Viewport viewport, double screenX, double screenY)
+    {
+        var result = new Dictionary<string, IEnumerable<IFeature>>();
```
pauldendulk, line 245: This is a lot of code. This make uneasy. Makes me think we should perhaps go for an entirely different approach. 

I introduced the wrapping strategy, where you put a rasterizing provider around another provider. I liked the simplicity  of how you could combine providers like that, like stacking Lego blocks. But maybe that was a bad idea. Rasterizing is a performance optimization, and should perhaps be the responsibility of the renderer. Perhaps a Layer should have a field to indicate whether it should use Tiled, Rasterized or Direct rendering. Anyway, a thought, not something for now.

The fact that you need the renderer in a layer is already uncomfortable (something I introduced in the RasterizingLayer).

pauldendulk (changes requested):
Thanks a lot. The feature info has been half implemented for a very long time. Great you took it on. I have two fairly big requests for change though.
- As mentioned in the comments, we should not introduce .Result, it is considered a bad practice by Microsoft and I have personally struggled with very hard to find bugs caused by it. I think the best solution for now is to add a separate handler for feature info, and a separate IsFeatureInfoLayer property. Not beautiful, but clear, so easy to change into something better later on.
- Then a tough one for me to decide: I don't want add INotifyPropertyChanged change for widgets. I know it is me alone against the rest of the world on this one, but I still don't want it. I understand it makes sense because the widgets are controls like we all know from the UI frameworks and all these frameworks implement INotifyPropertyChanged, but still. Reasons:
  - Performance problems caused by what I call 'notification storms' are hard to track.
  - If we implement it we need to implement it very every single widget, this is lot of work to do, but also a lot of work to maintain. 
  - Every new widget that is added should also implement it.
  - I don't want INotifyPropertyChanged for features either. There my argumentation is stronger, because features are not at all like UI controls, they are more like primitives in a render engine. They come and go at a high pace. It is not a good idea to listen to them individually. You have to bind and detach all the time. Since we do not use it for features it makes more sense to not use it for widgets (although they are different I have to admit). Also everything we do is drawn by the same render cycle, in that sense features and widgets a the same, so it makes sense to treat them the same way.

Side note, INotifyPropertyChanged is not really related to feature info, so could have been in a separate MR. Smaller MRs are better anyway. I expect to be quite active on Mapsui the coming four months, so, hope to respond to PRs quicker, which makes it easier to make progress with smaller PRs. Also, sorry, I kept this one waiting for so long.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -263,6 +268,21 @@ private void Render(object canvas, Viewport viewport, IEnumerable<IWidget> widge
                         Logger.Log(LogLevel.Error, "Unexpected error in the code detecting if a feature is clicked. This uses SkiaSharp.", exception);
                     }
                 });
+
+                foreach (var infoLayer in featureInfoLayers)
+                {
+                    var layer = ((ILayer)infoLayer);
+#pragma warning disable VSTHRD002 // synchronously waiting
+                    var features = infoLayer.GetFeatureInfoAsync(viewport, x, y).Result;
```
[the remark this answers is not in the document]
inforithmics: I'll fix this.

Mapsui/Layers/BaseLayer.cs
```diff
@@ -182,7 +182,7 @@ protected set
     public virtual IReadOnlyList<double> Resolutions { get; } = new List<double>();
 
     /// <inheritdoc />
-    public bool IsMapInfoLayer { get; set; }
+    public virtual bool IsMapInfoLayer { get; set; }
```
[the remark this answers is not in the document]
inforithmics: Fixed

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -4,6 +4,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 using Mapsui.Extensions;
+using Mapsui.Fetcher;
```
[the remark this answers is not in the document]
inforithmics: fixed

inforithmics:
>     * Then a tough one for me to decide: I don't want add INotifyPropertyChanged change for widgets. 
I Reverted this now.

Mapsui/Layers/ImageLayer.cs
```diff
@@ -189,6 +189,24 @@ public void ClearCache()
         }
     }
 
+    public override bool IsMapInfoLayer
+    {
+        get
+        {
+            if (base.IsMapInfoLayer)
```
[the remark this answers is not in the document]
inforithmics: Is Reverted.

Mapsui.Rendering.Skia/MapRenderer.cs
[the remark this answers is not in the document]
inforithmics: I'm going with Option 2.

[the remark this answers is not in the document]
pauldendulk: > GetMapInfoAsync with .Result

Let us not use .Result anywhere. It's easier if this is just a hard rule, `No .Result.`.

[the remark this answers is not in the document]
inforithmics: So I will make here a breaking change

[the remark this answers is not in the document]
pauldendulk: Maybe I don't have the full picture but I would think a breaking change is not needed. I would keep the GetMapInfo like it currently is, and add a new GetFeatureInfoAsync. In the future we may want this to fall under one method, but I am not sure of that. For hover info you want to have direct access to what is in memory. WMS feature info is not suitable for hover info (too many requests to the server) but if you have the data in memory, it may be a good option. So, this would mean we really have to distinguish between the two kinds of info.

inforithmics:
It seems the Wms Info Sample does not work in Blazor have to Investigate why.

Mapsui/Styles/RasterStyle.cs
```diff
@@ -12,7 +13,7 @@ public class RasterStyle : IStyle
     private const int MinimumTilesToKeep = 32;
     private long _lastIteration;
     private readonly IDictionary<object, IBitmapInfo?> _tileCache =
-        new Dictionary<object, IBitmapInfo?>(new IdentityComparer<object>());
+        new ConcurrentDictionary<object, IBitmapInfo?>(new IdentityComparer<object>());
```
inforithmics, line 16: During Testing it crashed here because it was accessed by several threads.

pauldendulk: I was not aware we had a cache inside a style. This should be moved out of there. Let's use style just for defining the style. In general I am worried about the number of caches we have in all sort of places. A requirement of each cache should be that we should be able to replace it with a non-caching alternative, we will need that to track down the obscure bugs it will cause.

pauldendulk:
> I think the CustomCalloutSample does not behave correctly on all plattforms have to investigate

If this was already the case before this PR then it would be better to leave it for another PR.

inforithmics:
The Custom callout Sample works as expected. I misunderstood the code.

pauldendulk (changes requested):
Please leave GetMapInfo synchronous and and add separate logic for GetFeatureInfoAsync. This means users will have to take both into account. That is problem but let's fix that in a later PR. 

### On async and 'IO'
In general we want to avoid the use of async where we can. When async was introduced I was surprised at how it was viral and infected all it's callers. I did not like that aspect. Later I realized the main problem was with my design. IO (web request and db access) should not be on the inside (core) but the outside of your application, because if do it on the inside everything that depends on it suffers from the fuzziness of possibly failing calls. In a programming language like Haskell you are forced to put an IO tag on everything that is directly or indirectly involved in IO. And as a Haskell developer you are trying to keep as much code as you can in the part without IO. In C# async should be treated in the same way. If all you code becomes async you are doing something wrong. In Mapsui the IO is in the providers. The Map and Layers do the calls to it, becoming async as well. This is not a great design. Fetching data should be the responsibility of a single component which is on the outside (perhaps a member of the MapControl). The majority of the code (Styles, Layers, Map) should be just data, or methods operating on in memory data, preferably pure functions.

### Async in this PR
In this PR the editing controls now becomes async while it is not actually modifying remote data, this is a step in the wrong direction. Creating EventArgs is now async as well, that is something that should not be necessary. 

### Let the user fetch the feature info
A possible solution for this is to remove the feature data from the EventArgs, but provide a method that users could call to retrieve the feature data. This idea is something that I had already in mind for a problem we had with the mouse over event (hovering). When hovering you sometimes also want to listen to an event. In Mapsui we had that at some point but removed it because this could easily cause performance problems, because on every mouse move we are querying the data. Even it if is in-memory data this can become slow. If we do not retrieve the data on our side but let the users retrieve it there are some advantages: 
- In some cases users do not need the feature info but just want to know the coordinates in the mouse over. This would now be possible without a performance impact.
- The users understand better what is going on. They explicitly call the data. If this causes performance problems it is easier for them to see this is caused by their call. It is more visible what is happening, less magic under the hood.
- We could leave the call to the async methods to the user, leaving our own code synchronous.

pauldendulk:
I added this issue to discuss the proposal above https://github.com/Mapsui/Mapsui/issues/2250

robloo:
> In general we want to avoid the use of async where we can. When async was introduced I was surprised at how it was viral and infected all it's callers. I did not like that aspect. 

Yes, I agree with you that it's not great it is viral and propagates everywhere. Unfortunately, like overuse of the 'var' keyword, it has become the de-facto standard in .NET to just propagate async without thinking. .NET (and Windows App SDK) realized async everywhere probably wasn't the greatest too so has sync methods for almost everything now. That leaves it up to app developers to architect things correctly (lots of legacy code was doing synchronous IO on background threads already so introducing async was hacky and a waste).

pauldendulk:
> > In general we want to avoid the use of async where we can. When async was introduced I was surprised at how it was viral and infected all it's callers. I did not like that aspect.
> 
> Yes, I agree with you that it's not great it is viral and propagates everywhere. Unfortunately, like overuse of the 'var' keyword, it has become the de-facto standard in .NET to just propagate async without thinking. .NET (and Windows App SDK) realized async everywhere probably wasn't the greatest too so has sync methods for almost everything now. That leaves it up to app developers to architect things correctly (lots of legacy code was doing synchronous IO on background threads already so introducing async was hacky and a waste).

Yes, but in Mapsui the async calls are mostly because of HttpClient calls and those should really be asynchronous. The problem with the current design of Mapsui is that those calls or on the inside. So, the DataSource does the calls, but this is inside the Layer, which is inside the Map, which is inside the MapControl, making them all async. The core code should consist of pure functions, the provide the information and classes to do the calls, but the calls themselves should be in some component which is directly in the MapControl. This would reduce the problem of the viral cascade of async.

inforithmics:
Things done:
1. I added a Task to Mapinfo event and a GetMapInfoAsync() Method, So when the Async Info is although requested, it waits for the Async Tasks to finish and repopulates the Mapinfo with the Async information.
2. Improved MapInfoWidget to Call first the synchronous MapInfo and afterwards the Async MapInfo. Which can take longer.

Mapsui/Styles/RasterStyle.cs
inforithmics: Should I move this out in this pull request or in a seperate pull request.

Mapsui.Tiling/Provider/RasterizingTileProvider.cs
inforithmics: I'll have to think about a better method.

inforithmics: Currently the Renderer is needed for Hit detection, Because a Color change is used for finding out if a Feature is placed at a certain point, This is not very efficient with a lot of Features. So the Hit Algorithm needs to be improved for finding intersecting Features with the Point. But Symbols need to be handled too. Maybe there it could be when the Color is not transparent.

inforithmics:
@pauldendulk So this pull request can be reviewed again.

Things done:
1) Updated Unit Tests nugets (for aligning with Visual Studio Version)
2) dotnet format whitespace (for formatting code)

Mapsui/Styles/RasterStyle.cs
pauldendulk: So you did, and it is already merged.

Mapsui.Tiling/Provider/RasterizingTileProvider.cs
pauldendulk: In the past we had a GetMapInfo implementation that calculates intersections. The color based hit detection, suggested by charlenni, was an immense improvement. A mathematical method to calculate an intersection on proper geometric shapes is straightforward (we had such methods in Mapsui, maybe still have), but that is not the main problem. When using calculations you need to do that for all feature types. For NTS we could use the methods that NTS provides but we also need to add it for MPoint and MRect. When we introduce a new feature type we need to implement intersection methods for it. And when someone adds a custom renderer (like the 'bears' in the custom renderer sample) they should also implement the intersection methods for it. Then there is the width of the outlines you need to take into account, you need to grow your geometry accordingly to adjust for that, for that you need to take into account the current resolution. And indeed the symbols, which is the hardest part. You need to take into account the scale, rotation, RotateWithMap = true|false, and the shape of the symbol itself. If a bitmap is used you still need to do some kind of pixel color detection to know if it is actually hit. Other symbol types may need other calculations. I enjoy the color based detection every day :)

Perhaps this was not your main point, but I felt it was important to mention that here.

pauldendulk: For hover tooltips we may need a faster GetMapInfo. I was thinking about a 'pixel map' solution, where we use a bitmap which is the same size as the screen. The pixels are not used to store regular colors but colors to represent a feature ID (instead of a bitmap we could use a two dimensional array as well, whatever performs better). There are all sorts of things to take into account related to performance. I have thoughts about that, but lets not go into this right now.

pauldendulk: Now you main point is about hit detection in the context of the RasterizingTileLayer. I don't have a solution for that in mind, but I also did not study the code. This suggestion is a big change which will affect rendering which I think we should postpone for a later release. So, for now lets keep it like it is.

Directory.Packages.props
```diff
@@ -65,7 +65,7 @@
     <PackageVersion Include="NetTopologySuite" Version="[2.5.0,3.0.0)" />
     <PackageVersion Include="NetTopologySuite.IO.GeoJSON4STJ" Version="[4.0.0,5.0.0)" />
     <PackageVersion Include="Newtonsoft.Json" Version="[13.0.2,14.0.0)" />
-    <PackageVersion Include="NUnit" Version="[4.0.1,5.0.0)" />
+    <PackageVersion Include="NUnit" Version="[3.14.0,4.0.0)" />
```
pauldendulk: Yes, I ran into this in another project as well. It is a very big change. I have been told that a stepwise migration is possible using nunit legacy https://docs.nunit.org/api/NUnit.Framework.Legacy.html.

pauldendulk (approved):
🙏

Issue #2189, opened 2023-09-30
When clicking the zoom button the map pans while mouse is up


pauldendulk:
**Mapsui Version**
4.0.0-preview.1

**Mapsui Platform**
Blazor

**Device**
Chrome on Windows

**Describe the bug**
When clicking the zoom button (mouse down and then up) and then moving away from the zoom button, the map pans with the moving mouse even though the mouse is up. 

Also while the mouse is down on the zoom button and you move within the button, the map also moves. So, the problem is probably on the mouse down.

**To Reproduce**
1. Go to the Mapsui online Blazor samples
2. Select 'OpenStreetMap' under demo
3. Click on the + button.
4. Move the mouse away
5. The map pans while it should not on mouse up

**Expected behavior**
The map should not pan on mouse up

pauldendulk:
I tested this on Avalonia as well (no other platforms apart from Blazor and Avalonia) and there is works correctly, but when stepping through the code in Avalonia I saw that in MapControlMouseMove _MouseDown was true. It did not pan because _previousMousePosition was null, but that is not how it is supposed to work in general. This is the code in Avalonia:

```csharp
        if (_mouseDown)
        {
            if (_previousMousePosition == null)
            {
                // Usually MapControlMouseLeftButton down initializes _previousMousePosition but in some
                // situations it can be null. So far I could only reproduce this in debug mode when putting
                // a breakpoint and continuing.
                return;
            }
```

Could it be there is an error in the shared HandleTouching? See the `return false` in the `else` in the code below.

@inforithmics You added the `else` in the [simplify editing widget PR](https://github.com/Mapsui/Mapsui/pull/2133/files) but it does not change the behavior (it return false already in that case), so, it looks to me like an attempt to fix exactly this problem, but you only wrote false instead of true. Could that be?

```csharp
private bool HandleTouching(MPoint position, bool leftButton, int clickCount, bool shift)
{
    var touchableWidgets = GetTouchableWidgets();
    var touchedWidgets = WidgetTouch.GetTouchedWidget(position, position, touchableWidgets);

    foreach (var widget in touchedWidgets)
    {
        if (widget is IWidgetExtended extendedWidget)
        {
            var widgetArgs = new WidgetArgs(clickCount, leftButton, shift);
            if (extendedWidget.HandleWidgetTouching(Map.Navigator, position, widgetArgs))
                return true;
        }
        else
        {
            // Handle the touched avoid duplicated touched events
[...]

PR #2190, opened 2023-09-30
Fixed Zoom Button Handling

inforithmics:
fixes this https://github.com/Mapsui/Mapsui/issues/2189

### DIFF
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -615,17 +615,14 @@ public partial class MapControl : INotifyPropertyChanged, IDisposable
 
     private bool HandleTouchingTouched(MPoint position, bool leftButton, int clickCount, bool shift)
     {
-        if (HandleTouching(position, leftButton, clickCount, shift))
-        {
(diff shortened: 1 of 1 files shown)


pauldendulk (approved):
Thanks!

The code makes me think a bit, but I don't know a way to make it easier to skim yet.

The result could perhaps be renamed to handed. That would be an appropriate description right? Still it would not be ready easy to read.

Also, it takes some effort to distinguish Touched from Touching. Microsoft also uses this way of indicating a before and after event in their apis, but I also find it hard read there. Perhaps we could use something with Up and Down, which are easy to grasp for everyone (PointerUp and PointerDown). Btw, I understand that this came about because of the already existing `HandleWidgetTouched`. Adding a `HandleWidgetTouching` was the logical thing to do.

If we would change the api this would be a breaking change and should be in v5. To help users with these kind of migrations you can keep the old version and mark it as obsolete and indicate what the alternative is. Like: `[Obsolete("Use PointerUp instead", true)]`. We would have to implement an empty method in the Widget with a NotImplemented Exception.

PR #2191, opened 2023-10-01
Update Topten.RichTextKit, Svg.Skia, NetTopologySuite.IO.GeoJSON4STJ

inforithmics:


pauldendulk (approved):
The shared Directory.Package.props saves us so much time.

PR #2194, opened 2023-10-04
Fix: Imagelayer invisible and no warning when constructor with name parameter is used

pauldendulk:
Things done:
- Make parameterized constructor call default constructor, which fixes the actual bug.
- Reorder the methods/fields in ImageLayer somewhat. We pay little attention to the order usually but in this case I think it caused the bug because the constructors were not next to each other.
- Add logging in the VectorStyleRenderer if the feature type is unknown. This might cause logging in existing scenarios which are not bugs, but I still think it is better to log.

### DIFF
--- a/Mapsui/Layers/ImageLayer.cs
+++ b/Mapsui/Layers/ImageLayer.cs
@@ -20,21 +20,6 @@ namespace Mapsui.Layers;
 
 public class ImageLayer : BaseLayer, IAsyncDataFetcher, ILayerDataSource<IProvider>, IDisposable, ILayer
 {
-    public ImageLayer()
-    {
-        Style = new RasterStyle();
-    }
-
-    protected override void Dispose(bool disposing)
-    {
-        if (disposing)
-        {
-            _startFetchTimer?.Dispose();
-        }
-
-        base.Dispose(disposing);
-    }
-
     private class FeatureSets
     {
         public long TimeRequested { get; set; }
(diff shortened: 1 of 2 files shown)


Mapsui/Layers/ImageLayer.cs
```diff
@@ -49,6 +34,18 @@ private class FeatureSets
     private IProvider? _dataSource;
     private readonly int _numberOfFeaturesReturned;
 
+    public ImageLayer()
+    {
+        Style = new RasterStyle();
+    }
+
+    public ImageLayer(string layerName) : this()
```
inforithmics, line 42: Yes calling the base constructor fixes the bug.

Mapsui/Layers/ImageLayer.cs
```diff
@@ -49,6 +34,18 @@ private class FeatureSets
...
+    {
+        Name = layerName;
+        _startFetchTimer = new Timer(StartFetchTimerElapsed, null, Timeout.Infinite, Timeout.Infinite);
```
inforithmics, line 45: Looking at this I'm wondering if this shouldn't be done in the ImageLayer Constructor. Because when an ImageLayer is created without name it does not have a Fetch Timer initialized

inforithmics (approved):

Mapsui/Layers/ImageLayer.cs
pauldendulk: ah, that is actually another bug. Good catch! I pushed the fix.

Issue #2087, opened 2023-06-23
Mapsui.Samples.Maui does not work properly on iOS (and Mac)


janusw:
**Mapsui Version**
master as of today with modifications from PR https://github.com/Mapsui/Mapsui/pull/2084, which are necessary to restore the sln with VS4Mac (branch janusw:building_xamarin_forms_on_mac)

**Mapsui Platform**
MAUI

**Device**
iOS (iPhone + iPad), MacOS (MacBook Pro)

**Describe the bug**
When building and running the MAUI sample on iOS (or Mac), the app starts up alright and I can select a sample category via the picker, but when tapping on a specific sample name, nothing happens. No map is shown.

pauldendulk:
I can reproduce this when running Mapsui.Samples.Maui Android on Windows Subsystem for Android.

janusw:
> I can reproduce this when running Mapsui.Samples.Maui Android on Windows Subsystem for Android.

so you say that Android has the same issue as iOS and Mac, and you don't see #2088 on Windows?

janusw:
> Fixed with #2087

Actually it is not fully fixed for me. The effect I see from #2087 is that now the sample name is highlighted when tapping on it. But on the iPhone and iPad I still see no map. The reason seems to be that the controls for selecting the sample are simply too large (and take up the whole screen on "small" devices - even on an 11" iPad).

When running the MAUI sample on MacOS, I see only a small map on the right side of the screen (see screenshot below), so at least the layout of the controls should be improved.

Note that in the Xamarin.Forms sample, the controls for selecting a sample are hidden when selecting a control and the map is shown almost full-screen (at least on Android and iOS - on Mac the sample list is shown permanently).

<img width="1685" alt="Screenshot 2023-06-25 at 10 15 55" src="https://github.com/Mapsui/Mapsui/assets/484108/4d61afb7-af02-4f1e-a364-3fba2c15731b">

pauldendulk:
The behavior you describe on iOS and MacOS is indeed unacceptable. Thanks for the feedback. I will add further improvements. Probably by adding a flyout, like mentioned above.

pauldendulk:
Fixed with https://github.com/Mapsui/Mapsui/pull/2199#pullrequestreview-1662726444

janusw:
Thanks for fixing the layout. Looks pretty good on an iPad now (on a small iPhone, the layout is still not optimal in portrait mode, but that is bearable).

[...]

PR #2199, opened 2023-10-06
Limit width of sample button list in MAUI sample

pauldendulk:
This is intended to fix the issue reported here https://github.com/Mapsui/Mapsui/issues/2087#issuecomment-1605939314. I could not test this because I do not run on Mac. I now hard limited the width op the controls. It seems on mac the auto size column takes up so much width because it thinks some of it's children are this wide. By explicitly setting the width I hope to counter that.

I thought about a better solution.
- A flyout menu is supported in MAUI but this only applies to tabbed pages.
- There is a MenuView, but this only works on Windows and Mac.
- What seems a fair option would be to add a toggle button on the top left of the map which could set the width of the sample menu to zero or back to the original size.


inforithmics (approved):
I tested it and it works now on macos (does now only use a small part of the window)

PR #2201, opened 2023-10-07
Fix hyperlink click

pauldendulk:


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -664,14 +664,14 @@ private bool HandleTouched(MPoint position, bool leftButton, int clickCount, boo
             }
             else
             {
-
                 if (widget.HandleWidgetTouched(Map.Navigator, position))
+                    return true;
+                else if (widget is Hyperlink hyperlink && !string.IsNullOrWhiteSpace(hyperlink.Url))
```
pauldendulk, line 669: @inforithmics Please look at this carefully, I am not sure this now works as was intended. This is weird also because the hyperlinks implementation is odd.

**How the hyperlink is supposed to work**
Because the OpenBrowser functionality is platform specific we call OpenBrowser in the MapControl specifically for Hyperlink. Perhaps there is a better solution for this where the platform specific code is still called from the hyperlink event handler itself. The hyperlink HandleWidgetTouched has an OnTouch method that users could use to create their own implementation (perhaps this should be removed, it customization is needed they could create their own). If the users event handler sets the args.Handled = true we should not call the hyperlink in our code. So, the code here should only do something when the method return false. Once we have called the hyperlink is considered handled and return true.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -664,14 +664,14 @@ private bool HandleTouched(MPoint position, bool leftButton, int clickCount, boo
...
                 {
-                    if (widget is Hyperlink hyperlink && !string.IsNullOrWhiteSpace(hyperlink.Url))
-                    {
-                        OpenBrowser(hyperlink.Url!);
-                    }
-
+                    // The HyperLink is a special case because we need platform specific code to open the
+                    // link in a browswer. If the link is not handled within the widget we handle it
```
inforithmics, line 672: nit: browswer => browser

inforithmics (approved):
Good catch: I broke this with my widget simplification. So the handling of the hyperlink here fixes this.

Issue #2197, opened 2023-10-06
Touch behavior issue after widget tap


janusw:
**Mapsui Version**
4.1.0-preview.1

**Mapsui Platform**
Xamarin.Forms

**Device**
iOS (v16+17)

**Describe the bug**
I'm seeing a regression with widget behavior in version 4.1.0-preview.1 (compared to 4.0.0). I use custom widgets in a MapView that trigger a particular action when tapping. The action is still triggered alright with the preview version, but afterwards there are issues with the touch behavior (it seems like a TouchEnded event is missing or something like that): Panning with a single finger results in two-finger pinch behavior, and after a second tap on the widget it does not work at all any more. Probably the MapView keeps some extraneous touch points that don't exist any more.

**Expected behavior**
Touch gestures should work like usual.

**Additional context**
It seems the bug was introduced with PR #2133 by @inforithmics (most probably due to the MapView changes in commit 91b24b316f40eac431060c6f52a7c7346081c951).

pauldendulk:
Thanks for the clear report. 

From reading the description I think the underlying problem is the same as in this issue:
https://github.com/Mapsui/Mapsui/issues/2189
Which should be fixed with this PR:
https://github.com/Mapsui/Mapsui/pull/2190

Although your description includes some more problems, so I am not sure. Perhaps you can test this with these artifacts:
https://github.com/Mapsui/Mapsui/actions/runs/6365272919

janusw:
> From reading the description I think the underlying problem is the same as in this issue: #2189 Which should be fixed with this PR: #2190

Ah, good. I did not see the other ticket, but it really sounds like it's the same problem.


> Although your description includes some more problems, so I am not sure. Perhaps you can test this with these artifacts: https://github.com/Mapsui/Mapsui/actions/runs/6365272919

I tried the merge commit of #2190, but unfortunately none of my widgets show any reaction any more ☹️
(neither on iOS nor on Android, with Mapsui.Forms)

janusw:
Just tried Mapsui.Samples.Forms.Android on current master. The zoom buttons also don't seem to work there. Was #2190 only tested with Blazor? Do the widgets work there?

pauldendulk:
[...]

PR #2203, opened 2023-10-08
fix: Widget Touch does not work on Maui/Forms on Touch devices

inforithmics:
doing this by Setting the handled to true if touch is handled

Fixes this

https://github.com/Mapsui/Mapsui/issues/2197


janusw (approved):
I verified that this makes the widgets work again in Mapsui.Samples.Maui.MapView on Android and iOS, and also in Mapsui.Samples.Forms.Android and Mapsui.Samples.Forms.iOS.

So, ok with me, thanks for fixing!

PR #2205, opened 2023-10-09
Fix jagged background for labels and widgets

pauldendulk:


pauldendulk:
> Are there maybe more SkPaint Colors that need antialising?

Likely. But I think not every draw method benefits from it, and there is a performance penalty we may not always want to take. For instance, for drawing bitmaps there is the FilterQuality and we use FilterQuality.Low, while there is also High and Medium. For my purpose (some years ago) High was too slow. Btw, if you do not set the FilterQuality the default None is used, which is significantly worse than Low.

inforithmics:
Yes, I thought about a configurable option could be helpful. I didn't see the problem because I have a 4K Monitor so Rounded Corners are smooth even without AntiAliasing and Smartphones have high Resolution too so you don't see much of a difference. But the Performance could be noticable. 

Maybe the MapControl could have two Properties Antialiasing and FilterQuality with Antialiasing set to true and FilterQuality set to FilterQuality.Low.

Or maybe this can be set with Properties on the Renderer.

charlenni:
Is this nothing, that should be configurable for each layer or widget and not globally?

pauldendulk:
@charlenni @inforithmics I created an issue to discuss this https://github.com/Mapsui/Mapsui/issues/2213

Issue #2209, opened 2023-10-09
MouseCoordinatesWidget does not display the mouse coordinates.


pauldendulk:
**Mapsui Version**
4.1.0-preview.1

**Mapsui Platform**
Tested on WPF but could be on all

**Device**
Tested on WPF but could be on all

**Describe the bug**
The MouseCoordinatesWidget shows the text 'Mouse Position' and not the mouse position.

**To Reproduce**
1. Open the MouseCoordinatesWidget sample under Widgets in WPF
2. Move the mouse over the map
3. The widget still displays 'Mouse Position'.

**Additional context**
In the code of the `MouseCoordinatesWidget.HandleWidgetMoving` the coordinates are updated but this method is never called. This seems to be caused by MapControl.HandleMoving the method is GetExtendedWidgets used which only return widgets with Touchable is true.

pauldendulk:
This other bug seems to be additional problem for the one reported here
https://github.com/Mapsui/Mapsui/issues/2208

@inforithmics Could we do with the Touchable boolean? The simplest solution would be to leave it up to the implementations. Just call them with the information if we are moving or dragging? On the other hand the boolean check will be better for performance, but if it causes enough confusion to introduce bugs it should perhaps be removed for now.

inforithmics:
Good point, I looked at the code when I introduced it and it was for fixing a Bug when the BoxWidget would capture the mouse, But because Widget and ExtendedWidgets are handled by the same code this can be removed now. 
Should I make a Pull request for that?

inforithmics:
I did a pull request, But I removed the touchable only for the IExtendedWidget because there it wasn't needed and caused a bug
But for IWidgets it is still needed. Because when a IWidget reacts to a Touchable Event I have to stop the evaluation in the Down event so that it can cancel the up Event.

pauldendulk:
At the moment there is `IWidget.HandleWidgetTouched(Navigator navigator, MPoint position);` and `IExtendedWidget.HandleWidgetTouched(Navigator navigator, MPoint position, WidgetArgs args);` which looks very similar. Is there a reason for them to exist both apart from history and not wanted to break existing code?

inforithmics:
The only reason for that is that I didn't want to break code so late into the Release. I plan to break it on the next major release.
Removing IWidget.HandleWidgetTouched(Navigator navigator, MPoint position);

pauldendulk:
[...]

Issue #2208, opened 2023-10-09
Switching between the widget samples causes hyperlink and buttons samples to become unresponsive


pauldendulk:
**Mapsui Version**
4.1.0-preview.1

**Mapsui Platform**
Tested on WPF but could be on all

**Device**
Tested on WPF but could be on all

**Describe the bug**
Switching from the Button sample to Custom Widget sample and back makes the 'click me' button become unresponsive. If you back to the Custom Widget sample and drag the map a bit and then back. It is fixed again.

**To Reproduce**
See above

**Additional context**
- Stepping through the code this it seemed the cause was a malfunction of the code in the shared MapControl pasted below. Changes in the number of widgets are not updated. This code looks like a performance optimization. Perhaps we should just iterate through all the widgets on every touch event for now. We could use `yield return` to avoid having to create a List every time. GetExtendedWidgets() seems to have the same problem.

```csharp
  private List<IWidget> GetTouchableWidgets()
  {
      if (_updateTouchableWidget != Map.Widgets.Count || _touchableWidgets == null)
      {
          _updateTouchableWidget = Map.Widgets.Count;
          _touchableWidgets = new List<IWidget>();
          var touchableWidgets = Map.GetWidgetsOfMapAndLayers().ToList();
          foreach (var widget in touchableWidgets)
          {
              if (widget is IWidgetTouchable { Touchable: false }) continue;

              _touchableWidgets.Add(widget);
          }
      }

      return _touchableWidgets;
  }
```

pauldendulk:
@inforithmics I think this is related to the changes to touch logic that you worked. I propose to remove the optimization to fix this. Is that possible or are there other complications?

inforithmics:
I think the culprit is here

` if (_updateTouchableWidget != Map.Widgets.Count || _touchableWidgets == null)`

if there is the same number of updateTouchableWidget and map widgets it won't update the touchable widgets.

But as you discovered this can be removed.
https://github.com/Mapsui/Mapsui/issues/2209#issuecomment-1753298739

pauldendulk:
Fixed with https://github.com/Mapsui/Mapsui/pull/2210
Tested on MAUI, Avalonia and Uno.WinUI.

PR #2210, opened 2023-10-09
Fix: Mouse Pointer Widget and Widget Updates after Map Change

inforithmics:
Fixes this
https://github.com/Mapsui/Mapsui/issues/2209
https://github.com/Mapsui/Mapsui/issues/2208

### DIFF
--- a/Mapsui.UI.Shared/MapControl.cs
+++ b/Mapsui.UI.Shared/MapControl.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
@@ -64,6 +65,8 @@ public partial class MapControl : INotifyPropertyChanged, IDisposable
     private readonly System.Diagnostics.Stopwatch _stopwatch = new System.Diagnostics.Stopwatch();
     // saving list of extended Widgets
     private List<IWidgetExtended>? _extendedWidgets;
+    // old widget Collection to compare if widget Collection was changed.
+    private ConcurrentQueue<IWidget>? _widgetCollection;
     // saving list of touchable Widgets
     private List<IWidget>? _touchableWidgets;
     // keeps track of the widgets count to see if i need to recalculate the extended widgets.
@@ -682,6 +685,7 @@ public partial class MapControl : INotifyPropertyChanged, IDisposable
 
     private List<IWidgetExtended> GetExtendedWidgets()
     {
+        AssureWidgets();
         if (_updateWidget != Map.Widgets.Count || _extendedWidgets == null)
         {
             _updateWidget = Map.Widgets.Count;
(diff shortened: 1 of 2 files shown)


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -699,8 +703,20 @@ private List<IWidgetExtended> GetExtendedWidgets()
         return _extendedWidgets;
     }
 
+    private void AssureWidgets()
+    {
+        if (_widgetCollection != Map.Widgets)
```
inforithmics, line 708: If the widget Collection changes set the cached Widgets (touchable and extended to null) to recreate them. The widgetcollection changes normally when a new map is assigned to the control

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -682,14 +685,15 @@ private bool HandleTouched(MPoint position, bool leftButton, int clickCount, boo
...
     {
+        AssureWidgets();
         if (_updateWidget != Map.Widgets.Count || _extendedWidgets == null)
         {
             _updateWidget = Map.Widgets.Count;
             _extendedWidgets = new List<IWidgetExtended>();
             var widgetsOfMapAndLayers = Map.GetWidgetsOfMapAndLayers().ToList();
             foreach (var widget in widgetsOfMapAndLayers)
             {
-                if (widget is IWidgetExtended extendedWidget && extendedWidget.Touchable)
```
inforithmics, line 692: Extended Widget do not need to handle touchable this is only for IWidget Widgets because they can only handle the Up event the Down event has to be marked as handled for IWidget widgets.

Mapsui/Widgets/IWidgetExtended.cs
```diff
@@ -1,5 +1,5 @@
 ﻿namespace Mapsui.Widgets;
-public interface IWidgetExtended : IWidget, IWidgetTouchable
```
inforithmics, line 2: Remove the IWidgetTouchable because an IWidgetExtended handles all Widgets anyway

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -64,6 +65,8 @@ public partial class MapControl : INotifyPropertyChanged, IDisposable
     private readonly System.Diagnostics.Stopwatch _stopwatch = new System.Diagnostics.Stopwatch();
     // saving list of extended Widgets
     private List<IWidgetExtended>? _extendedWidgets;
+    // old widget Collection to compare if widget Collection was changed.
+    private ConcurrentQueue<IWidget>? _widgetCollection;
```
pauldendulk, line 69: One general remark about the ConcurrentQueue. If we need a thread save collection one of the things we need to ask ourselves is why is it needed. Not everything needs to be accessed from multiple threads and if we can make cross thread access impossible with our design this a better solution.

pauldendulk (approved):
Muchas gracias!

PR #2216, opened 2023-10-13
Explicitly ignore Avalonia samples to build as nugets

pauldendulk:
The test build showed that nuget packages were generated for the Avalonia samples. This should not happen.

The current configuration for packing is that the Directory.Build.props has it set to false and all projects that need to be packed have IsPackable set to true.

For the Avalonia samples this system apparently failed because it has it's own Directory.Build.props. So, this fix also set IsPackage to false in that file.


pauldendulk:
I tested this by running the pack command locally.

PR #2219, opened 2023-10-13
Use widgets in animation samples

pauldendulk:
Update the six viewport animation samples. Two get buttons to start the animation (rotation, and plain zoom), the other four get an instruction label.

I also remove the xaml mouse coordinate box in wpf because we should do that with a widget now.

There are only changes in the samples in this PR.


pauldendulk:
Buttons in Zoom Animation Sample

image

Instruction in Zoom Around Locations Sample

image

PR #2222, opened 2023-10-13
Use TextBox in InfoLayersSample and remove platform specific code to show info

pauldendulk:
To show the data of a feature when using the Info event we had platform specific code. Not on all platforms, and with differences between platforms. This PR adds a TextBox in the InfoLayersSample so, that all the platform specific code could be removed.

I will test the changes platforms to check if I did not remove to much, or broke something otherwise.

The PR only affects the samples.


pauldendulk:
I work on this because:
- We released a  stable version, I don't want to change to much in that version right, so do some things around the core.
- By working on the samples I dog feed Mapsui and this way I have found 4 (minor) bugs last week.
- I want to remove platform specific code and replace it with generic code to make the code easier to maintain.
- By going through the samples it set my mind in motion thinking about a better design, important for the v5.

pauldendulk:
This PR removes the map info field in the platform specific UI and adds it to the InfoLayerSample, but there are other samples that use info as well. What to do with those? Below is a list of search for IsInfoLayer = true. I went through it, not all are important. I could add sample specific info for a few, but for debugging and testing it is handy to have for all. Especially for solutions where we use intermediates, like TiledRasterizingLayer. An alternative I am now thinking about is to add a default feature info textbox on the shared MapControl, which could be turned on and off with a boolean. For users this would be useful too during development. But when you start with that it becomes logical to think about other widgets as well, like the zoomInOutButtons, and there are too many choices to make in that regard, which I don't want to make for v4. Perhaps a better option would be to add a MapInfoTextBox which should be added explicitly but has more off the shelf functionality, so it becomes easy to add.

image

pauldendulk:
After a bit of thinking I prefer the last option the most. User would have to explicitly add it, but the ShowMapInfoInMap option would probably be off by default, so it would also need:
```map.ShowMapInfoInMap = true'```
Instead of:
```map.Widgest.Add(new MapInfoWidget());```
This is not a real advantage so, I prefer to keep the MapControl empty by default. And as mentioned elsewhere, we could provide a builder for some common off the shelf configurations.

**Comment by pauldendulk** [comment]
About building the MapInfoWidget. It would need to react to events all over the map so would have to be a transparent map sized widget, with the TextBox as a widget within the widget.

File: Samples/Mapsui.Samples.Common/Maps/Demo/InfoLayersSample.cs [hunk 1]
```diff
@@ -33,6 +35,18 @@ public static Map CreateMap()
         map.Layers.Add(new WritableLayer());
         map.Layers.Add(CreateLineLayer());
 
+        var textBox = CreateTextBox();
+        map.Widgets.Add(textBox);
```
inforithmics, line 39: Good idea to centralize the display of the mapinfo reduces the platform specific sample code and tests mapsui widgets.

Samples/Mapsui.Samples.Common/Maps/Demo/InfoLayersSample.cs
```diff
@@ -33,6 +35,18 @@ public static Map CreateMap()
...
+
+        map.Info += (s, a) =>
+        {
+            if (a.MapInfo?.Feature is null) 
+                textBox.Text = "";
+            else
+                textBox.Text = $"Feature Info - {a.MapInfo.Feature.ToDisplayText()}";
+            map.RefreshGraphics();
```
inforithmics, line 47: With this pull request changes on the widgets trigger automatically a redraw


https://github.com/Mapsui/Mapsui/pull/2183

inforithmics (approved):
Nice improvement on maintainability

inforithmics:
I too would keep the MapControl slim. But I think there is a need for something like MapView where there is a default project set and utility layers and widgets per default instantiated. Could have a different name than mapview.

So on MapView there could be a showmapinfo property and on mapcontrol you add the mapinfowidget.

PR #2225, opened 2023-10-16
Add sample for custom SVG color

pauldendulk:
This sample involves quite a lot of complicated code but I think this is important functionality.

### What does it do?
It takes an SVG and draws it with different colors, but unlike the BlendColorMode used in the DynamicSvgStyleSample it does not draw the entire SVG in one color, but only the fill color. 

### Why is this important?
Thematic coloring is important for many purposes. Mapsui can use the BlendColorMode but this paints the entire symbol in one color. This is not pretty, and when many symbols are clustered together it becomes ugly. The custom SVG technique allows us to preserve the outline, and could also be used for other custom coloring.

image

### How does it work?
We load the SVG with the Svg.Skia library, and the SvgDocument of Svg.Skia is used to manipulate the SVG. We do this by iterating through the document tree, and changing specific elements.

### What should be improved
Svg.Skia should be abstracted away. We should not have to know about SvgDocument. A SVG should be loaded as string or stream and it should be possible to specify which fill or stroke should get which specific color.

image

### DIFF
--- a/Samples/Mapsui.Samples.Common/Maps/Demo/CustomSvgColorSample.cs
+++ b/Samples/Mapsui.Samples.Common/Maps/Demo/CustomSvgColorSample.cs
@@ -1,0 +1,155 @@
+﻿using Mapsui.Layers;
+using Mapsui.Samples.Common.DataBuilders;
+using Mapsui.Styles.Thematics;
+using Mapsui.Tiling;
+using Mapsui.Utilities;
+using Svg.Model;
+using Svg.Skia;
+using Svg;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using SkiaSharp;
+using Mapsui.Styles;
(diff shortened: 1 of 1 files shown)


inforithmics (approved):
Looks good for me.

inforithmics:
Maybe you could upgrade svg Skia which removes the Fizzler dependency

inforithmics:
The coloring can be done with css but for that something like this pull request is needed.

https://github.com/svg-net/SVG/pull/1094

pauldendulk:
> Maybe you could upgrade svg Skia to 1.0.0.3 which removes the Fizzler dependency

I will in a separate PR. I like the fact that this PR only changes samples so it is easy to see this can not introduce a bug.

pauldendulk:
> The coloring can be done with css but for that something like this pull request is needed.
> https://github.com/svg-net/SVG/pull/1094

That would be great.

PR #2231, opened 2023-10-21
Update README.md

saakshii12:
Changed gettings started to Quickstart guide


pauldendulk (commented):
Ah, thanks! we should be consistent in our naming of such things. Perhaps you could also change the 'getting started' in the samples section. You could also try to search for other use of getting started on https://mapsui.com/.

pauldendulk:
> Ah, thanks! we should be consistent in our naming of such things. Perhaps you could also change the 'getting started' in the samples section. You could also try to search for other use of getting started on https://mapsui.com/.

I updated it in the open documentation pr instead https://github.com/Mapsui/Mapsui/pull/2232

PR #2232, opened 2023-10-22
Reorganize documentation regarding contributor guidelines and design choices

pauldendulk:


pauldendulk:
We had a separate page on development constraints and renderers, I've put them now all in contributor guidelines. Part of what is in there could be on a separate page, but since I did not see a really good solution I've put it all together ordered it from more practical to more about design choices. For comparison, the avalonia guidelines is much more about practical things   https://github.com/AvaloniaUI/Avalonia/blob/master/CONTRIBUTING.md

PR #2241, opened 2023-11-04
Geo tiff performance to 4.1

pauldendulk:
I created this new pull request to push this into develop/4.1. The existing pull request is merging into master.


inforithmics:
Should I target Future Pull Requests to 4.1 Branch?

pauldendulk:
Yes, I think that would be better for small MRs like the geotiff and callout-on-wasm. If those are merged to develop/4.1 we could merge develop/4.1 into master too. It is not possible to do it the other way around (like I was trying here) because the newer features in master would come along into develop/4.1. At some point master will become sufficiently different that this won't work either, I guess we then really need to create to separate PRs. .... but alone if it sufficiently important. I am currently getting into the mind set of going fast forward to v5.

PR #2244, opened 2023-11-05
Remove DEBUG code in benchmarks which did not compile

pauldendulk:
A tiny MR

### DIFF
--- a/Benchmarks/Mapsui.Rendering.Benchmarks/Program.cs
+++ b/Benchmarks/Mapsui.Rendering.Benchmarks/Program.cs
@@ -1,15 +1,9 @@
 ﻿using BenchmarkDotNet.Running;
-using Mapsui.Rendering.Benchmarks;
 
 public class Program
 {
     public static void Main(string[] args)
     {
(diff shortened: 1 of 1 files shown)


Benchmarks/Mapsui.Rendering.Benchmarks/Program.cs
```diff
@@ -1,15 +1,9 @@
 ﻿using BenchmarkDotNet.Running;
-using Mapsui.Rendering.Benchmarks;
 
 public class Program
 {
     public static void Main(string[] args)
     {
-#if DEBUG
```
inforithmics, line 8: Can be removed I used this to Debug some Performance/Memory Problems

inforithmics (approved):

Benchmarks/Mapsui.Rendering.Benchmarks/Program.cs
pauldendulk: It did not compile in it's current form, so something else must have changed to break it.

PR #2245, opened 2023-11-05
Remove AsyncLock

pauldendulk:
Some thoughts.
- This is old code. Originally there was a lock on the provider, added long ago. 
- Later the AsyncLock was added because the method was made async.
- The FeatureFetcher is the only place where the AsyncLock is used.
- The FeatureFetcher itself is only called in the ImageLayer. ImageFetcher would have been a more appropriate name.
- It is not really clear what the purpose of the lock was. To limit the number of simultaneous request seems a logical answer but such lock would result in queing of the various requests, in the context of Mapsui having a queue of items to fetch often does not make sense. This is because the map viewport changes quickly through panning and scrolling and data needed in a previous step is no longer relevant in the next. It is often better to replace an existing pending request. The best solution for Mapsui would be a central fetcher. We have an item on the backlog for this somewhere.
- The timer in the ImageLayer would prevent simultaneous requests from one ImageLayer. So, the lock would not do anything in that scenario. With multiple ImageLayers it could prevent simultaneous requests. It is not clear if this was ever a real problem.


Mapsui/Fetcher/FeatureFetcher.cs
```diff
@@ -30,10 +25,7 @@ public FeatureFetcher(FetchInfo fetchInfo, IProvider provider, DataArrivedDelega
 
     public async Task FetchOnThreadAsync()
     {
-        using (await _providerLock.LockAsync())
```
inforithmics, line 33: I think this was useful before the Timer Triggered Map Referesh. Now it is just Cargo Programming. (Code without useful functionality).
It could only go into the lock if the request takes longer than 1 second. But then it could hinder Performance because first it waits for the previous request to finish causing to queue all requests. The beauty of async is that several requests could be executed in parallel. And I think the Provider should be Thread Safe anyway.

Mapsui/Utilities/AsyncLock.cs
inforithmics, line 1: If a good Implementation is used DotNext.Threading could be used there is an AsyncLock Implmentation.

DotNext nuget packages are a Testing Ground for new .NET Framework Classes. That might be integrated into the .NET Framework.

inforithmics (approved):

PR #2246, opened 2023-11-05
Add github actions to sln folder

pauldendulk:
Tiny MR but this had bothered me for a while


pauldendulk:
image

inforithmics (approved):
Visual Studio some times automatically adds the GitHub actions into a custom GitHub Action folder. But sometimes it appears and sometimes not I haven't found a rule out when it appears or not so good improvement

pauldendulk:
> Visual Studio some times automatically adds the GitHub actions into a custom GitHub Action folder. But sometimes it appears and sometimes not, I haven't found a rule out when it appears or not so good improvement

Yes, the default folder was gone on my machine and I decided it was better not to depend on it.

PR #2247, opened 2023-11-05
Remove Avalonia-V0

pauldendulk:


inforithmics (approved):
Nice one target less (and Avalonia.V0 wasn't used so much).

Question: Do want to keep Avalonia.V0 nuget package in the Readme.md file or not?

pauldendulk:
> Question: Do want to keep Avalonia.V0 nuget package in the Readme.md file or not?

Good question, when doing the search voor remaining Avalonia.V0 references I wondered about this too. It is still mentioned in the readme and in the docs. What I was thinking:
- v4 is the current released version, so that is now our main thing, although we as developers love to think about the next version for our user this is just **the** version. So we should just promote what we have right now. 
- Documentation should be separately maintained for each main version (or better, for minor versions). In our setup the docs are part of the branch, which makes this possible, good. We should also deploy those separately. So we should have different folders for each version. https://mapsui.com/documentation/v4/home.html and https://mapsui.com/documentation/v5/home.html. The default should be our current released v4. This means we can remove it from our documentation.
- But then the readme is part of the master branch. It is one of the most visible things of the project. So, I guess we have to keep it there anyway. A possible solution would be to move the nuget references to a page in the documentation, but perhaps it will not be as visible. Also, there has to be some reference to the current version in the master readme. So, I think we can not work around it completely.

inforithmics:
I agree we should keep it in the Readme.md When a new Mapsui Version is Released. the User sees that Mapsui.AvaloniaV0 is still on Version 4.0

So when V5 or Beta is released we could may seperate V4 UI nugets from V5 nugets.

The maintaining of documentation by major version is a good idea.

PR #2248, opened 2023-11-05
Remove Uno UWP

pauldendulk:


inforithmics (changes requested):
Mapsui.Uno.slnf shouldn't be deleted because it contains Uno and Uno.WinUI Sample projects. Only the Uno Samples and Mapsui.Uno should be deleted

inforithmics (approved):

PR #2249, opened 2023-11-05
Run dotnet format and add to github action

pauldendulk:


inforithmics:
Good idea to check the Formatting in the Github Action so every pull request gets first checked before it is merged.

Issue #2275, opened 2023-11-12
CenterOnAndZoomTo Not Working Unless Delayed


nm4568:
**Mapsui Version**
Mapsui Version v4.1.1
Mapsui.ArcGis v4.1.1

**Mapsui Platform**
.NET MAUI

**Device**
iPhone 14 Pro Max

**Describe the bug** 
CenterOnZoom only works when delayed.

**To Reproduce**
1) Launch Sample App with MainPage provided.

**Expected Result** - CenterOnAndZoomTo should work within Map_Loaded event handler of MainPage.xaml.cs

**Actual Result** - CenterOnAndZoomTo does not work - only works when task is delayed by adjusting the ZOOM_DELAY constant at the top of MainPage.xaml.cs.  Change constant from 0 to 5000 for a workaround.

**Additional context**
Sample Repo - https://github.com/nm4568/MapsuiSamples/tree/bug/zoom-issue

pauldendulk:
Thanks a lot for the sample project! This makes things so much easier, and I learn from seeing your code. I like the image layer you use.

In Mapsui changing the viewport depends on the size of the canvas and the extent of the data and this not known during startup. This is why the Map.Home method was added. This method is called ones those are known. I could fix your problem by writing it like this:

```csharp
var layer = await CreateLayerAsync(default);
mapControl.Map?.Layers.Add(layer);

double x = -9188151.36056;
double y = 3235144.74039;
var point = new MPoint(x, y);
mapControl.Map.Home = (n) => n.CenterOnAndZoomTo(point, resolution: 1000, 500, Mapsui.Animations.Easing.CubicOut);

MyMapControl = mapControl;
Content = mapControl;
```

It now starts like this:

![image](https://github.com/Mapsui/Mapsui/assets/963462/f8107890-da01-4118-b916-6bea8d451cac)

###Notes:
- I tested on Windows, not on Mac or iOS, but I assume the fix also works there.
- Perhaps we should add the to the FAQ.
- I expect the Mapsui logger to log a warning when CenterOnAndZoom is called too early. I did not see that warning.
- You do not always need to the canvas size, and in this particular case (where resolution and center are set). I think you don't need it. Perhaps we can improve this.

pauldendulk:
Thanks a lot for the sample project! This makes things so much easier, and I learn from seeing your code. I like the image layer you use.

[...]

Issue #2115, opened 2023-07-16
Center map doesn't work after a WMS map was loaded


gpsaliola:
**Mapsui Version**
4.0.0

**Mapsui Platform**
Xamarin - Android

**Device**
Samsung A53

**Describe the bug**
Centering the map doesn't work after a WMS map was loaded.
Some WMS maps works but others fails centering the map.

**To Reproduce**
This is the code I've used to center the map.

```
                MPoint point = new Position(42, 12).ToMapsui();
                mapView.Map.Home = n =>
                {
                    n.CenterOn(point); 
                };
```

**Expected behavior**
Map should be centered on provided coordinates.

gpsaliola:
I've found the problem but seems an issue.
If the map is XYZ this statement works but it doesn't work for some WMS maps:

```
                MPoint point = new Position(lat, lon).ToMapsui();
                mapView.Map.Home = n =>
                {
                    n.CenterOn(point); 
                };
```
If the map is WMS this statements works but doesn't works for all XYZ maps:

```
                MPoint point = new Position(lat, lon).ToMapsui();
                mapView.Map.Navigator.CenterOnAndZoomTo(point, resolutions[AppGeneral.cZoomFind - 1]);

```
This is in contradiction with what is written in #2081

pauldendulk:
@gpsaliola What do you mean with 'map is XYZ'?

gpsaliola:
Yes you have reason. I meant tiled map (HttpTileSource).

pauldendulk:
Do you have some sample code to build the map so I could reproduce it?

gpsaliola:
> @gpsaliolaCosa intendi con "la mappa è XYZ"?

Tiled map (x, y, z in the url).

pauldendulk:
What are the coordinates of the lan lon position you zoom into?

gpsaliola:
This is an example code:

```
        public static double[] cResolutions = {
            156543.03392804103,
            78271.5169640205,
            39135.75848201024,
            19567.879241005125,
            9783.939620502562,
            4891.969810251281,
            2445.9849051256406,
            1222.9924525628203,
            611.4962262814101,
            305.74811314070485,
            152.87405657035254,
            76.43702828517625,
            38.218514142588134,
            19.109257071294063,
            9.554628535647034,
            4.777314267823517,
            2.3886571339117584,
            1.1943285669558792,
            0.5971642834779396,
            0.29858214173896974,
            0.14929107086948493,
[...]

PR #2281, opened 2023-11-13
Remove Map.Home

charlenni:
Remove Map.Home

I think, it is not very clear for users to set a Map.Home function (e.g. see #2275), that is called after initialization and not the normal functions. It needs much of explanation and new user do it often wrong.

What I did is, to save all the calls of any Navigator calls while Viewport has no size. This functions then called in the same order when Viewport gets a size.

All tests work beside the native Navigator tests for Pinch. They are designed to work without Viewport set to a size. I assume, that there must be a SetSize before and then the values are different. Any ideas why they work before?


pauldendulk:
This is clever! My initial thought was, 'but this is not possible!'. The solution you came up with had not occurred to me.

I will need to take a closer look though. It will have to work for all scenarios,  and it should not lead to confusion in other scenarios. 

I see you took into account Pan/ZoomLock. Was there a specific scenario where you noticed this was a problem? So, you store the Pan/ZoomLock as local state in a method. Could you explain a bit more what the logic is now? 

How about RotationLock?

How will it work for the limiter? Should we also capture the current limiter? Perhaps it is not really necessary, limiting is usually not something that changes.

Is there any other state like lock or limiter?

Perhaps we would want to put all these constraints in one readonly record, like the Viewport.

You mention the Home method is called when the size is known, but there is also the extent of the data to consider. ZoomToBox depends on this. This aspect is even more complicated because different layers can be loaded at different moments, and you don't know when the last one is loaded. I suspect this does not even work correctly in the current implementation. If we go to this alternative route we should also have a solution for that. 

When I first worked Home this my idea was that both the size and the data extent should be known, but because the data extent is even more problematic (because of the different timing of different layers) it may be better to see it as two different problems. 

The idea of Home was that users could call Home later on, to return to the default extent. This now disappears. Instead user have explicitly call a navigator method. Perhaps this is better. It fits more in the 'less magic' approach that I like.

I will look at the failing test.

charlenni:
> This is clever! My initial thought was, 'but this is not possible!'. The solution you came up with had not occurred to me.
> 
> I will need to take a closer look though. It will have to work for all scenarios, and it should not lead to confusion in other scenarios.

Take your time.

> I see you took into account Pan/ZoomLock. Was there a specific scenario where you noticed this was a problem? So, you store the Pan/ZoomLock as local state in a method. Could you explain a bit more what the logic is now?

It occurred as problem with a failing test, where first is ZoomToPanBounds and after that a PanLock = true is set. If you later call the ZoomToPanBounds, the PanLock == true returns without doing nothing. So I saved the state of the flags at the moment you call the ZoomToPanBounds.

> How about RotationLock?

You are right, missed this. There was no test, that test RotationTo.

> How will it work for the limiter? Should we also capture the current limiter? Perhaps it is not really necessary, limiting is usually not something that changes.

The limiter isn't used in any of these functions. It is used to set the viewport, but not, e.g. in CenterTo(). So I see no use case, where this could be a problem.

> Is there any other state like lock or limiter?
> 
> Perhaps we would want to put all these constraints in one readonly record, like the Viewport.

Perhaps this is overdone, but possible, if you like it in this way.
 
> You mention the Home method is called when the size is known, but there is also the extent of the data to consider. ZoomToBox depends on this. This aspect is even more complicated because different layers can be loaded at different moments, and you don't know when the last one is loaded. I suspect this does not even work correctly in the current implementation. If we go to this alternative route we should also have a solution for that.

But what is the work flow? The user create all the layers and then goes to a given point. What does it mean, when someone creates a layer, goes to a special point, creates the next layer and goes to another point? And that all in the creator of the app.

> When I first worked Home this my idea was that both the size and the data extent should be known, but because the data extent is even more problematic (because of the different timing of different layers) it may be better to see it as two different problems.
> 
> The idea of Home was that users could call Home later on, to return to the default extent. This now disappears. Instead user have explicitly call a navigator method. Perhaps this is better. It fits more in the 'less magic' approach that I like.

Better to use the normal functions you use later too, then use one special function while init the map. For me, it is much clearer. And it seems, that the users like to have it in this way, because they often ask and don't find Home by themselves.

> I will look at the failing test.

Thank you.

pauldendulk:
> The limiter isn't used in any of these functions. It is used to set the viewport, but not, e.g. in CenterTo(). So I see no use case, where this could be a problem.

As a consequence of CenterTo a new viewport will be set. Every time the viewport changes the limiter is called.  So, changing the limiter before or after a call could make a difference. 

A further complication is that the limiter depends on the PanBounds and the PanBound depend on the map extent.

> Perhaps this is overdone, but possible, if you like it in this way.

This is not something for now, I was considering the option, please ignore it.

> But what is the work flow? The user create all the layers and then goes to a given point. What does it mean, when someone creates a layer, goes to a special point, creates the next layer and goes to another point? And that all in the creator of the app.

What I am trying to say is that there is a problem in the current implementation and I wonder how this would workout in the new implementation. I will explain the problem in a code comment in this PR.

> Better to use the normal functions you use later too, then use one special function while init the map. For me, it is much clearer. And it seems, that the users like to have it in this way, because they often ask and don't find Home by themselves

I agree.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -411,23 +411,22 @@ private void Map_PropertyChanged(object? sender, PropertyChangedEventArgs e)
...
             Refresh();
         }
     }
     // ReSharper restore RedundantNameQualifier
 
-    public void CallHomeIfNeeded()
+    public void CallInitializedIfNeeded()
     {
-        if (!Map.HomeIsCalledOnce && Map.Navigator.Viewport.HasSize() && Map?.Extent is not null)
+        if (!Map.Navigator.IsInitialized && Map.Navigator.Viewport.HasSize() && Map?.Extent is not null)
```
pauldendulk, line 427: In this line of the code there is a check if Map?Extent is not null. The problem is that we have several layers. If just one layer has an extent which is not null the Map?.Extent will not be null. To first layer to have an extent is not known. It could even be different from run to run, with a different startup extent as a consequence. 

This problem existed before and still exists. I have to think a bit about how it works out in the new setup, and what an eventual solution would look like.

pauldendulk: This problem may be related https://github.com/Mapsui/Mapsui/issues/2115

pauldendulk:
I looked at the unit test. In the new setup you need a size or OnCenter will have no effect. Also, it makes more sense to set the size. Settings the size in the old setup has the same results as in the new setup, so I suggest to just update the values in this PR

```
  [TestCase(0.5, 40, -10)]
  [TestCase(1, 20, 10)]
  [TestCase(2, -20, 50)]
```

pauldendulk:
I have an idea how to solve the Map.Extent problem. The initialization should work with PanBounds instead of the Map.Extent. This also makes the code more logical because inside the Navigator we only work with PanBounds not with Map.Extent.

This is how the PanBounds are defined:
`public MRect? PanBounds => OverridePanBounds ?? DefaultPanBounds;`
So, the user can always override it by setting OverridePanBounds, that is one advantage. The DefaultPanBounds are now directly set each time the Map.Extent changes. Instead it should be set by the extent of one layer, in that case the problem is solved. I think this should be the bottom layer. If it needs to be determined by the extent of a number of layers every individual layers needs to have an extent before the DefaultPanBounds is assigned.

This is not something to fix in this PR but we can start by checking null on the PanBounds instead of the Map.Extent.

Benchmarks/Mapsui.Rendering.Benchmarks/RenderToBitmapPerformance.cs
```diff
@@ -100,7 +100,7 @@ private static Map CreateMap(RenderFormat? renderFormat = null)
         map.Layers.Add(layer);
 
         var extent = map.Layers[0].Extent!;
-        map.Home = n => n.ZoomToBox(extent);
+        map.Navigator.ZoomToBox(extent);
```
pauldendulk, line 103: Much better to see a regular call in all these places. Reads like normal code.

Mapsui/Navigator.cs
```diff
@@ -15,7 +16,13 @@ public class Navigator
 {
     private Viewport _viewport = new(0, 0, 1, 0, 0, 0);
     private IEnumerable<AnimationEntry<Viewport>> _animations = Enumerable.Empty<AnimationEntry<Viewport>>();
-    
+    private List<Action> _initialization;
+
+    public Navigator()
+    {
+        _initialization = new List<Action>();
```
pauldendulk, line 23: Does it need to be a list or could you also store the most recent call? Not sure if there are cases where you need the list.

Mapsui/Navigator.cs
```diff
@@ -79,6 +86,29 @@ private set
         }
     }
 
+    public bool IsInitialized { get; private set; } = false;
+
+    public void Initialization()
+    {
+        if (_initialization is not null && !IsInitialized)
```
pauldendulk, line 93: We work with nullable enabled. That means that if the field is not nullable (which it is in this case) you should not check. Assigning it becomes mandatory.

Mapsui/Navigator.cs
```diff
@@ -15,7 +16,13 @@ public class Navigator
 {
     private Viewport _viewport = new(0, 0, 1, 0, 0, 0);
     private IEnumerable<AnimationEntry<Viewport>> _animations = Enumerable.Empty<AnimationEntry<Viewport>>();
-    
+    private List<Action> _initialization;
```
pauldendulk, line 19: It is simpler to initialize it in the declaration. You do not need to add the constructor.

Mapsui/Navigator.cs
```diff
@@ -114,7 +142,12 @@ public void MouseWheelZoom(int mouseWheelDelta, MPoint centerOfZoom)
     /// <param name="easing">The type of easing function used to transform from begin tot end state</param>
     public void ZoomToBox(MRect? box, MBoxFit boxFit = MBoxFit.Fit, long duration = -1, Easing? easing = default)
     {
-        if (!Viewport.HasSize()) return;
+        if (!Viewport.HasSize())
```
pauldendulk, line 145: As mentioned in the thread we need to check on the PanBounds as well, otherwise ZoomToPanBounds could fail and the limiter does not work. Perhaps !IsInitialized is the most appropriate check.

Mapsui/Navigator.cs
```diff
@@ -551,6 +619,34 @@ public void SetViewport(Viewport viewport, long duration = -1, Easing? easing =
...
     }
 
+    /// <summary>
+    /// Add a call to a function called before initialization of Viewport to a list
+    /// </summary>
+    /// <param name="action">Function called before initialization</param>
+    private void AddToInitialization(Action action)
+    {
+        // Save state when this function is originally called
+        var panLock = PanLock;
```
pauldendulk, line 629: Nice simple way to store the locks.

pauldendulk (commented):
I added some comments but also made a PR to your PR. I will put some comments there too.

Mapsui/Navigator.cs
charlenni: I think, it needs to be a List. If not, then calls like CenterTo() with a following ZoomTo() could be lost.

Mapsui/Navigator.cs
charlenni: You are right. Still starting again ;-)

Mapsui/Navigator.cs
charlenni: That's right. I was adding the default behavior at the beginning of the list. But that isn't good. So I checked, if the List is empty and, if yes, call the default behavior. Then I forgot to remove the constructor.

Mapsui/Navigator.cs
charlenni: But in your example above with the Map.Home you also don't check/save the PanBounds. At the moment you set the Home function, you use the old extent, even if it had changed in between.

charlenni:
It didn't run anymore. The problem is, that you replaced the check to Viewport.HasSize() with IsInitialized, which is at this moment false. So it adds another entry to the _initialization and then crashes.

pauldendulk:
I probably broke the tests again with the check on PanBounds. I will try to look into this tonight.

charlenni:
No problem, I solved it already. I set IsInitialized to true before I call all the stored functions. Then it works again.

pauldendulk:
> Perhaps I didn't understand it right, but I thought the problem here is, that Viewport gets its size very late (when the MapControl is added to the screen). So the problem has for me nothing to do with the content of the map. I'm wrong?

There are two things to take into account. The canvas size and the PanBounds. In the initialize method we call ZoomToPanBounds. That means you need to know the PanBounds and this is often determined by the data source. For instance for a WMS you only know the capabilities url at the start. When initializing it will fetch the capabilities and read the extent from it. 

For the limiter to work (which is called on any change of the Viewport) you also need to know the PanBounds. Although perhaps it would also be possible to apply limiting afterwards, once the PanBounds it known. Not sure if that works correctly for all cases. It would be nice if we could thread the two aspects as two separate problems.

Mapsui/Navigator.cs
pauldendulk: ah yes!

Mapsui/Navigator.cs
pauldendulk: I don't know exactly what you are referring to, but PanBounds is updated when things change. It is not very easy to see how the notifications flow, but if the extent of a layer changes then the extent of the map changes which will result in a change in the PanBounds.

charlenni:
Ok, then let’s save the PanBounds in the same way then PanLock/ZoomLock/RotationLock. Do you think, that this should remove the problem?

pauldendulk:
> Ok, then let’s save the PanBounds in the same way then PanLock/ZoomLock/RotationLock. Do you think, that this should remove the problem?

I don't think it is necessary to save the PanBounds.

pauldendulk:
Sorry but I have caused conflicts to this PR by merging some big changes.

charlenni:
Solved them

pauldendulk:
I now realize I changed the results of the unit test again by setting the initial bounding box. This is what the results should be:
```
  [TestCase(0.5, 70, -40)]
  [TestCase(1, 30, 0)]
  [TestCase(2, -50, 80)]
```

pauldendulk:
@charlenni I made another PR to this PR to resolve the conflicts and fix the test. With that I would like to merge this PR.

There is still a problem with initializing the extent (not introduces in this PR). Currently the initialize method is called after both Size and PanBounds are known. PanBounds is set as soon as the Map.Extent is not null, but it could be initialized by any layer. The first layer that is initialized will trigger the Initialize method. So, depending on the time it takes a layer is initialized the PanBounds could be different. 

Also if PanBounds is not null it could become null again later on (for instance when the app user removes the current layer to replace them with others) which further complicates things. 

I think ZoomToPanBounds should be removed from the navigator. Currently ZoomToPanBounds is called from the Navigator when Size and PanBounds are known. This functionality should move out of the Navigator to the Map. The implementation should have various options to trigger ZoomToPanBounds. Perhaps only if all layers are initialized, or only if the base layer is initialized, or with a custom function.

pauldendulk:
@charlenni Could you look at the PR I created to your branch? I doing cleanups all over the code and will introduce conflicts again easily.

charlenni:
I hope, it is now up to date ;-)

Mapsui/Navigator.cs
```diff
@@ -14,6 +14,13 @@ public class Navigator
     private Viewport _viewport = new(0, 0, 1, 0, 0, 0);
     private IEnumerable<AnimationEntry<Viewport>> _animations = Enumerable.Empty<AnimationEntry<Viewport>>();
 
+<<<<<<<<< Temporary merge branch 1
```
pauldendulk, line 17: Something went wrong when merging. Perhaps just this line needs to be deleted.

pauldendulk (approved):
Thanks a lot. This will make it a lot easier for the users. The initialization of the map after the extent becomes known remains problematic but this was an already existing problem which should be dealt with later.

Issue #2288, opened 2023-11-16
Drawing a polygon with one hole


gpsaliola:
Mapsui Version: 4.1.1
Device: Android, iOS

I'm trying to draw a polygon (in this case a circle) with a hole (in this case a smaller circle).
I would like to fill the polygon with a color and a pattern but leave the hole transparent.
What I get is that the hole is filled with the same color as the polygon.
This is the code used:

```
		VectorStyle lineStyle = new VectorStyle
		{
			Fill =
			{
				Color = Mapsui.Styles.Color.Red,
				FillStyle = FillStyle.DiagonalCross
			},
			Outline =
			{
				Color = Mapsui.Styles.Color.Red,
				Width = 3.0,
				PenStyle = PenStyle.Solid
			}
		};

		List<GeometryFeature> featureLinesList = new List<GeometryFeature>();
		List<Coordinate> lineMin = new List<Coordinate>();
		List<Coordinate> lineMax = new List<Coordinate>();

		for (int step = 0; step < 360; step += 1)
		{
			(double lon, double lat) = GetCirclePoint(12.1, 42.1, step, 30000);
			(double pointX, double pointY) = SphericalMercator.FromLonLat(lon, lat);
			lineMin.Add(new Coordinate(pointX, pointY));
		}

		for (int step = 0; step < 360; step += 1)
		{
			(double lon, double lat) = GetCirclePoint(12.1, 42.1, step, 100000);
			(double pointX, double pointY) = SphericalMercator.FromLonLat(lon, lat);
			lineMax.Add(new Coordinate(pointX, pointY));
		}

		lineMin.Add(lineMin[0]);
		lineMax.Add(lineMax[0]);

		LinearRing ringMin = new LinearRing(lineMin.ToArray());
		LinearRing ringMax = new LinearRing(lineMax.ToArray());
		LinearRing[] holes = new LinearRing[] { ringMin };

		GeometryFeature lineFeature = new GeometryFeature()
		{
			Geometry = new Polygon(ringMax, holes, GeometryFactory.Default),
			Styles = { lineStyle }
		};
		featureLinesList.Add(lineFeature);
		lineFeature.Dispose();

		MemoryLayer memoryLayer = new MemoryLayer
		{
			Features = featureLinesList,
			Name = "polygon",
			Style = null
		};
		mapView.Map.Layers.Add(memoryLayer);

		...
				
[...]

PR #2289, opened 2023-11-16
Added check for CCW of outer ring and CW for inner rings

charlenni:
Checks, if the rings of a polygon have the correct direction and, if not, turn them.


charlenni:
Is all about #2288

PR #2290, opened 2023-11-16
Added check for CCW of outer ring and CW for inner rings

charlenni:
This is regarding #2288.

Checks don't run, even if the resulting image looks totally the same. Could be, that there are problems, because the not solid line around, which could be slightly different from the original one.


charlenni:
Test don't run, because of the dashed line around. Now the direction is vice versa, so the line looks different from the original test. So it should be updated.

pauldendulk (approved):
Great fix.

You also reverse the exterior ring to how it is supposed to be, good. I just wondered if this was really required because if it was then I think we would have found out earlier, but I gather from you comment '(if this is the other way around it also seems to work)', I gather that it is not. Weird the inner and outer rings are different in skia.

charlenni:
Skia needs this to distinguish between outer and inner ring/holes. It seems, that it is only important to have different orientation for holes than for the outer ring.

pauldendulk:
This fix is also needed for master btw

Issue #2253, opened 2023-11-06
IntArrayExtension.IsNaturalOrder returns wrong result in .NET-Standard 2.0


der1Mac:
**Mapsui Version**
4.1.1

**Mapsui Platform**
WPF, .Net 4.8

**Device**
Windows

**Describe the bug**
I think the code in Mapsui.Extensions.IntArrayExtension:

        if (axisOrder.Length == 2 && axisOrder[0] == 1 && axisOrder[1] == 1)

should be replaced by

        if (axisOrder.Length == 2 && axisOrder[0] == 0 && axisOrder[1] == 1)

Actual version always returns false for IsNaturalOrder no matter if axisOrder is [0, 1] or [1, 0] for netstandard.

**Expected behavior**
for axisOrder [0, 1] true should be returned

Nevertheless, the project evolved very nice. Like it very much (despite not having much time for using it at the moment).

pauldendulk:
It looks like you are right. What do you think @inforithmics?

@der1Mac I would be great if you could create a MR for this. If you do please point it to the develop/4.1 branch.

der1Mac:
> It looks like you are right. What do you think @inforithmics?
> 
> @der1Mac I would be great if you could create a MR for this. If you do please point it to the develop/4.1 branch.

Sorry, I will not have time to do this before mid December (maybe later). I hope I can contribute even more starting from January.

pauldendulk:
@der1Mac Thanks for the reply. In the meantime @charlenni has fixed this, so this can be closed. Everything ok! ✅

PR #2292, opened 2023-11-17
Removed a bug regarding #2253

charlenni:


pauldendulk (changes requested):
Thanks for picking this one up! One request for change. This change should be on develop/4.1. So, the the target branch should be develop/4.1 but you also need to make this change on a clone of the develop/4.1 branch.

In master this change is not needed because we will remove support for NETSTANDARD2_0 entirely, and it removed already in the pending .NET PR.

charlenni:
You are right. Did this already wrong yesterday. Should learn this ;-)

PR #2293, opened 2023-11-17
Removed a bug regarding #2253

charlenni:


pauldendulk (approved):
oh, that was quick! Thanks. The build errors are related to changes in the checks themselves (version not supported anymore, and a vulnerability in compatibility). I will look at that later.

pauldendulk:
Something I did not use at all so far is to put the link between the PR and the issue through the Development setting at the right. That is something I intent to do consistently and this is something PR submitters could do. I mean this:

image

I should put this in the guidelines somewhere.

pauldendulk:
Something else. I generate release notes from the PRs, so it is nice to have PR titles that read as release notes right away. Last month I added this section to the guidelines https://mapsui.com/documentation/contributors-guidelines.html#pr-titles-should-be-written-as-release-notes. 

If a change is needed I prefer to do this in the source (the PR) so that release notes and PR titles are identical. I will need to do this frequently.

Issue #2280, opened 2023-11-13
Add previous Viewport state to ViewportChanged event


pauldendulk:
**The problem**
Users report that sometimes they need to know the previous Viewport state for their logic.

**The solution**
Add the previous Viewport state tot he ViewportChanged event. The `value` in the property needs to be passed into the OnViewportChanged() event. We need to make changes to the event args and can not use the standard PropertyChangedEventHandler anymore.

Current Viewport property:
```csharp
    public Viewport Viewport
    {
        get => _viewport;
        private set
        {
            if (_viewport == value) return;
            _viewport = value;
            OnViewportChanged();
        }
    }
```

**Additional context**
- This originated here https://github.com/Mapsui/Mapsui/issues/928
- If users want to know which fields in the Viewport have changed they need to do some comparison. This should be done by the user. They could use an extension method for this `if (Viewport.HasCenterChanged(previousViewport)) DoYourThing()`
- There should be unit test for this.

PR #2294, opened 2023-11-17
Add old viewport in ViewportChanged event arguments

charlenni:
Until now, the old Viewport wasn't available after changing some of the settings. With this change, the ViewportChanged event has its own EventArgs, which contain the Viewport from before the change. So it is possible to check, what changed in between.

This is regarding #2280.


pauldendulk:
Sorry but I have caused conflicts to this PR by merging some big changes.

charlenni:
Solved them

pauldendulk (approved):
Great.

I guess logical next request will be to have the current viewport state as well. Should we? Lets just take this step first.

charlenni:
But why? I don't see any advantage for this.

pauldendulk:
You could do a comparison locally with no need to have access to the current viewport, but indeed, there is no real need because there is an alternative solution. So, let's keep it like it is now.

Issue #2235, opened 2023-10-25
Performance optimization


345813473:
https://github.com/Mapsui/Mapsui/blame/654daac40359b4f31fe5aaf2e8898f36ae0f2ee6/Mapsui/Features/BaseFeature.cs#L25

 public virtual object? this[string key]
 {
        get => _dictionary.TryGetValue(key，out var value) ? value : null;
        set => _dictionary[key] = value;
  }

pauldendulk:
Thanks, this is a small thing and there are many other things to look at wrt performance, but I don't see a reason not to improve this.

PR #2295, opened 2023-11-17
Improve performance when retrieving value from Dictionary in BaseFeature

charlenni:
This was done by suggestion in #2235


pauldendulk (approved):
Thanks a lot for picking of the not so exiting chores as well.

PR #2297, opened 2023-11-17
Add a widget to show logging messages in map

charlenni:
This Widget displays the given last logs to the Map.


charlenni:
@pauldendulk Not sure, if this is the way you want to have it. Perhaps you could look into it and suggest, what you want to have changed.

Mapsui/Widgets/LoggingWidget/LoggingWidget.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// <summary>
+    ///  Event handler for logging
+    /// </summary>
+    public void Log(LogLevel level, string description, Exception? exception)
+    {
+        if (_listOfLogEntries.Count >= _maxNumOfLogEntries) 
+        { 
+            // List is full, so delete the last one
+            _listOfLogEntries.RemoveAt(_maxNumOfLogEntries - 1);
```
inforithmics, line 51: Is there maybe some sort of lock needed, Because there are some MultiThreaded parts in Mapsui mostly in Fetching the Data. So there could be race conditions without a lock

Samples/Mapsui.Samples.Common/Maps/Widgets/LoggingWidgetSample.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        var widget = new LoggingWidget(mapControl.Map, 7);
+
+        // Add event handle, so that LoggingWidget gets all logs
+        Logger.LogDelegate += widget.Log;
+
+        widget.WidgetTouched += OnClick;
+
+        mapControl.Map.Widgets.Add(widget);
+        mapControl.Renderer.WidgetRenders[typeof(LoggingWidget)] = new LoggingWidgetRenderer(new MRect(10, 10, 210, 110), 12, Color.Black, Color.Orange, Color.Red, Color.White);
```
inforithmics, line 41: Should there be a default WidgetRenderer for LoggingWidget already in the MapRenderer?

inforithmics (approved):
Good improvement, so the Log Display could be unified around all samples platforms.

I had a look which samples already display Log Information.

And it was Blazor/Eto/Maui/Wpf

So the question is shouldn't be this widget be added by default to all samples so that Logs are always displayed? I don't have a strong opinion on that only asking for your opinion on that.

Mapsui/Widgets/LoggingWidget/LoggingWidget.cs
pauldendulk: Prefer to use thread save collections instead of lock(). In that case Microsoft solves the locking problem for us. It is easy to make mistakes with locks. Also, they may use better performing options than we would. In WPF logging logic we use our own LimitedQueue, so we don't have to do the cleanup.

Samples/Mapsui.Samples.Common/Maps/Widgets/LoggingWidgetSample.cs
```diff
@@ -0,0 +1,49 @@
...
+
+public class LoggingWidgetSample : IMapControlSample
+{
+    private IMapControl? _mapControl;
+
+    public string Name => "LoggingWidget";
+
+    public string Category => "Widgets";
+
+    public void OnClick(object? sender, WidgetTouchedEventArgs args)
```
pauldendulk, line 21: Is the OnClick needed? It does not seem to be called.

Mapsui/Widgets/LoggingWidget/LoggingWidget.cs
```diff
@@ -0,0 +1,104 @@
...
+        set
+        {
+            if (_opacity == value)
+                return;
+            _opacity = value;
+            OnPropertyChanged();
+        }
+    }
+
+    public override bool HandleWidgetTouched(Navigator navigator, MPoint position)
```
pauldendulk, line 91: HandleWidgetTouched is not called.  We do not really need this right now but perhaps it some point. It is not called because the Envelope is not set. This is something we do in the Draw method, which is a bit of a hack, but we don't have a better solution yet.

pauldendulk (changes requested):
This is great! Three small requests for change:
- Use the LimitedQueue to store the log lines. That would result in an exception at some point. 
- Register the widget renderer in the Renderer itself and not in the sample.
- Use ISample instead of IMapControlSample

Wishes that could be done later (could be someone else).
- Make the background fully transparent by default.
- Make the widget full screen, except for the margins.
- This logger should be turned on by default in debug mode. Also for users that build and app with our nuget package. This would be the first widget added by default (not entirely sure if we should. Alternative is to use a builder method which add it).
- It should be possible to turn it on in Release mode.
- It should be possible to turn it off in Debug mode.
- Some filter determining at which level to log. The default should be Info.
- Fancy, for easier reading: Perhaps we do not want to scroll the messages but always write a new one below the existing one. Once we reach the bottom we should overwrite the older message at the top. Those older messages could be faded out to make the difference between old and new clear. Or perhaps just show 10 lines, removing the older ones at the top while we go down.
- Super fancy: Some button inside the component to reset the log lines.

pauldendulk:
> So the question is shouldn't be this widget be added by default to all samples so that Logs are always displayed? I don't have a strong opinion on that only asking for your opinion on that.

I think it should be part of all samples and we should remove the existing ones. I don't want to add it explicitly to all samples. One option is to add this Widget to the Map by default. It would be nice to keep the default Map super clean though, and add no widgets at all. So, I am not sure about that. Another option is to introduce a MapBuilder which we would then use in all of our samples. So, currently all our samples have a `var map = new Map();` line. This should be replaced with `var map = MapBuilder.Create();`. The MapBuilder could have several overloads for different scenarios. For instance  `var map = MapBuilder.CreateDiagnostics();` to turn on everything that helps for debugging.

Samples/Mapsui.Samples.Common/Maps/Widgets/LoggingWidgetSample.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        args.Handled = true;
+    }
+
+    public void Setup(IMapControl mapControl)
+    {
+        _mapControl = mapControl;
+
+        //I like bing Hybrid
+        mapControl.Map = BingSample.CreateMap(BingHybrid.DefaultCache, BruTile.Predefined.KnownTileSource.BingHybrid);
```
pauldendulk, line 31: Because of this sample this line depends on another sample. This reduces the lines of code, not bad, but one thing I had in mind is to try to make it possible that the samples could be copy-pasted and compiled if only the nuget package was added. So, anyone that would have run the quickstart should be able to include the file and their project and it would compile. This is currently not possible, but something I would like to investigate. Some limitations I am already aware of:
- The ISample is in the samples so, would not compile. Instead we could use reflection searching for interfaces that implement `CreateMapAsync()` (or a more unique name if that is necessary).
- The `using Mapsui;` namespace is not in our samples because it already is in that namespace. 
- There are lots of samples that have code in other files, but perhaps this can be included as nested classes within the main sample.

charlenni:
Change the most what you expect. Some parts are not possible:

- LimitedQueue isn't possible, because it is in namespace of WPF.
- Cleared in the sample by clicking on the widget. 
- Not possible to make the widget "full screen", because we don't know the correct size. Isn't stored in Map, but in MapControl.
- Didn't add it automatically to the Map. Should be discussed.
- Printing is difficult. If the widget is at the top, it would be nice to have the newest at the top, older going down. If you have the widget at the bottom, it would be the best to have the newest at the bottom, older going down. But I would say, this is too much for such a small widget. Or should I add a flag for orientation?

pauldendulk:
> * LimitedQueue isn't possible, because it is in namespace of WPF.

Fine to use ConcurrentQueue. We could move LimitedQueue to Mapsui core, but perhaps we there are better options nowadays.

> * Cleared in the sample by clicking on the widget.

For the sample that is a good improvement. 

> * Not possible to make the widget "full screen", because we don't know the correct size. Isn't stored in Map, but in MapControl.

In other widgets the Draw method which takes the Viewport is used for that. It is a bit of a hack, but since this is already used in other widgets it will do for now. Could also changed after this PR.

> * Didn't add it automatically to the Map. Should be discussed.

I think it is very important to have this on by default in debug but we need to think about the best way to do this. I was thinking about a MapBuilder https://github.com/Mapsui/Mapsui/issues/2328. Also, I think enabled should have three states: On, Off, DebugOnly (which should be default).

> * Printing is difficult. If the widget is at the top, it would be nice to have the newest at the top, older going down. If you have the widget at the bottom, it would be the best to have the newest at the bottom, older going down. But I would say, this is too much for such a small widget. Or should I add a flag for orientation?

A nice to have, not needed. 

I think we could merge this. If you think so too then remove the WIP.

pauldendulk:
@charlenni You did something that is new to me. In GitLab there is a Draft checkbox, but in GitHub I had to use the WIP plugin. But now this PR is in Draft mode and the WIP is removed. So, apparently GitHub now added this feature. How did you put it in Draft?

pauldendulk (approved):

pauldendulk:
oh, it is this:

pauldendulk:
oh, it is here:

image

pauldendulk:
After you've put it in Draft you need to click the 'ready for review' button.

charlenni:
Oh, I only removed the WIP in the headline. Sorry. Try to do it next time correct.

pauldendulk:
> Oh, I only removed the WIP in the headline. Sorry. Try to do it next time correct.

Just trying to figure out how this works. Perhaps we could work with the 'convert to draft' button instead of the WIP tag and plugin.

PR #2298, opened 2023-11-17
Disable building uno samples

pauldendulk:


pauldendulk:
Disabled building the samples. There might be real problems underlying the fail. This needs to be fixed but we are far away from a release. The version overrides of the samples may need to be changed.

Issue #2262, opened 2023-11-10
[4.1 Regression] More touch issues after widget interaction


janusw:
**Mapsui Version**
4.1.1

**Mapsui Platform**
MAUI / Xamarin

**Device**
iOS 17

**Describe the bug**
Follow-up to #2197. After interacting with a widget in a specific way (see below), map panning is broken.

**To Reproduce**
1. Run Mapsui.Samples.Maui.MapView on an iOS device (iPhone/iPad).
2. Select a sample, let's say Demo -> OpenStreetMap.
3. Touch the map with one finger, drag onto one of the widgets (zoom in/out or center) and lift off the finger.
4. -> The widget is triggered and map panning is broken now: single-finger dragging zooms now.
5. Repeat step 3.
6. -> Now map panning is broken even more: single-finger dragging has no effect any more.

**Expected behavior**
Single-finger dragging should continue to work as before (panning the map). 4.0.0 works well, 4.1.1 fails.

**Additional context**
The issue is similar to #2197, and therefore probably related to the changes in PR #2133, #2190 and #2203.

janusw:
>     3. Touch the map with one finger, drag onto one of the widgets (zoom in/out or center) and lift off the finger.
> 
>     4. -> The widget is triggered and map panning is broken now: single-finger dragging zooms now.

Note: With 4.0.0, the gesture in step 3 does not trigger the widget.

charlenni:
@janusw What do you expect in this case? Is it, when you enter the widget in pan mode, you still want to pan? So if you enter the widget with an already touched state, you like no event for the widget.

I assume, that this should be checked [here](https://github.com/Mapsui/Mapsui/blob/83d393f2a2b8bd6aee2cf5169f4ea4ffd43e84ef/Mapsui.UI.Shared/MapControl.cs#L653). There should be a distinguishing flag, if the widget likes to have all events (click and move) or only that events, that corresponds to clicks.

When I look through the code, we are not very clear with the naming. What is a click, a touch, a move or a pan? Sometimes we use the mobile version, sometimes the desktop version. So is a touchable widget on, that could get only clicks or clicks and pans? And if I'm understanding right, then @inforithmics introduced this because of the edit functionality. Is this something belonging to widgets? For me, widgets are always overlays, a layer fixed on the screen coordinates on top of the map. But in this case the definition seems not to be right, correct?

janusw:
[...]

PR #2300, opened 2023-11-17
WIF Fix bug where touch detection ignores touch start position

pauldendulk:


janusw:
Thanks for looking into this, @pauldendulk.

I just tested the branch at revision afd2d4198d9d1cf85261c5b0a855a7eabdad51fc, and it seems that actually none of the widgets show any reaction (it is not possible to trigger them by tapping/clicking).

You wrote that you did not think everything through yet. Apparently more adjustments are needed.

pauldendulk:
> just tested the branch at revision [afd2d41](https://github.com/Mapsui/Mapsui/commit/afd2d4198d9d1cf85261c5b0a855a7eabdad51fc), and it seems that actually none of the widgets show any reaction (it is not possible to trigger them by tapping/clicking).
> 
> You wrote that you did not think everything through yet. Apparently more adjustments are needed.

Thanks for testing. What happened: The _pointerDownPosition was set after the check for widget touch. If the pointer down was on a widget the method returned and _pointerDownPosition was not set. I now moved this before the handling check for all platforms. I tested this only with mouse in WPF. Perhaps you could test this with touch on Android.

With all the different flows of touch in the different platforms it is likely there is still some problem somewhere. Perhaps we should set _pointerDownPosition to null in some scenarios. I guess every time there is a multi touch somewhere between touch down and touch up we should set _pointerDownPosition to null.

janusw:
> What happened: The _pointerDownPosition was set after the check for widget touch. If the pointer down was on a widget the method returned and _pointerDownPosition was not set. I now moved this before the handling check for all platforms. I tested this only with mouse in WPF. Perhaps you could test this with touch on Android.

Thanks for the update. I just tried 5fb4984e070507e87a2e6a25cda33dff998cbd32 on Android, and it seems to work pretty well. The widgets are functional again, and the bug reported in #2262 appears to be gone. Will try to do some more testing later.

janusw:
On iOS (via Mapsui.Forms), the widgets work as well, but after tapping a widget (in a 'normal' fashion), the touch-navigation is still broken (or again?) in the same way described in #2262. After tapping a widget for the first time, single-finger dragging performs a zoom operation. After tapping a widget for the second time, single-finger dragging does not do anything any more.

memsom:
We are keen to see this fix as I think we have been hit by this bug. The 4.0.0 release worked as expected (but had the rotation bug and the tiling spacing bug), but 4.1.0 and 4.1.1 (which fixed those bugs) seem to have broken the long touch events for us under Android.

memsom:
So a colleague has tested this and it seems to only partially work for us. We will hopefully be able to explain more shortly - trying to pin down what is going wrong.

memsom:
@pauldendulk yes - Xamarin Forms, but specifically only Android. We currently don't have an iOS version of the app in question.

memsom:
To us - it looks like in the `MapControl`, where the `SKTouchAction.Released` is handled, there are no touches on a long press. So the fix we have is to remove the check from the outer if, then we inject a `TouchEvent` by setting the `releaseTouched` to a value if the `TryRemove(..)` fails. Don't know if that helps. My colleague has demonstrated it just now and it seems to work.  So - circa 

(L118) We added a new field to record the time the initial Pressed event took place.

```csharp
    private long _pointerDownTicks;
```

(L310) We store the time when we also set the Pressed location.

```csharp
            if (e.ActionType == SKTouchAction.Pressed)
            {
                _pointerDownPosition = location;
                _pointerDownTicks = ticks;
            }
```
 
{L343) Moved the 'touchRemoved' check from the if statement so we can take action when it fails.
(L346 - L351) If we didn't receive a 'releasedTouch' TouchEvent from the _touches.TryRemove we create one with the Id of the Released, the  _pointerDownPosition and _pointerDownTicks; 
We don't add this to _touches collection as we still want to enter the if on L353, but as we now have the 'releasedTouch', the code to check for long press should work.

```csharp
           else if (e.ActionType == SKTouchAction.Released)
            {
                var touchRemoved = _touches.TryRemove(e.Id, out var releasedTouch);
                if (!touchRemoved)
                {
                    // There was no touch to remove so create a releasedTouch event to use in long tap checking
                    releasedTouch = new TouchEvent(e.Id, new MPoint(_pointerDownPosition?.X ?? 0, _pointerDownPosition?.Y ?? 0),
                        _pointerDownTicks);
                }

                if (_touches.Count == 0)
                {
                    // Is this a fling?
```

This seems to solve the issue we had at any rate.

memsom:
Is there any update on this bug?

pauldendulk:
@memsom The long press seems a separate problem. I created a new issue for it: https://github.com/Mapsui/Mapsui/issues/2313.

What happens is that the ticks used to compare for long press are set on every move, not just on touch down. This was indeed in line with the fix that you suggested.

pauldendulk:
I will merge this one, as it is at least an improvement. The long press issue will be in a separate fix. Any remaining issues need to be fixed later.

PR #2302, opened 2023-11-19
Move nullable to root Directory.Build.props

pauldendulk:
As part of the cleanup of v5


inforithmics (commented):
Generally I like the idea to enable nullability for all Projects centrally.

When this message occours.
The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.

Somhow the nullable setting is overriden in this project, don't know what causes it maybe bugs in the sdk.

The only solution I found so far is to add

#nullable enable on top of the file

pauldendulk:
I will look into any remaining warnings. The 'files changed' page shows warnings for Avalonia https://github.com/Mapsui/Mapsui/pull/2302/files. Is that what you were referring to?

PR #2314, opened 2023-12-03
Fix the v4 build by removing net6.0 from Maui, Uno, Uno.WinUI and iOS

pauldendulk:
With the release of .NET 8 there have been big changes to the gitlab build agents which break the build in many ways without any changes on our side. This PR is the struggle to fix it.


pauldendulk:
In order to get v4 to build again with the new build agents I did not only had to update the build scripts but also had to remove target frameworks. This is definitely not something you want to do in a patch, but I don't see a way around it:
- Removed net6.0 from mapsui.UI.Maui
- Removed net6.0-ios from Mapsui.UI.Uno.WinUI
- Removed net6.0-ios from mapsui.UI.Uno
- Removed net6.0-ios15.4 from Mapsui.UI.iOS

PR #2326, opened 2023-12-09
Fix widget touch

pauldendulk:


pauldendulk:
This fixes the issues I can reproduce. Worked on:
- For MAUI and for Android we used a central method for handling of all three type (down, move, up), this complicated the logic.
- The widget touch up needs to know the widget touch down position.
- After a widget touch down the map should not respond to down, move or up

It is unfortunate these changes are necessary in the code that should undergo a major refactoring in master.

stevegdavis:
Just updated my project to Mapsui 4.1.2 and now I find that when I press my user defined widget on the map the MapClicked event fires. Is this intentional? Prior to this version the event doesn't fire except when clicking on the map as excepted.

PR #2330, opened 2023-12-10
Introduce a TileCache (and Remove it from RasterStyle)

inforithmics:


pauldendulk (approved):
This is an improvement. Styles should just define the style and should hold no state other than that. By moving the TileCache to the RenderCache it become more comprehensible.

PR #2334, opened 2023-12-11
Fix several warnings

pauldendulk:


pauldendulk:
Done:
- Avalonia to 11.0.6. This fixed warnings on Android
- All samples and tests to .NET 8, this fixed several warnings
- Add `<Nullable>enable</Nullable>` to the Avalonia specific Directory.Build.props to fix null annotation warnings (which should not have been there).
- Remove duplicate references for Microsoft.AspNetCore.Components.Web*
- Return the instance instead of an abstraction from a method in two places
- Use await cancelationTokenSource.CancelAsync() instead of cancelationTokenSource.Cancel() in method returning Task.
- Suppress a Warning CA1416 (Call site reachable by all platforms) in the WPF samples
- Made a method static following suggestion.
- Add script delete bin and obj

PR #2340, opened 2023-12-12
Fix System.AccessViolationException in Unit Tests

inforithmics:


inforithmics:
Things done: 
1) Upgraded to NUnit 4 (has .NET 6 Target)
2) Fixed VectorStyleRenderer can not render feature of type 'Mapsui.Layers.RasterFeature', Layer: ...
3) The Access Violation can only be reproduced when running without attached debugger.

inforithmics:
I looked at the Code and saw that when the Renderer isn't used anymore the RendererCache isn't disposed. Which should happen because there a lot of SkiaSharp objects are located, (SkPath, ...)

inforithmics:
It seems that the Bug is in the Symbols Sample, which makes sense because it crashes during rendering of Symbols

inforithmics:
What I discovered is that when a Style is defined on the Feature and on the Layer the Feature is Rendered two Times. The question is now if a Feature defines a Style should the Layer Style still be used for Rendering?

pauldendulk:
> What I discovered is that when a Style is defined on the Feature and on the Layer the Feature is Rendered two Times. The question is now if a Feature defines a Style should the Layer Style still be used for Rendering?

This is a very old problem. And we have this item in FAQ related to it https://mapsui.com/documentation/faq.html#how-can-i-get-rid-of-the-white-dots-or-black-lines. Not drawing the Layer.Style is not what you want in many cases. Often there is a default style on the layer and feature specific styles on top of that. Stacking styles is used a lot in the applications that I work on. For instance, you can have a general outline of a vehicle, you could have color symbols on top of that to indicate which compony it is. There could be custom symbols to indicated status (driving, parked, breakdown), there could be symbols indicating the kind of load it is carrying. We have the SymbolCollection to make this possible. So, when stacking is a common thing I don't see a reason why you should not be able to stack of feature symbol on top the layer symbol.

A solution that I had in mind would be to not have a default style on the layer. The downside is that users just starting out would see nothing in the map if they do not add a style. This would be confusing. Some default does make sense. For debugging purposes we could draw some grayed out test version of the style, so users would at least see something.

Then again they would still need to add a style to get it property working. So a default style is not that bad. For the RasterStyle it would even weirder because the style does not add any additional information, why would a user need to add it.

Another thing I considered to change was to replae the Layer.Style with a collection: `ICollection<IStyle> Styles { get; }`, so that it is identical to the Features.Styles. In general I like list more than single values because you do not need to null check the list.

General remark: I prefer the layer style and for advanced scenarios the ThemeStyle. Users often choose the feature style when starting because that seems easier, but in the long run this is often more complicated, especially if you have a lot of features.

pauldendulk:
> 2\. Fixed VectorStyleRenderer can not render feature of type 'Mapsui.Layers.RasterFeature', Layer: ...

So, how was it solved?

Mapsui/Extensions/CatchExceptions.cs
```diff
@@ -1,12 +1,14 @@
...
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using Mapsui.Logging;
 
 namespace Mapsui.Extensions;
 
 public static class Catch
 {
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
pauldendulk, line 11: Why were these attributes added?

Tests/Mapsui.Rendering.Skia.Tests/RegressionMapControl.cs
```diff
@@ -10,6 +10,7 @@
 using Mapsui.Utilities;
 
 #pragma warning disable CS0067 // The event is never used
+#pragma warning disable IDISP008 // Don't assign member with injected and created disposables
```
pauldendulk, line 13: I think for v5 we should go all the way and fix these problem properly instead of to suppress them. The issue makes sense. So, if you have a disposable you need to choose, either always pass it in, or always create it internally.
https://github.com/DotNetAnalyzers/IDisposableAnalyzers/blob/master/documentation/IDISP008.md

Tests/Mapsui.Tests/Layers/FetchInfoTests.cs
```diff
@@ -18,6 +19,6 @@ public void GrowFetchInfo(double amount, double resolution, double expectedWidth
         var grownFetchInfo = fetchInfo.Grow(amount);
 
         // Assert
-        Assert.AreEqual(expectedWidth, grownFetchInfo.Extent.Width);
+        ClassicAssert.AreEqual(expectedWidth, grownFetchInfo.Extent.Width);
```
pauldendulk, line 22: Thanks for upgrading nunit, that is a lot of work.

Mapsui.Nts/Providers/Wfs/Xml/XPathQueryManager.cs
```diff
@@ -355,7 +360,10 @@ private void FindXPath(XPathExpression? xPath)
     {
         xPath?.SetContext(_paramContext!);
         if (xPath != null)
-            _xIter = _xNav?.Select(xPath);
+        {
+            Catch.Exceptions(() => _xIter = _xNav?.Select(xPath));
```
pauldendulk, line 364: See the comment above.

Mapsui.Nts/Providers/Wfs/Xml/XPathQueryManager.cs
```diff
@@ -192,8 +193,12 @@ public IXPathQueryManager Clone()
             _paramContext.AddParam(queryParameters);
         string? result = null;
         FindXPath(xPath);
-        if (_xIter?.MoveNext() ?? false)
-            result = _xIter?.Current?.Value;
+        Catch.Exceptions(() =>
```
pauldendulk, line 196: Why is this catch necessary? Could add be avoided somehow but first checking the data? Is this xml invalid? Is this a threading problem and should we fix that. In general, if there is this weird problem that you discovered while debugging that a comment needs to explain why this empty catch was added.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -23,10 +23,11 @@
 
 namespace Mapsui.Rendering.Skia;
 
-public class MapRenderer : IRenderer
+public class MapRenderer : IRenderer, IDisposable
```
pauldendulk, line 26: I understand why this is necessary. Lets do it like this for now, but in general I would like to reduce the number of disposable classes.

pauldendulk (commented):
Thanks a lot! I added a few questions to the code.

Mapsui.Nts/Providers/Wfs/Xml/XPathQueryManager.cs
inforithmics: During testing this threw exceptions, I couldn't really find a way to detect the reason. Generally the WFSProvider throws to many exceptions in normal operating mode. I have to investigate it if I can reduce them, because when the exceptions are thrown at the wrong moment. Than the WFS Layer sample fails.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -23,10 +23,11 @@
...
 {
     private readonly IRenderCache _renderCache;
     private long _currentIteration;
+    private readonly bool _ownsRenderCache;
```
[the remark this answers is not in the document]
inforithmics: The owner ship of the Cache I introduced, because in the TileRasterizingProvider several renderes are used that share the same cache. So when the Renderer creates the Cache he owns it, when he gets injected by the constructor he isn't allowed to dispose it. I think for that is the IDISP0008 rule

Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
```diff
@@ -168,7 +168,7 @@ public static bool DrawSymbol(SKCanvas canvas, Viewport viewport, ILayer layer,
         var path = vectorCache.GetOrCreatePath(symbolStyle.SymbolType, CreatePath);
 
         if (fillPaint != null && fillPaint.Color.Alpha != 0) canvas.DrawPath(path, fillPaint);
-        if (linePaint != null && linePaint.Color.Alpha != 0) canvas.DrawPath(path, linePaint);
+        if (linePaint != null && linePaint.Color.Alpha != 0) canvas.DrawPath(path, linePaint);
```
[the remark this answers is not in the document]
inforithmics: Here the Exception occured so I was experimenting with some code changes, appearently I didn't fully revert the changes. So doing it now.

Mapsui.Rendering.Skia/MapRenderer.cs
inforithmics: I could remove the IDisposable by requiring an IRenderCache parameter.

Mapsui/Fetcher/FetchWorker.cs
```diff
@@ -68,5 +68,9 @@ private async Task FetchAsync(CancellationTokenSource? cancellationTokenSource)
             // cancellationTokenSource.Cancel() is called.
             // Logger.Log(LogLevel.Error, e.Message, e);
         }
+        catch (Exception e)
```
pauldendulk, line 71: hmmm, This catch should have been here from the start.

Mapsui.Nts/Providers/Wfs/Xml/XPathQueryManager.cs
pauldendulk: Okay, it is important to understand why.

Mapsui.Rendering.Skia/MapRenderer.cs
pauldendulk: Lets keep it like it is.  This is the SkiaSharp renderer, the cache holds objects specific for SkiaSharp and they are intended to make the renderer faster. So, let's make the renderer responsible.

Mapsui.Rendering.Skia/MapRenderer.cs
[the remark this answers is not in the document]
pauldendulk: But why could the Renderer not always own the RenderCache and let the TileRasterizingProvider use the RenderCache owned by the Renderer?

inforithmics:
Fixed now the WFSProvider. Doesn't throw anymore a lot of exceptions. 
1. InitAsync was called parallel causing a lot of threading exceptions.
2. InitAsync was called on every query but it is only needed once ore when the settings have changed.

Mapsui/Extensions/CatchExceptions.cs
inforithmics: I will revert this, because I fixed the wfsprovider so that it doesn't throw anymore exceptions. And besides I read that methods with try catch aren't inlined anyway so this wouldn't work.

Mapsui.Nts/Providers/Wfs/WFSProvider.cs
```diff
@@ -76,6 +77,7 @@ public enum WFSVersionEnum
     private string? _proxyUrl;
     private ICredentials? _credentials;
     private readonly CrsAxisOrderRegistry _crsAxisOrderRegistry = new();
+    private static SemaphoreSlim _init = new(1, 1);
```
pauldendulk: One more thing about the lock. If a second call reached the lock, it will wait until the first call finishes initializing and after that will start to initializing for a seconds time (I could be mistaken but that is how it looks to me from this diff). Perhaps it should check the _initialized boolean within the lock.

pauldendulk (approved):

Mapsui.Nts/Providers/Wfs/WFSProvider.cs
inforithmics:
```
  try
        {
            // test again could be already initialized
            if (_initialized)
                return;
```

after the lock it checks if it is _already initialized and if so it exists without initializing.

PR #2341, opened 2023-12-14
MapBuilder

charlenni:
Create a first attempt for a MapBuilder.

@pauldendulk Is this the way you thought about it?


pauldendulk (approved):
Thanks a lot. This will make it a lot easier for the users. The initialization of the map after the extent becomes known remains problematic but this was an already existing problem which should be dealt with later.

pauldendulk:
Oops, merged the wrong merge request. First time ever for me. Reverted it just now.

PR #2348, opened 2023-12-15
Compile all samples in one go

inforithmics:


pauldendulk (approved):
Yes, this is much better. I think we could do this in more places. In v4 it was often not possible because we needed msbuild for some projects. Now it is `dotnet` all the way which makes life easier for us.

PR #2349, opened 2023-12-16
AOT

inforithmics:


inforithmics:
Currently it is failing because of Reflection getting the Samples. I think I will write a Source Generator for that, than this becomes AOT Compatible.

pauldendulk:
I don't know how much work it would be, and more importantly how maintainable the source generators will be, but perhaps we do not need the the samples to be AOT. Or perhaps just one sample demonstrating that it works. ... Then again it would be nice if our own online samples are AOT, to see if it is actually faster.

inforithmics:
> I don't know how much work it would be, and more importantly how maintainable the source generators will be, but perhaps we do not need the the samples to be AOT. Or perhaps just one sample demonstrating that it works. ... Then again it would be nice if our own online samples are AOT, to see if it is actually faster.

I do a lot of Mobile Development and there because of AOT and Performance issues I try to avoid Reflection, The solution is Source Generators, which do the Refelection during compilation time. Making the App faster and AOT safe.

pauldendulk:
Okay, lets go ahead with code generation.

inforithmics:
Things done: 
1) Samples are now Registered with SourceCodeGenerator. Avoiding Reflection speeds up the inital Sample Display quite a lot
2) Blazor is Compiled with publish AOT and Strip Symbols for faster startup and loading

inforithmics:
There are still 2 AOT Warnings, but in my opinion they shouldn't matter, or I don't know how to fix them.

inforithmics:
Additional Things done:
1) Delete Mapsui.Mac.Legacy.sln (isn't needed anymore because there is no Xamarin.Forms)
2) Fixed current Dispose Warnings

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -436,14 +443,24 @@ private void Map_PropertyChanged(object? sender, PropertyChangedEventArgs e)
 
     public Map Map
     {
-        get => (Map)GetValue(MapProperty);
+        get
+        {
+            var map = GetValue(MapProperty) as Map;
+            if (map == null)
+            {
+                _ownsMap = true;
```
pauldendulk, line 451: Complicated 🤨

Mapsui/Extensions/FeatureExtensions.cs
```diff
@@ -3,24 +3,64 @@
...
+    private static readonly Dictionary<Type, Func<IFeature, IFeature>> _copyFeature = new();
+
+    public static void RegisterFeature<T>(Func<IFeature, IFeature> copy)
+    {
+        _copyFeature[typeof(T)] = copy;
+    }
+
+    public static IFeature StandardCopy(IFeature feature)
+    {
+        if (feature is PointFeature pointFeature)
```
pauldendulk, line 22: So what if a user adds a custom renderer and uses a custom feature type?

Mapsui/Fetcher/FetchWorker.cs
```diff
@@ -59,7 +59,11 @@ private async Task FetchAsync(CancellationTokenSource? cancellationTokenSource)
                 if (_fetchDispatcher.TryTake(out var method))
                     await method().ConfigureAwait(false);
                 else
+#if NET8_0_OR_GREATER
```
pauldendulk, line 62: I never like the #ifs. Perhaps we can make them less ugly if we create a CancelAsync extension method for 6. That method should still have a #if !NET8_0_OR_GREATER, but at least it will be in one place, and in code you do not read that often.

Samples/Avalonia/Mapsui.Samples.Avalonia/Views/MainView.axaml.cs
```diff
@@ -15,8 +15,8 @@ public partial class MainView : UserControl
 {
     static MainView()
     {
-        // todo: find proper way to load assembly
-        Mapsui.Tests.Common.Utilities.LoadAssembly();
+        Mapsui.Tests.Common.Samples.Register();
```
pauldendulk, line 18: Thats an easy fix.

pauldendulk (commented):
Great you took this up! Quite a lot to learn. I suspected this would be something for you.

You have a [WIP PR](https://github.com/BruTile/BruTile/pull/225) at BruTile. So, this needs to be reviewed, merged, released and updated in Mapsui before it can be used. What will not work as long is that is not the case? 

This is a new technology so a have some questions about the changes to help me understand. Please take a look.

Also, I need to read up on source generator. Any recommendations?

pauldendulk:
@inforithmics I resolved a conflict in the browser, you may want to check it https://github.com/Mapsui/Mapsui/pull/2349/commits/ddc9e2969625ff339786ecb67ddff0a6bed38997

Directory.Build.props
```diff
@@ -18,6 +18,7 @@
     <!--Disable Run Api Compat Task causes Build failure on github actions with visual studio 2022 17.5 and .net Compilers Toolset 4.6-->
     <RunApiCompat>false</RunApiCompat>
 	<Nullable>enable</Nullable>
+<IsAotCompatible Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net7.0'))">true</IsAotCompatible>
```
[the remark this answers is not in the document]
inforithmics: According to this
https://devblogs.microsoft.com/dotnet/creating-aot-compatible-libraries/
IsAotCompatible is available on .NET 7 Framework and later, that's why it is definied this way. So this needs addtional Targets for mapsui. This is the reason why I added the .NET 7 and .NET 8 Targets to Mapsui Projects

Directory.Packages.props
```diff
@@ -52,7 +52,7 @@
     <PackageVersion Include="Mapsui.Uno.WinUI" Version="4.0.0-beta.8" />
     <PackageVersion Include="Mapsui.WinUI" Version="4.0.0-beta.8" />
     <PackageVersion Include="Mapsui.Wpf" Version="4.0.0-beta.8" />
-    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="[7.0.0,8.0.0)" />
+    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="[8.0.0,9.0.0)" />
```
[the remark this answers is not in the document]
inforithmics: good point will remove it

Mapsui.ArcGIS/Mapsui.ArcGIS.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
```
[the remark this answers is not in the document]
inforithmics: Yes that's right, if something doesn't compile with .NET 6 it will get a compilation error. Or can be conditional compiled.
with 
```
#if NET6_0
#else
#endif
```
or
```
#if NET7_0_OR_GREATER
#else
#endif
```

Mapsui.Nts/GlobalSuppressions.cs
```diff
@@ -13,3 +13,7 @@
 [assembly: SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP008:Don't assign member with injected and created disposables", Justification = "<Pending>", Scope = "member", Target = "~F:Mapsui.Providers.Wfs.Utilities.GeometryFactory.XmlReader")]
 [assembly: SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP001:Dispose created", Justification = "<Pending>", Scope = "member", Target = "~M:Mapsui.Providers.Wfs.Utilities.HttpClientUtil.GetDataStreamAsync~System.Threading.Tasks.Task{System.IO.Stream}")]
 [assembly: SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP003:Dispose previous before re-assigning", Justification = "<Pending>", Scope = "member", Target = "~P:Mapsui.Nts.Providers.GeoJsonProvider.FeatureCollection")]
+[assembly: SuppressMessage("Trimming", "IL2111:Method with parameters or return value with `DynamicallyAccessedMembersAttribute` is accessed via reflection. Trimmer can't guarantee availability of the requirements of the method.", Justification = "Wrong warning", Scope = "member", Target = "~M:Mapsui.Nts.Providers.Shapefile.DbaseReader.GetSchemaTable~System.Data.DataTable")]
```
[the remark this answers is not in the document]
inforithmics: The IL messages are warnings and not errors that's why they can be surpressed, and sometimes the Analyzer has a Bug so it warns on something that works. Because in my opinion here is no reflection involved here (the Interface was to broadly tagged as DynamicCode).

Mapsui.UI.Maui/MapControl.cs
```diff
@@ -102,6 +102,9 @@ public TouchEvent(long id, MPoint screenPosition, long tick)
 
     private TouchMode _mode;
 
+    private bool _ownsMap;
```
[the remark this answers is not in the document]
inforithmics: I think with the removal of the Skia Objects, the Map can be made again non disposable. I think I will revert this here and make a seperate pull request for that.
The reason I did this was to fix the warnings but It seems not to be a quick fix.

Directory.Packages.props
```diff
@@ -92,6 +92,7 @@
     <PackageVersion Include="Svg.Skia" Version="[1.0.0.7,2.0.0)" />
     <PackageVersion Include="System.Linq.Async" Version="[6.0.1,7.0.0)" />
     <PackageVersion Include="System.Memory" Version="[4.5.5,5.0.0)" />
+    <PackageVersion Include="System.Text.Json" Version="[8.0.0,9.0.0)" />
```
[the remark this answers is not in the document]
inforithmics: I have to look if I can remove this

inforithmics:
> Great you took this up! Quite a lot to learn. I suspected this would be something for you.
> 
> You have a [WIP PR](https://github.com/BruTile/BruTile/pull/225) at BruTile. So, this needs to be reviewed, merged, released and updated in Mapsui before it can be used. What will not work as long is that is not the case?

AOT has some Basic support for reflection, so when there is a warning about not beeing AOT compatible it might work or not. So this has to be tested on the resulting Program and cannot be done on the Library level. Making an assembly AOT compatible means removing all scenarios that involve reflection and certain Code patterns so that it always works.

I already use AOT on Mobile Applications with mapsui this means I have had to annotate Code as beeing used so that it wasn't removed.

In the Case of Brutile I think the WMTS Support uses XmlSerializer that uses reflection. So this could be working or not. If the AOT compiler is to aggressive and removes Things that are needed for the XmlSerializer than WMTS wouldn't work in an AOT scenario.

> 
> This is a new technology so a have some questions about the changes to help me understand. Please take a look.
> 
> Also, I need to read up on source generator. Any recommendations?

Is a good introduction
https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/source-generators-overview

Mapsui/Fetcher/FetchWorker.cs
inforithmics: I'll try that

Mapsui/Extensions/FeatureExtensions.cs
inforithmics: I could add as fall back the old Activator.CreateInstance method.

inforithmics: Or I could add a source generator to Mapsui that does this autmatically.

Mapsui.UI.Shared/MapControl.cs
inforithmics: Maybe I could introduce a DisposableWrapper that is always disposed but has a member with the value ownsDisposable and when it is true then it disposes the disposable else it sets only the member to null.
So the reading the code is more straight forward.

inforithmics:
Things done: Clean up nuget packages and removed appveyer.testloggger (isn't used anymore)

Mapsui/Fetcher/FetchWorker.cs
inforithmics: Done

Directory.Packages.props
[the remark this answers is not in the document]
inforithmics: I did remove all unneded nuget packages from Directory.Packages.props so simplifiying it quite a bit.

Directory.Packages.props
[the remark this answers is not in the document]
inforithmics: Done

Mapsui.UI.Maui/MapControl.cs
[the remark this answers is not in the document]
inforithmics: The Dispose in the map is used to stop the Fetching of Data and Deregister EventHandlers

Mapsui/Extensions/FeatureExtensions.cs
inforithmics: I'll try adding a SourceGenerator

Mapsui.ArcGIS/Mapsui.ArcGIS.csproj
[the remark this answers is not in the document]
pauldendulk: I would like to avoid #if's as much as possible.

Mapsui.UI.Shared/MapControl.cs
pauldendulk: Let keep it like this and start thinking about making the Map not-disposable.

Mapsui/Extensions/FeatureExtensions.cs
pauldendulk: Did you manage to fix this with a source generator?

inforithmics: Yes, in Mapsui the SourceGenerator Project is referenced and in the mapsui nuget package a compiled version of the SourceGenerator is included, executing it during the build on librarie that reference the mpasui nuget package.

Mapsui.Rendering.Skia/GlobalSuppressions.cs
```diff
@@ -12,3 +12,10 @@
...
 [assembly: SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP001:Dispose created", Justification = "<Pending>", Scope = "member", Target = "~M:Mapsui.Rendering.Skia.Extensions.GeometryCollectionExtensions.ToSkiaPath(NetTopologySuite.Geometries.GeometryCollection,Mapsui.Viewport,SkiaSharp.SKRect,System.Single)~SkiaSharp.SKPath")]
 [assembly: SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP003:Dispose previous before re-assigning", Justification = "<Pending>", Scope = "member", Target = "~M:Mapsui.Rendering.Skia.Extensions.GeometryCollectionExtensions.ToSkiaPath(NetTopologySuite.Geometries.GeometryCollection,Mapsui.Viewport,SkiaSharp.SKRect,System.Single)~SkiaSharp.SKPath")]
+[assembly: SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP003:Dispose previous before re-assigning", Justification = "<Pending>", Scope = "member", Target = "~M:Mapsui.Rendering.Skia.MapRenderer.#ctor(Mapsui.Rendering.IRenderCache)")]
```
pauldendulk, line 15: We really need to come up with a  redesign where these suppressions are not needed.

inforithmics: Analyzers are sometimes not 100 % accurate and give sometimes more hints. So to avoid pulluting the Source Code a add the regressions to GlobalSuppressions.cs so that the reading flow of the source code isn't hindered.
Maybe an update to the Anlyzer improves things.

inforithmics:
Brutile IsAatCompatible is now available
https://github.com/BruTile/BruTile/pull/225

Mapsui.Rendering.Skia/GlobalSuppressions.cs
pauldendulk: In case of the ownership issue the code may be correct, but if the analyzer gets confused about who is owning the disposable then it is also confusing for us. So, in case of the ownership of the Map I would think the analyzer is correct and we should choose to either make the MapControl own or the user. It is simpler, even if the current code is correct. (This is just an example, in case of the Map control we should look into making it non-disposable).

Mapsui/Extensions/FeatureExtensions.cs
```diff
@@ -8,19 +8,45 @@ namespace Mapsui.Extensions;
...
     {
-        return (T)Activator.CreateInstance(typeof(T), original)!;
+        if (_copyFeature.TryGetValue(feature.GetType(), out var registeredCopy))
+        {
+            return registeredCopy(feature);
+        }
+        
+        // Fall Back if Type is not registered
+        var type = feature.GetType();
+        return (IFeature)Activator.CreateInstance(type, feature)!;
```
pauldendulk, line 28: This is an improvement compared to the earlier version but my understanding is that custom features will not work in AOT. I don't think this is a problem, but it is important that the user gets a clear error message in that case.  So underneath the '// Fall Back ...'  line there should be a check:

```csharp
if (RunningInAotMode) 
   throw new NotSupportedException("Custom feature types are not supported when running in AOT"). 
```

Samples/Mapsui.Samples.MapView/AnimatedMyLocationSample.cs
```diff
@@ -12,7 +12,7 @@ public sealed class AnimatedMyLocationSample : IMapViewSample, IDisposable
 {
     private readonly PeriodicTimer _timer = new PeriodicTimer(TimeSpan.FromSeconds(1));
     private Position _newLocation;
-    private MapView? _mapView;
+    private UI.Maui.MapView? _mapView;
```
pauldendulk, line 15: These namespace changes seem unrelated to AOT, why is the now necessary?

Directory.Packages.props
```diff
@@ -115,6 +96,5 @@
     <PackageVersion Include="Uno.WinUI.Skia.Wpf" Version="[5.0.48,6.0.0)" />
     <PackageVersion Include="Uno.WinUI.RemoteControl" Version="[5.0.48,6.0.0)" />
     <PackageVersion Include="Uno.WinUI.WebAssembly" Version="[5.0.48,6.0.0)" />
-    <PackageVersion Include="Xam.Plugin.Geolocator" Version="[5.0.0-beta,6.0.0)" />
```
pauldendulk, line 118: Great you removed all the package not in use. This was an item on the roadmap https://mapsui.com/documentation/roadmap.html#mapsui-5

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -427,14 +444,21 @@ private void Map_PropertyChanged(object? sender, PropertyChangedEventArgs e)
 
     public Map Map
     {
-        get => (Map)GetValue(MapProperty);
+        get
+        {
+            if (GetValue(MapProperty) is not Map map)
+            {
+                _map ??= new DisposableWrapper<Map>(new Map(), true);
```
pauldendulk, line 451: I guess this is an improvement. I will have to look at this a bit better later one. As discussed elsewhere, a better solution would be to make the Map non-disposable.

Mapsui.sln
```diff
@@ -2777,6 +2784,118 @@ Global
 		{A8808FA9-B3FC-4C13-8C22-22DC2C9A1EA7}.Release|x86.ActiveCfg = Release|Any CPU
 		{A8808FA9-B3FC-4C13-8C22-22DC2C9A1EA7}.Release|x86.Build.0 = Release|Any CPU
 		{A8808FA9-B3FC-4C13-8C22-22DC2C9A1EA7}.Release|x86.Deploy.0 = Release|Any CPU
+		{AE094043-B696-49D4-8F58-2523047A6210}.Ad-Hoc|AnyCPU.ActiveCfg = Debug|Any CPU
```
pauldendulk, line 2787: What a horrible bloat of sln entries. Added by by VS I guess, but perhaps based on the configurations we have defined elsewhere in the project. I don't expect this bloat if we would create a new project from scratch. This is not something to fix in this PR, and perhaps this should have no priority at all, but if anyone knows the way to simplifies this I would be interested.

pauldendulk (commented):
I added a few more comments and have two additional things I would like you to look at. 
- Could a if (RunningInAotMode) exception be added?
- Do we really need the namespace in UI.Maui.MapView? I have been removing such namespaces in the cleanup in previous weeks and would like to have a 'Only add namespace in code if really needed' policy.

Mapsui.sln
pauldendulk: It would be nice if we could reduce the configuration bloat one day. But not today 😟

Mapsui/Extensions/FeatureExtensions.cs
pauldendulk: @inforithmics I think it is important to have an exception here. If not this could be very frustrating for users. Could you look into this?

pauldendulk (changes requested):
I have one question left, and one remark about adding an exception. If fixed I will merge this asap.

pauldendulk:
@inforithmics I caused a conflict, tried to solve it in the browser but I am not quite sure about how, so please take a look yourself.

Samples/Mapsui.Samples.MapView/AnimatedMyLocationSample.cs
inforithmics: The reason this is needed is because of this SourceGenerated Code so MapView is now a namespace and therefore I have to specifiy more to say I want the MapView Control

```
using System;
namespace Mapsui.Samples.Maui.MapView
{
    public static class Samples
    {
        // Avoid double registration
        private static bool _registered;
        
        public static void Register() 
        {
            if (_registered)
                return;
                
            _registered = true;
            Mapsui.Samples.Common.AllSamples.Register(new Mapsui.Samples.Maui.PolylineSample());
            Mapsui.Samples.Common.AllSamples.Register(new Mapsui.Samples.Maui.CircleSample());
            Mapsui.Samples.Common.AllSamples.Register(new Mapsui.Samples.Maui.MyLocationSample());
            Mapsui.Samples.Common.AllSamples.Register(new Mapsui.Samples.Maui.PinSample());
            Mapsui.Samples.Common.AllSamples.Register(new Mapsui.Samples.Maui.PolygonSample());
            Mapsui.Samples.Common.AllSamples.Register(new Mapsui.Samples.Maui.ManyPinsSample());
            Mapsui.Samples.Common.AllSamples.Register(new Mapsui.Samples.Maui.AnimatedMyLocationSample());
            Mapsui.Samples.Common.AllSamples.Register(new Mapsui.Samples.Maui.SnapshotSample());

        }
    }
}
```

Mapsui.Nts/GeometryFeature.cs
```diff
@@ -7,6 +8,11 @@ namespace Mapsui.Nts;
 
 public class GeometryFeature : BaseFeature, IFeature
 {
+    static GeometryFeature()
+    {
+        FeatureExtensions.RegisterFeature<GeometryFeature>(f => new GeometryFeature((GeometryFeature)f));
```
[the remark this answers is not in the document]
inforithmics: The Copy Feature does not work. But this Code is now created by a Source Generator so no need to do it by hand.

Mapsui/Extensions/FeatureExtensions.cs
inforithmics: This shouldn't happen, because I wrote a Source Generator that creates a static Constructor in which the Custom Feature gets registered. So that the Copy works for CustomFeatures to. I tested the Source Generator for Types that are already in Mapsui.
Writing this I discovered that I don't Source Generate for Features that inherit from another feature for example PointFeature the Copy does not work.

inforithmics: I wrote this code I don't know if I can ask if I'm in AOT Mode, and return (IFeature)Activator.CreateInstance(type, feature)!; Could still work but is not guaranteed in AOT Mode.

```
try
{
    // Fall Back if Type is not registered
    var type = feature.GetType();
    return (IFeature)Activator.CreateInstance(type, feature)!;
}
catch (Exception ex)
{
    Logger.Log(LogLevel.Error, ex.Message, ex);
    throw new NotSupportedException($"Register {feature.GetType().Name} so that it works in AOT Mode");
}
```

pauldendulk: I think that this is an acceptable fix. As soon as an inherited feature is introduced in AOT mode the user sees the error (when running in debug mode).

I think we should think about the need to copying in general. It would also be the responsibility of the user to implement a copy function. Or perhaps we should remove the need for copying entirely. We need copying at the moment in projecting and editing. There may be other ways to deal with this. 

Is there any other reason why the PR is WIP? Is don't see an objection to merge.

Samples/Mapsui.Samples.MapView/AnimatedMyLocationSample.cs
pauldendulk: "so MapView is now a namespace", I don't understand why, but maybe I need to learn more about source generators.

inforithmics:
Things done:
I use now ICloneable to Clone a Feature so a Custom Feature needs to implement IClonable for copying

inforithmics:
> I have one question left, and one remark about adding an exception. If fixed I will merge this asap.

It is done now with IClonable so no need for an exception

Mapsui/Layers/AnimatedLayers/AnimatedPointFeature.cs
```diff
@@ -7,6 +7,13 @@ internal class AnimatedPointFeature : PointFeature
 {
     long _startTime;
 
+    public AnimatedPointFeature(AnimatedPointFeature animatedPointFeature) : base(animatedPointFeature)
```
pauldendulk, line 10: I like that you still implement the copy constructor although the main purpose is to implement ICloneable.

Mapsui/Extensions/FeatureExtensions.cs
```diff
@@ -3,24 +3,20 @@
...
 using System.Text;
+using Mapsui.Logging;
 
 namespace Mapsui.Extensions;
 
 public static class FeatureExtensions
 {
     public static T Copy<T>(this T original) where T : IFeature
     {
-        return (T)Activator.CreateInstance(typeof(T), original)!;
```
pauldendulk, line 13: Activator.CreatInstance is now fully removed here. Good, it was problematic in other situations as well.

Doing a search on the code base I see there is still a Activator.CreatInstance in GradientTheme. Not used a lot but we use it in the ShapefileSample. What will users experience when they try to use it in Blazor AOT? I guess we can find out soon. For me it is important that the user understands what is going on. We do not have to fix all the scenarios. Let't not implement ICloneable for style. Perhaps we will convert all styles to records some day to get this functionality for free.

pauldendulk (approved):
Good decision to revert the source generators. It is always hard to make such a decision if you have already invested in that solution. The downside of ICloneable is that users have to implement it if they want to add a custom feature. That ix some work they need to do, but it also clear.

inforithmics:
I plan to do 2 follow up pull requests
1. Remove Activator from GradientTheme
2. Remove IDisposable from Map/Layer

pauldendulk:
Please create issues to discuss a possible solution. 

In the case of number 2 I think this could be a really big change which we should be preceded by a number of smaller PRs that prepare for it.

pauldendulk:
Here is an issue about making the Map non-disposable. https://github.com/Mapsui/Mapsui/issues/2372

PR #2354, opened 2023-12-17
Update scripts for rename of master to main

pauldendulk:


pauldendulk:
I did a search of 'master' in all files and changed to 'main' where appropriate. The only instances where we use 'master' is in urls pointing to other repositories.

PR #2365, opened 2023-12-27
GHA: build MAUI stuff on Linux as well

janusw:
This extends the GitHub-Actions build on Linux to include the MAUI parts as well. `Mapsui.UI.Maui` builds on Linux without a problem. The corresponding samples can only target Android on Linux (so the iOS and Mac targets are disabled in that case).

`Mapsui.Linux.slnf` is extended a bit with the Maui projects, but also some Uno.WinUI projects are removed (which can not actually be built on Linux).

Also the GHA config on Linux is simplified quite a bit (by using the slnf to determine which parts to build on Linux, instead of building multiple projects explicitly).


pauldendulk (approved):
Great, cool MAUI also builds on linux. Thanks for the simplifications.

PR #2371, opened 2023-12-28
WIP Build the general documentation with mkdocs

pauldendulk:


pauldendulk:
This uses mkdocs instead of docfx to build the website. Advantages:
- mkdocs, and particularly the material theme this PR uses, has great support and is a very active project with many options and plugins.
- The code samples now have a scrollbar, so all code can be seen without having to use wrap around.
- The code samples now have an option to copy the code when hovering over it.
- For the quickstarts we can now use a tabbed content block, making it more compact, and the start page becomes the single page for everyone to get started. My plan is to add more documentation after the quickstarts (on the same page) taking the users along with some more samples (the basics of navigation, adding points, adding widgets, etc).

The mkdocs site is currently deployed here: https://mapsui.com/documentation2/

At the moment I do not have plans to port the api documentation to mkdocs. There are quite a number of plugins for api documentation in mkdocs (look [here](https://github.com/mkdocs/catalog?tab=readme-ov-file#-api-documentation-building)), so I guess it is possible for Mapsui as well, but because we have a C# project it might be problematic. This means we have to deploy mkdocs and docfx side by side and integrate it somehow or just link from one to the other.

pauldendulk:
This is half done but I will merge it because (as far as I know) it does not break the existing website.

PR #2378, opened 2024-01-02
Delete PCLTimer

pauldendulk:
It was not used anymore. PCL refers to Portable Class Libraries, which preceded .netstandard, which we dropped and main, so very old code.


charlenni (approved):
This is useful, because things that no longer used should be removed from code.

PR #2377, opened 2024-01-01
Rework on Widgets

charlenni:
Second try. I now have made all changes in small parts with their own commits. I hope, I explained it all very well.

Run all test. They work without any error. I run WPF, MAUI, MapView and Blazor samples. Work. Found a bug in Blazor and removed it (last commit), which shouldn't have done here. Sorry.

PR #2380, opened 2024-01-02
Add logging debug

charlenni:
Add a LoggingWidget to Map if a debugger is attached. If the Map already contains a LoggingWidget do nothing. It is created one widget for different maps, so that the debug information isn't lost.

There are bugs in the renderer, where Opacity isn't respected correct in colors. Should be looked in after #2377.


charlenni:
A comment to the second commit: The problem is, that the LoggingWidget don't respect the opacity of the background color, but calculates it by `.WithAlpha(255 * Opacity)`. Correct would be `.WithAlpha(Color.Alpha * Opacity)`. If I set now the background to `Color.Transparent` which is `Color(255, 255, 255, 0)` and the Opacity to `0.8`, then I get `Color(255, 255, 255, 255*0.8)`, which isn't transparent anymore.

Will correct this, when working on the `LoggingWidgetRenderer` later.

pauldendulk:
> A comment to the second commit: The problem is, that the LoggingWidget don't respect the opacity of the background color, but calculates it by `.WithAlpha(255 * Opacity)`. Correct would be `.WithAlpha(Color.Alpha * Opacity)`. If I set now the background to `Color.Transparent` which is `Color(255, 255, 255, 0)` and the Opacity to `0.8`, then I get `Color(255, 255, 255, 255*0.8)`, which isn't transparent anymore.
> 
> Will correct this, when working on the `LoggingWidgetRenderer` later.

My idea (when commenting on an earlier PR) was to make the background fully transparent (and also full screen), so you would just see the logging message itself. A benefit would be that the map looks exactly the same apart from the text. A drawback could be that it is hard to read against the map background (depending on what background it is). What made you choose semi-transparant?

pauldendulk (approved):
Thanks, this is a very important feature. 

I think I would have gone for a solution where you create the widget in the map constructor. You need to check debug mode, but nothing else. The code would be simpler. 

A drawback would be that you loose the logging message if you switch to another Map, but perhaps that is not a bad thing. I could imagine switching several times between two Maps. If you switch to the previous one the last log message will be relevant to the currently selected Map. Perhaps this is better.

charlenni:
I used the semi-transparent background to check, where the widget is displayed. And while doing this, I found the problem, that the Opacity isn't set correct. I only made the comment, so that I remember this later and change it.

I'm not sure, if adding to MapControl or MapControl is better. For both could be found pros and cons. If you like a more Map centric version, we can move it.

One problem is the full screen. You have to take into account, that the size of the MapControl changes. And if you think further about it, you will find, that this is also a good feature for other widgets. With this in mind I start the refactoring of widgets.

charlenni:
But: Thank you for looking into this PR!

PR #2381, opened 2024-01-02
[WIP] Marker

charlenni:
First efforts to integrate MapView into MapControl.

The idea is to lower the threshold for using Mapsui. One of the most common requests for maps is to set a marker. Marker because Google Maps and Mapbox use this term.

Along with the marker come extensions for the MemoryLayer to create and set these markers quickly and easily. The handling of the style is done at the same time.

In the end, in combination with the extensions for the map or the MapBuilder, it should be possible to easily create basic maps. If more in-depth options are required, then Mapsui functions can still be used directly.

Any further ideas or comments?


charlenni:
@pauldendulk I know, the next long PR, but I started this time with small steps and tried to describe what I did very precise. And the most is new code, no changes.

Meanwhile, I had to derive an own layer for Markers: MarkerLayer. I had to do this, because the clicked marker, that show the callout was behind other markers. So I had to change `GetFeatures()` to change the order: first all markers without a callout, then all markers with a callout. With this, the callouts always on top. A problem is the wrong order of features because of the ConcurrentBag usage in MemoryLayer (see #2389).

Mapsui/Extensions/MapExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// Add a layer for <see cref="Marker">
+    /// </summary>
+    /// <remarks>
+    /// This layer should be the topmost <see cref="Layer"> in a <see cref="Map">, so that the <see cref="Callouts">
+    /// are always on top.
+    /// </remarks>
+    /// <param name="map">Map to add this layer too</param>
+    /// <param name="name">Name of layer</param>
+    /// <returns>Created MemoryLayer</returns>
+    public static MemoryLayer AddMarkerLayer(this Map map, string name)
```
pauldendulk, line 19: An extension method approach. I like it.

Mapsui/Features/Marker.cs
```diff
@@ -0,0 +1,265 @@
...
+using Mapsui.Styles;
+using Mapsui.Utilities;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Color = Mapsui.Styles.Color;
+
+namespace Mapsui.Features;
+
+public class Marker : PointFeature
```
pauldendulk, line 11: I want to avoid having adding another feature type. Also, it contains a lot of code, this needs to be maintained.

Mapsui/Features/MarkerType.cs
```diff
@@ -0,0 +1,222 @@
...
+
+namespace Mapsui.Features;
+
+[DefaultValue(Pin_Circle)]
+public enum MarkerType
+{
+    /// <summary>
+    /// Pin could change color
+    /// </summary>
+    Pin_Circle,
```
pauldendulk, line 11: All these different types depend on images that now have to be in core. This is something to avoid.

Mapsui/Features/Marker.cs
```diff
@@ -0,0 +1,265 @@
...
+            UpdateMarker();
+        }
+    }
+
+    /// <summary>
+    /// Title of callout
+    /// </summary>
+    public string? Title
+    {
+        get => _calloutStyle.Title;
```
pauldendulk, line 108: This kind of wrapping is a lot of extra code. Instead you could set the properties of the callout direct. Perhaps you did it like this to trigger the property changed events.

Samples/Mapsui.Samples.Common/Maps/Special/ExtensionsSample.cs
```diff
@@ -0,0 +1,123 @@
...
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Mapsui.Samples.Common.Maps.Special;
+
+public class ExtensionsSample : ISample
+{
+    public string Name => "Extensions";
```
pauldendulk, line 19: This is called Extension, should it not be called Marker?

Mapsui/Extensions/MemoryLayerExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+/// </summary>
+public static class MemoryLayerExtensions
+{
+    /// <summary>
+    /// Add a <see cref="Marker"/> to the layer
+    /// </summary>
+    /// <param name="layer">Layer to use</param>
+    /// <param name="x">X position</param>
+    /// <param name="y">Y position</param>
+    public static MemoryLayer AddMarker(this MemoryLayer layer, double x, double y, MarkerType type = MarkerType.Pin_Circle, string? title = null, string? subtitle = null, Styles.Color? color = null, double scale = 1.0, Offset? anchor = null, Offset? calloutAnchor = null, Action<ILayer, Marker, MapInfoEventArgs>? touched = null)
```
pauldendulk, line 21: So, it is only possible to add Markers to extension layers. I understand that it is hard to fix this with the current layer setup - you tackled this elsewhere - but if so perhaps we should not add this functionality now.

pauldendulk (commented):
I don't want to add Markers in this form. There is a lot specific code in the Marker feature type itself that would become part of core. I also have a problem with the BelongsTo field, coupling styles. We should look at other options to add the same functionality, and if that fails maybe we have to do without this functionality in core. I will close this PR. We should discusses an alternative in a new issue.

An alternative approach is to work with StyleCollections and ThemeStyles. I am not sure you can achieve the same thing. Perhaps we need to rewrite core to make it possible. Also related is the `IFeature.Data<T>` field that I have in mind. I should add a US for it.

pauldendulk:
Somewhat related is my post here: https://github.com/Mapsui/Mapsui/issues/2452.

Mapsui/Styles/BitmapRegistry.cs
```diff
@@ -46,6 +46,15 @@ public int Register(object bitmapData, string? key = null)
     {
         _register.TryGetValue(id, out var val);
         _register.Remove(id);
+
+        KeyValuePair<string, int>[] temp = new KeyValuePair<string, int>[_lookup.Count];
```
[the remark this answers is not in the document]
charlenni: If you remove the bitmap from the register, it no longer exists. Then you have to remove the bitmap also in the _lookup, because some entry of _lookup now point to not existing bitmaps.

It's making things slightly better, but is far from perfect.

Mapsui/Styles/CalloutStyle.cs
```diff
@@ -297,6 +297,15 @@ public float ShadowWidth
...
     }
 
+    /// <summary>
+    /// SymbolStyle to which this Callout belongs
+    /// </summary>
+    /// <remarks>
+    /// The callout is positioned relative to this symbol. 
+    /// This is needed for calculation of relative SymbolOffset. 
+    /// </remarks>
+    public SymbolStyle? BelongsTo { get; set; }
```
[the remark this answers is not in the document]
charlenni: I could remove this, but callouts are placed to another object/style. In this case, the image of the pin.

I try to remove this and add the calculation to the Marker.

Mapsui/Features/Marker.cs
charlenni: That could get difficult. To replace the whole MapView Pin implementation with no new code is not trivial. And to make such thing only with extension methods make things not easier.

By the way, then lets make a Mapsui.MapView. Or a Mapsui.Extensions. As I said always: MapView is a convenience layer around Mapsui. All what Marker does could be explained by documentation. Nothing new. But it seems, that users don't read documentations and want to use things out of the box.

If you don't want to add new code, but want to get rid of MapView, then delete MapView. There is nothing, that couldn't be done with Mapsui by its own.

Mapsui/Features/MarkerType.cs
charlenni: That is the thing, that convenience layers do. Needed two hours to make, but could be done by everyone else.

Mapsui/Features/Marker.cs
charlenni: Yes, don't need this. Could have done this by exposing CalloutStyle or by accessing Marker.Styles[1], but then it wouldn't be a convenience layer. There isn't any magic in Marker. Nothing you couldn't do by yourself.

Samples/Mapsui.Samples.Common/Maps/Special/ExtensionsSample.cs
charlenni: Could be. But there is also a MemoryLayer extension. And at the end there will also be a MapExtension. It is WIP, so you could choose whatever you want. We can also make an own group. You decide.

charlenni:
Ok, @pauldendulk, you have to decide, what you want. One milestone of Mapsui 5 is to get rid of MapView and integrate it into MapControl. MapView added no functionality to Mapsui, that wasn't there before. The pin layer adds some special PointFeatures to a layer with some added functionality. Nothing else special. You could make this by your own with only using Mapsui.

So I see three ways to get rid of MapView:
1. Integrate its functionality to MapControl
2. Add more documentation
3. Delete it and do nothing

All three ways would work. As you see, I don't have any problems with all this three ways. I know how things work in Mapsui. I don't need Markers, because I could implement them very fast by my own. But it seems difficult for our users.

Up to now, if I understand it correct, you don't want to have more code. Ok, then let's transfer MapView into documentation. Make a good documentation, how a user could display a "marker" on the map. Make good documentation about usage of styles, renderers and so on. And say to the users, they have to read it. No problem with this.

But to be clear, all extension methods for Map are also new, not needed code, because you could do this all by your own.

inforithmics:
I didn‘t look to closely on the implementation. But the part I really liked was the extension method for adding Markers. If it should use a Marker Feature or existing other Features I don‘t have a strong opinion. But it should be convenient and easy to add a marker. Because this is one of the first things that a user of mapsui might use. I‘m saying this because I used first the MapView Control and it was easy to start with.

Mapsui/Styles/BitmapRegistry.cs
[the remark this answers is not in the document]
pauldendulk: ah, so we need this fix anyway.

pauldendulk:
My suggestion, hinted at in the comment, is to add AddMarkerLayer Map extension which creates a layer (perhaps a MemoryLayer). Adds a StyleCollections which contains a regular SymbolStyle and on top of that a ThemeStyle for the callouts. The ThemeStyle has a method which can be used to either return a CalloutSymbol or not, depending on the state of the Feature. Perhaps we need to look into `IFeature.Data<T>` to make this easier. Perhaps it is not this easy, if so, we should consider adjustments to make it easier.

Mapsui/Styles/BitmapRegistry.cs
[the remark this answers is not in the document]
charlenni: Yes, this is a bug and should be removed by its own. I didn't see it, when I tried to make own PR for bugs.

charlenni:
But this is a totally different approach. You suggest a layer, that holds a SymbolStyle. This style holds perhaps the color of the marker. On this, the theme style holds the function, that could access the marker image and if the callout is shown. This callout is then in a third style. A style for the layer or the feature? Have to test this.

Ok, then I try this.

pauldendulk:
> I didn‘t look too closely on the implementation. But the part I really liked was the extension method for adding Markers. If it should use a Marker Feature or existing other Features I don‘t have a strong opinion. But it should be convenient and easy to add a marker. Because this is one of the first things that a user of mapsui might use. I‘m saying this because I used first the MapView Control and it was easy to start with.

@inforithmics Yes, the extension method approach is what we are striving for. In the extension methods we compose the core components to something that implements the users requirements in a single statement. Ease of use is important.

But I also want to emphasize that if I have to choose between **ease of use** and **ease of maintenance** I will choose the latter. Also if everyone else has a different opinion. 

I said something about ease of use [here](ttps://github.com/Mapsui/Mapsui/issues/2452). Some more on that. @inforithmics You mention it was easy to get started with MapView. So, with ease of use we could get more active users. Users will be happy with Mapsui on day one and start using it in their project. But if later it turns out that their app crashes in production because Mapsui has cross thread exceptions they will not be happy at all, and will regret to choice for Mapsui. So, I don't want a lot of users that way. **Instead, I want a lot of happy users**.

If I choose **ease of maintenance** over **ease of use** this is not just to make things easy for myself (although I am committed to make things easy for myself) but it is also to make sure our users are happy in the long run.

charlenni:
@pauldendulk I have to apologize, if you really think, that the extensions approach is easier to maintain. We will get a huge file with many big functions, that check again and again if this is the right thing to do with the given object. Will be horrible to maintain.

pauldendulk:
Pure functions in combination with immutable data is easier to maintain than loads of stateful objects communicating their state with each other. This is not just my opinion but a widely shared view. It is not the best solution for every problem however, and statefulnes is necessary in almost every application. If you need to check that state again and again there is something wrong in the design, and we need to tackle that. 

Past weeks if have spend a lot of time trying to explain the you the general ideas behind good software design. I see you do try to comply with my requests - thank you for that - but at some point you yourself need to see the benefit of this yourself. It does not look like this is happening. If you only do this because I ask you too in every PR this is going to be a very laborious process. When you start on a PR using a certain design you need to understand yourself why this is a good design. If you don't it is better you do not start at all and focus on other things.

charlenni:
I think I have at least understood the last hint correctly.

I would like to apologise for wasting your time. Unfortunately, I seem to have more enthusiasm than skill, which doesn't always lead to the best results. I therefore think it's best not to waste any more valuable resources. I'm putting my PRs on hold for now. Then there will be more time for the important problems.

PR #2382, opened 2024-01-02
MapControl code cleanup

pauldendulk:


Mapsui.UI.Blazor/MapControl.cs
```diff
@@ -15,7 +14,7 @@ public partial class MapControl : ComponentBase, IMapControl
     public static bool UseGPU { get; set; } = false;
 
     private readonly SKCanvasView? _viewCpu;
-    private readonly SKGLView? _viewGpu;
+    protected readonly SKGLView? _viewGpu;
```
charlenni, line 17: I'm not sure, what the problem was. Do Blazer need access to this fields? Any problems with the samples?

charlenni (approved):
All things make sense. For the last commit I'm not sure about the reason, but it turns things back, so I have no problem with it.

Mapsui.UI.Blazor/MapControl.cs
pauldendulk: This is because our Blazor solution uses a MapControlComponent.razor which inherits from the MapControl. The MapControlComponent is what the users should use and not the MapControl. So Blazor is differs from the other platforms in that way. I am not sure if there is another solution, where all code is a MapControl class. At least we could rename MapControlComponent.razor to MapControl.razor and the Blazor base class to something else.

I broke the build inadvertently because I did not have the inheritance clear. Also ctrl+f12 in visual studio does not show the inherited class.

PR #2383, opened 2024-01-02
Replace float with double in MapControl

pauldendulk:
Changed it where it made sense but that was everywhere expect PixelDensity. This was because otherwise we had to do a lot of casts in the skia Scale methods.

### DIFF
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -48,9 +48,9 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
     private readonly SKPicture _pictZoomIn;
     private readonly SKPicture _pictZoomOut;
     private readonly SKPicture _pictNorthing;
-    private readonly ObservableRangeCollection<Pin> _pins = new ObservableRangeCollection<Pin>();
-    private readonly ObservableRangeCollection<Drawable> _drawables = new ObservableRangeCollection<Drawable>();
-    private readonly ObservableRangeCollection<Callout> _callouts = new ObservableRangeCollection<Callout>();
+    private readonly ObservableRangeCollection<Pin> _pins = [];
+    private readonly ObservableRangeCollection<Drawable> _drawables = [];
+    private readonly ObservableRangeCollection<Callout> _callouts = [];
 
     public MapView()
     {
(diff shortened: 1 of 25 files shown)


Mapsui.UI.WinUI/MapControl.cs
```diff
@@ -251,14 +251,11 @@ public void OpenBrowser(string url)
...
-    private float ViewportHeight => (float)ActualHeight;
-
-    private float GetPixelDensity()
-    {
-        return (float)(XamlRoot?.RasterizationScale ?? 1f);
-    }
+    private double ViewportWidth => ActualWidth;
+    private double ViewportHeight => ActualHeight;
 
+    private double GetPixelDensity() => XamlRoot?.RasterizationScale ?? 1f;
```
charlenni, line 257: `GetPixelDensity()`is here a double, but if `RasterizationScale` is null, it is filled with float.

charlenni:
Most of the code is ok. Some parts are not possible for me to check (GlobalSuppressions), because I don't know anything about this.

Regarding the PixelDensity: I would say, that we should change this too, even if there are many changes. It is a one time problem, but then we have a strict line in Mapsui to use everywhere double.

Regarding the build error: I don't see any changes in the part that fail. Strange.

So from my side it is ok.

pauldendulk:
> Most of the code is ok. Some parts are not possible for me to check (GlobalSuppressions), because I don't know anything about this.

The GlobalSuppressions are just another way to suppress a warning. The advantage is that it does not clutter the code. I removed the global suppressions of which the IDE says they were not needed. In some cases such messages can be ambiguous. For instance, you can remove a warning for MAUI but not for UNO. So, sometimes that is something I find out later. Yet, I remove them anyway because I need to know if they are needed only in one context. In such case we should add a comment or another way to tell this to other developers. Perhaps put the suppression in a `#if _UNO_`.

> Regarding the PixelDensity: I would say, that we should change this too, even if there are many changes. It is a one time problem, but then we have a strict line in Mapsui to use everywhere double.

I am not yet fully convinced yet we should go for all float. In my past code bases there were good reasons to use part double and part float. It looks like Mapsui could be all double but I want to hold out that decision a bit until I had a closer look.

> Regarding the build error: I don't see any changes in the part that fail. Strange.

It was the same reason my earlier PR failed. The Blazor MapControl inheritance.

Mapsui.UI.MapView/MapView.cs
```diff
@@ -480,9 +478,9 @@ private void HandlerViewportChanged(object? sender, ViewportChangedEventArgs e)
 
     private void HandleLayersChanged(object? sender, LayerCollectionChangedEventArgs args)
     {
-        var localRemovedLayers = args.RemovedLayers?.ToList() ?? new List<ILayer>();
-        var localAddedLayers = args.AddedLayers?.ToList() ?? new List<ILayer>();
-        var movedLayers = args.MovedLayers?.ToList() ?? new List<ILayer>();
+        var localRemovedLayers = args.RemovedLayers?.ToList() ?? [];
+        var localAddedLayers = args.AddedLayers?.ToList() ?? [];
+        var movedLayers = args.MovedLayers?.ToList() ?? [];
```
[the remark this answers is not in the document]
pauldendulk: Yes I had a bit of hesitation but I expect this is the kind of thing that we will get used to. Almost all C# changes that made the code more succinct I ended up liking after a while. The IDE suggest this, that makes a difference too. This seems be more 'go with the flow'.

Mapsui.UI.WinUI/MapControl.cs
pauldendulk: Good observation. It makes more sense to have this 1d. I changed it.

pauldendulk:
Also, changed Convert.ToSingle to (float).

We used 9 `Convert.ToSingle` in our code and 128 `(float)`. I am not used to Convert.ToSingle and it seems it does the same thing and has no advantages, so I made it (float) as well.

https://stackoverflow.com/questions/34090397/convert-double-to-float-by-cast-or-convert-tosingle

charlenni (approved):

PR #2388, opened 2024-01-04
Even more cleanup on MapControl

pauldendulk:


Mapsui.UI.Maui/GlobalSuppressions.cs
```diff
@@ -13,4 +13,4 @@
...
 [assembly: SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP001:Dispose created", Justification = "<Pending>", Scope = "member", Target = "~M:Mapsui.UI.Objects.MyLocationLayer.UpdateMyLocation(Mapsui.UI.Maui.Position,System.Boolean)")]
 [assembly: SuppressMessage("IDisposableAnalyzers.Correctness", "IDISP001:Dispose created.", Justification = "<Pending>", Scope = "member", Target = "~M:Mapsui.UI.Maui.Pin.CreateFeature")]
-[assembly: SuppressMessage("Usage", "CA2211:Non-constant fields should not be visible", Justification = "<Pending>", Scope = "member", Target = "~F:Mapsui.UI.Maui.MapControl.UseGPU")]
+[assembly: SuppressMessage("Usage", "CA2211:Non-constant fields should not be visible", Justification = "In general we should adhere to this rule but there can be exceptions. This static property is currently used a lot and not causing problems. Let's keep it", Scope = "member", Target = "~F:Mapsui.UI.Maui.MapControl.UseGPU")]
```
pauldendulk, line 16: In another thread we talked about the suppressions and that it is hard to understand why it is needed. We 'Justification' field of the suppression can be used to explain why suppression was needed.

Mapsui.UI.Maui/MapControl.cs
```diff
@@ -48,13 +34,6 @@ static MapControl()
         DeviceInfo.Platform != DevicePlatform.MacCatalyst &&
         DeviceInfo.Platform != DevicePlatform.Android;
 
-    private class TouchEvent(long id, MPoint screenPosition, long tick)
```
pauldendulk, line 51: Only the Location of the TouchEvents was used, so simpler to work with only the location.

Mapsui.UI.Wpf/MapControl.cs
```diff
@@ -46,7 +46,6 @@ private void Initialize()
             else RunOnUIThread(InvalidateCanvas);
         };
 
-        Children.Add(WpfCanvas);
```
pauldendulk, line 49: Some old XAML rendering relics that could be removed.

Mapsui.UI.Wpf/MapControl.cs
```diff
@@ -115,7 +102,7 @@ private static SKElement CreateSkiaRenderElement()
         };
     }
 
-    [Obsolete("Use Info and ILayerFeatureInfo")]
+    [Obsolete("Use Info and ILayerFeatureInfo", true)]
```
pauldendulk, line 105: By adding the 'true' to the FeatureInfo using it will result in a compile error which shows the message of the attribute.

pauldendulk:
These are things I run into when I sit down to figure out how to reorganize the MapControl.

PR #2392, opened 2024-01-06
Cached LoggingWidget improvement and Removing Logging TextBoxes

inforithmics:
Cached LoggingWidget improvement by updating the map of the cached logging widget
And removed Logging Textboxes from samples

see https://github.com/Mapsui/Mapsui/pull/2377#issuecomment-1879821981


charlenni:
I will look into this in the next days. But not before the Widget rework isn't done. The greater idea is to remove Map from all widgets. I'm not sure, if this is possible.

Would it be possible to to remove the widget from rendering?

charlenni:
Would it be easier to create the LoggingWidget in the constructor for the Map instead of MapControl?

inforithmics:
The setting of the map isn't necessary with the widget improvements pull request. But the widget creation should stay in the current position because only when a debugger is attached the widgetlogging is created.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -68,7 +68,7 @@ public partial class MapControl : INotifyPropertyChanged, IDisposable
     // keeps track of the widgets count to see if i need to recalculate the touchable widgets.
     private int _updateTouchableWidget;
     // Default LoggingWidget when the debugger is attached
-    private LoggingWidget _loggingWidget;
+    private LoggingWidget? _loggingWidget;
```
pauldendulk, line 71: I already saw the warning in the build log. Good fix.

pauldendulk:
> Would it be easier to create the LoggingWidget in the constructor for the Map instead of MapControl?

I would prefer that. I expect we could still check if the debugger is attached if we do it like that.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -344,6 +344,8 @@ private void CheckForLoggingWidget(Map map)
                     Opacity = 0.0f,
                     LogLevelFilter = LogLevel.Trace,
                 };
+            else
+                _loggingWidget.Map = map;
```
pauldendulk, line 348: As mentioned elsewhere we should come up with a different mechanism for refresh, but this is the fix in the current state.

Mapsui/Widgets/LoggingWidget/LoggingWidget.cs
```diff
@@ -283,6 +283,15 @@ public Color InformationTextColor
         }
     }
 
+    public Map Map
```
pauldendulk, line 286: Ugly but necessary for now.

pauldendulk (approved):
All good improvements. I want to review smaller merge requests. This is small merge request, great. It would be even better if it was still smaller. Just removing the old logging widget would have resulted in no discussion at all, such a PR could be merged from the phone in the train.

pauldendulk:
> I will look into this in the next days. But not before the Widget rework isn't done.

I am going to merge the PR. This is small PR, that is a good thing. By blocking a smaller PR with a bigger PR we go in the opposite direction.

Issue #2375, opened 2023-12-31
GeometryFeature disappears while panning


nm4568:
**Mapsui Version**
Mapsui.Maui (4.1.2)

**Mapsui Platform**
MAUI

**Device**
iPhone 15 Pro Max iOS 17.2 Simulator

**Describe the bug**
GeometryFeature dissappears when dragged north.

**Video**
https://github.com/Mapsui/Mapsui/assets/3853981/0b81aa4a-3ef5-495d-b911-1dadaf862382

**To Reproduce**
1) Launch sample app.  Initial state of the app has a Polygon that is draggable.
2) Without zooming, drag the polygon up the East Coast of the United States with the goal of centering the polygon over Washington D.C.  This will require a little bit of dragging/panning.

Expected Result - The end result should be a polygon centered over Washington D.C.
Actual Result - The polygon disappears around Charleston, South Carolina.

**Minimal Reproducible Sample**
https://github.com/nm4568/MapsuiSamples/tree/bug/polygon-drag-disappearing

pauldendulk:
Thanks for you excellent issue. I will not have time to check this out today. A possible cause is that something goes wrong with caching. @inforithmics Do we have a way to turn off caching? Or could we fully clear the render cache on each iteration?

One more thing you could do is to attach a logger to see if an error is logged when the polygon disappears.

nm4568:
You're quite welcome.  I don't see anything logged to the debug output window.  I also turned on CLR Exceptions within Exception Settings to see if maybe an exception was being swallowed but  no exception is thrown.

pauldendulk:
I would investigate this by creating IRenderCache that does no caching. But that takes some figuring out and perhaps some changes to the code to make this possible. We already have a NonCachingVectorCache, but is only the VectorCache, the RenderCache takes a number of caches:

```csharp
public interface IRenderCache : ILabelCache, ISymbolCache, IVectorCache, ITileCache, IDisposable
{
    IVectorCache? VectorCache { get; set; }
    ILabelCache LabelCache { get; set; }
    ISymbolCache SymbolCache { get; set; }
    ITileCache TileCache { get; set; }
}
```
This is not something I can look at soon.

nm4568:
I started investigating this issue and I am seeing the anatomy of the GeometryFeature which represents the orange draggable polygon in the minimal reproducible sample app are all stored in the Vector Cache.

**Anatomy of Draggable Polygon**
- Pen - Mapsui.Styles.Pen
[...]

PR #2393, opened 2024-01-07
Call GeometryChanged in Modified to make editing easier

pauldendulk:
This may help avoid some users frustration. When updating the NTS coordinates you need to call GeometryChanged and GeometryChangedAction. This is something most users will not know about. This will not be necessary anymore if we do this in the Modified() method, which they need to call anyway.


pauldendulk:
This would prevent the problem encountered in this issue: https://github.com/Mapsui/Mapsui/issues/2375

PR #2394, opened 2024-01-07
Some modifications of the EditingWidget

pauldendulk:
I am working on this as part of a bigger effort to improve and touch/mouse handling. 

In this PR I am working towards more shared code regarding editing. Editing on mobile is currently not fully supported. Currently the shift key is needed in some scenarios. This should be reworked into a more general state without not depending on shift. How to determine the state should not be part of the MapControl but of the EditingWidget itself. The EditingWidget may need additional information to determine this.

This first push is a very early stage where I made the ShiftPressed field private.

When working on this I ran into several issues and questions.
- On some platforms, like Eto, editing does not work at all.
- An Avalonia KeyDown/Up does not seem to be triggered in this branch (perhaps the original works correctly)
- Blazor has a more generic way to determine the key state (including shift/alt/ctrl). Perhaps this should be applied to all platforms. On the other hand, if we want to support mobile as well, we should perhaps move away from key modifiers entirely.
- All platforms have different ways to determine the key state. 
- Only WPF uses a global state (which seems such an obvious simple solution, but perhaps there is a good reason other platforms do not have this). 
- On the other platforms you need to determine the key modifiers by listening to events but not all events have the key modifier information, and this is why we need to store the shiftpressed state.
- Avalonia 'add line' and 'add polygon'  removes the last point (this is because of a difference in the number of touch events triggered).
- And a bigger thing: We have to straighten out how to work with events. The different platforms listen to different types of events and this can result in a different number of widget touch events. For instance: MouseUp/TouchUp/PointerUp (pointer or touch up), raw touch events like received from the skia control, Tapped, DoubleTapped.
- A thought: If MAUI, Uno and Avalonia app developers do not have key modifier in the DoubleTapped event perhaps we do not need that either. Our widgets do not need to be better than those platforms.


pauldendulk (commented):
Sorry I could not make time for this PR yet. I glanced through the code, I see a lot of good improvements, and some great (like reducing the number of renderers needed). So far I have two concerns. 

I am still hesitant about INotifyPropertyChanged in the widgets. We discussed this and decided to merge this including the INotifyPropertyChanged and will we revert it later if we choose some other mechanism.

This changes the widget event logic. This worries me because the current main branch has bugs in it. In the previous PR there we still many bugs that had to be merged from 4.1. I am working on this problem in this PR https://github.com/Mapsui/Mapsui/pull/2394. The reason it has bugs is that it is not clear how it should work. Our touch handing code is the messiest part of the project, and it is one of the main things to improve in v5. So, I would rather sit back and investigate, figure out how it works and then decide how it should work and only then change it. Then again if this PR introduces no fundamental changes in the widget events handling there is no problem. I want to investigate this by checking out this branch and to step though the event code. This is something I can not do before Friday. So, this will take more time.

charlenni:
@pauldendulk I assume, that you are not talking about this PR made by yourself, but about #2377, correct?

If you select a good name for the flag to set this widget to a dirty state, then I will do it right after the merge of #2377.

Regarding the event handling: I told you, that I changed nothing in respect of handling of events. All the problems are still there, but much more concentrated in one interface. If you had problems with the widgets handling before, you will still have the same problems after this PR.

Take your time, no problem for me. I will work on other PRs ;-)

pauldendulk:
Oh, I made a mistake. I will add it to the widgets PR.

Mapsui.Nts/Editing/EditManipulation.cs
```diff
@@ -5,13 +5,12 @@
 
 namespace Mapsui.Nts.Editing;
 
-public enum MouseState
+public enum PointerState
```
pauldendulk, line 8: Using 'Pointer' to refer to both Touch and Mouse (where Tap is used to refer to both a 'click' or 'touch tap').

Mapsui.Nts/Editing/EditManipulation.cs
```diff
@@ -87,14 +93,9 @@ public class EditManipulation
 
                     return false;
                 }
-            case MouseState.Moving:
+            case PointerState.Hovering:
                 editManager.HoveringVertex(mapControl.GetMapInfo(screenPosition));
                 return false;
-            case MouseState.DoubleClick:
```
pauldendulk, line 93: DoubleClick was not used in the current code.

Mapsui.Nts/Editing/EditManager.cs
```diff
@@ -53,9 +49,6 @@ public bool EndEdit()
         }
         else if (EditMode == EditMode.DrawingPolygon)
         {
-            // Polygon needs to be at least 3 vertices
-            if (_addInfo.Vertices.Count > 3)
-                _addInfo.Vertices.RemoveAt(_addInfo.Vertices.Count - 1); // correct for double click
```
pauldendulk, line 58: The correction for double click is not needed when using shift.

Mapsui.UI.Blazor/MapControl.cs
```diff
@@ -383,7 +383,10 @@ public async void OpenBrowser(string url)
 
     }
 
-    public bool ShiftPressed => _pressedKeys.Contains("ShiftLeft") || _pressedKeys.Contains("ShiftRight") || _pressedKeys.Contains("Shift");
+    private bool GetShiftPressed()
+    {
+        return _pressedKeys.Contains("ShiftLeft") || _pressedKeys.Contains("ShiftRight") || _pressedKeys.Contains("Shift");
```
pauldendulk, line 388: When storing the shift state for later use in events we are trying to do better than the platforms themselves. We should just pass the information that we have, if there is no shift we pass on no shift. Perhaps we should use an enum instead of bool with: True, False, NotSupported.

Mapsui.UI.Maui/MapControl.cs
```diff
@@ -248,7 +248,7 @@ private async void OnTouch(object? sender, SKTouchEventArgs e)
                     _pointerDownTicks = DateTime.UtcNow.Ticks;
                 }
 
-                if (HandleWidgetPointerDown(location, true, Math.Max(1, _numOfTaps), ShiftPressed))
+                if (HandleWidgetPointerDown(location, true, Math.Max(1, _numOfTaps), false))
```
pauldendulk, line 251: In MAUI shift is not support by Mapsui.

pauldendulk (commented):
There are so many questions wrt to the editing control I now think we should leave it aside for now to focus on more fundamental things. Some remarks:

- The editing can only be used properly with shift and this is not supported on many platforms.
- Why use shift specifically, many other options are imaginable. Should we send all key states if we know them?
- MAUI does not support shift although you would expect it on Windows.
- Avalonia and WPF have a different number of touch events with the same actions.
- We now have three event types up/down/moving. Down is only used in editing. Moving in editing and mouse control. Up is somewhere in between an Up and a Tap event.
- We should not have to depend specifically on the shift key. There should be some general state, and using shift is one way to set that state. Perhaps the events args should have an action which calls back to the MapControl. So, there can be a platform specific implementation of the 'special state'.
- The shift state is used for two different cases. 1. Deleting a vertex in a polygon. Perhaps it should be seen as the delete state.  2. Closing a polygon. 
- Perhaps a button can also be used to set the special state.

PR #2400, opened 2024-01-08
Remove RectFeature

charlenni:
Why do we still have RectFeature?

RectFeature inherits from BaseFeature and is used only in StackedLabelProvider and one sample to draw a frame around an area. So I would say it isn't useful and should be dropped. It could be replaced by a GeometryFeature with a Polygon Geometry. That is the reason for this PR.

Btw. what is with PointFeature? This is more widely used, but is now a double for `GeometryFeature(Point())`. The only problem is, that this could only access, if Mapsui.NTS is added. So I would say, that this takes more effort to realize.


charlenni:
Perfect. I expected this and I'm fully with you regarding PointFeature. With this in mind, could you look also into #2399?

PR #2401, opened 2024-01-08
Use RelativeOffset instead of Offset

charlenni:
There is a RelativeOffset that derives from Offset and could be used better instead of the IsRelative flag of Offset. So I added a `CalcOffset()` function, that returns in Offset the offset and in RelativeOffset the new calculated offset from width and height and use this in all renderer calculations for the offset. So the code is easier and better readable.


pauldendulk (changes requested):
Clearly a good improvement. This is the kind of thing to fix in this phase. I added one comment. Ideally you would want to abstract the implementation away. Perhaps this can be improved. If not I will merge it in this form.

One other thing. The CalcOffset method is now optional. In a new renderer the developer could forget to call it. It would be better if there is no way around it. I did not think of a good solution for that.  I can think of ways to enforce it, but not nice ways.

charlenni:
Sorry @pauldendulk, but I couldn't see any comment.

Regarding the abstraction: I assume, that you talk about the `CalcOffset()` function in `Offset`. If yes, then this isn't possible. If we do this, then we have no implementation in `Offset` and have to use `if`'s everywhere. More code.

Regarding that it could be forgotten: Yes, could be. But I couldn't think of any good strategy to force the developer to use it.

pauldendulk:
oh, I indeed, my comment is missing. I was talking about the fact that `is RelativeOffset' is still necessary in the renderer. A further improvement would be to abstract that away entirely behind some interface. That will probably be possible in some form or other (but not always better). It may need more basic changes. This is not something that you should look into. 

I am striving for smaller PRs so I will merge this in it's current form. It is correct and an improvement. There are still some questions I have which could be investigated later. I will look into those myself.

### Possible follow ups
- Investigate if we can use just an interface in the renderer
- There are (before and after this PR) two Offsets, one absolute and one relative. The relative one is now derived from the absolute one, which is a bit skewed. Better if there was an AbsoluteOffset derived from a base class.

pauldendulk (approved):

Issue #2402, opened 2024-01-09
Regression tests crash on GeoTiff


pauldendulk:
**Mapsui Version**
Current main

**Mapsui Platform**
Regression tests

**Device**
github build server

**Describe the bug**
In more than 50% of the cases the build fails with an error related to geotiff.

This is part of the logging. The full log is in the link below:
```console
ReadDirectory: 
Warning, 
C:\Users\runneradmin\AppData\Local\Mapsui.Samples\example.tif: unknown field with tag 33550 (0x830e) encountered
ReadDirectory: 
Warning, 
C:\Users\runneradmin\AppData\Local\Mapsui.Samples\example.tif: unknown field with tag 33922 (0x8482) encountered
ReadDirectory: 
Warning, 
in-memory: unknown field with tag 33550 (0x830e) encountered
ReadDirectory: 
Warning, 
in-memory: unknown field with tag 33922 (0x8482) encountered
The active test run was aborted. Reason: Test host process crashed : Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
Repeat 2 times:
```

https://github.com/Mapsui/Mapsui/actions/runs/7444661574/job/20251467790

To it seems this error started occurring after this PR (but I did not check the dates):
https://github.com/Mapsui/Mapsui/pull/2240/files

@inforithmics Is this something that you could take a look at?

inforithmics:
When I debugged it last time error like
in-memory: unknown field with tag 33922 (0x8482) encountered happen in geotiff but don't crash it. So my guess it is another process, I'll have a look at it.

charlenni:
I had the same feeling. If you look into the report, it has problems to draw a path for a SymbolStyle. And there is no SymbolStyle in the GeoTiff sample. 

That was the reason for PR #2403. And I couldn't reproduce it local.

pauldendulk:
It is not impossible that the geotiff error logging has nothing to do with the failed build and something else is flaky, but I have seen this message many times and not always in combination with other errors (Although it is easy to miss something in those big build logs). So, I still suspect geotiff, also because I think to have seen an increase in failed builds after that PR I pointed to.

[...]

PR #2403, opened 2024-01-09
Add output of which MapRegressionTest is started

charlenni:
For me it is not clear which test is running before a crash happens. So I added a log entry with the name of the test.


charlenni:
Added a log entry before and after the test.

This PR is in conjunction with #2402. For me it is not clear, if the crash happens while running th GeoTiff sample or a sample later. The crash occurs when a SymbolStyle should be drawn. But there isn't a SmbolStyle in the GeoTiff sample. So I added a start and end flag for each sample that runs to make it easier to find the right sample. If I have overseen this information and this PR is useless, then we can close it.

inforithmics:
It crashed after "End MapRegressionTest MbTilesSample"

The next test would have been "Start MapRegressionTest RasterizingTileLayerGeoJsonSample"

So there are two possible Areas.
1. A Memory Leak/Corruption in Sqlite (Because of MbTiles)
2. A Memory Leak/Corruption in RasterizingTileLayer

charlenni:
@inforithmics Where did you see this? Could you reproduce this locally or do you have a protocol where it could be seen?

inforithmics:
I saw this on the last GitHub action build. I couldn‘t reproduce locally. I‘m thinking about running the tests locally 10 times or more so if there is a problem in one test it should sow up

charlenni:
Ok, I got the error locally by running test "RasterizingTileLayerGeoJsonSample" in debug mode.

grafik

I would say, that the line 168 of file "SymbolStyleRenderer.cs", which is

        var path = vectorCache.GetOrCreatePath(symbolStyle.SymbolType, CreatePath);

make the trouble. All could be expected except `path` when the test crashes.

charlenni:
Ok, seems to me as a racing condition. From 10 times to run, it fails around 6 times.

inforithmics:
I tried this branch and made a loop with 100 calls and it didn't crash on my computer. I tested it after I merged main into this branch. So it is using now SkiaSharp 2.88.7.

charlenni:
I run only this test in debug mode. Will try it again with the new version.

charlenni:
I run it 7 times with no problems. The 8th crashed.

charlenni:
Run it in debug mode and do something beside this. When I close the test tree after I start the test, then it crashes.

inforithmics:
I could now reproduce it. Open the Editing Sample in Avalonia and move the Shape around then it crashes. So I can fix it from there.

charlenni:
How can I help?

inforithmics:
I found out why it doesn't work every drawn Element in SkiaSharp should have it's own skpath.
1. Because of caching the All Symbols had the same SkPath so giving every symbol it's own skpath improved things considerably.
2. Clipping shouldn't be done on a cached path, but the clipped path should be cached.
```
               canvas.ClipPath(linePath);
```

pauldendulk:
Great you are looking into these hard to reproduce issues and not giving up on it. The bugs that are the hardest to find cause the most problems in the long run. Often people tend to ignore them because they don't happen very often.

charlenni:
@pauldendulk in our case they happen very often.

@inforithmics When you look into SkPath of the Skia [documentation](https://api.skia.org/classSkPath.html), go down to the "Detail description". The last line suggest that SkPath is threadbare if you call `SkPath::updateBoundsCache` (read also [this](https://api.skia.org/classSkPath.html#a6de7de42175fe3ce5ff92f3d2267ec2d)). Perhaps we should try this when writing the SKPath to the cache?

charlenni:
Ok, looked into SkiaSharp implementation, but couldn't find the `UpdateBoundsCache` function. So it isn't there and if we like to have it, we must create it. Perhaps a PR?

By investigating further, the `SkPath::updateBoundsCache()` is internal only a `SkPath->getBounds()` call (see [here](https://github.com/google/skia/blob/39bb9c35b840a4dedfc3c86316a3c77c21e905eb/include/core/SkPath.h#L501)). So it would be enough to add a call like this

        if (path is SKPath p) _ = p.Bounds;

after creating the path. But it doesn't work.

Perhaps related with [this](https://github.com/mono/SkiaSharp/issues/571)? And an [message](https://groups.google.com/g/skia-discuss/c/-G1cyl1QD9E) (quite old) or [this](https://issues.skia.org/issues/40036954) or [this](https://issuetracker.google.com/issues/36925032#comment5) about thread safeness of Skia.

I would say, caching for same pathes to one path isn't a good idea.

inforithmics:
Pull request with my learnings and Bugfixes (does not work yet 100%)
https://github.com/Mapsui/Mapsui/pull/2425

pauldendulk:
I am going to merge this PR which just adds console logging. We might want to go back to this thread to see what is discussed.

pauldendulk (approved):
Yes, good idea, its a puzzle for me every time too.

Issue #2389, opened 2024-01-05
Drawing order different to map info order


charlenni:
**Mapsui Version**
Main

**Mapsui Platform**
WPF

**Device**
Windows 10

**Describe the bug**
If you add `PointFeatures` to a `MemoryLayer` one after one, they are drawn in wrong order. The last added will be drawn last, the first added will be drawn first. The problem is, that the map info works the other way. The function is checking the last added first and the first added last. That is no problem until the features overlap. If this is the case, the feature later added is drawn below the earlier added, but gets the map info event earlier.

**To Reproduce**
See video below

**Expected behavior**
Drawing and touching should be consistent. I'm not sure, what is the idea behind drawing in this case. Should the first added feature be drawn first (as expected by touching) or should it be drawn last (then we have to change touching order).

**Screenshots**
I added a video showing the problem: 
https://github.com/Mapsui/Mapsui/assets/4381223/3b7e3360-3f49-4351-9d7d-25f9bf5435be

charlenni:
One info I missed in the description of the video is the order of adding the markers to the map. I added first the @ marker, then the A marker and so on. Z marker was the last added one.

charlenni:
Looked into this problem. Seems in conjunction with ConcurrentBag, which isn't an [ordered list](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentbag-1?view=net-8.0). This shouldn't be used here, because when features overlap, it isn't possible to say, which is the upper and which is the lower.

That said, we have to find something else as list for MemoryLayer.

pauldendulk:
Is this issue now fixed with the [id sorting fix](https://github.com/Mapsui/Mapsui/pull/2404), or is there something else to fix? We could have a more general discussion about how to determine the order but lets have that here: https://github.com/Mapsui/Mapsui/issues/2421

PR #2404, opened 2024-01-09
Add ZOrder to IFeatures

charlenni:
This PR is regarding #2389.

Add a ZOrder property for features to order features while retrieving them.

What is the problem? Up to now, MemoryLayer uses a ConcurrentBag to store features. ConcurrentBag is an unordered collection, so that you could not decide, which feature is created before or after another feaure. This is problematic, if you want to have access to the creation of the map. With ZOrder it is possible to say, which feature will be print over which other feature.

Additionally, the list of features returned by `GetFeatures()` is first ordered by the ZOrder and then ordered by the Id of the feature, because this Id is a consecutive number. If the Id of a feature is lower than another, then it is created earlier. Other map libraries use the y position to sort features. See for detailed discussion issue #2389.

### DIFF
--- a/Mapsui.Nts/GeometryFeature.cs
+++ b/Mapsui.Nts/GeometryFeature.cs
@@ -46,7 +46,7 @@ public class GeometryFeature : BaseFeature, IFeature
     /// <summary>
     /// Order of feature
     /// </summary>
-    public int ZOrder => throw new NotImplementedException();
+    public int ZOrder { get; set; } = 0;
 
     /// <summary>
     /// Implementation of visitor pattern for coordinates
--- a/Mapsui.Nts/GeometryFeature.cs
+++ b/Mapsui.Nts/GeometryFeature.cs
@@ -1,10 +1,13 @@
-using System;
 using Mapsui.Layers;
 using Mapsui.Nts.Extensions;
 using NetTopologySuite.Geometries;
+using System;
 
 namespace Mapsui.Nts;
 
+/// <summary>
+/// Feature representing a NTS geometry on the <cref="Map"/>
+/// </summary>
 public class GeometryFeature : BaseFeature, IFeature
 {
     public GeometryFeature()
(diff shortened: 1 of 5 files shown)


pauldendulk (approved):
So, by sorting on Id you already fix the problem for all but a few use cases. Setting the ZOrder is an option, we do not use it currently anywhere in our code base yet (we may want to add it to our samples on some point).

The sorting is done for MemoryLayer, but the ZOrder property is there for features of all layers. It won't work on other layers. That might be confusing. Other layers have a deterministic order, so don't have such a big problem. Still if the ZOrder is assigned they would expect a change. I propose not to fix this problem right now because the fetching mechanism might change. Perhaps this is the kind of things we want to have arranged centrally. Perhaps the fetcher should do the ordering, not the layers themselves.

pauldendulk:
We need to follow up on this. This PR contains an important fix by sorting on f.Id. A simple elegant solution. This fixes the mixed feature order of the MemoryLayer, which I suspect has been causing quite some flaky tests.

But setting the ZOrder will have no effect on all other layers, this will confuse the user. I have been thinking about this, implementing ZOrder everywhere is more code, and a burden for new implementations. I now have a solution in mind which (I hope) would require not too many changes:
https://github.com/Mapsui/Mapsui/issues/2421

charlenni:
I'm not very happy with this PR, but I fear we need such a thing.

For me the problem is, that we sort the features each time we draw the map twice. I assume, that this is very bad for the performance. It would be better, if the features have already an order (perhaps by retriving them in the order they are added).

And yes, you are right, that this isn't respected in all the other layers. I tried to make a first attempt to go in this direction. On the other side, I expect, that the providers of the other layers return from GetFeatures with the right order. So there should be a comment, that this function should provide a list features in the order they should be drawn. If they order it by ZOrder or something totally different is beyond our scope. I assume, that it isn't very important for other providers. They are mostly providing faster features, which don't need ZOrder.

PR #2405, opened 2024-01-09
Move LoggingWidget from MapControl to Map

charlenni:
Move LoggingWidget from MapControl and add it to Map, so that each map has its own LoggingWidget and not one, that is used for all.

After the first version of adding the LoggingWidget to MapControl, @pauldendulk and @inforithmics both suggest bringing it to Map. This PR should do this.


pauldendulk (approved):
I think this is simpler.

PR #2407, opened 2024-01-10
Add TagExtender to handle warning messages of GeoTiffProvider

charlenni:
Add TagExtender to handle warning messages of GeoTiffProvider loading GeoTiff files with unknown tags

In our samples are GeoTiff files, that have unknown tags. If Bitmiracls LibTiff.Net library reads one of this tags, it produces a warning. These warnings pollute our logs. 

Solution is to add this tags to the library. This is done by this PR.

Helps with searching of problems in #2402.


inforithmics (approved):
I didn't know this Feature existed, it solves the warning problem

PR #2414, opened 2024-01-11
install android sdk on linux Fix Main Branch

inforithmics:


pauldendulk (approved):
I just noticed this problem and then saw that you already fixed it. Fantastic.

PR #2418, opened 2024-01-12
Update LoggingWidget

charlenni:
Change the LoggingWidget after the bid rework of Widgets. Now it is possible to use stretch for vertical and horizontal alignment. Also it is possible to save more log entries, so that the filter or size could be changed, so that old/more log entries are visible.

### DIFF
--- a/Mapsui.Rendering.Skia/SkiaWidgets/LoggingWidgetRenderer.cs
+++ b/Mapsui.Rendering.Skia/SkiaWidgets/LoggingWidgetRenderer.cs
@@ -65,6 +65,14 @@ public class LoggingWidgetRenderer : ISkiaWidgetRenderer, IDisposable
 
         foreach (var entry in loggingWidget.ListOfLogEntries)
         {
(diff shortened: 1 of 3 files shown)


Mapsui.Rendering.Skia/SkiaWidgets/LoggingWidgetRenderer.cs
```diff
@@ -65,6 +65,14 @@ public void Draw(SKCanvas canvas, Viewport viewport, IWidget widget, float layer
 
         foreach (var entry in loggingWidget.ListOfLogEntries)
         {
+            if (entry.LogLevel > loggingWidget.LogLevelFilter)
+                continue;
+
+            var top = marginY + (paddingY * line) + loggingWidget.TextSize * (line + 1);
```
pauldendulk, line 71: Perhaps line can be renamed to lineNumber. It was confusing for me just now (even though in normal language you can use 'line' to indicate line number, e.g.: "See line 12")

pauldendulk (approved):

PR #2419, opened 2024-01-12
Fixing Blazor build

inforithmics:


pauldendulk (approved):
Thanks for the fix. Too many refactorings of the same code going on at the same time.

PR #2422, opened 2024-01-13
Remove INotifyPropertyChanged from widgets

charlenni:
As promised, when the new widget handling is merged, the INotifyPropertyChanged interface could be easily removed. This is done with this PR.

Widgets are now redrawn when their flag NeedsRedraw is set to true. After a redraw, this flag is set to false by the renderer. For easy use, this flag is set in the `Invalidate()` function.

### DIFF
--- a/Mapsui/Widgets/Widget.cs
+++ b/Mapsui/Widgets/Widget.cs
@@ -4,7 +4,7 @@ using System.Runtime.CompilerServices;
 
 namespace Mapsui.Widgets;
 
-public abstract class Widget : IWidget, INotifyPropertyChanged
+public abstract class Widget : IWidget
 {
     private HorizontalAlignment _horizontalAlignment = HorizontalAlignment.Right;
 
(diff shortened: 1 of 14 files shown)


Mapsui/Widgets/Widget.cs
```diff
@@ -157,9 +157,9 @@ public void UpdateEnvelope(double maxWidth, double maxHeight, double screenWidth
         Envelope = new MRect(minX, minY, maxX, maxY);
     }
 
-    public virtual void OnPropertyChanged([CallerMemberName] string name = "")
+    public virtual void Invalidate([CallerMemberName] string name = "")
```
pauldendulk, line 160: The name argument could be removed now.

pauldendulk (approved):
Thanks for coming back to this. This way it works without notifications.

Mapsui/Widgets/Widget.cs
charlenni: No, that isn't possible. I override the Invalidate() function to check, if the LoggingWidget Enabled has change to add or remove the subscription to the logging event. So I would keep it. Perhaps the nameof() while calling Invalidate() in the LoggingWidget could be removed.

PR #2423, opened 2024-01-13
Change Padding for IconButtonWidget

charlenni:
Change padding from PaddingX and PaddingY to a MRect Padding for consistency with other widgets.
Change also background color of IconButtonWidget to transparent.

### DIFF
--- a/Mapsui.Rendering.Skia/SkiaWidgets/IconButtonWidgetRenderer.cs
+++ b/Mapsui.Rendering.Skia/SkiaWidgets/IconButtonWidgetRenderer.cs
@@ -24,8 +24,8 @@ public class IconButtonWidgetRenderer : ISkiaWidgetRenderer
 
         // Calc Envelope by Width/Height or, if not set, by size of content
         button.UpdateEnvelope(
-            button.Width != 0 ? button.Width : picture.CullRect.Width + button.PaddingX * 2, 
-            button.Height != 0 ? button.Height : picture.CullRect.Height + button.PaddingY * 2, 
+            button.Width != 0 ? button.Width : picture.CullRect.Width + button.Padding.Left + button.Padding.Right, 
+            button.Height != 0 ? button.Height : picture.CullRect.Height + button.Padding.Top + button.Padding .Bottom, 
             viewport.Width, 
             viewport.Height);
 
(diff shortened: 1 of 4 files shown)


pauldendulk:
@charlenni The rendering tests keep failing. I did not look into it, but it could be that some rendering tests actually detect a difference. I saw you changed something in transparency, perhaps that was the cause.

If the tests need fixing please take a look at the documentation on rendering tests again https://mapsui.com/documentation/rendering-tests.html.

The current rendering test mechanism tests all the samples. That is great. However, if a reference (original) image does not exist it will ignore that test. It is likely that there are some recent samples that do not have a reference image yet. If you run `scripts\test-image-copier.cmd` those images will show as added in the diff. It is okay to push them.

charlenni:
I tested it before upload and now again. All tests worked. So it is the normal SkiaSharp crash because of threading (see #2425).

pauldendulk:
The regression tests have been flaky since we added tests for all samples using discovery (It added a huge number of tests at once and was a very important improvement). A number of the flaky issues have been fixed (some real bugs). Currently the cross-thread rendering issue is still open, so it can crash sometimes. 

I need to know if a PR introduces a failing test or not. How I work with this flakyness is to re-run the windows tests to fix it. I try not to dig into the build log because I am already short on time. Sometimes I need to re-run it twice.  At some point it is green and I can assume it failed because of flakiness. But this PR is still red after running it for about four times, if it does not turn green ever I will have to look into the details of the log.

pauldendulk:
Later PRs were green right away:

image

pauldendulk:
btw, I the past I have noticed that the regression tests, which do a pixel by pixel compare with a small tolerance, can have different result from time to time. That was why the tolerance was introduced. I can not explain this difference, I would expect identical results. Perhaps it is a lurking bug in Skia, SkiaSharp or Mapsui. The difference is bigger between mac/linux/windows runs. So, if you run the tests on a mac they could fail on the windows build server, and you may need to update an image. btw, many images will be different, but only update the ones that actually fail.

charlenni:
@pauldendulk I have to apologize. It was my fault. The problem was, that I missed to push two commits that I made after creating the PR. When I checked it the last time, all the tests running correct, because the two commits solved the problem. So on my PC all worked perfect.

I now pushed the two missing commits and expect, that it will running.

charlenni:
I had one time, where the map had changed. There vanished a town, so that the image was different.

I would say, that the regression tests work perfect and are a very good for testing. But I found, that some of them were not correct. They missed the attribution.

pauldendulk:
Yes, if the tests depend the osm service the data could just change. I think Thomas partially resolved this by working with caching.

charlenni:
Ok, Mac fails because it needs a newer version of XCode :-( @inforithmics Anything we can do?

charlenni:
Ok, now the regression tests work

pauldendulk (approved):
Makes sense.

PR #2425, opened 2024-01-14
Fix render crash

inforithmics:
Currently it crashes more, but I discovered that two threads are accessing SKCanvas which is not Thread Safe. So I have to introduce Locking so that only one Thread at a time Access SKCanvas and then the crashes should go away.


inforithmics:
The current shape of the Code could change as I discover more things. So the caching could go back to less fine grained things.

inforithmics:
Currently I'm debugging RasterizingTileLayer with ShapeFile. Several renderers access the same RenderCache and Render simultanously.

pauldendulk:
@inforithmics Thanks for looking into this. It is important to get this completely straightened out. I like to help. I have some questions/remarks.
- My first suggestion is to invest in some rewriting to make debugging easier, without having the intent the solve the problem. 
  - When two threads access the same resource they crash only occasionally. You could store the first thread in a static field. If a second thread comes in throw an exception, so you know right away things have gone wrong.
  - A thread in C# has an ID. Store the ID in a dictionary, how many entries do you get in the dictionary?  
  - Check if a second thread comes in and in that case let the code stop on a breakpoint so you could check the stacktrace. 
- You mention MapInfo being on another thread accessing the SKCanvas. But I would think in our tests we don't usually do a MapInfo, we mostly just render a map. Perhaps this is seen in manual tests, but we need another explanation for the test fails?
- The SKCanvas is not thread save. The MapControl renderer runs on the SkiaSharp Paint events (which is always on the same thread). A second thread could be MapInfo. Any others?
- Could this be an phenomenon caused by the tests? Tests can run in parallel? Do they? Or can you reproduce the phenomenon running one test?
- You talks about multiple renderers. Are those MapRenderer instances? If yes put a breakpoint in the constructor to check where those are created.
- I am not a fan of locks. If it looks like we need them I like to look into a possible rewrite to avoid having to use them. 
- What is the best way for me to reproduce the problem?

inforithmics:
> I think, that problem would be much easier to solve, if we have different draw and paint routines. A draw could be done in multiple threads, a draw should be done always in the same thread.

This could solve it too. The first Bug I fixed with the lock was that the same shape was rendered at two places at the same time. And this caused an exception.
During editing the drawing of the map was active and the editing itself triggered a mapinfo request which too triggered a drawing operation causing the same skpath to be drawn at two canvases at the same time, which it didn't like it. So the current fix was a lock that only one Render operation is done in the renderer at a time.

inforithmics:
> @inforithmics Thanks for looking into this. It is important to get this completely straightened out. I like to help. I have some questions/remarks.
> 
>     * My first suggestion is to invest in some rewriting to make debugging easier, without having the intent the solve the problem.
>       
>       * When two threads access the same resource they crash only occasionally. You could store the first thread in a static field. If a second thread comes in throw an exception, so you know right away things have gone wrong.
Good idea I'll try that, my first approach was to use the Thread Id as a cache key but didn't work at the first time, Intrestingly If I disable the caching than it works too. So it definitly has to do with the shared resource beeing used on two threads simultanously. The current crashing RasterizingTileLayerWithShapeFile has several threads accessing the same skpath objects rendering them on several seperate Canvases. And this causes crashes. So I plan to see what happens when I lock the shape/paint to see if it still crashes.
>       * A thread in C# has an ID. Store the ID in a dictionary, how many entries do you get in the dictionary?
In the Editing sample there where 2 threads the ui thread (mapinfo request) and the render thread because of Fetch.
In the RasterizingTileLayer with Shapefile, there are more threads (and not locking yet) so it crashes. What I plan to do here is investigate if every thread has it's own Vector Cache if things work better. Was previously the case until I made the Cache Thread Safe. 
>       * Check if a second thread comes in and in that case let the code stop on a breakpoint so you could check the stacktrace.
When the crash happens with the option view all threads I see that two threads are rendering at the same time maybe accessing the same object.
> 
>     * You mention MapInfo being on another thread accessing the SKCanvas. But I would think in our tests we don't usually do a MapInfo, we mostly just render a map. Perhaps this is seen in manual tests, but we need another explanation for the test fails?
The failing tests come because of RasterizingTileLayer uses more than one thread to faster render the tiles. I guess If I limit this to one thread, it will be slower but does not have anymore crashes.
> 
>     * The SKCanvas is not thread save. The MapControl renderer runs on the SkiaSharp Paint events (which is always on the same thread). A second thread could be MapInfo. Any others?
The RasterizingTileLayer and RasterizingLayer Render images that run on other threads, because of Async Fetch. Somehow using the same SkObject at the same time on different threads isn't a good idea in Skiasharp it seems.
> 
>     * Could this be an phenomenon caused by the tests? Tests can run in parallel? Do they? Or can you reproduce the phenomenon running one test?
Intrestingly with my changes, It crashes now very reliable meaning every time. Before it was not every time.
> 
>     * You talks about multiple renderers. Are those MapRenderer instances? If yes put a breakpoint in the constructor to check where those are created.
Yes every Thread in RasterizingTileLayer has it's own MapRenderer but not it's own RenderCache.
> 
>     * I am not a fan of locks. If it looks like we need them I like to look into a possible rewrite to avoid having to use them.
Me too. When I learn more I try to remove these.
>     * What is the best way for me to reproduce the problem?

1. In Main, you can open the Editing Sample and move around the Shape it crashes very quickly.

2. In this Branch you can open the RastizeringTileLayer with ShapeFile Sample and the it crashes immediatily, That's why the Unit Tests fail now.

Maybe to avoid locking I have to make a differnt Cache Scheme, When I use a SkPath object I have to remove it from cache so that it isn't used at the same time und when I'm finished I return it. But I first want to verify that the parallel usage of one SkObject is the problem. Or I could mark it as in use and if I need it I create one more. So that under the same key objects that are reused often have more than one cached instance.

pauldendulk:
About the lock. I now think a lock may be the best solution in the current architecture. There is no async/await within the renderer so it can be a simple lock (no SemaphoreSlim).

An alternative would be to sync everything to the same thread using queues and messages. That might be an option for MapInfo but it might be more complicated to get this to work for the tile rasterizer. 

I would suggest to put the lock in the root methods of the renderer to keep the code simple to understand. In the alternative (using fine grained locks) there is more bloat added by the locking code and it is more likely to introduce errors in the locking, like forgetting to lock one render type, or making errors in the nesting of  locking. So, after having a short glance I think it should be in the public methods Render and RenderToBitmapStream.

When rendering is redesigned this might all work different, so lets keep the code easy to understand and not try to optimize for performance.

pauldendulk:
>  if every thread has it's own Vector Cache

This worries me. Caching is always a problem in all the apps I work with. We have too much growth of caches in different places already. We should get this under control. A problem with a cache per thread is that this could grow too big. The same data could be cached in different vector caches. If new thread are created every time. This causes the kind of bugs that you only find after running the app for an hour.

inforithmics:
With the lock I discovered that the Skia objects can only used once on a Rendering surface and not in parallel. So I have an idea to improve the Cache so that it gives back a skia object and if it needs another one it creates a new one. With disposing I keep track of how many Skia objects are out there. So Objects that are only used once have one cache item and objects that are used more often like colors have one or more cached items. This would avoid the lock. and the several caches.

inforithmics:
I managed to fix it without global locks, I track if a cached instance is in use if it is in use I create temporary a new one that is disposed after use. I tested caching more than one instance but, the caches didn't reuse so not worth the extra effort.

inforithmics:
I'm trying to reduce the number of casts, in the caches.

inforithmics:
I got more stable with my changes but crashes now at another test. When I introduce an Iteration lock the crashes go away. So I think I let this lock there until I have a better idea to get rid of it.

pauldendulk:
@inforithmics Sorry, for causing a conflict.

inforithmics:
@pauldendulk no problem, I only changed the Regression Image because it changed showhow. I guess it could be something Framework related.

Mapsui.Rendering.Skia/Cache/RenderCache.cs
```diff
@@ -1,19 +1,20 @@
...
 
 #pragma warning disable IDISP008 // Don't assign member with injected and created disposables
 
 namespace Mapsui.Rendering.Skia.Cache;
 
-public sealed class RenderCache : IRenderCache
+public sealed class RenderCache : IRenderCache<SKPath, SKPaint>
 {
-    private IVectorCache? _vectorCache;
+    private IVectorCache<SKPath, SKPaint> _vectorCache;
```
pauldendulk, line 12: Is there a specific reason to add the generics? Or is it just good to have it a bit more strictly typed?

Mapsui.Rendering.Skia/Cache/RenderCache.cs
```diff
@@ -1,19 +1,20 @@
...
+using SkiaSharp;
 
 #pragma warning disable IDISP008 // Don't assign member with injected and created disposables
 
 namespace Mapsui.Rendering.Skia.Cache;
 
-public sealed class RenderCache : IRenderCache
+public sealed class RenderCache : IRenderCache<SKPath, SKPaint>
 {
-    private IVectorCache? _vectorCache;
```
pauldendulk, line 11: Always better if you do not need the nullability.

Mapsui/Extensions/BrushExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+public static class BrushExtensions
+{
+    public static bool IsVisible(this Brush? brush)
+    {
+        if (brush == null)
+        {
+            return false;
+        }
+
+        if (brush.Color?.A == 0)
```
pauldendulk, line 13: Not for this PR, but this could call an IsVisible extension method in Color again and this method could become a one liner. 

btw I like the new MyMethod() => syntax.

Mapsui/Rendering/NonCachingVectorCache.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+
+namespace Mapsui.Rendering;
+
+public sealed class NonCachingVectorCache(ISymbolCache symbolCache) : IVectorCache
```
pauldendulk, line 5: What happens when you use the NonCachingVectorCache. No crashes at all?

Mapsui/Rendering/VisibleFeatureIterator.cs
```diff
@@ -20,7 +22,12 @@ public static class VisibleFeatureIterator
             if (layer.MinVisible > viewport.Resolution) continue;
             if (layer.MaxVisible < viewport.Resolution) continue;
 
-            IterateLayer(viewport, layer, iteration, callback);
+            // somehow it crashes when more then one iteration or rendering is done in parallel 
+            // TODO: find out which Caching path causes this. Because when the caching is disabled it works.
+            lock (_iterationLock)
```
pauldendulk, line 27: This is coarse grained lock (opposed to fine grained). So if multiple threads are used they have to wait a lot. Our primary use case is however fast performance while panning and zooming. Those calls will mostly be on one thread. So no problem there. When rasterizers are used the background rasterizer would get a lock here and the draw thread would have to wait. That is a downside. This solution is however super simple. While the underlying cause of this problem is that we do not fully understand what is happening. The complexity is the problem. So keeping the lock this simple is great.

pauldendulk (approved):
Thanks a lot for investigating this. The PR looks good. 

A question indirectly related to this PR. Caching is a common cause of problems. When there is some obscure error I often first want to disable the cache. Even if the problem is not in the cache it helps to rule it out, because that makes it easier to find the actual problem. Is there currently a way (preferably a one liner) to disable the cache. Often when users have a problem I say 'enable logging', the next thing I would like to say it 'disable the cache' and here is how to do it ....

Mapsui.Rendering.Skia/Cache/RenderCache.cs
inforithmics: The reason I typed it specifically is to reduce the number of castings. I might make a follow pulltequest to simplify this again. With the removal of the global iteration lock.

Mapsui/Rendering/VisibleFeatureIterator.cs
inforithmics: I tried to remove this lock. But it still crashes without this lock. It works without caching. But I‘m not sure that it works because of slower execution File. Because I saw some exceptions in Skiasharp that track the used items for disposability. 
I tried to use the Mange’s thread id as cache and it still crashed. So I think I will try to make a reproduction sample for Skia that reuses skiaobjects on different threads, where every thread has it‘s own skiasharp objects.

Mapsui.Rendering.Skia/Cache/RenderCache.cs
pauldendulk: No need to change. I just try to understand things and ask some questions. Having a conversion helps in the thinking. I use a lot of generics in some other projects. Sometimes I like it, sometimes not, sometimes I go too far. Every time you have to decide what is the best option.

Mapsui/Rendering/VisibleFeatureIterator.cs
pauldendulk: If you feel like it. You could could also consider this stable.

Issue #387, opened 2018-04-18
DataSource informing Layer about DataChanged


charlenni:
Is there a possibility, that a IProvider could inform a Layer, which DataSource it is, that the data has changed?

Problem: I have a an IProvider which collects data and than wants to update the Layer with this new data. For this, it should inform the Layer, that an update is needed and than Layer fetches the data from DataSource at GetFeaturesInView(). Is this possible? If not, how could this be done? Wait in GetFeaturesInView()?

pauldendulk:
No there isn't. For the WritableLayer I added a SignalDataChanged. I needed it for editing but I am not very happy with this solution. You have to have access to the layer while sometimes you only have access to the provider. We could add a DataChanged EventHandler on IProvider that would trigger the layer. All layers need to register to it. Not sure if this is a good idea.

pauldendulk:
I now think we should add it.

starnutoditopo:
I also think that some more information would be handy in the DataChangedEventHandler. It would be useful, for the subscribers, to know what has changed: what has been added/removed/modified.

PR #2426, opened 2024-01-15
Add virtual methods in EditManager

starnutoditopo:
Add virtual methods in EditManager, so that inherited classes can intervene on modified features.
See: https://github.com/Mapsui/Mapsui/issues/387#issuecomment-1890995785

### DIFF
--- a/Mapsui.Nts/Editing/EditManager.cs
+++ b/Mapsui.Nts/Editing/EditManager.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using Mapsui.Layers;
 using Mapsui.Nts.Extensions;
+using NetTopologySuite.Features;
 using NetTopologySuite.Geometries;
 
 namespace Mapsui.Nts.Editing;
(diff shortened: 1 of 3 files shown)


pauldendulk (commented):
Thanks you for you contribution. However in the current phase that we are in I don't want to work on extensions of the EditManager. 

The edit functionality was something I added for my specific purpose. It was at that point in the samples. This was because it was not well designed and not thought through well. 

The edit sample was a WPF app. This was becoming increasingly impractical because we support so many platforms. So, the editing code was converted to a platform independent solution, introducing the widgets, which makes it easier to maintain for us. It was however still a first version and in retrospect I think we should have marked it as experimental. Currently it does not function correctly for Avalonia for example. So, this is something we need to revisit so I don't want to add something on top of it.

Also, if we will support something like this it will probably through callbacks and not overrides of virtual methods. 

Related, this answer I posted on stackoverflow:
https://stackoverflow.com/questions/77807110/how-to-use-two-editingwidgets-in-c-sharp-mapsui-wpf-simultaneously/77807240#77807240

pauldendulk:
See, my comment. I will close this PR for now. At some point we may want to add something like this. In general we appreciate feedback. A clear explanation of the use case you want to support is very welcome. We can take that into account rewriting the current code.

starnutoditopo:
Thanks for your answer: it's useful to understand in which direction to go.

starnutoditopo:
Then, `EditManager` could marked as `[Obsolete]`?

PR #2431, opened 2024-01-15
Add remark to not commit file to git if there are changes but the test succeeds

pauldendulk:


charlenni:
Btw. your test fail too. I assume, that we get not ride of this problem until Thomas has made his PR ready.

pauldendulk:
This changes docs only. I am pretty sure this does not break the build. I am not blindly following procedures but dealing with real world problems that occur with some frequency. Ideally the doc change should not trigger a build at all, but our docs include api docs and the api docs depend on a build the code.

PR #2436, opened 2024-01-16
Formatting Fixes "dotnet format" run

inforithmics:
Helps to reduce the changes in https://github.com/Mapsui/Mapsui/pull/2436 because here I did a dotnet format and it found some more formatting problems in other files


pauldendulk (approved):
Thanks a lot. That really helps me understand the changes in the other PR.

I still do not fully understand dotnet format, I would expect the  github action (which runs dotnet format) to fail if these whitespace changes are necessary.

Issue #2434, opened 2024-01-15
Remove the number prefix from all our samples


pauldendulk:
**The problem**
At some point we added a number as prefix to our sample titles. For example in "1 Single Callout". This was so that we could control the order of the samples. The samples are discovered and ordered alphabetically. If a new sample is added we have to change the prefix of the other items to place the new one in the desired place. This is a hassle and we have come to neglect it.  

**The proposed solution**
Remove the prefix number in all samples and just accept the alphabetic ordering as it is.

charlenni:
See PR #2437

PR #2437, opened 2024-01-16
Remove numbers for demo captions

charlenni:
See issue #2434. Many files changed, but not very complicated.


pauldendulk (approved):
Thanks for picking this up! This changed 32 files but very easy to review. It changes one kind of thing.

Issue #2421, opened 2024-01-13
Add feature sorting mechanism


pauldendulk:
**The problem**
A common requirement for map display is to be able to specify the order in which features are drawn. For instance, when drawing cities you may want to put the biggest cities on top.

**The proposed solution**
Not decided yet. 

**Option: Provide custom sorting function**
We just introduced a ZOrder. This is used in the MemoryLayer but not anywhere else. In order to make this consistent we need to start sorting on ZOrder everywhere else. This puts a burden on all layer implementations. Also use of the ZOrder may not always be the easiest solution. In some cases the user already has some object with a field indicating the order. Now the ZOrder has to be assigned in the user code. An alternative that I am thinking about is to provide a function to do the sorting. It has a default implementation, or perhaps no implementation, in which case the provider determines the order. The user can provide a function which takes the list of features as input on which to apply custom sorting logic. Any field (or fields) could be used. Advantages: We can remove ZOrder (one property less). We do not have to implement sorting on ZOrder for all layers (and perhaps providers?).

Note, Currently the IFeature has some fixed properties and has the option for 'custom properties' through the dictionary. This solution was introduced in SharpMap, long ago. Nowadays there are alternative options, like generics. This is a separate feature request but may be related.

charlenni:
You are right, there could be a better way of sorting. And much better, if there is no need for sorting at all (faster).

I looked into other map libraries, and they used ZOrder or ZIndex for this. So it was the most obvious way.

charlenni:
See PR #2438.

Perhaps we should remove a special sorting function for all existing layers (`MemoryLayer` included) and let the user decide, which is the right one to use. This would be possible with the given implementation in PR #2438.

PR #2438, opened 2024-01-16
Add sorting of features before drawing

charlenni:
See issue #2421.

This PR introduce a `SortFeature()` function, which is used to sort features before drawing. In the `BaseLayer` implementation, this functions return the given features untouched. In the `MemoryLayer` implementation this function is replaced by a function, that orders the features by ZOrder and Id by default (perhaps this should be reverted and the default function should use). With this, it is possible to use whatever sorting algorithm you like.

### DIFF
--- a/Mapsui/Layers/BaseLayer.cs
+++ b/Mapsui/Layers/BaseLayer.cs
@@ -187,7 +187,7 @@ public abstract class BaseLayer : ILayer
     public abstract IEnumerable<IFeature> GetFeatures(MRect box, double resolution);
 
     /// <inheritdoc/>
-    public virtual Func<IEnumerable<IFeature>, IEnumerable<IFeature>> SortFeatures => (feature) => feature;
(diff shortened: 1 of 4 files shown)


pauldendulk (approved):
Nice! By letting this VisibleFeatureInterator do the sorting you have to call it only once (instead of calling it in the layer).

With this change setting the IFeature.ZOrder on a non-MemoryLayer will still have no effect. This will surprise users. One solution is to move the default sort function to the BaseLayer. Another would be to remove the ZOrder altogether. I am thinking about the generic Data field on the IFeature which could contain a ZOrder, or any other field that the users could use for sorting.

charlenni:
I found a little problem here. The MapInfo result is not in the right order. We check the layers in reverse order (topmost first), but don't respect the order of the features inside the layers. The first drawn feature should be the last one to check. So the sort order of the features inside a layer has to be reversed.

PR #2442, opened 2024-01-17
Change order of features of each layer in MapInfo

charlenni:
The sort order is respected when drawing the map, but not when the map is touched. All features, that are hit in a layer are in sorted order, but should be in reversed order like the layers, which are already reversed. The topmost feature is the last drawn feature, but should be the first hidden feature. See also #2438. 

This PR reverses all features for all layers.

### DIFF
--- a/Mapsui.Rendering.Skia/MapRenderer.cs
+++ b/Mapsui.Rendering.Skia/MapRenderer.cs
@@ -252,7 +252,7 @@ public sealed class MapRenderer : IRenderer, IDisposable
                                 mapList = new List<MapInfoRecord>();
                                 // get information from ILayer Feature Info
                                 var features = await layerFeatureInfo.GetFeatureInfoAsync(viewport, x, y);
-                                foreach (var it in features)
+                                foreach (var it in features.Reverse())
                                 {
                                     foreach (var feature in it.Value)
                                     {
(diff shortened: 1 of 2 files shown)


pauldendulk (changes requested):
Thanks for finding this bug! It has been in here from the start.

I propose a different solution though, which is to fix in the MapRenderer.GetMapInfo (which is the root of the MapInfo request). I have code for it locally. I will paste it in the comments so you can apply it. 

The problem that you were facing that in your fix the layers were already in reverse order, while the feature were not. This made reversing complicated. 

With the fix in the root this can be done simpler.  The layers are not reversed at first (we did this be inverting the index). Then add everything to the result in reverse order. At the end reverse the entire result.

The important part is that you found the bug and found the root cause, which is that the renderer draws in reverse order. Thanks!

pauldendulk:
btw, this is the kind of thing that is very suitable for a unit test.

charlenni:
Did you think of a solution like this? But how to reverse the task's results? And btw. I don't understand, what the tasks do. A bit more comments would have been helpful.

pauldendulk:
That was the local code I had, bit late in pasting, got distracted.

pauldendulk:
But perhaps it is easier if I just pushed it myself.

charlenni:
And what is going on with the tasks? Is/must their order not reversed?

And yes, add it by your own and we close this PR.

PR #2448, opened 2024-01-19
Fix order of features within a layer in MapInfo

pauldendulk:


charlenni (approved):

charlenni:
It is a part that isn't tested with any of our tests. So I didn't expect any problems. And I read every line of your changes and it was nearly the same as I did some minutes before.

But yes, I will wait a little longer next time.

PR #2451, opened 2024-01-20
Update logging.md

pauldendulk:
Add the exception to the log message. This is what users would always want.


charlenni (approved):
Make totally sense

PR #2453, opened 2024-01-20
Add new constructor to MPoint with (double, double) as parameter

charlenni:
Add a new constructor to MPoint. This is useful in conjunction with `SphericalMercator.FromLatLon()` which returns a `(double, double)` if it is called with `double, double` as a parameter. So it is possible to create a MPoint directly.

Additionally, add some comments if I have currently hands on.


pauldendulk (approved):
Nice. I was very happy when the tuple support was improved in C#, because it avoids having to construct another point type. I never used it as much as liked. This is another step in the good direction.

PR #2454, opened 2024-01-20
Remove bugs in CalloutStyleRenderer

charlenni:
Remove some bugs in CalloutStyleRenderer where padding isn't calculated correct.

In this case, the map regression test image has changed. It now has attribution (don't know why this wasn't the case before). But also the callouts have changed, and I don't know, why. The callouts are created by random numbers and should always be the same.


pauldendulk:
> In this case, the map regression test image has changed. It now has attribution (don't know why this wasn't the case before). But also the callouts have changed, and I don't know, why. The callouts are created by random numbers and should always be the same.

I had seen this before and assumed it was fixed by the sorting on the ConcurrentBag, but apparently it wasn't. Perhaps there is a ConcurrentBag involved somewhere else?

pauldendulk (approved):
A petty the image totally changed as you described. Now it is not possible to see how this fixes the padding. Thanks for the fix!

PR #2455, opened 2024-01-20
Add (4.1.2) .net 7  .net 8 target to avalonia for .net 8 Wasm support

inforithmics:
Investigating .Net 8 Avalonia Wasm support. It seems not to work yet. In Main and in 4.1.2 Branch have to investigate.


pauldendulk (approved):
Great. We should release a patch soonish.

PR #2459, opened 2024-01-21
Add:  (4.1.2) an iteration Lock so that only one Thread renders at the sa…

inforithmics:
…me time.

I discovered this during this pull request

https://github.com/Mapsui/Mapsui/pull/2425

That when SkiaObjects are used in more than one threads exceptions occour. This was the simplest fix to the Skia Crashes. I'll still investigating If I can do without a global Render/Iteration Lock But for the 4.1.2 Branch this should be sufficient to Fix Threading issues.


pauldendulk (approved):
Splendid!

PR #2460, opened 2024-01-21
[WIP] Add marker extensions

charlenni:
Next try for an extension for a marker. Now it is complete as extensions implemented and in one new file.

Idea is to have an easy possibility to add markers to a map.


pauldendulk (commented):
I had a quick glance and will give some comments without fully understanding everything.
- Lots of extension methods. That is great. 
- Extension methods of a type [should be in a file for that type](https://mapsui.com/documentation/contributors-guidelines.html#extension-methods). A downside may be that the functionality for Markers gets distributed over many files Too bad, lets just stick to what is in the guidelines. Also, even if some functionality was added for Markers it may also be used in other scenarios. So, perhaps there is mostly an upside in doing it like this..
- Static fields should be avoided. Static methods are great as long as they are pure functions, but as soon as static state is used within the static methods they are not pure functions anymore. I suspect the code in this PR will misbehave if you start using multiple marker layers (perhaps not, but I don't want to think about the details of what happens, better to just not use static fields). So, what is the alternative here? This is something I did not think about yet 😬. Perhaps the fields can be replaced with defaults in the main extension method, and that method should have some overloads to add extra parameters (just doing some guessing here).

A better review will follow later, I have some other priorities right now.

charlenni:
I will do a rework, so that the extension methods are in the right files. I fear, that at the end it would be difficult to read the code.

The static fields are `readonly` and never change after set in the constructor. They are used for the image of the marker, which never change and isn't replaceable. The others are for the height of the image and the regex to extract this height from the image. None of them changes over the lifetime of the app. But it is surely possible to load this image each time the color changes.

PR #2461, opened 2024-01-22
Install .NET 8 SDK in github actions in develop/4.1

pauldendulk:


pauldendulk:
@inforithmics I had problems building Maui in the 'Release Nugets' action with both net7 and net8. I removed net8 to release at least part of our fixes. Perhaps you can help me figure out what the problem is. I had the same build problem locally. I mostly do not understand why it does build on the server in the main dotnet.yml.

PR #2469, opened 2024-01-27
Remove net6 from MAUI

pauldendulk:
We should try to simplify more so that we can move forward faster (Support less and support it better). So, I this kind of odd configuration for older version should be removed. Also MAUI 6 is not supported anymore.

@inforithmics I like to have your opinion on this.


pauldendulk:
Merging this but still like to know what you think @inforithmics

PR #2475, opened 2024-01-27
Added MVVM support to Map property of MapControl

mr-const:
Added StyledProperty (AvaloniaProperty) for Map property of the MapControl to allow using databinding with ViewModel.


pauldendulk (approved):
Great thanks. I can not properly judge the correctness of the Avalonia implementation for this, but I assume this will be tested as part of the v5 betas.

Some thoughts about future changes: It is unfortunate we have to solve this with an `#elif` and I guess there are more to come. Perhaps we should move this to the platform implementation and only use a shared field in the shared MapControl. Perhaps _map., or a SharedMap property, which is assigned in the platform implementation.

pauldendulk:
Perhaps this should be patched in 4.1 as well.

PR #2479, opened 2024-01-29
WIP Move Map property to platform implementation

pauldendulk:
This follows up on this PR. We will probably have more platform specific code for other platforms. We need less `#if`s. It is cleaner to move the Map property to the platforms even though there is now more code duplication.

I expect changes will come to WinUI (Uno) and WPF as well.

Binding implementation for our eight platforms:
- [x] MAUI
- [x] Avalonia
- [ ] WinUI / Uno Platform
- [ ] WPF
- [ ] Blazor - what kind of implementation do we want here?
- [ ] iOS
- [ ] Android
- [ ] Eto.Forms


pauldendulk:
I can re-do this later.

PR #2481, opened 2024-01-31
Made Avalonia MapControl's Map property DirectProperty

mr-const:
Made Avalonia MapControl's Map property direct, because it doesn't need styling and thus we can remove necessity to access it in UI thread only.

Using StyledProperty, as it was made in previous PR requires access to Map property through UI thread and it was not really good idea.
DirectProperty, on the other hand, allows us to use existing Map property without any modifications.
(More details can be found [here](https://docs.avaloniaui.net/docs/guides/custom-controls/how-to-create-advanced-custom-controls))

Removed databinding from sample, because it needs a little bit more advanced DI to use (currently we assign DataContext after construction, but constructor needs ViewModel's Map property to initialize)

### DIFF
--- a/Mapsui.UI.Avalonia/MapControl.cs
+++ b/Mapsui.UI.Avalonia/MapControl.cs
@@ -46,6 +46,9 @@ public partial class MapControl : UserControl, IMapControl, IDisposable
 
     private bool _shiftPressed;
 
+    public static readonly DirectProperty<MapControl, Map> MapProperty =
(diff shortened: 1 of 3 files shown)


pauldendulk (approved):
Again I can not properly review this and will accept it as is. The [link to the documentation](https://docs.avaloniaui.net/docs/guides/custom-controls/how-to-create-advanced-custom-controls) is helpful. We value a minimal approach in Mapsui so if a simpler solution suffices this is better. I like the fact that this direct property is now in the Avalonia MapControl and needs no `#if`. 

One question: You do not use data binding in the sample because it would need more changes to the sample, fine. But it is possible to use data binding with the DirectProperty?

PR #2484, opened 2024-02-05
Move rotation snap and more Pinch handling to the shared Navigator

pauldendulk:


Mapsui.UI.WinUI/MapControl.cs
```diff
@@ -224,25 +225,22 @@ private static void OnManipulationInertiaStarting(object sender, ManipulationIne
...
+    }
 
-        Map.Navigator.Pinch(center, previousCenter, radius / previousRadius, rotationDelta);
-        e.Handled = true;
+    private PinchState GetPinchState(ManipulationDeltaRoutedEventArgs e)
+    {
+        // Todo: Get center relative the MapControl. The center is a application wide location. If the MapControl is not
+        // on the top left this will malfunction.
+        var position = e.Position.ToMapsui().Offset(e.Delta.Translation.X, e.Delta.Translation.Y);
+        return new PinchState(position, e.Delta.Scale, _totalPinchRotation + e.Delta.Rotation);
```
pauldendulk, line 243: I could use help here. I can't find the way to calculate the relative position to the MapControl.

pauldendulk:
Although this is still not perfect this is a good example of the kind of thing I like to work on in v5. Removed a lot of platform specific code and moved it to the shared navigator. The complications of snapping are now hidden inside the navigator.

Unfortunately it is hard for me to specify what needs to be done in a PR like this in advance. I need to go through the code myself,  first I make the code in all platforms similar, and then I move it to the Navigator.

pauldendulk:
Some things to improve later:
- In Blazor we still track the previous state locally. It would make sense to treat this in the same way but in Blazor it is tied to dragging, so to fix this would require a further reorganization of dragging. And I am not sure what is the best way to do this. Should drag be completely separated or is drag like pinch with a rotation of 0 and radius change of 1
- The state is now tracked in the Navigator. Better than to track it in all MapControls. But it is a but obscure (What does ClearPinchState do exactly?). The Navigator already does a lot. Perhaps there should be a PinchTracker field in the MapControls shared code.
- There are still some differences in the way pinch is called. This is because WinUI/WPF differs from the other platforms because it works with deltas and there is no access to the touch positions. This is a bit annoying, but it is enough information for our purpose. Perhaps all platforms should adopt the WinUI/WPF approach and work with a PinchManipulation argument instead of the PinchState.

pauldendulk:
Tested:
- [x] Avalonia
- [x] Uno
- [x] MAUI Android
- [x] .NET Android
- [ ] Blazor - It is a bit harder because you have to debug the website on Android. Did a few weeks ago, but skipped it this time
- [ ] iOS - I don't have a Mac and the alternatives are a hassle
- [ ] WPF - I don't have a touch screen on my laptop.  Are there other options?
- [ ] Eto - Touch not supported

PR #2488, opened 2024-02-10
Improvement: Remove CreateInstance from GradientStyle

inforithmics:
Follow Up of Pull Request https://github.com/Mapsui/Mapsui/pull/2349
Here I remove the last CreateInstance Call in Mapsui.

### DIFF
--- a/Mapsui/Styles/CalloutStyle.cs
+++ b/Mapsui/Styles/CalloutStyle.cs
@@ -473,4 +473,9 @@ public class CalloutStyle : SymbolStyle
             SubtitleFont.Invalidated = value;
         }
     }
+
+    public override IStyle CreateInstance()
+    {
+        return new SymbolStyle();
+    }
 }
--- a/Mapsui/Styles/IStyle.cs
+++ b/Mapsui/Styles/IStyle.cs
@@ -30,4 +30,8 @@ public interface IStyle
     /// Gets or sets the objects overall opacity
     /// </summary>
     float Opacity { get; set; }
+
+    /// <summary> Creates an Empty Instance For Gradient Style Calculation </summary>
+    /// <returns></returns>
+    IStyle CreateInstance();
 }
--- a/Mapsui/Styles/LabelStyle.cs
+++ b/Mapsui/Styles/LabelStyle.cs
@@ -209,5 +209,9 @@ public class LabelStyle : Style
         return Text;
     }
 
-
+    /// <inheritdoc/>
+    public override IStyle CreateInstance()
+    {
+        return new LabelStyle();
+    }
 }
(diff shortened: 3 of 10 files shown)


pauldendulk (changes requested):
Adding the CreateInstance to the interface forces all styles to implement the interface. Also new styles or custom styles. And this is just because of some code somebody wrote 20 years ago (I think this is from Mortens first version). Instead I think it can be solved locally. 

```csharp
        var style = (IStyle)Activator.CreateInstance(MinStyle.GetType())!;
        if (MinStyle is LabelStyle && MaxStyle is LabelStyle)
            CalculateLabelStyle(style as LabelStyle, MinStyle as LabelStyle, MaxStyle as LabelStyle, attr);
        if (MinStyle is VectorStyle && MaxStyle is VectorStyle)
            CalculateVectorStyle(style as VectorStyle, MinStyle as VectorStyle, MaxStyle as VectorStyle, attr);
        if (MinStyle is SymbolStyle && MaxStyle is SymbolStyle)
            CalculateSymbolStyle(style as SymbolStyle, MinStyle as SymbolStyle, MaxStyle as SymbolStyle, attr);
        return style;
```

The first line creates a generic style, but after that only three type of styles are used in the logic. So instead of copying it first, we could check the type first and then create the specific instances through the regular consturctor (new LabelStyle(), new VectorStyle(), new SymbolStyle(). 

```csharp
        if (MinStyle is LabelStyle && MaxStyle is LabelStyle)
        {
            var labelStyle = new LabelStyle();
            CalculateLabelStyle(labelStyle, MinStyle as LabelStyle, MaxStyle as LabelStyle, attr);
            return labelStyle;
        }
        if (MinStyle is VectorStyle && MaxStyle is VectorStyle)
        {
            var vectorStyle = new VectorStyle();
            CalculateVectorStyle(vectorStyle, MinStyle as VectorStyle, MaxStyle as VectorStyle, attr);
            return vectorStyle;
        }
        if (MinStyle is SymbolStyle && MaxStyle is SymbolStyle)
        {
            var symbolStyle = new SymbolStyle();
            CalculateSymbolStyle(symbolStyle, MinStyle as SymbolStyle, MaxStyle as SymbolStyle, attr);
            return symbolStyle;
        }
        throw new NotSupportedException($"Style type '{MinStyle.GetType()}' and/or '{MinStyle.GetType()}' can not be used in the GradientTheme");
```
btw, the shapefile sample test failed. Supposedly because of the change in this PR. Looking af the code I don't see why it should fail.

pauldendulk:
In a more compact variant the CalculateXxxStyle methods could be rewritten to InterpolateXxsStyle. They could create the instance  inside the method and return the new instance so that the three lines would become one line.

inforithmics:
I'll make a new pull request

pauldendulk:
@inforithmics Sorry, I got carried away and started rewriting this in my local version. So, no need to create a new PR. I will make one.

PR #2493, opened 2024-02-11
Rewrite GradientTheme and remove Activator

pauldendulk:


Mapsui/Styles/Thematics/GradientTheme.cs
```diff
@@ -119,130 +84,118 @@ public GradientTheme(string columnName, double minValue, double maxValue, IStyle
...
-
-        var style = (IStyle)Activator.CreateInstance(MinStyle.GetType())!;
-        if (MinStyle is LabelStyle && MaxStyle is LabelStyle)
-            CalculateLabelStyle(style as LabelStyle, MinStyle as LabelStyle, MaxStyle as LabelStyle, attr);
-        if (MinStyle is VectorStyle && MaxStyle is VectorStyle)
-            CalculateVectorStyle(style as VectorStyle, MinStyle as VectorStyle, MaxStyle as VectorStyle, attr);
-        if (MinStyle is SymbolStyle && MaxStyle is SymbolStyle)
-            CalculateSymbolStyle(style as SymbolStyle, MinStyle as SymbolStyle, MaxStyle as SymbolStyle, attr);
-        return style;
+        return (MinStyle, MaxStyle) switch
```
pauldendulk, line 87: We can remove the activator because we do not deal with any type, the three types we handle in the following code, so we can create the interpolated instance inside that method. Also writing it like this the min and max can not be null and we do not need to take this into account in our methods.

Mapsui/Styles/Thematics/GradientTheme.cs
```diff
@@ -119,130 +84,118 @@ public GradientTheme(string columnName, double minValue, double maxValue, IStyle
...
     }
 
-    private void CalculateVectorStyle(VectorStyle? style, VectorStyle? min, VectorStyle? max, double value)
+    private VectorStyle ToInterpolatedVectorStyle(VectorStyle min, VectorStyle max, double value)
     {
-        if (style is null)
-            return;
-        var fraction = Fraction(value);
-        style.Enabled = fraction > 0.5 ? min?.Enabled ?? false : max?.Enabled ?? false;
+        var result = new VectorStyle();
```
pauldendulk, line 98: I often work with such a single create result line at the start and a return result at the end of the method. At a glance you can see what this method is about, altering this result. Often no other state is changed.

Mapsui/Styles/Thematics/GradientTheme.cs
```diff
@@ -119,130 +84,118 @@ public GradientTheme(string columnName, double minValue, double maxValue, IStyle
...
 
-    private bool InterpolateBool(bool min, bool max, double attr)
-    {
-        var frac = Fraction(attr);
-        return frac > 0.5 ? max : min;
+        return result;
     }
 
-    private string? InterpolateString(string? min, string? max, double attr)
+    private static double Fraction(double attr, double min, double max)
```
pauldendulk, line 160: All methods from here on down are now static pure functions. I am a big fan or pure functions. It took just a little rewriting, calculate the fraction up front, not on the inside.

Mapsui/Styles/Thematics/GradientTheme.cs
```diff
@@ -119,130 +84,118 @@ public GradientTheme(string columnName, double minValue, double maxValue, IStyle
...
-    private Brush InterpolateBrush(Brush min, Brush max, double attr)
-    {
-        return new Brush { Color = InterpolateColor(min.Color, max.Color, attr) };
-    }
+    private static bool InterpolateBool(bool min, bool max, double fraction) => fraction > 0.5 ? max : min;
 
-    private Pen InterpolatePen(Pen min, Pen max, double attr)
-    {
-        return new Pen
+    private static string? InterpolateString(string? min, string? max, double fraction) => fraction > 0.5 ? max : min;
```
pauldendulk, line 170: I like the `=>` syntax to make methods really short. I guess you could over-use it. Although it is also a matter of getting used to it.

Mapsui/Styles/Thematics/GradientTheme.cs
```diff
@@ -119,130 +84,118 @@ public GradientTheme(string columnName, double minValue, double maxValue, IStyle
...
+    private VectorStyle ToInterpolatedVectorStyle(VectorStyle min, VectorStyle max, double value)
     {
-        if (style is null)
-            return;
-        var fraction = Fraction(value);
-        style.Enabled = fraction > 0.5 ? min?.Enabled ?? false : max?.Enabled ?? false;
+        var result = new VectorStyle();
+
+        var fraction = Fraction(value, Min, Max);
+        result.Enabled = fraction > 0.5 ? min.Enabled : max.Enabled;
```
pauldendulk, line 101: A whole lot of these null checks could be removed. It is often a lot better to check null before passing an argument into a method. Then again there are also cases where there are a lot of calls to an method and it is easier if the method checks null so the caller does not have to. In this case there was just one call and around ten null checks inside the method, sometimes passing the problem even deeper into the call stack.

Mapsui/Styles/Thematics/GradientTheme.cs
```diff
@@ -119,130 +84,118 @@ public GradientTheme(string columnName, double minValue, double maxValue, IStyle
...
-        var style = (IStyle)Activator.CreateInstance(MinStyle.GetType())!;
-        if (MinStyle is LabelStyle && MaxStyle is LabelStyle)
-            CalculateLabelStyle(style as LabelStyle, MinStyle as LabelStyle, MaxStyle as LabelStyle, attr);
-        if (MinStyle is VectorStyle && MaxStyle is VectorStyle)
-            CalculateVectorStyle(style as VectorStyle, MinStyle as VectorStyle, MaxStyle as VectorStyle, attr);
-        if (MinStyle is SymbolStyle && MaxStyle is SymbolStyle)
-            CalculateSymbolStyle(style as SymbolStyle, MinStyle as SymbolStyle, MaxStyle as SymbolStyle, attr);
-        return style;
+        return (MinStyle, MaxStyle) switch
+        {
```
pauldendulk, line 88: I use a switch expression here. I am still not quite comfortable with this myself but David Fowler likes it, so I guess I should too.

Mapsui/Styles/Thematics/GradientTheme.cs
```diff
@@ -119,130 +84,118 @@ public GradientTheme(string columnName, double minValue, double maxValue, IStyle
...
+            (LabelStyle minLabelStyle, LabelStyle maxLabelStyle) => ToInterpolatedLabelStyle(minLabelStyle, maxLabelStyle, attr),
+            (SymbolStyle minSymbolStyle, SymbolStyle maxSymbolStyle) => ToInterpolatedSymbolStyle(minSymbolStyle, maxSymbolStyle, attr),
```
pauldendulk, line 90: I changed the order of the checks here because the compiler warned me that the SymbolStyle would never be reached, because it is derived from VectorStyle. I got this warning in the switch-expression, not in the previous if-statement, but I guess it applied there too. So I assume the SymbolStyle method was never called in the past.

pauldendulk: This is not true. The previous if checked for the exact type and the SymbolStyle logic was applied. The switch written like this has more options for error. The order of the types makes a difference, not nice, but the compiler warns so no real big problem. I will leave it like this.

Mapsui/Styles/Thematics/GradientTheme.cs
```diff
@@ -119,130 +84,125 @@ public GradientTheme(string columnName, double minValue, double maxValue, IStyle
...
-            CalculateLabelStyle(style as LabelStyle, MinStyle as LabelStyle, MaxStyle as LabelStyle, attr);
-        if (MinStyle is VectorStyle && MaxStyle is VectorStyle)
-            CalculateVectorStyle(style as VectorStyle, MinStyle as VectorStyle, MaxStyle as VectorStyle, attr);
-        if (MinStyle is SymbolStyle && MaxStyle is SymbolStyle)
-            CalculateSymbolStyle(style as SymbolStyle, MinStyle as SymbolStyle, MaxStyle as SymbolStyle, attr);
-        return style;
+        return (MinStyle, MaxStyle) switch
+        {
+            (LabelStyle minLabelStyle, LabelStyle maxLabelStyle) 
+                => ToInterpolatedLabelStyle(minLabelStyle, maxLabelStyle, attr, this),
```
pauldendulk, line 90: Here I pass in the instance of the GradientTheme. This is so that I can make the method a static pure function as well. The instance functions to pass in the configuration fields. Since the GradientTheme is immutable now (all fields are 'init') nothing can be changed from the outside or inside the pure function. 

Something to improve would be if I had some way to formally pass it in as immutable (I think it would need c# changes). With the current solution I could add an immutable field to the GradientTheme to break the immutability. I would like to have a compile error if that happened.

This is the first time I use this way to pass in configuration properties. I think there are many other situations where I could use this pattern.

inforithmics (approved):
Good Improvements. I like that the settings are init only so it is guaranted to not change after creation.
And with the removal of the Activator it is much less magic code and more straight forward

PR #2494, opened 2024-02-11
Improvement: Simplifying IVectorCache Interface

inforithmics:
I removed the SkPath and SkPaint Type constraints from IVectorCache, because they are not really needed and it simplify things. (Less casts when using VectorCache)


pauldendulk (approved):
Great, cleaning up is our highest priority in this phase.

Issue #2372, opened 2023-12-29
Make the MapControl disposable and not the Map


pauldendulk:
*Note: This is an idea which might be hard to realized, perhaps impossible. Still I think it is good way to think about this problem because it is a clear objective. The solution that comes out of this may be different.*

**The problem**
We need to deal with a lot of IDisposable classes in our code base. There are a lot of issues cause by this. We need to deal think about it in most of our PRs. If I search for 'SuppressMessage("IDisposableAnalyzers' I get 100 hits. 😬

**The proposed solution**
Rewrite our code so that the user only needs to dispose the MapControl. Of course many of it's children need to be disposed, but that that is not the users problem. The Map and its children (Layers) should not need dispose.

**What do we need to dispose?**
My idea right now is that we have a couple of categories that will always need dispose. Those are:
- SkiaSharp object. This is part of the rendering. The Renderer is a field in the MapControl. Those need to be cached and the renderer should be responsible to clear and dispose the cache. 
- Timers. Those are mostly related to fetching. This could be resolved with a [centralized fetcher](https://github.com/Mapsui/Mapsui/issues/2269) in the MapControl.
- HttpClient and database access. Those are part of the providers. Perhaps we need to register the providers to a registry that is owned by the MapControl (I don't have a clear picture of how that should work. Perhaps a layers only specifies which kind of Provider is used, and the fetcher creates and disposes it.

PR #2496, opened 2024-02-11
WIP Improvement: Remove IDisposable from map

inforithmics:
I removed the IDisposable Interface from Map:
I avoid Memory Leaks by Implementing Weak Event Pattern for the Events in the Layer so The Dispose does not to be called To remove the Events.


inforithmics:
Could be that it introduces Merge conflicts with this pull request.
https://github.com/Mapsui/Mapsui/pull/2479

pauldendulk:
I like to have some better understanding of what is going on. 

The main thing I see when scanning the code is that event listeners now use weak references, but the layers themselves also need to be disposed. They have timers, and HttpClients. This is the responsibility of the user?

What is the downside of using weak references? If there are no references to an instance it will be garbage collected. Is this not happening when there are only weak references to an instance?

pauldendulk:
Also note this issue to discuss decisions https://github.com/Mapsui/Mapsui/issues/2372

inforithmics:
1. For disposing HttpClients and Timers I plan to make a RenderService Class that creates the HttpClients and Timers and keeps track of the usage with CacheHolder maybe I should rename this then to IServiceHolder or something more appropriate. I haven't decided yet on names. So disposing the RenderService disposes all Timers and HttpClients and stops them.
2. The WeakEvent Pattern, makes the Event Registration not a strong Garbage Collection dependency. So when you register a Class on a Weak Event and the Class isn't hold anywhere else it is garbage collected.
https://learn.microsoft.com/en-us/dotnet/desktop/wpf/events/weak-event-patterns?view=netdesktop-8.0
The advantage of this is that now Memory Leaks can occour because of Events. And so I don't need to Deregister from Events when I remove the Layer. So the Layer will be garbage collected when it isn't used anymore.

pauldendulk:
@inforithmics I was hoping to have time to take a closer look at this this weekend but I am in the flow with the touch logic and I want to continue with that (The TouchTracker will do the work for both pinch and touch, and now it becomes related to the widget touch. I have a plan for that and want to continue on that tomorrow).

In general I think we need weak references anyway. I was thinking about Messages to replace INotifyPropertyChanged in some areas, but Message would also benefit from weak references. I see you made a lot of unit tests, great, yet I have not looked into the details at all yet.

inforithmics:
Yes messages is another way.  And yes the Messages need to use Weak References too so that when a component is used otherwise it gets garbage collected.

pauldendulk:
I have been doing a little research related to event and disposing. I will past it here so that it does not get lost.

Usually I just assign an Action but it is also possible to use `+=` on an action. I did not know:

```
var c = new Class();
c.Action = () => { Console.Write("Hello "); return true; };
c.Action += () => { Console.Write("World"); return true; };
c.CallAction(); // Output: Hello World
Console.WriteLine();
c.Action = () => { Console.Write("Hi"); return true; };
c.CallAction(); // Output: Hi

class Class
{
    public void CallAction() => Action();
    public Func<bool> Action { get; set; } = () => { Console.Write("Default"); return true; };
}
```

It is not possible to just assign with `=` on a delegate. And removing the last delegate causes a null reference exception, where you could also expect it to call an empty list. 

```
public class Program
{
    public delegate void MyDelegate();

    public static void Main()
    {
        MyDelegate myDelegate = new MyDelegate(Method1);
        myDelegate.Invoke(); // Output: Method 1

        myDelegate += Method2;
        myDelegate.Invoke(); // Output: Method 1, Method 2

        myDelegate -= Method1;
        myDelegate.Invoke(); // Output: Method 2

        myDelegate -= Method2;
        myDelegate.Invoke(); // Null reference exceptions. Why would it not iterate over an empty list. As the invoker I do not need to know who is listening to me.
    }

    public static void Method1() => Console.WriteLine("Method 1");
    public static void Method2() => Console.WriteLine("Method 2");
}

```
An Action behaves the same. Perhaps they are not that different. On an action you can use the = operator, on a delegate you have to use: `= new MyDelegate(Method1);`.

inforithmics:
It seems my changes have changed something fundamentally because the Unit Tests seem to be broken. So I'm thinking about closing this pull request.

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
```diff
@@ -276,7 +276,7 @@ private static void DrawContent(CalloutStyle callout, SKCanvas canvas, ISymbolCa
             {
 
                 // Get size of content
-                var bitmapInfo = (BitmapInfo)symbolCache.GetOrCreate(callout.Content);
+                using var bitmapInfo = (BitmapInfo)symbolCache.GetOrCreate(callout.Content);
```
pauldendulk, line 279: I think you can not add this `using` here. I looked at this code too because there is a warning about disposing, but you can not dispose it here because the bitmap is added to the cache. The cache is responsible for disposing it. I like the dispose warnings and would like to rewrite this code so we have more useful warnings but I don't know how this should be done (also did not spend time on it).

PR #2499, opened 2024-02-12
Add ManipulationTracker that gets touch/mouse locations as input and translates it into a Manipulation object

pauldendulk:
This is a follow up of the previous Pinch PR. It deals with the 2nd and 2th bullet in this remark:

https://github.com/Mapsui/Mapsui/pull/2484#issuecomment-1937463171

The problem we had is that WinUI/WPF is different from the others in that it does not provide all touch state, but it does provide enough for what we need. By introducing PinchManipulation we go with what is available on WinUI/WPF, making those platforms very simple. In the other platforms we use the PinchTracker to help track the information we need construct the PinchManipulation state. 

Why is this better:
- The Navigator.Pinch now has a single interface that can be used that way from all platforms. There are platform differences but those are now dealt with by having different logic in the platforms.
- The preserved state is still a tiny bit obscure but it isolated to the PinchTracker, which is a class that is fairly easy to understand in itself.


pauldendulk:
In the later commits the PinchTracker is using touch locations instead of PinchState. This was because in Blazor Pinch and Drag were conflated. To resolve this the PinchTracker now also deals with both Pinch and Drag on Blazor. Logical steps to follow up on this are:
- Use the pinch tracker for touch on all platforms.
- Rename the PinchTracker to TouchTracker.
- Remove the PinchTracker.Restart and make the PinchTracker.Update detect a restart.  This can be done by checking the number of touch locations and/or comparing it to the previous touch locations. We may also need to call PinchTracker.Update in situations where there are no fingers touching, like on OnTouchUp, OnCancelled, OnExit.

Tests/Mapsui.Tests/Navigation/TouchTrackerTests.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    record Input(List<MPoint> Touches, TouchManipulation? TouchManipulation, string Message);
+
+    [Test]
+    public void ManipulationSequenceTests()
+    {
+        // Arrange
+        var inputs = new List<Input>
+        {
+            new([], null, "No touches and no previous input."),
```
pauldendulk, line 17: I did not use TestCase here because I wanted to work with a sequence of events. The messages help with detecting where and why things fail.

Mapsui.UI.Avalonia/MapControl.cs
```diff
@@ -59,28 +52,38 @@ private void Initialize()
...
-        PointerExited += MapControlMouseLeave;
-        PointerCaptureLost += MapControlPointerCaptureLost;
-        PointerWheelChanged += MapControlMouseWheel;
-        DoubleTapped += OnDoubleTapped;
+        PointerMoved += MapControl_PointerMoved;
+        PointerExited += MapControl_PointerExited;
+        PointerCaptureLost += MapControl_PointerCaptureLost;
+        PointerWheelChanged += MapControl_PointerWheelChanged;
+
+        Tapped += MapControl_Tapped;
```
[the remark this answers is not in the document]
pauldendulk: More about widget touch. All our 'Touched' events are 'Tapped' events, not TouchUp events. They could be called from the platform Tapped events. We have one 'Touching' (TouchDown) event for the EditingWidget. I think we really need that. We do not use the TouchUp anywhere but I think we will need it for drag and drop functionality at some point.

pauldendulk:
To test touch input from my input I connect it through usb and run:
`adb reverse tcp:<port-on-android> tcp:<port-on-windows>`

#docs-need-update

Mapsui.Nts/Widgets/EditingWidget.cs
```diff
@@ -58,15 +58,13 @@ public bool HandleWidgetMoving(Navigator navigator, MPoint position, WidgetTouch
 
         if (args.LeftButton)
         {
-            EditManipulation.Manipulate(
+            return EditManipulation.Manipulate(
```
pauldendulk, line 61: This was a bug, the widget is responsible to inform the caller if the events is handled or not.

Mapsui.UI.Android/MapControl.cs
```diff
@@ -118,22 +116,26 @@ public SkiaRenderMode RenderMode
         }
     }
 
-    private void OnDoubleTapped(object? sender, GestureDetector.DoubleTapEventArgs e)
+    private void OnSingleTapped(object? sender, GestureDetector.SingleTapConfirmedEventArgs e)
```
pauldendulk, line 119: I would like to use platform gestures for MapInfo and widget Taps but that is not always possible.

Mapsui.UI.Android/MapControl.cs
```diff
@@ -169,87 +171,30 @@ public void OnFling(object? sender, GestureDetector.FlingEventArgs args)
...
-                    Map.Navigator.ClearPinchState();
-                    Map.Navigator.Pinch(GetPinchState(touchPoints));
-                }
-                else
-                {
-                    _mode = TouchMode.Dragging;
-                    _previousTouch = touchPoints.First();
-                }
-                Refresh();
+                _manipulationTracker.Restart(locations);
```
pauldendulk, line 185: The amount of code to deal with touch in the MapControls is strongly reduced.

Mapsui.UI.Android/MapControl.cs
```diff
@@ -169,87 +171,30 @@ public void OnFling(object? sender, GestureDetector.FlingEventArgs args)
...
-                            if (touchPoints.Count < 2)
-                                return;
-
-                            Map.Navigator.Pinch(GetPinchState(touchPoints));
-                        }
-                        break;
-                }
+                if (HandleWidgetPointerMove(locations[0], true, 0, false))
+                    return;
+                _manipulationTracker.Manipulate(locations, Map.Navigator.Pinch);
```
pauldendulk, line 192: This line does a lot `_manipulationTracker.Manipulate(locations, Map.Navigator.Pinch);`. The Manipulation tracker updates its state, and based on the current updates and the previous one creates a Manipulation object. The object is then passed into the action which is the second parameter (it is a method group, so not easy to recognize as an event).

pauldendulk:
There is still some work to do on the TapGesture (and perhaps others), but after this is merged I am going to focus on widget touch first. Some things I have in mind:

- The current Up/Touched event is a Tap event (not just on Up). I am going to rename it.
- I am going to rename Touching/Touch to Down/Up
- I am going to introduce an Up event, even though we do not need it at the moment, but we will on some point, and it will make sense to have all three. That is what people expect.
- The Tap event will get only one parameter, the is-tap-gesture check needs to be done in the MapControl, or by a component used in the MapControl
- I am going to try add an info log event that works for all widgets to write a line on up, down and tapped (not move).

pauldendulk:
Note, I made big changes, this will introduce bugs, and there will be things that are just unfinished. oh, I did not test .NET for iOS at all.

PR #2500, opened 2024-02-16
WIP No caching in Blazor

pauldendulk:
Depends on https://github.com/Mapsui/Mapsui/pull/2499

I do not intent to merge this. But use this for discussion. I had a lot of trouble testing Blazor touch from my Android because it kept on caching on older version, no matter what I tried. The tricks in this PR improved this problem, but I am not entirely sure if it fixed it.


pauldendulk:
I don't see any caching problems when working with Blazor on my Windows desktop. Perhaps I need better tools on my Android. I will try firefox next time.

PR #2503, opened 2024-02-20
Fix: End Of Stream Exception on PointZ Shapefile

inforithmics:
Because Points have only 2 Values reading 4 values causes an end of stream exception on the last Point in a Shapefile with PointZ values


pauldendulk (approved):
Great, thanks for the fix

PR #2504, opened 2024-02-20
[4.1] Fix End Of Stream Exception on PointZ Shapefile

inforithmics:
Fix: End Of Stream Exception on PointZ Shapefile on Develop 4.1.2


pauldendulk (approved):
Marvelous

Issue #2033, opened 2023-06-07
Highly sensitive zoom when using touchpad 


pauldendulk:
**Mapsui Version**
4.0.0-rc.2

**Mapsui Platform**
Blazor

**Device**
Mac (perhaps also on Linux)

**Describe the bug**
Highly sensitive zoom when using touchpad 

**To Reproduce**
1. Start Mapsui.Samples.Blazor
2. Zoom in using touchpad
3. See zooming in will be way too fast

**Expected behavior**
Similar behavior as on Windows

**Additional context**
This could be the issue as we had in Avalonia:
https://github.com/Mapsui/Mapsui/pull/1984

pauldendulk:
This bug report was submitted to me through a user. I do not use mac so can not test this. Could one of you please test this @inforithmics @janusw? You do not need to open the IDE, just go to the samples and use the mouse pad to zoom in. https://mapsui.com/samples/. Do you have any additional information? What could it be? Does it look similar to this issue https://github.com/Mapsui/Mapsui/pull/1984? Does mouse wheel zoom behave properly?

inforithmics:
I have unfortunatly only a MacMini (without Touchpad), but the Mouse Zoom behaves correctly.

inforithmics:
I think it is the same issue as in https://github.com/Mapsui/Mapsui/pull/1984? Because in Blazor the DeltaY is defined as double so it could provide fracations of a int value causing the to fast zoom. So the solution is the same as in https://github.com/Mapsui/Mapsui/pull/1984 make an mousewheel aggregator variable.

inforithmics:
I think it is the same issue as in https://github.com/Mapsui/Mapsui/pull/1984? Because in Blazor the DeltaY is defined as double so it could provide fracations of a int value causing the to fast zoom. So the solution is the same as in https://github.com/Mapsui/Mapsui/pull/1984 make an mousewheel aggregator variable.

PR #2507, opened 2024-02-21
[4.1] Add option to configure mouse-wheel-zoom-step-size in Blazor and Avalonia

pauldendulk:


RobertWatchful:
I looked at the variable speed of the mouse wheel event, and while it is quite a good feature for the mouse wheel, It doesn't seem to be so good for the trackpad users -so it is just an idea.
  private DateTime lastScrollTime = DateTime.Now;
    private double accumulatedDeltaY = 0;
    protected void OnMouseWheel(WheelEventArgs e)
    {
        TimeSpan elapsed = DateTime.Now - lastScrollTime;

        // Check if the time elapsed is quick or not
        // This doesn't work well with a touchpad for some reason?
        if (elapsed.TotalMilliseconds < 500)
        {
            Console.WriteLine($"MouseWheel speed < 500 (: {elapsed.TotalMilliseconds}");
            Map.Navigator.MouseWheelZoomContinuous(Math.Pow(2, e.DeltaY > 0 ? 0.25 : -0.25), e.ToLocation(_clientRect));
        } else
        {
            Console.WriteLine($"MouseWheelDelta speed > 500: {elapsed.TotalMilliseconds}");
            Map.Navigator.MouseWheelZoomContinuous(Math.Pow(2, e.DeltaY > 0 ? 0.08 : -0.08), e.ToLocation(_clientRect));
        }

        // Update last scroll time for the next event
        lastScrollTime = DateTime.Now;

        //if (UseContinuousMouseWheelZoom)
        //{
        //    var mouseWheelDelta = Math.Pow(2, e.DeltaY);
        //    Console.WriteLine($"MouseWheelDelta: {mouseWheelDelta}");
        //    var currentMousePosition = e.ToLocation(_clientRect);
        //    Map.Navigator.MouseWheelZoomContinuous(mouseWheelDelta, currentMousePosition);
        //}
        //else
        //{
        //    var mouseWheelDelta = (int)e.DeltaY * -1; // so that it zooms like on windows
        //    var currentMousePosition = e.ToLocation(_clientRect);
        //    Map.Navigator.MouseWheelZoom(mouseWheelDelta, currentMousePosition);
        //}
    }

pauldendulk:
This fixes https://github.com/Mapsui/Mapsui/issues/2033. 

It makes the mouse wheel step configurable, which makes touchpad on Mac usable. For touchpad Windows it is also an improvement. Pads simulate the mouse wheel by triggering an event but are more frequent than a real mouse wheel. By making is configurable it becomes manageable, better would be if we could detect if it is a real mouse wheel or a pad.

pauldendulk:
This PR fixed touchpad but the mouse wheel now feels too slow. Perhaps we should use something to detect the device:
https://stackoverflow.com/questions/10744645/detect-touchpad-vs-mouse-in-javascript

RobertWatchful:
Hi Paul

Yes Paul. We discussed the same. We could look at this Wednesday?
Robert Greenwood

Quintham.com

On 15 Mar 2024, at 19:16, Paul den Dulk ***@***.***> wrote:

﻿

This PR fixed touchpad but the mouse wheel now feels too slow. Perhaps we should use something to detect the device:
https://stackoverflow.com/questions/10744645/detect-touchpad-vs-mouse-in-javascript

PR #2510, opened 2024-02-23
[4.1] Add option to zoom in on dblclick on blazor

pauldendulk:


pauldendulk:
Decided to not add this. It is not really needed for the project I work on. I assume that users can add their own zoom in on dblclick as well.

PR #2513, opened 2024-02-29
Some rework around WidgetEventArgs

pauldendulk:
Things done:
- Rename to WidgetEventArgs (it is not touched)
- Some renaming from Touched to Tapped
- Removed event and EventArgs from Tapped function and now using Func.
- Use e instead of args for EventArgs
- Use more pure static create functions


Mapsui/Widgets/ButtonWidgets/ButtonWidget.cs
```diff
@@ -3,12 +3,12 @@
...
 namespace Mapsui.Widgets.ButtonWidgets;
 
-public class TextButtonWidget : TextBoxWidget, ITouchableWidget
+public class ButtonWidget : TextBoxWidget, ITouchableWidget
 {
     /// <summary>
     /// Event which is called if widget is touched
     /// </summary>
-    public event EventHandler<WidgetTouchedEventArgs>? Touched;
+    public Func<ButtonWidget, WidgetEventArgs, bool> Tapped = (s, e) => false;
```
pauldendulk, line 11: The Tapped method is not an `event` anymore and not of type `EventHandler`. The advantage is that it now can return handled status and and the `WidgetEventArgs` can be immutable. You can now also assign Tapped with `=` instead of `+=` so that it can be initialized within a single line static create method.

PR #2515, opened 2024-03-03
Move call to OpenInBrowser from MapControl to HyperlinkWidget

pauldendulk:


pauldendulk:
This makes the MapControl a bit cleaner and it is easier to understand how the widget logic is supposed to work. I tested in MAUI, WPF, Blazor, WinUI/Uno and Avalonia. On Eto it is not working because the widget logic is not properly implemented.

PR #2516, opened 2024-03-04
More cleanup around MapControl manipulation

pauldendulk:


Mapsui.UI.Blazor/MapControl.cs
```diff
@@ -263,11 +261,11 @@ protected virtual void Dispose(bool disposing)
 
     public void OpenInBrowser(string url)
     {
-        Catch.TaskRun(async () =>
+        Catch.TaskRun(() =>
         {
             if (JsRuntime != null)
-                await JsRuntime.InvokeAsync<object>("open", [url, "_blank"]);
-        });    
+                _ = JsRuntime.InvokeAsync<object>("open", [url, "_blank"]);
```
pauldendulk, line 267: This OpenInBrowser did not work with the await. No error showed, just nothing happend. I don't know why.

Mapsui.UI.Eto/MapControl.cs
```diff
@@ -21,12 +21,8 @@ public partial class MapControl : SkiaDrawable, IMapControl
...
     public MapControl()
     {
-        CommonInitialize();
-        ControlInitialize();
-    }
+        SharedConstructor();
 
-    private void ControlInitialize()
-    {
         _invalidate = () => RunOnUIThread(Invalidate);
```
pauldendulk, line 26: Move the code to the constructor where possible. This helps the compiler to see that objects created in the constructor are cleaned up in the dispose. Also readonly can now be added. This was not possible for .NET Android and .NET iOS, unfortunately.

Mapsui.UI.Eto/MapControl.cs
```diff
@@ -21,12 +21,8 @@ public partial class MapControl : SkiaDrawable, IMapControl
 
     public MapControl()
     {
-        CommonInitialize();
-        ControlInitialize();
-    }
+        SharedConstructor();
```
pauldendulk, line 24: Renamed CommonInitialize to SharedConstructor to indicate it is really meant to be called from the constructor (and thus should be cleanup in the Dispose.

PR #2529, opened 2024-03-08
Add ScreenPosition record to use for all screen positions

pauldendulk:


pauldendulk:
For screen positions we now use the ScreenPosition record everywhere. That is in the manipulations, touch/mouse events, widgets and many places in the MapControls.

PR #2531, opened 2024-03-09
Add WidgetPointerReleased

pauldendulk:


pauldendulk:
It was necessary for releasing pointer capture in the editing tool, and will be necessary for other scenarios.

PR #2532, opened 2024-03-09
Add shared OnMapPointerPressed/Moved/Released methods which are called from the platforms

pauldendulk:
Another step towards more shared code. With respect to manipulation it looks like there is not much more to share.


pauldendulk:
- _tapGestureTracker is now called only twice in all code. 
- _manipulationTracker needs to stay in the platforms because it is not used everywhere.
- Simplified fling
- The SourceGenerator has a compile error in VS on windows when run with net8 (while it compiles on the server), reverted to netstandard2.0 again, a bit weird.

PR #2539, opened 2024-03-10
Move widget events to IWidget and add default implementation to Widget

pauldendulk:


pauldendulk:
Moving the events to IWidget and Widget makes you think a bit less (you just know all widgets have all events) and it reduces the code (you do not have to add default implementations on various places in the hierarchy). The only downside I could think of is performance, widgets with just default implementations will now be called as well. I think this will have a very small impact. If it really matters we could add an 'InputTransparent = true' property to filter them out early.

I renamed TouchAreaType to WidgetAreaType (it is not just about touch, alternatives for Widget could be 'manipulation' (but hover is not manipulation) or 'interaction'. I also rename Viewport to Map, I think this is slightly better for users who are not very familiar with Mapsuis concept (but they will always know there is a map and what its area is).

The EditingWidget now also used WidgetAreaType.Map, just like the MouseCoordinateWidget.

Toughts:
- Perhaps Widget should be renamed to BaseWidget.
- EditingWidget needs no renderer anymore (it is just interacting and using other widgets), but it does need a default renderer, so I derived from TextBoxWidget. There must we a better way. I think it is good that users get an error when they did not think about a renderer. Perhaps you should derive from InteractionOnlyWidget.

PR #2540, opened 2024-03-10
Improvement: Weak events Implementation in BaseLayer

inforithmics:
This adds Weak Events to the BaseLayer, this is Step into the direction of removing the IDisposable of the Map.

Is taken from this closed pull request. https://github.com/Mapsui/Mapsui/pull/2496

### DIFF
--- a/Mapsui/Layers/BaseLayer.cs
+++ b/Mapsui/Layers/BaseLayer.cs
@@ -6,11 +6,14 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel;
 using System.Threading;
+using Mapsui.UI;
 
 namespace Mapsui.Layers;
 
 public abstract class BaseLayer : ILayer
 {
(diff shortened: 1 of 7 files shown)


pauldendulk (approved):
Great. I have to do some stepping through the code to understand it better. 

The DataChangedEventHandler might become part of an event channel at some point. I have to think about that.

Love the tests.

pauldendulk:
@inforithmics When I run the WPF samples the map does not load when selecting the sample half of the time. I could not figure out how my EditingWidget changes could have caused this, and it turned out it also showed in other samples. I reverted the weak DataHasChanged logic to the old logic and now it works like before. I will push this revert as part of my EditingWidgetChanges.  

I did not investigate the problem. One thing that comes to mind: The weak references are faster and the event arrives too early, before the mapcontrol is initialized. If this is the case there is probably an existing bug underneath it that shows due to the change in performance.

PR #2541, opened 2024-03-10
Feature: Add CancellationToken to Fetchinfo to abort Fetches WIP

inforithmics:
Adding a CancellationToken to FetchInfo to abort Fetches. This is the second part to remove IDisposable from Map. Because then I can have a Central CancellationTokenSource and when this is aborted all fetches will stop. More than todays IAsyncDataFetchers. So when this is implemented The IDisposable can be removed.

This implementation isn't finished yet, I think there are other Places where I can add the CancellationToken to improve the abort.


pauldendulk:
I was thinking about System.Threading.Channels.Channel to replace the workers. It may still need a CancellationToken but it will probably work different.

pauldendulk:
The goals is to remove IDisposable from Map. How will this work eventually? The HttpCients still need to be disposed. I would think this first step is to make the ILayer not disposable.

My idea is to solve this with a central FetcherMachine which is not in the Layer but in the MapContol or the Map. But this is a big change with many complications which I want to work on myself.

inforithmics:
My idea behind is this. There are two things that are done in the Map Dispose.

1) Disconnecting Events. (This is now handled with WeakEvents)
2) Aborting the Fetching of Data. (For this I planned to use the Cancellation Token).
3) The Dispose can now be removed (Because the Map Behaves now the same.

Further improvements like Central Fetcher, Channels, Disposing of HttpClients, ....

The Central Fetcher would simplify this massively. Maybe I have to Recreate the pull request with CancellationTokens on ly on AsyncData Fetch Functions.

Like
```
public interface IProvider
....
Task<IEnumerable<IFeature>> GetFeaturesAsync(FetchInfo fetchInfo);
```

inforithmics:
I'll redoing this pull request

PR #2554, opened 2024-03-17
Implement GetShiftPressed on Eto

pauldendulk:
Added GetShiftPressed  for Eto but not for Blazor and MAUI. This is the status for now:

- WPF: Done
- Eto: Done
- Avalonia: Done
- WinUI/Uno: Done. I can't say anything about if this could work on Android or iOS
- Blazor: There was some code but it did not work and I did not see a way to fix it, so I remove the existing code.
- MAUI: It seems there is no support for this in .NET 8, but might be in .NET 9.
- .NET Android: Not implemented
- .NET iOS: Not implemented

My plan is to leave it like this is for the time being.

Currently the shift state is checked in the EditingWidget to:
- Finalize adding a linestring or polygon
- Delete a vertex in a linestring or polygon

Both can also be done with a double tap.

### Future work:
- Change EditingWidget into a generic ManipulationWidget and move the EditManipuation and EditManager back to the samples.
- In double tap check if the second tap is near the first tap.

### DIFF
--- a/Mapsui.UI.Blazor/MapControl.cs
+++ b/Mapsui.UI.Blazor/MapControl.cs
@@ -16,7 +16,6 @@ public partial class MapControl : ComponentBase, IMapControl
     protected readonly string _elementId = Guid.NewGuid().ToString("N");
     private SKImageInfo? _canvasSize;
     private bool _onLoaded;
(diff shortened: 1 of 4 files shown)


pauldendulk:
In my last test Avalonia did not work either.

PR #2556, opened 2024-03-18
Fix json parsing

pauldendulk:
Use JObject instead. See https://github.com/Mapsui/Mapsui/discussions/2549#discussioncomment-8825521

### DIFF
--- a/Samples/Mapsui.Samples.Common/Maps/Geometries/PointsSample.cs
+++ b/Samples/Mapsui.Samples.Common/Maps/Geometries/PointsSample.cs
@@ -72,18 +72,16 @@ public class PointsSample : ISample
     private static List<City> DeserializeFromStream(Stream stream)
     {
         using var streamReader = new StreamReader(stream);
-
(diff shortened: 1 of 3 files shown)


inforithmics:
Normally Newtonsoft.Json should be replaced by System.Text.Json which is AOT compatible.

pauldendulk:
> Normally Newtonsoft.Json should be replaced by System.Text.Json which is AOT compatible.

I tried that and got AOT related warnings which hinted me to add JsonTextInfo to the constructor, which needed a json options argument, which had in turn another argument, so I went with JObject instead. I would like to see how this should work with System.Text.

btw, Is it not odd we have this error while aot is turned off? Could this be caused by the FolderProfile.pubxml which has RunAOTCompilation as true?

inforithmics:
You have to use the system text json source generator to make it work.

https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation?pivots=dotnet-8-0

PR #2564, opened 2024-03-23
Improvement: Added CancellationToken to IProvider.GetFeaturesAsync

inforithmics:
For better Aborting of Features I added a CancellationToken to IProvider.GetFeaturesAsync

### DIFF
--- a/Benchmarks/Mapsui.Rendering.Benchmarks/RenderSkCanvasPerformance.cs
+++ b/Benchmarks/Mapsui.Rendering.Benchmarks/RenderSkCanvasPerformance.cs
@@ -1,12 +1,6 @@
 ﻿#pragma warning disable IDE0005
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using BenchmarkDotNet.Attributes;
 using BenchmarkDotNet.Engines;
-using ExCSS;
 using SkiaSharp;
 
 namespace Mapsui.Rendering.Benchmarks;
--- a/Benchmarks/Mapsui.Rendering.Benchmarks/RenderToCpuPerformance.cs
+++ b/Benchmarks/Mapsui.Rendering.Benchmarks/RenderToCpuPerformance.cs
@@ -1,5 +1,4 @@
 ﻿#pragma warning disable IDE0005
-using System.IO;
 using BenchmarkDotNet.Attributes;
 using Mapsui.Nts.Providers.Shapefile;
 using Mapsui.Providers;
@@ -14,11 +13,6 @@ using Mapsui.Rendering.Skia.Tests;
 using Mapsui.Styles.Thematics;
 using Mapsui.Nts.Providers;
 using SkiaSharp;
-using BenchmarkDotNet.Configs;
-using BenchmarkDotNet.Jobs;
-using BenchmarkDotNet.Toolchains.InProcess.Emit;
-using Mapsui.Rendering.Skia.Cache;
-using Mapsui.Utilities;
 
 #pragma warning disable IDISP001
 #pragma warning disable IDISP003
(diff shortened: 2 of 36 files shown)


pauldendulk (commented):
I am undecisive about this one. Making the Map none-disposable is an important improvement, but I feel uneasy about adding the cancellation token to an important interface. I hope we can think of an entirely different solution where we do not have to deal with disposing, async and cancelation at this level. First I will take a look at BruTile, where I intent to remove the HttpClient altogether. The caller should have the responsibility of managing the HttpClient (creating (using threadpool?)  and disposing). In this case it will be Mapsui, so that won't fix things for us, but maybe we can make similar changes in Mapsui.

inforithmics:
I agree, because mapsui heavily depends on Brutile, it's best to have a strategy for this in Brutile and when it's implemented use a similar strategy in mapsui.

In .net most of the Async IO Interfaces have a CancellationToken Parameter for cancelling IO operations.  So I simply made the same changes here so that the CancellationToken can be relayed to httpClient, File IO operations. Sqlite pcl net does not support cancellationtokens yet (Or I haven't found out how yet, so mbtiles-brutile cannot use it yet).

As I pointed out in the discussion the CancellationToken could be an Addtional Property in the FetchInfo, avoiding breaking changes. But then the Code Analyzers wouldn't detect that I forget to relay the Cancellation Token.

If there is concern about the removing of IDisposable from Map. I could do that in a seperate Pull request. Because with the WeakEvents the only thing that dispose does in the Map is aborting the fetches. Which I can call directly in the Map instead of calling Dispose.

Sample for Cancellation Token interfaces in the .NET Api
https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.getasync?view=net-8.0

Here every method has a variant with and without a CancellationToken.

For easier usage I changed the Interface to this.

```
public interface IProvider
{
    string? CRS { get; set; }

    MRect? GetExtent();

    Task<IEnumerable<IFeature>> GetFeaturesAsync(FetchInfo fetchInfo)
    {
        return GetFeaturesAsync(fetchInfo, CancellationToken.None);
    }

    Task<IEnumerable<IFeature>> GetFeaturesAsync(FetchInfo fetchInfo, CancellationToken cancellationToken);
}
```

Which means users don't have to recompile or change code because, if you don't use Cancellation Tokens it will internally call the new Interface with the CancellationToken.None.

Only implementers need to implement the GetFeaturesAsync(FetchInfo fetchInfo, CancellationToken cancellationToken); They need to recompile and change Source Code

inforithmics:
The failing ArcGisDynamicSample test is fixed here https://github.com/Mapsui/Mapsui/pull/2574

inforithmics:
I restore the Dispose of the Map (Less changes for better reviewing).

pauldendulk:
> First I will take a look at BruTile, where I intent to remove the HttpClient altogether. The caller should have the responsibility of managing the HttpClient (creating (using threadpool?) and disposing). In this case it will be Mapsui, so that won't fix things for us, but maybe we can make similar changes in Mapsui.

This has been done a while ago.

pauldendulk:
The review of the PR is way over due, sorry about that. And today I won't be able to review it properly either. But let me post a comment just to get my thoughts started.

> In my opinion CancellationToken and Disposability are somehow orthogonally connected.

Good point. And that was what I did not like initially. Maybe it should be a separate merge request. My thought now is that it just makes sense to add a cancelation token to a request that could really take long. Disposing is a separate topic, with a lot of complicated aspects. I did something about this here:  https://github.com/Mapsui/Mapsui/pull/2844 Your comments are welcome. 

> but keeps for example the httpClient alive

For HttpClient we could perhaps use an entirely different solution. See my comment on this PR https://github.com/Mapsui/Mapsui/pull/2844#issuecomment-2600843678

Also relevant: https://github.com/Mapsui/Mapsui/issues/2703
And: https://github.com/Mapsui/Mapsui/issues/2649

inforithmics:
I think to get this green this pull request needs to be first merged (because some Idisposable warnings came up).
https://github.com/Mapsui/Mapsui/pull/2844

pauldendulk:
I merged https://github.com/Mapsui/Mapsui/pull/2844. Sorry for all the conflicts.

PR #2568, opened 2024-03-28
Update: sqlite to version that has offical .net 8 support

inforithmics:
Update sqlite to Version that officially supports .NET 8


pauldendulk (approved):
Great, thanks

PR #2569, opened 2024-03-28
Replace: Newtonsoft.Json with System.Text.Json which is AOT Compatible

inforithmics:
The successor of NewtonSoft.Json is System.Text.Json which is AOT Compatible with the Source Generator.
Mapsui already has a dependency on System.Text.Json because of NetTopologySuite.IO.GeoJson4STJ. So with this replacement only 1 Json Library is used - System.Text.Json

System.Text.Json is twice as fast as Newtonsoft.Json
https://www.linkedin.com/pulse/benchmark-systemtextjson-vs-newtonsoftjson-net-8-2024-hussin-tgifc/

I added an ArcGisDynamic Sample too because no sample for this code path existed yet


Samples/Mapsui.Samples.Common/Maps/Geometries/PointsSample.cs
```diff
@@ -71,17 +72,7 @@ private class City
...
-        var jArray = JArray.Parse(str);
-
-        return jArray.Select(c => new City
-        {
-            Name = c[nameof(City.Name)]?.Value<string>(),
-            Country = c[nameof(City.Country)]?.Value<string>(),
-            Lat = c[nameof(City.Lat)]?.Value<double>() ?? 0,
-            Lng = c[nameof(City.Lng)]?.Value<double>() ?? 0
-        }).ToList();
+        return JsonSerializer.Deserialize(stream, PointsSampleContext.Default.ListCity) ?? [];
```
pauldendulk, line 75: The new code looks nice and compact but I don't understand how it works yet. Thanks for improving my previous jObject fix. I intended to fix it myself but now I don't have to.

Samples/Mapsui.Samples.Common/Maps/Geometries/PointsSample.cs
```diff
@@ -94,3 +85,8 @@ private static SymbolStyle CreateBitmapStyle()
         return new SymbolStyle { BitmapId = bitmapId, SymbolScale = 0.20, SymbolOffset = new Offset(0, bitmapHeight * 0.5) };
     }
 }
+
+[JsonSerializable(typeof(List<PointsSample.City>))]
+internal partial class PointsSampleContext : JsonSerializerContext
```
pauldendulk, line 90: This is part of the json trick. So, this is somehow related to source generators?

Directory.Packages.props
```diff
@@ -62,7 +62,6 @@
     <PackageVersion Include="Microsoft.Windows.SDK.BuildTools" Version="10.0.26031-preview" />
     <PackageVersion Include="NetTopologySuite" Version="[2.5.0,3.0.0)" />
     <PackageVersion Include="NetTopologySuite.IO.GeoJSON4STJ" Version="[4.0.0,5.0.0)" />
-    <PackageVersion Include="Newtonsoft.Json" Version="[13.0.2,14.0.0)" />
```
pauldendulk, line 65: Remove the entire reference, great!

pauldendulk (approved):
Great job! Fits within our clean up phase.

Samples/Mapsui.Samples.Common/Maps/Geometries/PointsSample.cs
inforithmics: 1.
On the PointsSampleContext a list of Attributes is added for which JSon Deserialization Code should be created.

The System.Text.Json  SourceGenerator generates PointsSampleContext.Default.ListCity and now I can use it to deserialize this type.

Samples/Mapsui.Samples.Common/Maps/Geometries/PointsSample.cs
inforithmics: Yes here I tell which Classes this Context can deserialize. For that System.Text.Json generates code with a SourceGenerator.

inforithmics: The important part is to set typeof(List<PointsSample.City>)) so that it creates a serializer for the a List<PointsSample.City>.

PR #2571, opened 2024-03-29
Remove a few of the null-forgiving operators (!)

pauldendulk:
The bang operator is an easy way to suppress the null warning. Perhaps this could be the best option in some cases but it should not be the general way to go about the null warnings, because that would defeat the purpose of turning nullable on. When working more with nullable=true, and working on new code, I feel less need to add the bang, perhaps it may not be needed at all. Also it may be better to have no bang, because if the compiler does not understand that something can not be null it is not easy to understand for the developer either. If we make it easy for the compiler we also make it easy for the developer.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -27,7 +27,7 @@ csharp_style_expression_bodied_accessors = true:silent
 csharp_style_expression_bodied_lambdas = true:silent
 csharp_style_expression_bodied_local_functions = false:silent
 csharp_style_inlined_variable_declaration = true:suggestion
-csharp_style_deconstructed_variable_declaration = true:suggestion
+csharp_style_deconstructed_variable_declaration = true:silent
 csharp_style_var_for_built_in_types = true:silent
 csharp_style_var_when_type_is_apparent = true:silent
 csharp_style_var_elsewhere = true:silent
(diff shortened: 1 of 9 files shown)


.editorconfig
```diff
@@ -27,7 +27,7 @@ csharp_style_expression_bodied_accessors = true:silent
 csharp_style_expression_bodied_lambdas = true:silent
 csharp_style_expression_bodied_local_functions = false:silent
 csharp_style_inlined_variable_declaration = true:suggestion
-csharp_style_deconstructed_variable_declaration = true:suggestion
+csharp_style_deconstructed_variable_declaration = true:silent
```
pauldendulk, line 30: I am inclined to go over all 'messages' and to inspect the ... suggestions in the IDE. When I think they are options but not general improvements I would prefer to make them silent. This is such an example, tuple deconstruction is not better in general in my option. The option to refactor is still available in the IDE when set to silent.

Mapsui.Extensions/Projections/DotSpatialProjection.cs
```diff
@@ -64,7 +61,7 @@ public static void Init()
         [DisallowNull] (ProjectionInfo From, ProjectionInfo To)? transform)
     {
         var pointsXy = new[] { x, y };
-        Reproject.ReprojectPoints(pointsXy, Array.Empty<double>(), transform.Value.From, transform.Value.To, 0, 1);
+        Reproject.ReprojectPoints(pointsXy, [], transform.Value.From, transform.Value.To, 0, 1);
```
pauldendulk, line 64: I love the new collection initializers.

PR #2574, opened 2024-03-30
Fix: Samples in Blazor

inforithmics:
Looking through the Code I found some instances where the Credentials were set for The HttpClientHandler without a try catch
PlatformNotSupportedException, this is necessary for Blazor because Blazor does not support this and throws a PlatformNotSupportedException

Clicking through the Blazor samples I discovered that following samples didn't work
1. WMTS
2. WMS called tiled
I fixed them (service url changed)
4. Fixed Wfs Caching, postData is now a cache criteria
5. Fixed WFS Points sample, because of CORS only HTTP Get is allowed and not HTTP Post.
6. Fixed ArcGisDynamicSample Unit Test (Missing Caching for Reproducability)


inforithmics:
In the process of fixing the WFS Point sample I discovered a caching Bug. I only cache the url but not depending on the post data, which is important because there the parameters are written

pauldendulk:
ah, it is important to fix this kind of thing and to be clear about what we support. What will the user experience be when users selects the 'wms called tiles' sample in Blazor?

inforithmics:
The user experience is that it works now. But the wms called tiled and wets sample didn’t work on other platforms either because, the wms and wets adresses changed.

pauldendulk:
> The user experience is that it works now. 

ah, that is a nice feature 😁

> But the wms called tiled and wmts sample didn’t work on other platforms either because, the wms and wmts addresses changed. The platform not supported fix is a Blazor specific fix.

In the current Blazor sample just nothing happens, that is not good. In debug you see an error, that's better. I think we should turn on error logging in map in our samples. Perhaps just all logging, if that is useful for the samples.

I don't see the 'ArcGIS dynamic' sample in the WPF sample. Do you happen to know why that is?

Note, better error messages should not be a part of this PR, but this is a situation where we should investigate this.

inforithmics:
That the sample isn't displayed in wpf I have to investigate. Have you tried a rebuild all after getting the Source Code. Could Be that the SourceGenerator that creates the sample calling Code wasn't run.

inforithmics:
I tried the wpf Sample and for me it displayed the dynamic Arcgis sample. You have to wait some time because it is not the fastest service.

inforithmics:
I'm still investigating why the wfs sample does not work in blazor. I progressed one step further.
1. If the Wfs service is asked with https:// for capabibilites other calls are expected to be called with https:// too or else it fails in blazor. (CORS)
2. Currently the wfs post for getting the points does not work yet in blazor, it works in wpf or other targets.

Mapsui/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+
+public static class StringExtensions
+{
+    public static string? UriScheme(this string? url)
```
pauldendulk, line 5: I think it is better to call the GetUriScheme.

Mapsui/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        if (string.IsNullOrEmpty(url))
+        {
+            return null;
```
pauldendulk, line 9: In general is better to throw early when something it happening that you do not expect. I don't understand the situation fully, but it seems to me here it would be better to always throw when the scheme can not be determined. If you know of a situation where it could not be determined then we need to discuss how to deal with it. If we both don't know when we should accept null but there is a real case where we should accept null than it is still be to throw now that we do not know and then think of a better solution when we find out.

Mapsui/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            return uri.Scheme;
+        }
+
+        return null;
+    }
+
+    public static string? AssureUriScheme(this string? url, string? scheme)
+    {
+        if (string.IsNullOrEmpty(scheme) || string.IsNullOrEmpty(url))
```
pauldendulk, line 22: Here too I think we should throw instead of return null. Also don't accept null instead there is a good reason for it.

Samples/Mapsui.Samples.Common/Maps/DataFormats/WfsPointsSample.cs
```diff
@@ -67,6 +67,7 @@ private static async Task<WFSProvider> CreateWfsProviderAsync()
             "laser_2002_04_punkte",
             WFSProvider.WFSVersionEnum.WFS_1_1_0);
 
+        provider.GetFeatureGetRequest = true;
```
pauldendulk, line 70: What does GetFeatureGetRequest do? What went wrong without this line?

pauldendulk (changes requested):
I have a few questions and propose to throw and not accept null in the uri scheme extensions.

Mapsui/Extensions/StringExtensions.cs
inforithmics: Done

Mapsui/Extensions/StringExtensions.cs
inforithmics: I throw now exceptions when the schema is not valid

Mapsui/Extensions/StringExtensions.cs
inforithmics: i throw now exceptions

Samples/Mapsui.Samples.Common/Maps/DataFormats/WfsPointsSample.cs
inforithmics: what I thought I could implment is a fallback on GetFeatureGetRequest when the post request does not work.

pauldendulk: Okay, I have worked with CORS before but is not easy to see from that code that this solves that problem. In general my reasoning is that you should be able to understand what code is doing by reading it, and if you don't you are allowed to remove it to find out (sometimes much later) why it is there. But I understand that this is all based on very old code which can not be easily rewritten. So fine for now.

pauldendulk (approved):

Issue #2572, opened 2024-03-29
PNG images not optimised


mikeirvingweb:
the PNGs in the project have not been optimised.

many could be losslessly compressed, resulting in a smaller repo, and smaller app footprints

PR #2577, opened 2024-04-01
Maximize png compression for reference images

pauldendulk:
After this PR https://github.com/Mapsui/Mapsui/issues/2572 it made sense to also change our compression when saving reference images to disk. However, even when using the maximum png compression of skia the images are bigger than they were after the previous PR: A total size of on disk of 72.5 MB instead of 62.5 MB.  This is not a big problem. Over times, when we need to update reference images, the total size on disk will grow again. Also, when we copy an image over an optimized image the binary will not be the same anymore.


pauldendulk:
When I use WebP the total size on disk is 55.9 MB. That might be worth it. But the should rename the files to .webp too.

inforithmics:
I agree with the webpage renaming the generated webpage images are lossless so no artefacts should be introduced

pauldendulk:
I am also thinking about not putting the images in the repository at all to reduce the data. One thing I am worried about is that if we checkout a branch we get all the diffs, also of the images, this means that over time if we keep updating the images the repository will grow even if we do not increase the number of images. I assume that is how git works, did not read up about it, and chatgpt agrees.

An alternative could be to write the reference images to a docker and in our code point to a specific version of that image. If we update the reference images we need to update the image. That way we won't have a history of image diffs in our repository.

inforithmics:
You can make a shallow checkout on hit which will only fetch the files of the requested branch without history.

https://stackoverflow.com/questions/68933323/git-shallow-checkout-a-remote-branch

pauldendulk:
But that is what every person would have to do individually, right? Random people cloning the repo would not do this.

PR #2578, opened 2024-04-02
WIP Use System.Drawing.Color

pauldendulk:
Ported to System.Drawing.Color but now I doubt if this is a good idea.

Why not:
- We have to add the System.Drawing.Point in many places where before we only had Mapsui.Style.
- There are many namespace conflicts between System.Drawing.Point and NetTopologySuite.Geometries.Point. Solvable, but they are in our most simple samples. Users will run into this all the time.
- There are also conflicts between System.Drawing.Pen and Mapsui.Styles.Pen. Originally the Mapsui Pen was converted from the System.Drawing.Pen when those things were removed from netstandard (or maybe PocketPC, don't remember). You could consider also porting Pen back to System.Drawing.Pen, but then maybe we should also port Brush (if that exists). Thinking about this I feel uneasy, I am not sure I want to depend on those. Perhaps it is better to stay independent. Their choices might not be the best for us.
- We have to use Color.FromArgb a lot instead of a regular constructor, which is a little simpler.
- Color.FromArgb has a overload that accepts only R, G and B, so it leaves out the first parameter, this is weird.

What to do instead:
- Turn System.Mapsui.Color into a struct, like System.Drawing.Color.
- Remove all the string names like "WhiteSmoke" and replace them with the constants.
- Add an implicit cast from System.Drawing.Color to System.Mapsui.Color.
- Add Color.FromArgb and Color.FromRgba to be more compatible.

Remarks:
- I had some doubt wrt the order of the A. In FromArgb it is up front, but I would think it is more logical to have it last (like we have in the styles). I would also think it is more common to have it last. ChatGPT agrees with that. Also SKColor has the A last. So, I conclude, that we got it right and System.Drawing.Color is just weird.


pauldendulk:
Closing this because we will use an updated version of Mapsui.Styles.Color instead https://github.com/Mapsui/Mapsui/pull/2593

PR #2582, opened 2024-04-07
Fix ArcGisDynamicProvider extent

pauldendulk:
The extent of the ArcGISDynamicProvider was in latlon. Now it is projected using the DefaultProjection.

### DIFF
--- a/Mapsui.ArcGIS/DynamicProvider/ArcGISDynamicProvider.cs
+++ b/Mapsui.ArcGIS/DynamicProvider/ArcGISDynamicProvider.cs
@@ -112,8 +112,15 @@ public class ArcGISDynamicProvider : IProvider, IProjectingProvider
     {
         if (ArcGisDynamicCapabilities.initialExtent == null)
             return null;
+        if (CRS is null)
+            return null;
+        var extent = ArcGisDynamicCapabilities.initialExtent.ToMRect();
+        if (extent is null)
+            return null;
 
-        return ArcGisDynamicCapabilities.initialExtent.ToMRect();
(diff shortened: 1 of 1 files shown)


Mapsui.ArcGIS/DynamicProvider/ArcGISDynamicProvider.cs
```diff
@@ -112,8 +112,15 @@ public async Task<IEnumerable<IFeature>> GetFeaturesAsync(FetchInfo fetchInfo)
     {
         if (ArcGisDynamicCapabilities.initialExtent == null)
             return null;
+        if (CRS is null)
+            return null;
+        var extent = ArcGisDynamicCapabilities.initialExtent.ToMRect();
+        if (extent is null)
+            return null;
+
+        ProjectionDefaults.Projection.Project("EPSG:4326", CRS, extent);
```
inforithmics, line 121: I think it should be something like this because the EPSG should be read from the capabilites

```
 var fromCRS = $"EPSG:{ArcGisDynamicCapabilities.spatialReference?.wkid ?? 4326}";
        ProjectionDefaults.Projection.Project(fromCRS, CRS, extent);
```

inforithmics (changes requested):

inforithmics (approved):

inforithmics:
Good fix

Mapsui.ArcGIS/DynamicProvider/ArcGISDynamicProvider.cs
pauldendulk: Thanks! fixed it. And some typos in comments.

PR #2583, opened 2024-04-08
Always log in Blazor samples

pauldendulk:
In our own sample we want logging enabled.

### Future work: LogLevel
btw, we currently have no filtering on log level anywhere in our own code. I would think this is up to the specific loggers. Perhaps the LoggingWidget should have it's own. Should it be a static (which I think was the best option for the ShowLoggingInMap field) or an instance field. Or perhaps a LoggingWidget.DefaultLogLevel which is static and an instance LoggingWidget.LogLevel.


charlenni:
[LogLevelFilter](https://github.com/Mapsui/Mapsui/blob/a60ce72cad612f52f4c72c6673dbbb7d7f59779a/Mapsui/Widgets/InfoWidgets/LoggingWidget.cs#L91)

PR #2586, opened 2024-04-10
Feature: Support Wfs 2.0

inforithmics:
This adds WFS 2.0 support

1. Mainly changing Namespaces because with WFS2.0 xml Namespaces have now version numbers.
2. Adding a Test Wfs 2.0 Sample too.
3. Fixed ArcGisDynamicSample Unit Test, the cache and unit test image seemed not to correspond, so I updated them


pauldendulk (approved):
Great, thanks for taking care of this.

PR #2594, opened 2024-04-15
Replace RenderCache with RenderService

pauldendulk:


pauldendulk:
@inforithmics To cut things down into smaller steps I first made a PR to introduce the IRenderService which does not implement the caches itself. Although this is a small PR there is already a problem for which I don't have a good solution for yet. The caches need to be disposed. Our dispose analysis tool complains about the setters of the cache because we dispose the caches on the inside while they could be created and assigned on the outside. Perhaps this is not always a real problem but we can make it much easier for ourselves to have clear responsibilities. The way I solve it now is to make the RenderService responsible for creating and disposing the caches. In itself an improvement, but I had to add a rather random DisableVectorCache() method to make this possible. There is room for improvement here, but I don't have time to think about that today.

Mapsui/Rendering/IRenderService.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Mapsui.Rendering;
+
+public interface IRenderService : IDisposable
+{
+    ILabelCache LabelCache { get; }
+    ISymbolCache SymbolCache { get; }
+    ITileCache TileCache { get; }
+    IVectorCache VectorCache { get; }
+    public void DisableVectorCache();
```
inforithmics, line 11: Good idea to make a disableVectorCache function, so the Vector Cache doesn't need to be settable.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -24,10 +24,10 @@ namespace Mapsui.Rendering.Skia;
 
 public sealed class MapRenderer : IRenderer, IDisposable
 {
-    private readonly DisposableWrapper<IRenderCache> _renderCache;
+    private readonly DisposableWrapper<IRenderService> _renderService;
```
inforithmics, line 27: Replacement of IRenderCache with IRenderService

Mapsui.Rendering.Skia/SkiaStyles/LabelStyleRenderer.cs
```diff
@@ -60,20 +60,20 @@ public bool Draw(SKCanvas canvas, Viewport viewport, ILayer layer, IFeature feat
             {
                 case (PointFeature pointFeature):
                     var (pointX, pointY) = viewport.WorldToScreenXY(pointFeature.Point.X, pointFeature.Point.Y);
-                    DrawLabel(canvas, (float)pointX, (float)pointY, labelStyle, text, (float)layer.Opacity, renderCache);
+                    DrawLabel(canvas, (float)pointX, (float)pointY, labelStyle, text, (float)layer.Opacity, renderService.LabelCache);
```
inforithmics, line 63: Now the member needs to be used instead of the renderservice itself

inforithmics (approved):
Simplifies the IRenderService(IRenderCache) interface and lays the foundations for further improvements

Mapsui/Rendering/IRenderService.cs
inforithmics: Maybe an EnableCache(ECache) and a DisableCache(ECache) could be made so that other caches could be enabled or disabled too Or better a propertie.  public ECache EnabledCaches {get;set;}
This are only other ideas I don't have a strong opinion about it.

Mapsui.Rendering.Skia/SkiaStyles/LabelStyleRenderer.cs
pauldendulk: Yes, it will result in a bit more code but I think it is clearer what is going on.

Mapsui/Rendering/IRenderService.cs
pauldendulk: Yes, it is weird we have this method for the VectorCache. I could add methods for all caches, but we would also have to add such a method if another cache is added. We could also and an Enabled property to the cache itself, and we would not work with another implementation of the cache.

pauldendulk: I decided to go for VectorCache.Enabled = false and remove the NonCachingVectorCache. In general I like the idea of overriding the functionality with another implementation. I use that a lot myself but in those applications I use dependency injection. Alternative implementations are injected through the constructor, but if we do it like that the injector would become responsible for disposing which does not fit will in our current design.

It now looks like this:
```csharp
_mapRendererWithoutCache.RenderService.VectorCache.Enabled = false;
```

inforithmics:
> Merging this. Sorry for causing a conflict @inforithmics.

git merge -X theirs main

helped (when conflicts arise the main version is taken). Afterwards I had to fix the build

PR #2598, opened 2024-04-18
Fix: Nullreference exception in MapView Constructor

inforithmics:
Fixes a Bug when a null Reference Exception occurs when new MapView Control is created.
This is because the Layer Fields are not initialized when when Layer Changed Events are called. 
So I initialized the Layers now in the Fields or Properties instead of the Constructor fixing this.
To reproduce before this pull request only the Mapview Test sample needs to be opened.


pauldendulk (approved):

pauldendulk:
Also broke the MapInfo functionality when working on the MapView. At some point while refactoring the manipulation the MapView took the majority of my time, given the number of things that needed fixing I ignored it. Currently I doubt if the MapView should be part of v5 at all.

inforithmics:
Maybe I Can write a MapInfo Widget, not sure if it will work only an idea.

PR #2600, opened 2024-04-20
Use Channels in feature fetch logic

pauldendulk:
This PR was triggered by lagging performance in case of many updating layers, with high update frequencies. 

In the new ManyMutatingLayersSample there are 5 timers calling Refresh() at intervals of 5, 6, 7, 8 and 9 milliseconds. Quite extreme, but it should not impact performance.


Mapsui/Fetcher/FeatureFetchMachine.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace Mapsui.Fetcher;
+
+public class FeatureFetchMachine
+{
+    readonly Channel<Func<Task>> _queue = Channel.CreateUnbounded<Func<Task>>();
+
+    public FeatureFetchMachine(int numberOfWorkers = 2)
```
inforithmics, line 12: From where the numberOfWorkers can be set and saw that it schould be a constructor parameter in the Layer class because else it will always be 2.

Samples/Mapsui.Samples.Common/Maps/Performance/ManyMutatingLayers.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Threading.Tasks;
+using System.Timers;
+
+
+namespace Mapsui.Samples.Common.Maps.Geometries;
+
+public class ManyMutatingLayers : ISample
+{
+    public string Name => "Many Mutating Layers";
+    public string Category => "Performance";
```
inforithmics, line 19: Nice sample I tried it and it works fluid on my windows machine

Mapsui/Fetcher/FeatureFetchDispatcher.cs
```diff
@@ -12,34 +11,35 @@
 
 namespace Mapsui.Fetcher;
 
-internal class FeatureFetchDispatcher<T> : IFetchDispatcher where T : IFeature
+internal class FeatureFetchDispatcher(ConcurrentStack<IFeature> cache)
```
inforithmics, line 14: I like primary constructors too

Mapsui/Layers/Layer.cs
```diff
@@ -77,23 +63,17 @@ public int FetchingPostponedInMilliseconds
...
-    {
-        if (propertyChangedEventArgs.PropertyName == nameof(Busy))
-        {
-            if (_fetchDispatcher != null) Busy = _fetchDispatcher.Busy;
-        }
-    }
 
     private void FetchDispatcherOnDataChanged(object sender, DataChangedEventArgs args)
     {
+        Busy = false; // Todo: Properly implement Busy
```
inforithmics, line 69: I saw that the Busy Property Notification was removed.

inforithmics (approved):

inforithmics:
I don't know why the builds are failing maybe some github infrastructure problems

Mapsui/Fetcher/FeatureFetchMachine.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        for (var i = 0; i < numberOfWorkers; i++)
+        {
+            Catch.TaskRun(() => AddConsumerAsync(_queue));
+        }
+    }
+
+    private async Task AddConsumerAsync(Channel<Func<Task>> queue)
+    {
+        await foreach (var action in queue.Reader.ReadAllAsync())
```
inforithmics, line 22: I think a Cancellation Token should be introduced hear so that it doesn't run forever

pauldendulk: Good point. We have to investigate this. My guess is that this is not necessary here (but it depends on the implementation). When there are no items in the Channel it is not really running but the consuming method is hanging on the ReadAllAsync. What I see in de debugger is that the different calls of the consumer are all on a different thread. So, what I think is happening that it uses no threads when inactive, like a HttpClient call waiting to return. That is how async is designed.

Another question is if the consumer block dispose somehow, but I don't think so, the channel implements no IDispose interface, but perhaps there is something we need to do to clean up.

What also plays a role is the duration of a call. If you want to switch layers you may want to cancel the pending http request. I wonder how important is to do this. The request will either return or time out at some point. If we want to add such canceling I think it should be at the level of the http request, not on the queue consumer. But this is all guessing.

Mapsui/Layers/Layer.cs
pauldendulk: I re-added it in a later commit. My intention is that it works like before, which is not great. It looks like it is not correctly implemented for all layers, and it was problematic during there refactorings. Perhaps we should think about entirely different mechanism.

Samples/Mapsui.Samples.Common/Maps/Performance/ManyMutatingLayers.cs
pauldendulk: Yes, but there was a real big problem in combination with the tile layers. Enable it in the sample, and zoom out to view the whole world and you should see it. Also turn on the Diagnostic Tools to see the garbage collection. I will make an issue for it although I have solved in my local code by replacing all Workers with the Channel consumers.

Issue #1998, opened 2023-05-26
Add SymbolStyle.IconPath as alternative for SymbolStyle.BitmapId


pauldendulk:
**The problem**
Currently the registration and assignment of BitmapId is not very handy. Users need to do a couple of things, load a bitmap, register it and assign it to the SymbolStyle.BitmapId. There are various helpers to make this easier but it is not very clear what should be used.

**The solution**
Add a SymbolStyle.IconPath which specifies the location of the icon, like:
- `SymbolStyle.IconPath = "file://myfolder/myimage.png"`
- `SymbolStyle.IconPath = "http://mywebsite.com/myimage.png"`
- `SymbolStyle.IconPath = "embeddedResource://myassembly.resourses.images.myimage.png"`

The images are loaded by Mapsui and added to BitmapRegistry. The IconPath string should be used as key to make sure they are loaded only once.

The BitmapId could still be used, but not in combination with IconPath.

A longer term goal is to serialize the entire layer definition to file. This would also need bigger changes to how layers are defined, and the fetching logic should be moved from the layers to the MapControl, Map, or a new Fetcher class.

**Alternatives**
I did not spend much time thinking about alternatives yet, but I am sure there are a lot.

pauldendulk:
Having a look of some of our code to see how this could work. The CustomCalloutSample has this string: `const string path = "Mapsui.Samples.Common.GeoData.Json.congo.json";` which is used to retrieve an embedded resource. This could be turned into: `embeddedResource://Mapsui.Samples.Common.GeoData.Json.congo.json`. And the renderer should use that string as the cache. If it is not in the cache the renderer should retrieve it from the assembly.

CustomCalloutSample  also has the CreateCallbackImage which creates an in-memory bitmap stream. For this case I don't have a good solution yet. In some cases we want to work with SKPicture references in memory. We do want to support this use case.

inforithmics:
I'll implement this but the IconPath is an Uri

```
  var file = new Uri("file://myfolder/myimage.png");
  var http = new Uri("http://mywebsite.com/myimage.png");
  var resource = new Uri("embeddedResource://myassembly.resourses.images.myimage.png");
```

tested it (in unit test and all 3 work).

pauldendulk:
This was implemented in a series of merge requests (and involved quite some refactoring).

mfranc28:
Hey where is this ? I don't see it in the last update.

pauldendulk:
[...]

PR #2601, opened 2024-04-20
Feature: Add Possibility to Register Uri as Bitmap

inforithmics:
Part of this
https://github.com/Mapsui/Mapsui/pull/2581

Things done:
1. Adds to the BitmapRegistry the possibility to Add Bitmaps as Uri with the format of this https://github.com/Mapsui/Mapsui/issues/1998 issue.
2. Made BitmapRegistry Threadsafe
3. Added more Unit Tests

The next pull request will be implementing this.
 https://github.com/Mapsui/Mapsui/issues/1998

And the https://github.com/Mapsui/Mapsui/pull/2581 pull request should then only be changing the samples to use the Icon/BitmapPath.


pauldendulk:
Without having to look at the code I would like to understand what it means that the registry is hierarchical . And what is the reason to add this functionality. Do do you have a sample where this would be helpful?

inforithmics:
> Without having to look at the code I would like to understand what it means that the registry is hierarchical . And what is the reason to add this functionality. Do do you have a sample where this would be helpful?

The Hierarchical part is needed. So that there is a static BitmapRegistry.Instance for registering Bitmaps for maps and a dynamic BitmapRegistry Instance in the RenderService. The one in the RenderService is disposed the static one not.

inforithmics:
An alternative implementation could be a RenderServiceBitmapRegistry that is only used in the RenderService.

inforithmics:
@pauldendulk for easier review I removed the Hierarchical part of the BitmapRegistry. Bitmap Registry is now threadsafe and if a get with an invalid bitmapid is done an exception is thrown.

Mapsui/Styles/BitmapRegistry.cs
```diff
@@ -39,13 +48,77 @@ public int Register(object bitmapData, string? key = null)
...
+        switch (bitmapPath.Scheme)
+        {
+            case "embeddedresource":
+                if (resourceCache.TryGetValue(bitmapPath.Host, out var found))
+                {
+                    stream = found.assembly.GetManifestResourceStream(found.realResourceName);
+                }
+                else
+                {
+                    foreach (var assembly in AppDomain.CurrentDomain.GetAssemblies())
```
pauldendulk, line 64: I realized just the problem that if you want to load an embedded resource you also need to know the specific assembly. I had just started thinking about the problem, but now I see you already solved this. Great.

pauldendulk (approved):
Great, thanks a lot.

PR #2602, opened 2024-04-20
Github: Maximum test duration 25s

inforithmics:
Set the maximum Test duration to 25s.

### DIFF
--- a/.github/workflows/dotnet-release-nugets.yml
+++ b/.github/workflows/dotnet-release-nugets.yml
@@ -33,8 +33,7 @@ jobs:
       uses: actions/setup-dotnet@v3
       with:
          dotnet-version: |
-           8.0.x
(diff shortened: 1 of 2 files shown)


pauldendulk (commented):
This PR sets `include-prerelease: false`. What problem did you run into that made you make this change? When I worked with the yml previous weeks I had warnings that `include-prerelease` was not known and I was about to delete. I just googled specifically for `include-prerelease` and get no results. Did you find any documentation about this?

inforithmics:
Include-prerelease can be deleted. It is used for getting preview .net sdks. But because .net 8 is released this isn’t necessary. I will change the pull request accordingly.

inforithmics:
include-prerelease: false is now removed

pauldendulk (approved):
👍

PR #2604, opened 2024-04-21
Remove: SkiaSharp.Views.Maui.Controls.Compatibility 

inforithmics:
SkiaSharp.Views.Maui.Controls.Compatibility is removed and the remaining SkiaSharp.Forms references.
SkiaSharp.Views.Maui.Controls.Compatibility isn't needed anymore. I discovered this while doing.

### DIFF
--- a/Samples/Mapsui.Samples.Maui.MapView/MauiProgram.cs
+++ b/Samples/Mapsui.Samples.Maui.MapView/MauiProgram.cs
@@ -10,7 +10,7 @@ public static class MauiProgram
     {
         var builder = MauiApp.CreateBuilder();
         builder
-            .UseSkiaSharp(true)
+            .UseSkiaSharp()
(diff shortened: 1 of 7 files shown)


Samples/Mapsui.Samples.Maui.MapView/MauiProgram.cs
```diff
@@ -10,7 +10,7 @@ public static MauiApp CreateMauiApp()
     {
         var builder = MauiApp.CreateBuilder();
         builder
-            .UseSkiaSharp(true)
+            .UseSkiaSharp()
```
pauldendulk, line 13: Do we still need the .UseSkiaSharp at all? They were thinking about that a year ago.

pauldendulk (approved):
Good idea, and always good to remove older stuff.

PR #2609, opened 2024-04-25
Upgrade: Uno to 5.2 and use single Project for Samples

inforithmics:
I upgraded to Uno 5.2 and use the new single Project Infrastructure for the Sample, All platforms are now in one project like maui. 

- [x] WinUI
- [x] Browser
- [x] Android
- [x] iOS 

Things done:
1. A lot of Code was removed mainly from the different Uno.WinUI Samples targets
2. Uno.Check runs now in the github actions so that all Uno dependencies are installed.
3. Uno Version is now managed in the global.json
```
 "msbuild-sdks": {
   "Uno.Sdk": "5.2.95"
 }
```
5. Skia is referenced with the UnoFeatures
```
<UnoFeatures>
  Skia;
</UnoFeatures>
```
4. Removed .NET 7 Target because .NET 7 is end of Life in 2024 May


Mapsui.UI.WinUI/Globals.cs
```diff
@@ -0,0 +1,4 @@
+﻿global using System;
```
pauldendulk, line 1: I don't like the use of Global.cs in Mapsui because we deal with many different namespaces and I like to know exactly what is used by looking at the usings. Namespace conflict can become extremely hard to understand when using globals. Although it may not be such a problem within the scope of this project I think it is just more clear to use Global.cs nowhere. Also the benefit here is very small, it affects only one class.

Mapsui.UI.WinUI/MapControl.cs
```diff
@@ -1,4 +1,3 @@
-// Copyright (c) The Mapsui authors.
```
pauldendulk, line 1: I think you deleted this line by accident.

Samples/Mapsui.Samples.Uno.WinUI/.gitignore
```diff
@@ -0,0 +1,403 @@
+## Ignore Visual Studio temporary files, build results, and
```
pauldendulk, line 1: Why a .gitignore is added at this level? We are in a cleanup phase and I would like to remove all the files we do not need, or do not understand. There may be a good reason to add it here but I like to understand why. Is there not a lot of overlap with the root one? Could what's in here not be added the root one? What worries me is if we add it here, why not in all other samples? I want to keep fighting against unnecessary growth of the repository.

global.json
```diff
@@ -8,5 +8,8 @@
     "allowPrerelease": false,
     "dotnet": "8.0.100",
     "rollForward": "latestMajor"
-  }  
+  },
+  "msbuild-sdks": {
```
pauldendulk, line 12: Why was this necessary. What step fails if this is not here?

pauldendulk (changes requested):
This is a great improvement. Lots of files removed. I have a few remarks and questions to look at.

Mapsui.UI.WinUI/Globals.cs
inforithmics: I’ll try to fix this with conditional compilation. In the mapcontrol

Mapsui.UI.WinUI/MapControl.cs
inforithmics: Thanks for finding this. Was a mistake

Samples/Mapsui.Samples.Uno.WinUI/.gitignore
inforithmics: This came with the uno winui template will remove it

global.json
inforithmics: With the uno.winui mechanism this needs in the global.json. I’ll try to move it to the uno.winui samples folder.

inforithmics: I tried moving it to Winui folders but didn't succeed I guess it has really to be in the global.json file and there can only be one at the root.

inforithmics:
@pauldendulk I adressed the comments, the  "msbuild-sdks":  I couldn't move downwards it really needs to be in the global.json file this value is used read by Uno to make make the single project work. It provides the Version Number for all Uno components.

Mapsui.UI.WinUI/Globals.cs
pauldendulk: oh, I think I do not understand the reason for the Global.cs. I thought it was just to have to include less usings, what needs to be fixed?

global.json
pauldendulk: No problem, I just need to understand these kind of things because they will need updating at some point.

pauldendulk (approved):
All great. I like all the cleaning up done in the later commits.

Mapsui.UI.WinUI/Globals.cs
pauldendulk: ok, so, they are really conditionally needed. I prefer the #if over Globals.cs. Fine like this.

.github/workflows/dotnet.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
       with:
-        fetch-depth: 0
+        fetch-depth: 1
```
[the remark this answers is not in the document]
inforithmics: Yes but I had to revert it because package with checkin hash needs history. Until the last version tag.

PR #2622, opened 2024-05-09
Fix unit tests for Busy updates and Delayer

pauldendulk:


Tests/Mapsui.Tests/Fetcher/FeatureFetcherTests.cs
```diff
@@ -23,28 +26,38 @@ public async Task TestFeatureFetcherDelayAsync()
         layer.Delayer.MillisecondsBetweenCalls = 0;
 
         var notifications = new List<bool>();
-        layer.PropertyChanged += (_, args) =>
+
+        // Because we use weak references for the PropertyChanged event handler it can get
+        // garbage collected if use a lambda directly assigned to the layer.PropertyChanged.
+        // So we need to create the variable below to prevent garbage collection.
+        PropertyChangedEventHandler propertyChanged = (_, args) =>
```
pauldendulk, line 33: @inforithmics See my comments above. This `PropertyChangedEventHandler propertyChanged` field is necessary to prevent garbage collection. Something to take into account in our design decisions.

pauldendulk: btw, If we want to make improvements with respect to this I would first invest in using more local data. So, use a slightly more detailed version of world.mbtiles. Or samples of local areas in an mbtiles file.

PR #2623, opened 2024-05-09
Enable the cross platform tests on all platforms

pauldendulk:


.github/workflows/dotnet.yml
```diff
@@ -230,17 +236,13 @@ jobs:
       run: dotnet build --no-restore --configuration Debug Tests/Mapsui.Tests/Mapsui.Tests.csproj
     - name: Mapsui.UI.Maui.Tests
       run: dotnet build --no-restore --configuration Debug Tests/Mapsui.UI.Maui.Tests/Mapsui.UI.Maui.Tests.csproj
```
pauldendulk, line 238: I intend to remove the Maui test and add a Mapsui.Dispose.Test project instead (or perhaps better to put it in Mapsui.Tests). We don't have platform specific tests for any other platform, and I don't want add them all. Also we should move as much code as possible to core so that we do not have to test a lot in the platforms.

Tests/Mapsui.Rendering.Skia.Tests/LabelStyleFeatureSizeTests.cs
```diff
@@ -14,7 +14,16 @@ namespace Mapsui.Rendering.Skia.Tests;
 public class LabelStyleFeatureSizeTests
 {
     // The Sizes are different on MacOs and Windows (windows it is 39.6 and macOS it is 42.6)
-    public readonly double LabelSize = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? 39.6 : 42.6;
+    const double labelSizeOnMac = 42.642578125d;
+    const double labelSizeOnWindows = 39.642578125d;
+    const double labelSizeOnLinux = 40.181640625d;
```
pauldendulk, line 19: I had some trouble getting this to work on linux, had to figure out the linux label size through the build server test log.

Tests/Mapsui.Rendering.Skia.Tests/LabelStyleFeatureSizeTests.cs
```diff
@@ -116,6 +125,6 @@ public void DefaultSizeFeatureSize_Offset_x_y()
         using var renderService = new RenderService();
         var size = LabelStyleRenderer.FeatureSize(feature, labelStyle, skPaint, renderService.LabelCache);
 
-        ClassicAssert.AreEqual(Math.Round(size, 0), Math.Round(LabelSize + Math.Sqrt(2 * 2 + 2 * 2) * 2, 0));
+        Assert.That(size, Is.EqualTo(LabelSize + Math.Sqrt(2 * 2 + 2 * 2) * 2).Within(Constants.Epsilon));
```
pauldendulk, line 128: The Linux label size rounded the wrong way for this test. Instead of rounding I now use a margin. Such a margin is part of all test frameworks. In the v4 nunit you can use `Within` for that. In general this is a more common solution than rounding.

PR #2624, opened 2024-05-09
Tests: Added live samples tests with [ExplicitAttribute]

inforithmics:
Adds TestLiveSmpleAsync with [ExplicitAttribute] to run unit tests against the Live Services in the samples to detect when a service changes url or Layers


pauldendulk (approved):
Lovely! I must say that I have never looked into the details of the DefaultCaches. As I have mentioned I am scared of caches, because they can cause hard to track problems. I think the DefaultCaches were mostly added for this purpose of these sample tests, correct? I wonder if we could limit it's use to this more.

inforithmics:
Yes, that's right they where added mostly for this purpose.

PR #2626, opened 2024-05-09
WIP Improvement: Put Symbols from BitmapPath into SymbolCache

inforithmics:


pauldendulk (approved):

pauldendulk (changes requested):
There is a formatting exception.

pauldendulk:
I already applied your changes [here](https://github.com/Mapsui/Mapsui/pull/2628) so will close this PR.

inforithmics:
Thanks for taking care of this

pauldendulk:
@inforithmics I had some doubts what was the best solution. The pattern proposed by Visual Studio (not sure which tool in VS generates this) is this:

```
{
    private bool _disposed;

    public void Dispose()
    {
        if (_disposed)
        {
            return;
        }

        _disposed = true;
    }

    private void ThrowIfDisposed()
    {
        if (_disposed)
        {
            throw new ObjectDisposedException(GetType().FullName);
        }
    }
}
```

So, it generates the `ThrowIfDisposed()` so apparently this is what we should call in each method. So, very similar to the code in this PR, only we call `ThrowIfDisposed()` instead of `if (_disposed) throw new ObjectDisposedException(nameof(RenderBitmapRegistry));`. So, let's stick to that. Earlier is doubted if we should have a check for this all, but this is more clear.

PR #2628, opened 2024-05-10
Use ImageSource instead of BitmapId everywhere

pauldendulk:
In previous PRs it was made possible to load image streams through a url. The underlying idea of working with such a url is that this url itself could serve as the key for the cache, not needing a bitmapId. The first implementation of bitmapPath did not work like that yet. BitmapPath's were also assigned a bitmapId. So in the case of bitmapPath's this was an extra layer on top of the bitmapPath. This PR attempts to remove that. I introduced a BitmapPathRegistry that should function like that. Eventually this may replace the BitmapRegistry entirely.

This is just te beginnen of a rewrite of the mechanism to cache bitmapPath. I split off a new BitmapPathRegistry which should only store bitmapPath items. So in todays version we have two mechanisms which does not make this better right away.


Mapsui.Rendering.Skia/Cache/RenderBitmapRegistry.cs
```diff
@@ -5,70 +5,90 @@
...
 
-public sealed class RenderBitmapRegistry(BitmapRegistry instance) : IBitmapRegistry
+public sealed class RenderBitmapRegistry(
+    BitmapRegistry bitmapRegistry, BitmapPathRegistry bitmapPathRegistry) : IRenderBitmapRegistry
 {
-    private ConcurrentHashSet<int>? _createdImages = new();
-
+    private readonly ConcurrentHashSet<int> _createdImages = [];
+    private readonly ConcurrentHashSet<string> _createdBitmapPathImages = [];
+    private bool _disposed;
```
pauldendulk, line 13: Now working with two caches here. This is temporary.

Mapsui.Rendering.Skia/Cache/RenderService.cs
```diff
@@ -9,15 +9,15 @@ public RenderService(int capacity = 10000)
...
+        BitmapRegistry = new RenderBitmapRegistry(Styles.BitmapRegistry.Instance, BitmapPathRegistry.Instance);
         VectorCache = new VectorCache(this, capacity);
     }
 
     public ISymbolCache SymbolCache { get; }
     public IVectorCache VectorCache { get; }
     public ITileCache TileCache { get; }
     public ILabelCache LabelCache { get; }
-    public IBitmapRegistry BitmapRegistry { get; }
+    public IRenderBitmapRegistry BitmapRegistry { get; }
```
pauldendulk, line 20: Splitting things up in the phase.

pauldendulk:
This ended up as a very big yak shaving. The purpose was to *use ImageSource instead of BitmapId everywhere* but this led to some additional work. 

In the previous situation the interface depended on skia objects. There was no explicit dependency but the skia objects were passed into the platform independent styles with object pointers and where cast back to skia objects in the renderer. This kind of dependency on skia should be in the custom renderers only. This was the case in a number of classes, and was reflected in the object pointers used in the SymbolCache. 

In the previous situation we used a singleton for the BitmapRegistry, which made it hard to track its logic. This has been replaced with regular instances used as cache. More work needs to be done on caches, but this is something for v6 where we will look at the rendering logic.

Caching of tiles is now separated from caching of symbols. Although they are both images they are treated in quite different ways.

The CustomSvgStyleSample also depended on such logic. To get the same functionality it is now possible to use the SymbolStyle.SvgFillColor and SymbolStyle.SvgStrokeColor, and it is a lot simpler for users to do so. This is in itself quite a big feature.

The CalloutStyle logic has been rewritten. It is now possible to create the callout outline and to put it around a custom SKPicture. Eventually I think that any style should be immutable, and that changes in the appearance should come from the feature data. This will take some more iterations, probably not before the release of v5.

Sprite logic has been rewritten. You can now specify a region to be sampled from the ImageSource.

PR #2630, opened 2024-05-10
Feature: Gpu Acceleration for Uno

inforithmics:
Added UseGpu to Uno so that it can run on Gpu on Android, Wasm and iOS


Mapsui.UI.WinUI/MapControl.cs
```diff
@@ -42,15 +55,29 @@ public MapControl()
             // The commented out code crashes the app when MouseWheelAnimation.Duration > 0. Could be a bug in SKXamlCanvas
             //if (Dispatcher.HasThreadAccess) _canvas?.Invalidate();
             //else RunOnUIThread(() => _canvas?.Invalidate());
-            RunOnUIThread(() => _canvas?.Invalidate());
+            RunOnUIThread(() =>
+            {
+                _canvas?.Invalidate();
+#if HAS_UNO_WINUI
```
charlenni, line 61: The idea behind UseGPU was to select CPU or GPU while runtime. If you set UseGPU before creating MapControl, then it would select the right view. If you do it by your way, you have to select it at compiletime.

inforithmics: This line is here to remove this code from the WinUI target. Because the Code is shared here with Uno.

The Conditional Compiliation is used only for setting the default values of useGpu.
Currently it is supported in Android iOS and WebAssembly on Windows it isn't supported yet.

charlenni: But this isn't correct. You use the conditional compilation also for adding code. Eg. in this case you add or remove the code for invalidation of the GPU view by conditional compilation and not by checking the flag UseGPU. This makes it harder to add the functionality later.

inforithmics: You are right, I have changed the Code now can you have again a Look at it.

#if HAS_UNO_WINUI is used for code that is only valid in Uno.Winui because the Control is shared between Uno.WinUI and WinUI.

Mapsui.UI.WinUI/MapControl.cs
```diff
@@ -29,29 +29,47 @@ namespace Mapsui.UI.WinUI;
...
-        };
+        // The commented out code crashes the app when MouseWheelAnimation.Duration > 0. Could be a bug in SKXamlCanvas
+        //if (Dispatcher.HasThreadAccess) _canvas?.Invalidate();
+        //else RunOnUIThread(() => _canvas?.Invalidate());
 
         Background = new SolidColorBrush(Colors.White); // DON'T REMOVE! Touch events do not work without a background
-
-        Children.Add(_canvas);
+#if HAS_UNO_WINUI
+        if (UseGPU)
```
pauldendulk, line 52: The `#if`s make the code hard to read. The indentation problem makes it worse. I propose to keep the UseGPU field outside the `#if` scope so that it is available on all platforms and would throw an exception if not available. The main code would be the same. Also from the users perspective this may be better, the exception can inform the developer. I was already considering moving UseGPU to the shared project so that it is available everywhere (but not implemented everywhere).  

Perhaps there are also other ways to rewrite this to reduce the invasiveness of the `#if`s. The way things is split up in methods may need to be altered. The main code should be easy to read, the different implementations should be some differences in the details.

pauldendulk (changes requested):
Great work. I have one remark wrt the #ifs

Mapsui.UI.Uno.WinUI/SKSwapChainPanel.cs
```diff
@@ -0,0 +1,24 @@
+#if !HAS_UNO_WINUI
```
pauldendulk: It's a lot better like this!

pauldendulk (approved):

PR #2631, opened 2024-05-13
Rewrite Sprites

pauldendulk:
This rewrites the sprites logic but depends on the BitmapRegistry changes so it merges to that branch. I've put this in a different merge request to explain a bit in the review comments.


Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
```diff
@@ -288,4 +279,7 @@ public static double FeatureSize(SymbolStyle symbolStyle, IRenderService renderS
 
         return size;
     }
+
+    public static string ToSpriteKey(string bitmapPath, Sprite sprite)
+        => $"{bitmapPath}?sprite=true,x={sprite.X},x={sprite.Y},width={sprite.Width},height={sprite.Height},pixelRatio{sprite.PixelRatio}";
```
charlenni: Here you use ˋx=ˋ twice

Mapsui/Styles/Sprite.cs
```diff
@@ -1,55 +1,11 @@
-﻿using System;
-using Mapsui.Logging;
+﻿namespace Mapsui.Styles;
 
-namespace Mapsui.Styles;
-
-public class Sprite
+public class Sprite(int x, int y, int width, int height, float pixelRatio)
```
charlenni: Could be a record

charlenni:
For me now the name 'Sprite' is strange. It is now only the area of the bitmap that is drawn to screen. So something as 'BitmapArea' or 'BitmapPart' (I don't like both of them) should be used.

Mapsui.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
pauldendulk: oh! that could have been a hard to find bug, thanks!

pauldendulk:
> For me now the name 'Sprite' is strange. It is now only the area of the bitmap that is drawn to screen. So something as 'BitmapArea' or 'BitmapPart' (I don't like both of them) should be used.

How about 'BitmapSection'?

charlenni:
Yes, sounds a bit better. Also, BitmapBounds, BitmapSource or BitmapSubset would work. I would prefer the last.

This should be used for all Bitmaps. Normally this subset should be by default (0, 0, width, height) of the bitmap (perhaps (0, 0, -1, -1) for whole bitmap?). For 'Sprites' this value isn't the default, but one with the right coordinates. 

With this, a SpriteCache isn't needed anymore. You only have a cache for the Bitmap and later use the subset coordinates to draw the bitmap. The most drawing engines have a special function for this.

If you remove the most parts of the Sprite, you could also use a MRect for the subset.

Mapsui/Styles/Sprite.cs
pauldendulk: I will make it a record.

Some thoughts about records:
- In this case the primary constructor makes the class already immutable, which is for me the most important reason to use records. By making it a records we have a better ToString and Equals.
- We have a lot of Equals implementations in our classes. In those cases we also implement GetHashCode. And if we do that we get warnings that it uses mutable fields, which is a real problem (I have struggled with very hard to find bugs when using such a flawed implementation for keys in a dictionary). I was just looking at the Brush and I wonder if it should be a record. I don't know how big the impact would be. We have a lot more classes like that.
- I also like structs for smaller classes. It would be especially useful for things in the render loop that get created on every iteration to reduce the load on the garbage collector. 
- If we use a record struct we need to add the readonly keyword to make it immutable. We have some record structs that need to keyword.

charlenni: What you use is not important. It should be always immutable after setting the bitmap source.

PR #2635, opened 2024-05-14
Fixed: MultiPolygons in WFS Services only loaded the First Polygon (4.1)

inforithmics:
Fixed: MultiPolygons in WFS Services only loaded the First Polygon (4.1)


pauldendulk (approved):
Great

PR #2639, opened 2024-05-18
Update Uno:  to Version 5.2.121 (for fixing build)

inforithmics:


pauldendulk (approved):
Great thanks. 

On my closed source projects where I use gitlab and build in dockers my builds never break through outside changes (well hardly ever). In the dotnet frontend world this seems to be an accepted thing. I think this is bad practice. Is there anything that we can do in our build configurations that reduces this problem?

Issue #2638, opened 2024-05-15
[UNO] MapInfo tapable area is off from symbol


mcocaroUX:
**Mapsui Version**
Mapsui.Uno.WinUI - 4.1.6

**Mapsui Platform**
UNO Platform

**Device**
This bug is only ocurring in iPhone Mini 12/13

**Describe the bug**
It seems that the tappable area that triggers MapInfo is slightly offset downwards. If you tap on the top half of the symbol, it doesn't trigger, but it does when tapping the bottom area or a few pixels below it.

**To Reproduce**
1. Tap the top half of the Red Circle
2. Notice how MapInfo doesn't trigger

**Expected behavior**
MapInfo should trigger when tapped on any part of the symbol.

inforithmics:
Is the red circle in the samples or can you provide a sample project which displays the behavior.

charlenni:
Please check, if you tap on a point, that is drawn by the symbol. If the symbol is transparent at that position, then Mapsui doesn't raise an event.

mcocaroUX:
@inforithmics Sorry about that, I missed to add the repro project. [MapsuiTapIssue.zip](https://github.com/Mapsui/Mapsui/files/15334664/MapsuiTapIssue.zip)

@charlenni The symbol isn't transparent.

pauldendulk:
You say 'only on iPhone Mini`. Did you only test it there or could you confirm it does not occur on other iPhones?

I have no time to dive into this issue, but [here](https://github.com/unoplatform/uno/issues/11944) is a (fixed) bug with some similarities. Don't think it is the same one, but it might be good to get some idea about what could be at play. Also a top bar  might be an issue, on iOS or in your app code. 

btw, do you see the same issue if you run our samples on the iPhone mini? (Mapsui.Samples.Uno.WinUI "Demo" "Map Info")

And can do you see the same problem if you run it with Mapsui Blazor in the browser:
- V4: https://mapsui.com/samples/
- V5: https://mapsui.com/v5/samples/

inforithmics:
@mcocaroUX I tried an iPhone 12 it worked i tried simulator iphone se and ipad mini and it worked too.

So my questsions: 
1. What iOS Version is on this iPhone mini. Is it updated to the latest iOS Version?
2. Have you used .NET 8.0.300 with newest iOS Workload dotnet `workload install maui ios` Because this/last week .NET 8.0.300 came out. (Visual Studio isn't updated yet to that Version).

[...]

PR #2640, opened 2024-05-19
Fix: Click Position and Hit Detection on iPhone Mini

inforithmics:
Things done:
1) Added a sample that shows where the mouse is clicked.
2) Added Fix for iOS Mini
3) Disabled GPU Rendering for iOS because in iPhone Mini it is totally broken.
4) Fixes this https://github.com/Mapsui/Mapsui/issues/2638


Samples/Avalonia/Mapsui.Samples.Avalonia.iOS/Info.plist
```diff
@@ -13,7 +13,7 @@
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>MinimumOSVersion</key>
-	<string>10.0</string>
+	<string>13.0</string>
```
inforithmics, line 16: Discovered that the min ios Version has changed to version 13

Tests/Mapsui.Tests.Common/Maps/TouchPointSample.cs
```diff
@@ -0,0 +1,112 @@
...
+using Mapsui.Widgets.BoxWidgets;
+using Color = Mapsui.Styles.Color;
+using HorizontalAlignment = Mapsui.Widgets.HorizontalAlignment;
+using VerticalAlignment = Mapsui.Widgets.VerticalAlignment;
+
+namespace Mapsui.Tests.Common.Maps;
+
+public class TouchPointSample : ISample
+{
+    private static Map _map;
```
pauldendulk, line 19: Why does the _map need to be static?

Tests/Mapsui.Tests.Common/Maps/TouchPointSample.cs
```diff
@@ -0,0 +1,112 @@
...
+    private TextBoxWidget _label;
```
pauldendulk, line 20: Not sure if you want this to be merged because it is mainly to show a bug. But if this is merged we should get rid of nullable and reassignment warnings.

pauldendulk (changes requested):
Not sure if you want to merge this, but if so try to fix the warnings.

Tests/Mapsui.Tests.Common/Maps/TouchPointSample.cs
inforithmics: fixed

Tests/Mapsui.Tests.Common/Maps/TouchPointSample.cs
inforithmics: fixed

pauldendulk (approved):

PR #2644, opened 2024-05-22
[4.1] Fix: Uno iPhone Mini Hit Detection

inforithmics:
Fixes this https://github.com/Mapsui/Mapsui/issues/2638


pauldendulk (approved):
Great you could fix this even though it probably originates in a bug in uno.

PR #2645, opened 2024-05-23
Separate tile image logic from symbol image logic 

pauldendulk:
Removed the object overload to replace it with type specific overloads. This made it already clear that some types were not used anymore so this was code that code could be removed.


Mapsui.Rendering.Skia/BitmapHelper.cs
```diff
@@ -1,82 +1,28 @@
...
 using System.IO;
 using Mapsui.Extensions;
 using Mapsui.Rendering.Skia.Extensions;
-using Mapsui.Styles;
 using Mapsui.Utilities;
 using SkiaSharp;
 
 namespace Mapsui.Rendering.Skia;
 
 public static class BitmapHelper
```
pauldendulk, line 9: In a first step I made separate overloads of the BitmapHelper methods so it was easy to see where each one was used. I could remove a few thanks for earlier refactoring. I moved the byte[] overloads out of here because the tile logic is quite different from the symbol logic. For instance it needs no Width/Height (because that is in the tile info).

Mapsui.Rendering.Skia/BitmapInfo.cs
```diff
@@ -15,17 +15,6 @@ public enum BitmapType
...
 {
     private object? _data;
-    private readonly bool _ownsBitmap;
-
-    public BitmapInfo()
-    {
-        _ownsBitmap = true;
-    }
-
-    public BitmapInfo(bool ownedBitmap)
```
pauldendulk, line 25: ownedBitmap was always true because of earlier refactoring and could be removed.

Mapsui.Rendering.Skia/Cache/TileCache.cs
```diff
@@ -1,8 +1,10 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using Mapsui.Extensions;
```
pauldendulk, line 5: Quite a few changes in the TileCache
- It now uses a TileCacheEntry instead of BitmapInfo. The entry points to the skia object to draw and keeps track of the iteration, which is specifically for the TileCache. The TileCacheEntry is a lot simpler than the BitmapInfo.
- Is uses an MRaster as key instead of an object (there was no reason for using an object).
- The Value is not null. This is rarely the case in a cache. Instead of making it null the entry is removed. So, we do not need to check for null.
- We don't check for disposable. The TileCache is responsible for disposing and only disposed after the entry is removed from the cache. If an object in the cache is disposed this means we have a bug in our code.
- The TileCacheEntry SKObject field is not nullable, so we do not need to check for null.

Mapsui.Rendering.Skia/Cache/TileCache.cs
```diff
@@ -46,6 +44,18 @@ public void UpdateCache(long iteration)
         }
     }
 
+    public static TileCacheEntry ToTileCacheEntry(byte[] data)
+    {
+        if (data.IsSkp())
```
pauldendulk, line 49: The skp is a serialized SKPicture, you mentioned this last week @inforithmics, I was not aware we used this serialization in our code. So, in theory our tile source on disk could also be skp this way. This is kind of cool but it makes it dependent on skia, which is not a good thing. But it does not complicate things a lot. Lets keep it like this.

Mapsui.Rendering.Skia/Cache/TileCache.cs
```diff
@@ -54,34 +64,33 @@ private void RemovedUnusedBitmapsFromCache()
         tilesToKeep = Math.Max(tilesToKeep, _minimumTilesToKeep);
         var tilesToRemove = _tileCache.Keys.Count - tilesToKeep;
 
-        if (tilesToRemove > 0) RemoveOldBitmaps(_tileCache, tilesToRemove);
+        if (tilesToRemove > 0)
+            RemoveOldBitmaps(_tileCache, tilesToRemove);
```
pauldendulk, line 68: I prefer if-then on two lines these day. I created a lot of one liners myself in past years.

Mapsui.Rendering.Skia/Cache/TileCacheEntry.cs
```diff
@@ -0,0 +1,9 @@
+﻿using SkiaSharp;
+
+namespace Mapsui.Rendering.Skia.Tiling;
+
+public class TileCacheEntry(SKObject skObject)
```
pauldendulk, line 5: This TileCacheEntry is the alternative for the BitmapInfo (in the context of tiling). It is called TileCacheEntry because the only reason it exists is because of the cache which needs to keep track of the iteration used.

Mapsui.Rendering.Skia/SkiaStyles/RasterStyleRenderer.cs
```diff
@@ -47,30 +50,22 @@ public bool Draw(SKCanvas canvas, Viewport viewport, ILayer layer, IFeature feat
...
-                switch (bitmapInfo.Type)
-                {
-                    case BitmapType.Bitmap:
-                        BitmapRenderer.Draw(canvas, bitmapInfo.Bitmap!, destination, opacity);
-                        break;
-                    case BitmapType.Picture:
-                        PictureRenderer.Draw(canvas, bitmapInfo.Picture!, destination, opacity);
-                        break;
-                }
+                if (tile.SKObject is SKImage skImage)
```
pauldendulk, line 53: The renderer now simply checks the type of SKObject, no need for any other metadata, enums or wrappers.

Mapsui.Rendering.Skia/SvgHelper.cs
```diff
@@ -38,13 +38,8 @@ public static class SvgHelper
     /// <param name="str">string</param>
     /// <returns>loaded svg image</returns>
     [return: NotNullIfNotNull(nameof(str))]
-    public static SKSvg? LoadSvg(this Stream? str)
+    public static SKSvg LoadSvg(this Stream str)
```
pauldendulk, line 41: if you check for null early you do not need to check for null later.

Mapsui/Rendering/ITileCache.cs
```diff
@@ -4,6 +4,5 @@ namespace Mapsui.Rendering;
 
 public interface ITileCache : IDisposable
 {
-    IBitmapInfo? GetOrCreate(MRaster raster, long currentIteration);
     void UpdateCache(long iteration);
```
pauldendulk, line 7: Had to remove this method because the implementation is now skia dependent. I would prefer to also remove the ITileCache interface from Mapsui core. Everything that belongs to the rendering (turning data into skia objects) should be encapsulated within the renderer. So, this applies to some other cache interfaces as well.

PR #2648, opened 2024-05-24
Replace BitmapInfo with IDrawableImage

pauldendulk:
- Not sure about IDrawableImage as name.
- The *if svg or bitmap* logic should be replaced with something better. Perhaps there should be logic to automatically select an ImageRenderer.


charlenni:
If you want to circumvent the if's and the interface, then you could use a SKDrawable. That is a Skia object, that knows how to draw itself on a canvas. It has 3 main functions that must be implemented: OnGetBounds(), OnDraw() and OnSnapshot(). OnSnapshot() is normally a call of OnDraw(). So this object provides all functions you implement in IDrawableImage or need to implement.

If you want to draw a SKDrawable, you could use SKCanvas.DrawDrawable(), which calls behind the scenes the Draw function of the object.

PR #2650, opened 2024-05-25
Let ImageFetcher return byte arrays

pauldendulk:
This way it is clear they do not need to be disposed. In some cases we need to do extra copies but in other we can avoid copies.

The ImageSourceCache does not have to be disposable anymore.

btw, We get a warning for not disposing a MemoryStream, but it does not need to be disposed. 
https://learn.microsoft.com/en-us/dotnet/api/system.io.memorystream?view=net-8.0
image


Mapsui.Rendering.Skia/Cache/RenderService.cs
```diff
@@ -11,23 +11,19 @@ public RenderService(int vectorCacheCapacity = 10000)
         LabelCache = new LabelCache();
         ImageSourceCache = ImageSourceCache.Instance;
         VectorCache = new VectorCache(this, vectorCacheCapacity);
-        SpriteCache = new SpriteCache();
```
pauldendulk, line 14: SpriteCache will be replaced with SymbolCache. SymbolCache will be renamed to DrawableImageCache.

Mapsui.Rendering.Skia/Cache/RenderService.cs
```diff
@@ -11,23 +11,19 @@ public RenderService(int vectorCacheCapacity = 10000)
...
     public ImageSourceCache ImageSourceCache { get; }
-    public SpriteCache SpriteCache { get; }
 
     public void Dispose()
     {
         LabelCache.Dispose();
         SymbolCache.Dispose();
         VectorCache.Dispose();
         TileCache.Dispose();
-        ImageSourceCache.Dispose();
```
pauldendulk, line 30: ImageSoruceCache needs no dispose because it contains only byte[] now.

Mapsui.Rendering.Skia/Cache/SymbolCache.cs
```diff
@@ -1,51 +1,30 @@
...
-        {
-            return null;
-        }
-        var loadBitmap = ImageHelper.LoadBitmap(imageStream) ?? throw new ArgumentNullException(nameof(key));
-        return _cache[key] = loadBitmap;
-    }
+        if (_cache.TryGetValue(key, out var value))
+            return value;
 
-    public Size? GetSize(string key)
```
pauldendulk, line 33: Removed the GetSize on the SymbolCache. It was weird. The cache is for caching. To get the size you can request the item from the cache.

Mapsui.Rendering.Skia/Cache/SymbolCache.cs
```diff
@@ -1,51 +1,30 @@
 ﻿using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using Mapsui.Rendering.Skia.Images;
-using Mapsui.Styles;
 
 namespace Mapsui.Rendering.Skia.Cache;
 
 public sealed class SymbolCache : IDisposable
```
pauldendulk, line 7: In the SymbolCache less checks are done. If there is no reason to suspect a null value or a disposed object then don't check for it. If there is such a thing there is bug and we will see the crash, but we should not check in all lines of every method for things that should never happen.

Mapsui.Rendering.Skia/Images/IDrawableImage.cs
```diff
@@ -4,7 +4,6 @@ namespace Mapsui.Rendering.Skia.Images;
 
 public interface IDrawableImage : IDisposable
 {
-    bool IsDisposed();
```
pauldendulk, line 7: There is no reason to expect a disposed object. The outside would should not have to worry about this. Our design should make it impossible, which makes all other code a lot simpler.

Mapsui.Rendering.Skia/SkiaStyles/CalloutStyleRenderer.cs
```diff
@@ -150,24 +148,25 @@ private static void DrawOutline(CalloutStyle callout, SKCanvas canvas, SKPath pa
...
     /// </summary>
-    public static SKPicture RenderContent(CalloutStyle callout, SymbolCache symbolCache)
+    public static SvgImage RenderContent(CalloutStyle callout, SymbolCache symbolCache)
     {
         if (callout.Type == CalloutType.Image && callout.ImageSource is not null)
         {
             using var recorder = new SKPictureRecorder();
-            var image = symbolCache.GetOrCreate(callout.ImageSource.ToString());
+            var image = symbolCache.GetOrCreate(callout.ImageSource,
+                () => SymbolStyleRenderer.TryCreateDrawableImage(callout.ImageSource));
```
pauldendulk, line 157: Because regions and customsvgs are now also in the symbolcache we can not depend on the ImageFetcher on the inside, so we need to pass in there create methods.

pauldendulk:
We are now going into the phase where things start to fall into place.

PR #2652, opened 2024-05-25
Samples: Add Wfs Ogc Filter Sample

inforithmics:
Adds an Ogc Filter Sample
Is based on this pull request. So review only after Added Ogc Filter Sample

https://github.com/Mapsui/Mapsui/pull/2633


Mapsui/Layers/BaseLayer.cs
```diff
@@ -38,7 +38,7 @@ protected BaseLayer()
         Id = NextId();
     }
 
-    internal static int NextId()
+    public static int NextId()
```
pauldendulk, line 41: The BaseLayer.NextId() is used in a providers. That could be confusing, you would expect it to be used in a layer. So, I was thinking about a provider id. But that would force provider builders, also custom provider to provide such an id. Then I wondered why did we use the layer id in the first place. We probably did discuss it but I can't remember. Was it perhaps because some provider had it's own ids? If so, perhaps we would should leave those as there are, and add our own ids, which I think are mainly for caching. What could be a problem if we do it like that?

pauldendulk (commented):
I added one question.

Mapsui/Layers/BaseLayer.cs
inforithmics: Good point. I looked at the Code and actually only the Id in the Providers is used for creating a new FeatureId, so that Every Feature gets an unique Id. Because in the Mapsui Architecture only Features are Created in the Providers, it is only used there to create a new FeatureId.
`Mapsui.Features.FeatureId.CreateId`

Because every Feature gets it's own unique Id only Providers that manipulate Featuers and therefore need a consistent way to change the new FeatureId need the LayerId (or better ProviderId).

So in conclusion it could be removed from the BaseLayer and renamed to a ProviderId, and when a BaseProvider implementation is done with a ProviderId an implementor wouldn't have to change much.

Explanation how it works, in the ShapeFile Provider.

In ShapeFiles every feature has an Id which is an int, this Id is only unique in a Shapefile, so different Shapefiles could have Features that have the same Id. To avoid that two Feature has the same Id the LayerId was introduced. 

```
// Encode the layer featureId and the feature featureId into a single long by using the negative values of long.
var result = -1 * uint.MaxValue * layerId - featureId;
return result;
```

So now because every Provider has a different Id two Features with the same Feature Id get a different Id.

pauldendulk (approved):

PR #2653, opened 2024-05-25
Rename SymbolCache to DrawableImageCache

pauldendulk:


pauldendulk:
Still not happy about the *DrawableImage* name but it has some meaning. The *drawable* means it is ready to draw to the skia canvas so it contains skia objects. The ImageSourceCache contains no skia objects. The other 3 (we currently have 5 in the renderservice) are skia specific. So should they be called DrawableTileCache, DrawableLabelCache and DrawableVectorCache? Or should we use Skia as prefix? I don't like SK as prefix because I like to be sure the SK classes are of skia itself.

Perhaps the *Skia* prefix is not bad. They are all caches that belon in the Mapsui.Rendering.Skia namespace.

PR #2665, opened 2024-06-08
Update: Eto.Forms to 2.8.0

inforithmics:
Updates Eto.Forms to 2.8.0 Because with this Version Eto.Forms has a .NET 6 Target and reduces dependencies.


pauldendulk (approved):
Wonderful

PR #2670, opened 2024-06-14
Add countries to GeoJsonSample

pauldendulk:


pauldendulk:
@inforithmics This may be relevant to you because you created the geojson provider. I worked on this because this was mentioned in a discussion and I had not worked with geojson before and wanted to do something hands-on. GeoJson is an important feature and will become more important, it is should be easy to work with. Mostly it worked just fine, so that is great. 

Some remarks:
- I also worked with a bigger countries.json (23 mb). Also wrapped in a RasterizingTileLayer, but it was more sluggish. I did not expect this because once the tiles are rasterized there should be no difference. This might be something silly, like it is recalculating the full extent on every change. We should not fix this now because that is work for v6.
- The string argument can be a file path or a geojson content string. I think it is better to be explicit about the choice for either filePath and geojson string.
- The sample uses deployment of the geojson to file. Loading the content string directly from embedded content might be slightly simpler. There is some benefit in having both, as test and to show how it works for users. Maybe we should do both, one layer using the file path and the other loading it as content string.

inforithmics:
I had a look at the sample, and on my computer it wasn't laggy, but I saw that a lot of Image Draws are done (Tiles) which took some time. For improving this - a mechanism would be needed to discover which tiles/part of the map have changed and need to draw. Avoiding redrawing the same tile/path again. Don't know if it is worth the complexity.

pauldendulk:
The current sample is not sluggish but it was a bit when I tested with this geojson https://github.com/datasets/geo-countries/blob/master/data/countries.geojson (replace "name" with "ADMIN"), even after the tiles were rasterized. The drawing of rasterized tiles will be as fast as on any other tile layer. But apparently there is still something going on in the background which consumes cpu, so there is still something we are doing with the un-rasterized features. Anyway, we should test with such data when optimizing performance for v6.

> Don't know if it is worth the complexity.

There should be less complexity, and there can be less, even if we make things faster. The current tile rasterizer is already more complex than it should. Mostly because logic which should be isolated in the MapRenderer needs to get exposed because we rasterize in the layer. Instead it should be a rendering option within the MapRenderer. The render strategy should be a property on a layer, and one of the options should be an adaptive strategy, which switches based on how long it takes to render a single iteration.

PR #2671, opened 2024-06-14
Fix: Map view Image Buttons

inforithmics:
Fixes the Images for Image Buttons.

Things done:
1. Set the ImageSource for the ImageButtons of MapView
2. Fixed ImageSourceCacheInitializer to Fetch Images of Widgets because they can have ImageSources too.


pauldendulk (approved):
Seeing the changes in the ImageSourceCacheInitializer I wondered how this worked for the ImageButtonWidget, so I checked it but it just doesn't work. Thanks for fixing this!

After my previous MRs I considered removing the MapView entirely. If we keep it I am not sure what the status should be. Even if it is functionally equivalent to v4 I would like to give a 'not supported' warning. If we keep it in v5.0.0 it should only be to allow an easy migration path for people using the MapView in v4.

Issue #2669, opened 2024-06-14
Mapsui Maui not working in ios 17.5


iliasrogaris:
**Mapsui Version**
4.1.6

**Mapsui Platform**
MAUI

**Device**
 iPhone simulator. ios 17.5

**Describe the bug**
in all iphone simulators on the ios 17.5  the application crashes when opening the map. Until 17.2 all the simulators work fine

inforithmics:
As far as I know iOS 17.5 support was only added recently https://github.com/dotnet/maui/issues/21057
And there was a bug with the iOS Simulator and localization. This means you have to use the latest .net Sdk and Maui workloads. `dotnet workload update`

iliasrogaris:
> Could it be a SkiSharp problem? Could you try to make a sample project with SkiaSharp but without Mapsui if this crashes too?

The app crashes when opening the map. I dont know of it SkiaSharp problem but I have to use map because the problem occurs when opening the map, not when the app starts

pauldendulk:
> The app crashes when opening the map. I dont know of it SkiaSharp problem but I have to use map because the problem occurs when opening the map, not when the app starts

Mapsui depends on SkiaSharp. This could still be a SkiaSharp problem, or SkiaSharp in combination with MAUI. Maybe we need an update of SkiaSharp. 

In the meantime you could try what Thomas is suggesting: 

> This means you have to use the latest .net Sdk 8.0.6 and Maui workloads. dotnet workload update

inforithmics:
Had a look at it, a Hello World Maui App works with 17.5 But not the Mapsui Samples project, I crashes because of a Satellite Assembly that is not found for MAUI, I think there are still Bugs in the 17.5 iOS Support in Maui.

inforithmics:
I debugged it further, on ios 17.5 the 3d accelerated rendering stopped working, so this needs an SkiaSharp update or fix or as a workround set UseGPU to false before the first control is rendered.
`MapControl.UseGPU = false;`

iliasrogaris:
I tested with MapControl.UseGPU = false and it work fine in ios 17.5
Thank you!

Issue #1896, opened 2023-03-20
MAUI with multiple pages only work first time


TFreudi1:
I have created a shell based MAUI project with to pages. The Map pages shows the map only the first time. If you go back to the other page and than choose the map-page again MapSui dont show a map. 

**To Reproduce**
Take my sample project
Choose Map Page from Hamburger menu
Chosse Home Page from Hamburger menu
Choose Map Page from Hamburger menu

**Expected behavior**
Map should be visible

This happens on all platforms
[MauiAppMapSui.zip](https://github.com/Mapsui/Mapsui/files/11019916/MauiAppMapSui.zip)

pauldendulk:
Thanks for the sample! I understand this is an important issue but I am now focused on beta 9 which will contain changes that may affect this issue.

charlenni:
Did you try to call ````MapControl.StopUpdates()```` before leaving the pages with the MapControl and call ````MapControl.StartUpdates()```` when coming back? Does this solve the problem?

TFreudi1:
Yes, the sample project is very simple, see link below. On the map-page the code is now:
`public partial class MapPage : ContentPage
{
    MapControl mapControl = new MapControl();
    public MapPage()
    {
		InitializeComponent();
        mapControl.Map?.Layers.Add(Mapsui.Tiling.OpenStreetMap.CreateTileLayer());
        Content = mapControl;
    }

    protected override void OnAppearing()
    {
        base.OnAppearing();
        mapControl.StartUpdates();
    }

    protected override void OnDisappearing()
    {
        mapControl.StopUpdates();
        base.OnDisappearing();
    }
}`

First time page open map correct, if you go back to home-page and than to map-page, map-page is white and shows nothing.
![error](https://user-images.githubusercontent.com/4739177/226624654-d16866d9-cb99-4ecb-9f66-f29e7b2ec756.png)

janseris:
I reported this as well and my issue somehow disappeared.
Basically navigating out of the page and visiting again only loads map background and nothing else.

inforithmics:
Seems to be some Bug with SKGlView

Workaround:
```
public static MauiApp CreateMauiApp()
{		
       Mapsui.UI.Maui.MapControl.UseGPU = false;
```

inforithmics:
An alternative approach would be to always create the Mapsui Control on Appearing.

```
  protected override void OnAppearing()
    {
        base.OnAppearing();

        MapControl mapControl = new MapControl();
[...]

PR #2673, opened 2024-06-16
[4.1] Fix Maui Rendering on iOs 17.5 (Disable GPU Rendering)

inforithmics:
Fixes this https://github.com/Mapsui/Mapsui/issues/2669

### DIFF
--- a/Samples/Mapsui.Samples.Maui.MapView/AndroidManifest.xml
+++ b/Samples/Mapsui.Samples.Maui.MapView/AndroidManifest.xml
@@ -1,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <uses-sdk android:minSdkVersion="23" android:targetSdkVersion="33" />
+  <application android:allowBackup="true" android:icon="@mipmap/appicon" android:roundIcon="@mipmap/appicon_round" android:supportsRtl="true"></application>
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+  <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+  <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+</manifest>
--- a/Mapsui.UI.Forms/MapControl.cs
+++ b/Mapsui.UI.Forms/MapControl.cs
@@ -73,7 +73,8 @@ public partial class MapControl : ContentView, IMapControl, IDisposable
         DeviceInfo.Platform != DevicePlatform.WinUI &&
         DeviceInfo.Platform != DevicePlatform.macOS &&
         DeviceInfo.Platform != DevicePlatform.MacCatalyst &&
-        DeviceInfo.Platform != DevicePlatform.Android;
+        DeviceInfo.Platform != DevicePlatform.Android &&
+        DeviceInfo.Platform != DevicePlatform.iOS; // 3d rendering does not work on ios 17.5 with 2.88.8 skiasharp.
 #else
     public static bool UseGPU = true;
 #endif
--- a/Mapsui.UI.Maui/Mapsui.UI.Maui.csproj
+++ b/Mapsui.UI.Maui/Mapsui.UI.Maui.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
 	<PropertyGroup>
-		<TargetFrameworks>net7.0</TargetFrameworks>
+		<TargetFrameworks>net7.0;net8.0</TargetFrameworks>
 		<!--Windows does not work because of this https://github.com/dotnet/maui/issues/6529-->
 		<TargetFrameworks Condition="$([MSBuild]::IsOSPlatform('windows'))">$(TargetFrameworks)</TargetFrameworks>
 		<UseMaui>true</UseMaui>
@@ -46,6 +46,11 @@
 		<PackageReference Include="SkiaSharp.Views.Maui.Controls" />
 		<PackageReference Include="SkiaSharp.Views.Maui.Controls.Compatibility" />
 	</ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)'=='net8.0'">
+    <PackageReference Include="Microsoft.Maui.Controls" VersionOverride="[$(MauiVersion),9.0.0)" />
+    <PackageReference Include="Microsoft.Maui.Controls.Compatibility" VersionOverride="[$(MauiVersion),9.0.0)" />
+  </ItemGroup>
 
 	<ItemGroup>
 		<ProjectReference Include="..\Mapsui.Nts\Mapsui.Nts.csproj" />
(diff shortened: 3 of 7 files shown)


Samples/Mapsui.Samples.Maui.MapView/AndroidManifest.xml
```diff
@@ -0,0 +1,8 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <uses-sdk android:minSdkVersion="23" android:targetSdkVersion="33" />
+  <application android:allowBackup="true" android:icon="@mipmap/appicon" android:roundIcon="@mipmap/appicon_round" android:supportsRtl="true"></application>
+  <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+  <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+  <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+</manifest>
```
pauldendulk, line 8: Did network not function properly before? Why was this added now?

pauldendulk (approved):

Samples/Mapsui.Samples.Maui.MapView/AndroidManifest.xml
inforithmics: This is a workaround because Maui looks in the wrong path for AndroidManifest.xml. So without this fix it doesn’t compile. Maybe I can copy it in a precompile event

pauldendulk: Perhaps you can mention in the xml that this is a workaround. If you have read about this in an issue you could point to the issue. In general, code comments should can often better be left out if they don't add anything that can be read from the code. It is this kind of unexpected thing that needs to be commented.

taublast:
Was surprised, now GPU support is just totally cut up for MAUI in this library, while Android had its MAUI renderer/handler long enough for SKGLView and SkiaSharp 2.88.8 shipped it's own Apple Metal handler.

What is currently possible:
image

janusw:
> Was surprised, now GPU support is just totally cut up for MAUI in this library

Same here. And unfortunately this can have a rather big impact on performance. Complex maps with multiple layers get pretty slow without GPU acceleration.

In any case, Mapsui 4.1.6 seemed to work well on physical iOS 17.5 devices for me, and #2669 seems to indicate that the crashes only occur in the simulator.

@pauldendulk @inforithmics  Maybe it would be better to disable `UseGPU` only when running in a simulator, but not on real devices?

janusw:
Btw, I also tried to enable `UseGPU` on Android (with a .NET MAUI 8 app), and this seems to work well initially, but after I navigate away from the map and then back, the map is all white (no content drawn any more). Would be great if this could be fixed as well (Android also profits greatly from GPU acceleration).

Is there an upstream SkiaSharp bug concerning GPU on Android, or is this a Mapsui-specific problem?

janusw:
> Is there an upstream SkiaSharp bug concerning GPU on Android, or is this a Mapsui-specific problem?

I found #1896 (fixed via #1934), but it does not seem to answer this question.

charlenni:
Isn't it possible to set UseGPU before the first use in any case you are sure, that the control doesn't crash? Inforithmics made a default setting, which works for each platform, so that new users have a satisfying experience. If you want to change something and you are sure, that this platform is capable to use GPU, then set MapControl.UseGPU to true before you instantiate the MapControl the first time.

taublast:
> Btw, I also tried to enable `UseGPU` on Android (with a .NET MAUI 8 app), and this seems to work well initially, but after I navigate away from the map and then back, the map is all white (no content drawn any more). Would be great if this could be fixed as well (Android also profits greatly from GPU acceleration).
> 
> Is there an upstream SkiaSharp bug concerning GPU on Android, or is this a Mapsui-specific problem?

SkiaSharp is not to blame here, quick guess is either GRContext changed, either the view Handler was somehow disposed when you navigated away.. Are you using Maui Shell?

janusw:
> Isn't it possible to set UseGPU before the first use in any case you are sure, that the control doesn't crash? Inforithmics made a default setting, which works for each platform, so that new users have a satisfying experience.

Yes, of course one can always override this default, and I do that for now to work around it.

But my point is that the default should be reasonable. So if there is no indication that GPU rendering fails on physical iOS devices (but only on the simulator) then it should be enabled for that case by default. This is also in line with the code snippet posted by @taublast above (which is from SkiaSharp, I assume?).

janusw:
> > Is there an upstream SkiaSharp bug concerning GPU on Android, or is this a Mapsui-specific problem?
> 
> SkiaSharp is not to blame here, quick guess is either GRContext changed, either the view Handler was somehow disposed when you navigated away.. Are you using Maui Shell?

Yes, Maui Shell with TabBar navigation. One of the tabs has a MapControl, and I'm calling MapControl.StopUpdates/StartUpdates when navigating to/from a different tab.

janusw:
> > > Is there an upstream SkiaSharp bug concerning GPU on Android, or is this a Mapsui-specific problem?
> > 
> > 
> > SkiaSharp is not to blame here, quick guess is either GRContext changed, either the view Handler was somehow disposed when you navigated away.. Are you using Maui Shell?
> 
> Yes, Maui Shell with TabBar navigation. One of the tabs has a MapControl, and I'm calling MapControl.StopUpdates/StartUpdates when navigating to/from a different tab.

I tried removing the StopUpdates/StartUpdates call, but the problem persists.

Note that if I navigate to a different page via `Shell.Current.CurrentPage.Navigation.PushAsync(new FancyPage())`, then the map keeps working upon return. But if I navigate to a different tab and back, the map is all white.

janusw:
> > Is there an upstream SkiaSharp bug concerning GPU on Android, or is this a Mapsui-specific problem?
> 
> SkiaSharp is not to blame here, quick guess is either GRContext changed, either the view Handler was somehow disposed when you navigated away..

@taublast Any idea how I can check which of the two is the problem? What can I do to debug this?

I already tried looking at the Mapsui log, but it doesn't seem to show any errors:

```
            Mapsui.Logging.Logger.LogDelegate += (level, message, ex) =>
            {
                Debug.WriteLine($"Mapsui: {level}: {message} - {ex}");
            };
```

charlenni:
@janusw Did you test this behavior with only SkiaSharp itself? I would assume, that it is a SkiaSharp problem and not a Mapsui one.

Replace the MapControl in your app with a native SkiaSharp control and add a line draw or anything else to it. Then try you usecase.

janusw:
> @janusw Did you test this behavior with only SkiaSharp itself?

An `SKCanvasView` still renders correctly after switching tabs, but I guess one needs to try an `SKGLView` (will try to look into that soon) ...

> I would assume, that it is a SkiaSharp problem and not a Mapsui one.

That's not clear, but it would be good to know. I rather suspect the opposite ;)

janusw:
> > @janusw Did you test this behavior with only SkiaSharp itself?
> 
> An `SKCanvasView` still renders correctly after switching tabs, but I guess one needs to try an `SKGLView` (will try to look into that soon) ...

Actually a plain `SKGLView` shows the same problem on Android: Works the first time, but does not render any more after switching tabs.

Seems to be this bug: https://github.com/mono/SkiaSharp/issues/2550

charlenni:
Did you checked the hack in the thread? To set MapControl.IsVisible = true when coming back to the tab containing Mapsui's MapControl?

janusw:
> Did you checked the hack in the thread? To set MapControl.IsVisible = true when coming back to the tab containing Mapsui's MapControl?

Yeah, that workaround actually works well (both for a plain SKGLView, as well as a MapControl). But note that setting `IsVisible = true` is not enough. You have to toggle it (set it to false and then true again).

PR #2679, opened 2024-06-20
Tests: Add widgets to FetchImagesInViewportAsync

inforithmics:
In this pull request I forgot to add the widgets in Tests so that in the Tests the widget images are loaded too.
https://github.com/Mapsui/Mapsui/pull/2671

### DIFF
--- a/Mapsui/Styles/ImageSourceCacheInitializer.cs
+++ b/Mapsui/Styles/ImageSourceCacheInitializer.cs
@@ -54,7 +54,7 @@ public static class ImageSourceCacheInitializer
     {
         var imageSources = GetAllImageSources(viewport, layers);
         if (widgets != null)
-            GetAllImageSources(imageSources, widgets);   
+            GetAllImageSources(imageSources, widgets);
 
         if (imageSources.Count == 0)
             return await Task.FromResult(false);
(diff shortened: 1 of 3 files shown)


Mapsui/Styles/ImageSourceCacheInitializer.cs
```diff
@@ -50,9 +50,11 @@ private static void GetAllImageSources(List<string> imageSources, IEnumerable<IW
     }
 
     public static async Task<bool> FetchImagesInViewportAsync(ImageSourceCache imageSourceCache,
-        Viewport viewport, IEnumerable<ILayer> layers)
+        Viewport viewport, IEnumerable<ILayer> layers, IEnumerable<IWidget>? widgets = null)
```
pauldendulk, line 53: There is a risk in the nullable arguments as you can forget the argument when it is needed. Perhaps it could be organized in such a way that you can not forget about it. I considered passing in the Map (which has both Layers and Widgets) that is possible except for the RasterizingTileLayer. Let's leave it like it is right now. The tile rasterizer should become part of the renderer instead of a layer in v6.

pauldendulk (approved):

PR #2680, opened 2024-06-20
Build:  use partial clone for Builds (Reduces checkout duration)

inforithmics:
A partail cone with  fetch with tree:0 is around 3 seconds. For a whole checkout it can be sometimes be minutes. 
For example a linux build an checkout on linux can be 54 seconds to 6m 14 seconds

Theory behind it.
https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

### DIFF
--- a/.github/workflows/dotnet-docs.yml
+++ b/.github/workflows/dotnet-docs.yml
@@ -36,8 +36,7 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
       with:
-        fetch-depth: 0 
-        fetch-tags: true
+        fetch-depth: 0         
         filter: tree:0 
     - name: Setup .NET 8 SDK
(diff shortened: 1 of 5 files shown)


pauldendulk (approved):

pauldendulk:
You tried this before and then reverted it for some reason. Why is the situation different now?

inforithmics:
filter: tree:0 was introduced with checkout@v4 and is the newer method for git to minmally clone.
The main difference is that fetch-depth:1 has only one version while  filter: tree:0  keeps the history but all not needed history is not fetched. and this reduces the clone speed drastically while not beeing so much slower.

pauldendulk:
Wow, advanced stuff. I know very little about this, thanks for the lesson.

PR #2683, opened 2024-06-21
Remove nullable widgets parameter in FetchImagesInViewportAsync

inforithmics:
follow up of: https://github.com/Mapsui/Mapsui/pull/2679


pauldendulk (approved):
Thanks, this is better than my thought to use the Map as argument. Providing an empty list in the RasterizingTileLayer is a simple solution. btw, I❤️ the simplicity of the collection initializers.

PR #2686, opened 2024-06-21
Fix: GeoTiffProvider avoid TagExtender recursion.

inforithmics:
In this pull request it is visible that the TagExtender recursively calls it self  
https://github.com/Mapsui/Mapsui/actions/runs/9618748477/job/26533287346?pr=2603
By setting the _parentExtender to null if it is the TagExtender I avoid this recursion


inforithmics:
closing this pull request recreating it

PR #2687, opened 2024-06-21
Fix: GeoTiffProvider TagProvider recursion

inforithmics:
Fixes this: Because the Error could happen in Main to so I make a sperate pull request.
https://github.com/Mapsui/Mapsui/actions/runs/9618748477/job/26533287346?pr=2603


Mapsui.Extensions/Provider/GeoTiffProvider.cs
```diff
@@ -130,7 +130,9 @@ private static TiffProperties LoadTiff(string location)
 
         // Register the custom tag handler
         Tiff.TiffExtendProc extender = TagExtender;
-        _parentExtender = Tiff.SetTagExtender(extender);
+        var previousExtender = Tiff.SetTagExtender(extender);
```
pauldendulk, line 133: What is the reason the _parentExtender is static? Could it be rewritten to make it not static? What would be the downside? If it is just a performance penalty which is not too big I would prefer that.

inforithmics: I made it an instance now and restore the original Value on disposing of GeoTiffProvider

inforithmics:
@pauldendulk can this be merged? I merged this pull request into skkasharp 3 branch and the tagorovider recursions are gone now

pauldendulk (approved):

PR #2689, opened 2024-06-22
Cleanup: Remove .net7 Because it is End Of Life

inforithmics:
I saw that there are still some .net 7 Targets. I remove them here because .net 7 is end of life since quite some time.


pauldendulk (approved):
Great thanks, I love cleanup. This will also make the build faster, locally and on the build server.

PR #2691, opened 2024-06-22
Samples: Add Wmts Zurich Sample and reorganize WMS and WMTS Samples

inforithmics:
Things done:
1) Added Wmts Zurich Sample 
2) reorgnized Samples to Wms and Wmts Categories.
3) Added missing Documentation for Samples (For Website)
4) Added missing Regression Images
5) Reenabled OpenSea Regression Test


pauldendulk (approved):
Thanks a lot.

btw, I am considering to put all samples in a single folder without subfolders. They will still be in categories, so that is how they will show up in the sample apps. 
- The disadvantage is that you can not see which sample belongs to which category when viewing the folder. But I rarely navigate the code that way.
- The advantage is that you do not have to move the samples when reorganizing them. This is something that happens a lot. 

I have the second strategy for documentation pages. Just dump them in one folder and reorganize them in a meta file (not sure if we need a meta file, it would mean we have to register them all, and it is nice we can now discover the samples).

The folder should be called 'Samples' instead of 'Maps'.

PR #2694, opened 2024-06-26
[4.1] Fix enabling widgets after app start

pauldendulk:
The extended widgets are cached. However, only the enabled widgets are cached. If later an extended widget is enabled this will not be in the cached list and its events will never be triggered.


pauldendulk:
@inforithmics Merging this because I want to release a beta to test in an app. Comments are still welcome.

charlenni:
I couldn't see the full source code, perhaps this is already handled, but what is when a Widget is added later?

pauldendulk:
There is a check on the number of widgets. If there is a difference the cache is refreshed. Then again, if in some scenario you go to a different mode when you add and remove a widget the count will be the same but the refresh should be triggered but is not. 

In v5 the caching of widgets is removed entirely.

PR #2698, opened 2024-06-29
Add MapBuilder and sample

pauldendulk:
Had another try at the MapBuilder.

Earlier I reconsidered and suggested to just use extension methods on the Map instead of a MapBuilder. There were some problems I did not know how to solve (without having to create a lot of additional code).

One such a problem is that you want to create the Map in the Build method, but that means you have to store all the settings before you apply it in Build. In this PR this kind of thing is stored in the factory and configure methods in the MapBuilder. This is not a lot of additional code.

A problem with the implementation of the first version: It is possible to modify what the factories are creating through the configure methods. These methods get the Map, Widget or Layer as parameter. This is flexible, but too flexible, because you can do anything with those classes, while it is supposed to be for configuration only. A solution would be to add classes like MapConfiguration, LayerConfiguration, WidgetConfiguration as proxy. These would have a lot of the same fields as their original classes, and all need to be explicitly assigned, not nice. Also not sure how this will work out for derived classes. For widgets we would need the same inheritance hierarchy for configurations as for the widgets themselves, meh.


pauldendulk:
Some thoughts after my previous commit.

- Perhaps use MapConfiguration but not  LayerConfiguration and WidgetConfiguration. For widgets and layers it is not such a problem users can access them.
- An alternative to the above. Don't use a configuration method but only specific methods, like WithCRS. It looks there are not that many of such methods needed.
- By wrapping configuration methods of derived classes in calls with the parent class as argument it is possible to change properties of derived classes. BTW, the configuration methods are not that useful if you create the widget or layer in your own code (because you can create them any way you like), but they are when using methods that create specific widgets or layers for you.

pauldendulk:
I am going to merge this stage. It adds only new files and changes no existing functionality. I will continue work on the MapBuilder in other PRs.

PR #2702, opened 2024-07-06
Update to Brutile-6.0.0-beta.3

pauldendulk:


pauldendulk:
There are some breaking changes in BruTile v6. 

### The motivation for the changes:
- Remove HttpClient from BruTile. It removes a big burden from BruTile. Things BruTile does not have to do anymore: Dispose the HttpClient, Come up with ways to configure the HttpClient. Come up with ways to add authentication to the request.
- Remove the (I)TileProvider(s). It was an older layer. In some case you could pass an ITileProvider into a TileSource in other cases a TileSource was itself an ITileProvider. It was conceptially hard to explain what was one or the other. Now there are just TileSources.

### Specific changes:
- Removed ITileProvider. Use IHttpTileSource for http requests and ILocalTileSource for files or MBTiles.
- Removed the HttpTileSource constructor which took a ITileProvider as argument. Use the other constructor where you need to specify the arguments that were previously passed into the TileProvider.
- Renamed WmtsParser to WmtsCapabilitiesParser
- Removed WmscTileSource. Use WmscUrlBuilder and HttpTileSource
- Removed ArcGisTileSource. Use a BasicUrlBuilder and regular HttpTileSource

PR #2710, opened 2024-07-13
Update: Use Release Version of Microsoft.Windows.CsWin32

inforithmics:
Microsoft.Windows.CsWin32 has now a non beta version use this now (is only used in the samples)


pauldendulk (approved):
👍

PR #2716, opened 2024-07-20
Improvement: Removed SkPaint Instance in LabelStyleRenderer

inforithmics:
Removed SkPaint Instance in LabelStyleRenderer.

Reasons:
1) It is not ThreadSafe, all render processes access the same instance
2) It is not disposed.
3) Cleanup before I replace Obsolete Apis in https://github.com/Mapsui/Mapsui/pull/2603
4) It seems I fixed a Bug Because Size on Linux and Windows increased by 1 and the size on the Unit Feature Size Tests is now in the range from 40 to 42 instead of 39 to 42.
5) Fixed the Measurement of Text with a Halo (Added Unit Test)


inforithmics:
Yes I agree. I think I can reduce the number of caches to one. Because the principle is always the same add key for identification of object (tuple) add a lambda that creates the object when it isn‘t in the cache. But this is for a seperate pull request. This would later help to for the splitting of rendering where Skia objects would be created and then rendered. Because with caching the same object does not create a new new Skia object.
The problem with caching and nts that an geometry can sometimes be manipulated without being noticeable from outside except comparing a cloned nts object which is quite expensive. Maybe I have to investigate there for a better solution.
I‘ll try in another pull request to remove the global render lock maybe this is fixed now with this pullrequest.

pauldendulk (approved):
Great and thanks for dealing with me comments

PR #2723, opened 2024-07-23
Dependencies: Remove upper Bounds from SkiaSharp

inforithmics:
Remove upper Bounds from SkiaSharp to make it easy upgradable to SkiaSharp 3.
Depends on this pull requests.
https://github.com/Mapsui/Mapsui/pull/2722
https://github.com/Mapsui/Mapsui/pull/2721
https://github.com/Mapsui/Mapsui/pull/2720


pauldendulk (approved):
Doing this update of step of smaller PRs makes it easy to review and easy to understand. Great. 

With the upper bounds you know users can only work with a version that was tested. Without those there will be bugs caused by incompatibilities. We have to accept that however, because there are too many dependencies which change too fast. We may not want to remove the upper bound on all packages though.

PR #2725, opened 2024-07-24
Feature: Add svg:// and image:// to imagesource

inforithmics:
Following schemas are added in this pull request:
svg:// svg string
image:// binary image encoded as base64 (could be an svg string encoded as base64)


pauldendulk:
I worked on the 'svg' option that is part of this PR myself in another PR and then decided to revert it then. I can not fully remember why. I think there were three reasons:
- You still have to escape characters in the string to be able to use it. So, it is not super easy to copy and paste the svg string (yes, sure there are ways, but it did not work for me at first, and it won't work for others). So, perhaps it is better to use the other options.
- I think there was a problem with invalid characters in the URL. So, you also have to url encode it (not sure about this). This PR removes the Uri this, so that problem is solved, but you also lose some of the Uri validation that might be useful.
- It is another option, every option added makse choosing the right one a little harder. The story becomes less simple. Personally I also don't see a great benefit either, but perhaps I miss out on something.
- The string itself is used as a key for the dictionary. The advantage of that is that if different symbols refer to the same image it will be instantiated as a single object in memory. If you use an entire base64 image as the key of the dictionary, and use it to search for the item it will impact performance. Not sure if it is significant, but I would rather not have that problem at all. A solution would be to add another layer that gets an integer key from the input string only once at the start, but I don't want to add another cache for this, so, I need to think about a good option.

I have to think about this one.

charlenni:
- For not escaping characters in a string use @ and ' instead of "
- No problem with invalid characters in the URL, because this type of strings are not handled as string. They look only as if they are URLs.
- If the number of options is the problem, then you should remove the http option, because one could download the file by itself and store it in a local file system.
And if you want a small graphic, that you don't want to fight how to handle local assets or offline, then it is useful.
- Don't use the string as a key, but a hash code (not the normal GetHashCode()). Or add a name for this like
    svg://image_name=<svg>...</svg>

charlenni:
One more thing: how often do you think this takes place? I assume, that no one is starting to added hundreds of SVG strings or images to the code. So timing problems for accessing the dictionary should be relative rare.

inforithmics:
In c# a Rae string can be used to avoid escaping https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/raw-string-literal

charlenni:
What is the use case? Small demos or apps, that don't want to have all the overhead to add a file to the source code. The code for a pin is a one liner as svg code. But if you have to save the code to a file and do all this things to copy the file to local file system, then it is very hard.

On the other side, it is a enlargement of the code about two additional switch entries and two functions shouldn't be the problem.

pauldendulk:
My current thoughts.

The option to work with an svg string is useful. Manipulating the string in memory is easy to do which makes it a flexible option. I talked to people who change specific colors to other colors or line types from dashed or dotted.

My main problem at the moment is that the ImageSource string is used as the key to the dictionary. If it gets longer the lookup will be slower (although this may not be show initially because as long as the hash returns a unique result it is just as fast). Possible solutions:
- Don't put it in a cache. This would mean every symbol will get it's own instance, even if the strings are identical, which would be a common scenario.
- Generate an integer key from the string at the first load and only use the integer key for the lookup. This needs some rewriting. I looked into a possible way to build this and my first impression was that it would become a bit hacky.
- Introduce different image classes for different image types. This already came to mind when working on the ImageSource feature. The SvgFillColor and SvgStrokeColor are now fields on the SymbolStyle but apply only when ImageSource is set to SVG. So this is not great, there are fields that you have to ignore if you use a png or use SymbolType.Ellipse/Triangle/Rectangle. Instead we could introduce a SymbolStyle.Image field which could be assigned with an instance of a specific image class. The instance would be the key in the cache. One such class would be an SvgSource with fields like Path, FillColor and StokeColor. Another would be SvgContent.
- Do not reuse the SymbolStyle but introduce a new Style for images with an image svg path, and a new Style for images with svg content. I have yet to investigate how much code needs to be duplicated between those. Conceptually it may be simpler. Instead of having to explain how the SymbolStyle should be configured we could say, assign the SvgContentStyle to your feature and set the path. Not sure if this solves the caching problem.
- Add an SvgImageContentRegistry, similar to the BitmapRegistry and let the ImageSource points to that: "svgcontent://keyoftheregistry";

pauldendulk:
I am releasing today. For the next release I want to include this PR.

pauldendulk:
Today I looked deeper into option 3, an image class instead of image source string. 

Pros: 
- No svg specific fields (SvgFillColor and SvgStrokeColor)  in SymbolStyle.
- It becomes possible to reuse the image in the other IHasImageSource classes (Brush, CalloutStyle, ImageButtonWidget).
- Image specific fields would also be available in other places.

Cons:
- My idea was that the image instance would be the key to the dictionary. However, the user should be aware of this, and should not create multiple instances with the same key. This will fail regularly.

I decided not to go for this. Eventually we will need something like an Image class, but not to solve the cache problem.

Option 4 is similar, maybe a better alternative to 3.

Option 5 is not hard to build, but the such a SvgImageContentRegistry is an extra step making it a bit more complicated for the user, and that registry would need cleaning as well, an an option to remove the content string.

Option 1 was never a real alternative.

Next time I will look into option 2.

pauldendulk:
To resolve the conflicts I created a new branch of the commits in this branch and continued work on it in this PR: https://github.com/Mapsui/Mapsui/pull/2894

PR #2735, opened 2024-08-25
Merge LabelCache into VectorCache

inforithmics:
Cache Simplification, I don't need a Label Cache, I can use the VectorCache.
Things done,
1. There are now only GetOrCreate Functions without specifying what is created or get so the cache gets universal usable - for example in the LabelStyleRenderer.


Mapsui.Rendering.Skia/Cache/RenderService.cs
```diff
@@ -4,24 +4,21 @@ namespace Mapsui.Rendering.Skia.Cache;
 
 public sealed class RenderService : IRenderService
 {
-    public RenderService(int vectorCacheCapacity = 10000)
+    public RenderService(int vectorCacheCapacity = 30000)
```
inforithmics, line 7: Because now 3 caches with average size of 10000 are consolidated into one i use now 30000 as the new default

Mapsui.Rendering.Skia/Cache/VectorCache.cs
```diff
@@ -5,19 +5,18 @@ namespace Mapsui.Rendering.Skia.Cache;
 
 public sealed class VectorCache(IRenderService renderService, int capacity) : IDisposable
 {
-    private readonly LruCache<object, ICacheHolder> _paintCache = new(Math.Min(capacity, 1));
-    private readonly LruCache<object, ICacheHolder> _pathParamCache = new(Math.Min(capacity, 1));
+    private readonly LruCache<object, ICacheHolder> _cache = new(Math.Min(capacity, 1));
```
inforithmics, line 8: I only need one cache simplifying code

Mapsui.Rendering.Skia/Cache/VectorCache.cs
```diff
@@ -5,19 +5,18 @@ namespace Mapsui.Rendering.Skia.Cache;
...
 
     public bool Enabled { get; set; } = true;
 
-    public CacheTracker<TPaint> GetOrCreatePaint<TParam, TPaint>(TParam param, Func<TParam, TPaint> toPaint)
+    public CacheTracker<TPaint> GetOrCreate<TParam, TPaint>(TParam param, Func<TParam, TPaint> toPaint)
```
inforithmics, line 12: I now use a GetOrCreate Function which can create any object simplifying cache

Mapsui.Rendering.Skia/SkiaStyles/LabelStyleRenderer.cs
```diff
@@ -57,20 +58,20 @@ public bool Draw(SKCanvas canvas, Viewport viewport, ILayer layer, IFeature feat
             {
                 case (PointFeature pointFeature):
                     var (pointX, pointY) = viewport.WorldToScreenXY(pointFeature.Point.X, pointFeature.Point.Y);
-                    DrawLabel(canvas, (float)pointX, (float)pointY, labelStyle, text, (float)layer.Opacity, renderService.LabelCache);
+                    DrawLabel(canvas, (float)pointX, (float)pointY, labelStyle, text, (float)layer.Opacity, renderService);
```
inforithmics, line 61: i now use the renderservice (with included VectorCache for rendering)

Mapsui.Rendering.Skia/SkiaStyles/LabelStyleRenderer.cs
```diff
@@ -225,8 +232,8 @@ private void DrawLabel(SKCanvas target, float x, float y, LabelStyle style, stri
         // If style has a halo value, than draw halo text
         if (style.Halo != null)
         {
-            using var paintHalo = CreateHaloPaintHolder(style, labelCache);
-
+            using var paintHaloHolder = vectorCache.GetOrCreate(style, CreateHaloPaintHolder);
```
inforithmics, line 235: The cache works as follow.
1. It creates a Holder object 
- the reason for that is that I can make a using around it, so that disposing anlyse is happy
- If the object is already in use in another part of mapsui, a new object is created and then disposed in the holder.
- Caching could be disabled and the objects are still disposed by the holder object.

Mapsui.Rendering.Skia/Cache/RenderService.cs
pauldendulk: At some point we need to improve caching. Ideally it should not depend on some specific number of items allowed in the cache.

- All features needed to draw the current screen should be allowed in the cache. This should not be limited by some maximum.
- We should track the render iteration in which they were used.
- We may want to specify the number of iterations we want to keep the items in the cache.
- There should be different caches for different canvases. If you have a map for details and and a map to show an overview, they should have different caches. This also applies if you rasterize a specific part of the map.

Mapsui.Rendering.Skia/SkiaStyles/LabelStyleRenderer.cs
pauldendulk: > the reason for that is that I can make a using around it, so that disposing analyse is happy

I found it often hard to please the analyzer and am considering removing the custom one and rely more on what Microsoft provides.

pauldendulk (approved):

pauldendulk:
Good improvement.

PR #2740, opened 2024-09-04
[4.1] MapControl: Reset widget cache

### LINKED ISSUE #2599: Wrong attribution link after switching map layer
I am building an app where the user can change the map type, for example switching from the Bing Aerial to the OSM.
The map control has 2 layers: the map and a GPS track.

When the user wants to switch map type, I remove the old map layer and replace it with the new one.
The new layer may have a different attribution text and URL.

So this is what happens:
- open the control with Bing Aerial; the attribution is Microsoft; clicking on it opens the Microsoft website
- switch to the OSM map; the attribution is OSM; clicking on it still opens the Microsoft website

I debugged the code and I see that when the layer is replaced, both the text and the URL of the attribution are correct, so I have no idea where the issue is coming from.
Is there maybe some sort of caching that I should take into account?

janusw:
This is my attempt to fix #2599. I'm adding a method that triggers the re-caching of widgets in `MapControl`. The method is protected, so that it can be called in all classes derived from MapControl (including MapView etc). This should be enough to fix the problem (for me).

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -35,7 +35,7 @@ jobs:
       uses: actions/setup-dotnet@v3
       with:
          dotnet-version: |
-           8.0.x
+           8.0.3xx
          include-prerelease: false
     # Android SDK Install
     - name: Set up Android SDK
@@ -92,7 +92,7 @@ jobs:
       uses: actions/setup-dotnet@v3
       with:
          dotnet-version: |
-           8.0.x
+           8.0.3xx
          include-prerelease: false
     - name: Setup Xcode version 14.2
       uses: maxim-lobanov/setup-xcode@v1.5.1
@@ -158,7 +158,7 @@ jobs:
       uses: actions/setup-dotnet@v3
       with:
          dotnet-version: |
-           8.0.x
+           8.0.3xx
          include-prerelease: false
     - name: install maui macos android ios maccatalyst wasm-tools
       run: dotnet workload install maui macos android ios maccatalyst wasm-tools
@@ -235,7 +235,7 @@ jobs:
       uses: actions/setup-dotnet@v3
       with:
          dotnet-version: |
-           8.0.x
+           8.0.3xx
          include-prerelease: false
     - name: Set up JDK 11
       uses: actions/setup-java@v4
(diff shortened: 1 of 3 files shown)


janusw:
> This should be enough to fix the problem (for me).

I just verified via the GHA-generated nupkgs that this does indeed fix the problem for me. And yes, it does!

All I do is call the new method `ResetWidgetCache` after switching layers.

@giobiondani Does this work for you as well, or do you need the method to be public?

@pauldendulk Is this PR ok for the `develop/4.1` branch? Or can we find a way to make this work automatically (without the additional call to `ResetWidgetCache`)?

giobiondani:
> > This should be enough to fix the problem (for me).
> 
> I just verified via the GHA-generated nupkgs that this does indeed fix the problem for me. And yes, it does!
> 
> All I do is call the new method `ResetWidgetCache` after switching layers.
> 
> @giobiondani Does this work for you as well, or do you need the method to be public?
> 
> @pauldendulk Is this PR ok for the `develop/4.1` branch? Or can we find a way to make this work automatically (without the additional call to `ResetWidgetCache`)?

In my case I had to create my own code to manage the attributions anyway because I realized that Stadia and Thunderforest require multiple separate URL links in the same attribution and as far as I can tell I can't do that with the standard Attribution object. So I guess this isn't relevant anymore to my case.

janusw:
> So I guess this isn't relevant anymore to my case.

Thanks for the comment :)

inforithmics (approved):
I think this is currently the Best solution Because change notification was removed or will be removed.

janusw:
> I think this is currently the Best solution Because change notification was removed or will be removed.

I'm not fully sure what you mean by "change notification was removed", @inforithmics, but in #2694 @pauldendulk mentioned that the caching of widgets is removed entirely in v5. So in any case, it seems like this problem only still exists on the stable v4.1.x, and we should try to apply a simple fix there without too much restructuring. My fix is really the least intrusive way of fixing it, I think.

Since @inforithmics has approved this PR (and I need the fix rather urgently) I will merge it tonight, unless anyone objects until then.

inforithmics:
In Mapsui 5.0 the Widget can indicate that it needs a Redraw NeedsRedraw So that it is again rendered. So this cache isn't needed there.

janusw:
> In Mapsui 5.0 the Widget can indicate that it needs a Redraw NeedsRedraw So that it is again rendered. So this cache isn't needed there.

Thanks for the comment! I merged this now.

@pauldendulk @inforithmics Do you think we can tag a version `4.1.8` or `4.1.8-beta.2` soon?

PR #2744, opened 2024-09-14
Net 9.0 Support

inforithmics:
Adds .NET 9 Targets to the Mapsui Nuget Packages.

- [x] Needs SkiaSharp2.88 nuget Release (Is already fixed).
- [x] Add .Net 9 Target to Blazor
- [x] Add .NET 9 Target to Avalonia Browser
- [x] Add .Net 9 Target to Uno Browser
- [x] Release Version of .NET 9 
- [x] Blazor Sample App does not work yet https://github.com/mono/SkiaSharp/pull/3064
- [x] Needs Skia 3.116 for Blazor .NET 9 (https://github.com/Mapsui/Mapsui/pull/2822)
- [x] .NET 9 Uno Browser does not work yet needs (Uno 5.6 currently in Development) https://github.com/dotnet/runtime/issues/109289
- [x] .NET 9 Avalonia Browser does not work yet (https://github.com/AvaloniaUI/Avalonia/pull/17507)
- [x] Release Version of Uno Sdk 5.5.49 (.Net 9) Targets


inforithmics:
@pauldendulk when .NET 9 is released .NET 6 becomes end of life (should I remove the .NET 6 Targets in this pull request or in another one?)

Mapsui.UI.Maui/Globals.cs
```diff
@@ -0,0 +1,5 @@
+﻿global using Microsoft.Maui;
```
pauldendulk, line 1: I had some problems because of globals within Mapsui in the past. Because we support many platforms we have to deal with class name collisions a lot. The global namespaces make this problem harder to understand. No need to change this in here right now, but let's keep it to a minimum. If it causes problems in the future perhaps we should have it as a rule to use no globals at all.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -74,12 +75,14 @@ public partial class MapControl : INotifyPropertyChanged, IDisposable
...
     /// The movement allowed between a touch down and touch up in a touch gestures in device independent pixels.
     /// </summary>
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Visible)]
     public int MaxTapGestureMovement { get; set; } = 8;
 
     /// <summary>
     /// Use fling gesture to move the map. Default is true. Fling means that the map will continue to move for a 
     /// short time after the user has lifted the finger.
     /// </summary>
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Visible)]
```
pauldendulk, line 85: Why was this tag necessary?

Mapsui/Extensions/StreamExtensions.cs
```diff
@@ -48,10 +48,10 @@ public static bool IsSvg(this Stream stream)
         var buffer = new byte[5];
 
         stream.Position = 0;
-        stream.Read(buffer, 0, 5);
+        var read = stream.Read(buffer, 0, 5);
```
pauldendulk, line 51: Where did this cause a problem?

Mapsui/Utilities/EmbeddedResourceLoader.cs
```diff
@@ -53,8 +53,11 @@ private static string ConstructExceptionMessage(string path, Assembly assembly)
         if (similarNames.Length <= 0) return stringBuilder.ToString();
         var nameLength = assembly.GetAssemblyName()?.Length ?? 0;
         similarNames = similarNames.Select(fullName => fullName.Remove(0, nameLength + 1)).ToArray();
+#if NET9_0_OR_GREATER
```
pauldendulk, line 56: Why is this specific code needed in .NET 9? What whet wrong without it?

pauldendulk (commented):
Looks good. I have some questions about things I don't understand.

Mapsui.UI.Maui/Globals.cs
inforithmics: Removed global usings (Was a workaround for .net 6 and .net 8 Compilation but .net 6 isn't supported anymore so it can be removed and normal usings can be used).

Mapsui.UI.Shared/MapControl.cs
inforithmics: Changed it to
#if __WINDOWSFORMS__
    [DefaultValue(true)]
#endif

So that it's clear that this is for Windows Forms without that an WOF1000 error is made and now the setting can be set in the Windows Forms designer

Mapsui/Extensions/StreamExtensions.cs
inforithmics: it didn't cause problems, an analyzer displayed a warning that the return value was not observed. And when the stream has only 4 bytes only 4 bytes are read and the it is definitly not an xml file.

` if (read < 5 || !buffer.IsXml())`

Mapsui/Utilities/EmbeddedResourceLoader.cs
inforithmics: Was an .NET 9 issue but this has been fixed so removing it.

Mapsui/Extensions/StreamExtensions.cs
pauldendulk: Good improvement.

PR #2746, opened 2024-09-14
Update vulnerable Nugets

inforithmics:
Following nugets have vulnerable Versions:
Updated: System.Formats.Asn1  to 8.0.1
Updated: System.Private.Uri       to 4.3.2


pauldendulk (approved):
I assume those are indirect dependencies.

PR #2748, opened 2024-09-18
Mark Public Properties with DesignerSerializationVisibility   (Part of .NET 9 Pull Request)

inforithmics:
Things done:
I marked public Properties with DesignerSerializationVisibility  Attribute.
Properties with simple types (int, float,,,) I marked as visible. 
Properties with complex Types (Performance, Map) I marked as hidden because for that BinarySerialization is needed which is removed in .NET 9 and wouldn't work.

Taken from this pull request.
https://github.com/Mapsui/Mapsui/pull/2744

I discovered this while enabling .NET 9 support because this is an error/warning there.


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -241,6 +244,7 @@ public int UpdateInterval
     /// <remarks>
     /// If this is null, no performance information is saved.
     /// </remarks>
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)] 
     public Performance? Performance
```
inforithmics, line 248: This property becomes problematic with .NET 9 because Binary Serialization is removed.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -440,6 +444,7 @@ public Map Map
     [Parameter]
     [SuppressMessage("Usage", "BL0007:Component parameters should be auto properties")]
 #endif
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
```
inforithmics, line 447: This property becomes problematic with .NET 9 because Binary Serialization is removed.

pauldendulk: @inforithmics  The Map property becomes problematic because Binary Serialization is removed. I don't see a direct relation. How does that work? Did you get a warning in your IDE?

pauldendulk (commented):
@inforithmics I don't fully understand the problem. Not any non-primitive property will be problematic. I guess it is somehow related to the designer. Is the core problem not in the designer tooling because it is using binary serialization? Do you have references to discussions about this?

Mapsui.UI.Shared/MapControl.cs
inforithmics: Luckely the Designer does not serialize the Property but generates invalid code without this Property.

Steps to reproduce.
1) Drag Control on Windows Forms.
image

2) Invalid Code behind is generated.
image
Because in Color the Colors are not settable.

pauldendulk (approved):

PR #2747, opened 2024-09-14
Update to WindowsAppSdk 1.6

inforithmics:
Upgrade to Windows App SDK 1.6

Fixes This
https://github.com/Mapsui/Mapsui/issues/2741

Bugs:
- [x] Currently Crashes on Startup https://github.com/mono/SkiaSharp/issues/2999

PR #2751, opened 2024-09-19
Windows app sdk cleanup.

inforithmics:
Things done:
1) Removed unused nuget packages.
2) Only set the WindowsAppSdk Version on Mapsui.WinUI project. 
3) When this is merged only the Version Number of WindowsAppSdk needs to change and a SkiaSharp Updated is needed.

https://github.com/Mapsui/Mapsui/pull/2747


pauldendulk (approved):
Thanks for cleaning up

PR #2753, opened 2024-09-30
Use XCode 16 so the MacOs Build passes

inforithmics:
Use XCode 16 on macOs so  that the build succeeds.


pauldendulk (approved):
Thanks for fixing this.

PR #2755, opened 2024-10-05
Add Known Colors to Color.Parse

inforithmics:
Color Parse can now parse Known Colors like for example Magenta (This was possible in Mapsui 4.0)

### DIFF
--- a/Mapsui/Styles/Color.cs
+++ b/Mapsui/Styles/Color.cs
@@ -1,11 +1,12 @@
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 
 namespace Mapsui.Styles;
 
 public readonly record struct Color
 {
-    public Color() { }
+    private static Dictionary<string, Color>? _namedColors;
 
     public Color(Color color) : this(color.R, color.G, color.B, color.A) { }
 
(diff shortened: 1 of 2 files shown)


pauldendulk (commented):
When working on color I intentionally replaced this with the enum for all known colors. What could be the advantage of doing `Color.FromString("Magenta")` over `Color.Magenta`? Perhaps *making the migration easier* could be an argument in favor. But I am not sure I want it because I prefer having less stuff en supporting it better. What was for you the reason to add it again?

inforithmics:
When migrating from v4 to v5 I discovered that this was possible in v4 and not in v5. Because I have written Logic werde Users can change colors using the Color Editor on Windows so it is possible that Users selected the magenta Color. I have implemented this Color passe in custom Code and thought that it could be usefull for others. If you think this Not necessary I can close the pull request. Because this Code can Radolfzell be Generated by Ai. 
1. and the sourcecode of the file to au
2. Generate a case insensitive Dictionary NamedColors with Name of Color as key and Color as Valeska

pauldendulk (changes requested):
Okay, lets add this for backward compatibility. You will not be the only one who used that method. One change request though, to reduce the code I would prefer a case insensitive parse https://learn.microsoft.com/en-us/dotnet/api/system.enum.parse?view=net-8.0#system-enum-parse(system-type-system-readonlyspan((system-char))-system-boolean), unless there are good reasons against this.

inforithmics:
The parse (lookup) should already be case Insensitive because the _namedColors Dictionary is created With Case Insensitive Keys.
return _namedColors ??= new(StringComparer.InvariantCultureIgnoreCase)

Or dou you suggest that I parse the Color with WindowsForms Color and than set the RGBA values with that?

pauldendulk:
@inforithmics Sorry, I messed up. In my memory the `Color.<Color>` fields were an enum, but they are static fields and there is no simple general way to parse them. I was thinking of a way in which we would not have to add another full list of colors.

If we reorganize color we could introduce an enum internally as the core definition of our colors. This would allow simplifying the FromColor parse, but we would still need to add a full list of all the static fields, so not less code. 

I am going to merge it like it is.

pauldendulk (approved):
Approving as is after feedback

PR #2756, opened 2024-10-05
WIP Track .png files with Git LFS

inforithmics:
Things done 
Use Git Lfs for .png Files


inforithmics:
Added to many png files

PR #2759, opened 2024-10-11
Add husky to run dotnet format whitespace on every commit

pauldendulk:
With this PR dotnet format whitespace will be run on every commit and the changes will be include automatically, so we cannot forget it. 

Also had to fix a vulnerability build fail.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,0 +1,13 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "husky": {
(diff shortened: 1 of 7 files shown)


inforithmics:
I like the idea of formatting before commit.

pauldendulk:
> I like the idea of formatting before commit.

Later we can also run the analyzers, and gradually get more strict, so that we will not get any new nullable and dispose warnings in our repository.

pauldendulk:
Merging this one because it also fixes the build

Issue #1491, opened 2022-02-15
Design proposal for better layer ordering


pauldendulk:
Below is a proposal originating from another thread. Other options can be discussed, but note, there are a lot of things to take into account.

---------------

```c#
    public interface ILayerCollection
    {
        void Clear(); // Removes all layers for all groups
        void Clear(int group); // Removes all layers in a group. No default here. If there is no group argument all layers will be cleared, see Clear()
        void AddToTop(ILayer layer, int group = 0);
        void AddToBottom(ILayer layer, int group = 0);
        void MoveToTop(ILayer layerToMove, int group = 0);
        void MoveToBottom(ILayer layerToMove, int group = 0);		
        void MoveUp(ILayer layerToMove, keepWithinGroup = false); // if it is the top layer in the group it will become the bottom layer in the group above, but not if keepWithinGroup is true
        void MoveDown(ILayer layerToMove, keepWithinGroup = false); // if it is the bottom layer in the group it will become the top layer in the group above, but not if keepWithinGroup is true
        void InsertAbove(ILayer referenceLayer, ILayer layerToInsert);
        void InsertBelow(ILayer referenceLayer, ILayer layerToInsert);
        ILayer Remove(ILayer layer); // Returns the removed layers so that it can be used for moving
    }
```

_Originally posted by @pauldendulk in https://github.com/Mapsui/Mapsui/issues/598#issuecomment-471181466_

pauldendulk:
Random thoughts:
- When referring to a layer to be moved we could use the layer pointer, the layer name, the id within the layer list, some global id. 
- Perhaps we should discuss the internal structure first.
- Another thing to keep in mind is how to persist this information.

janusw:
So, there is already a `Mapsui.Layers.LayerCollection` which has very similar methods, right?

```c#
public class LayerCollection : IEnumerable<ILayer>
{
    public void Clear();
    public void Add(params ILayer[] layers);
    public void Move(int index, ILayer layer);
    public void Insert(int index, params ILayer[] layers);
    public bool Remove(params ILayer[] layers);
}
```

The interface is somewhat different from what you propose, but the functionality is basically the same, right?

So, what exactly do you suggest? Replace this one by the new one? Expand it to include some convenience methods? Or just use it more consequently?

charlenni:
[...]

PR #2762, opened 2024-10-14
Add groups to the LayerCollection

pauldendulk:
Add groups to LayerCollection. This was done with a LayerEntry private class with Layer, Order and Group fields. This way the ILayer did not have to be altered. The order of the layers in the internal class does not matter anymore because we always sort on Order and Group when requesting the list.

~Open~ issues:

- [x] There was a Clear method that removed all. Now I added ClearGroup(). Perhaps it should be ClearAllGroups and Clear(int group = 0) instead.
- [x] GetLayersOfGroup is like the above. Perhaps it should be GetLayers(int group) and GetLayersOfAllGroups().
- [x] The group parameter is often optional. This is good because group 0 should be the default most of the time. For some methods a default may cause confusion. UPDATE: For all public methods it is now optional
- [x] I use Span on one method. IEnumerable on the other. The problem with IEnumerable is the 'possible multiple enumerations' which makes you need to convert to List. I prefer Span for that reason. I need to figure out if it is possible to use that everywhere. UPDATE: Can not use it that much because we are dealing with field members here, and mostly input parameters.
- [x] We use the params keyword in some cases, but the group parameter can not be added after it because that is never allowed after a params keyword. Now I added overload. Perhaps the params keyword should be removed altogether because this way it is only confusing. UPDATE: The params have some added value because you can add a list of parameters as separate parameters, like this: Method(layer1, layer2, layer3, layer4). Because collection initializers are now a good alternative I prefer to keep the interface simpler and remove the params.
- [x] I don't like the 'keepWithinGroup boolean that I suggested [here](https://github.com/Mapsui/Mapsui/issues/1491). I think it should be about moving within a group. Moving between groups may need a some other method.
- [x] Removed the CopyTo method. I don't remember what it was for.
- [x] Perhaps Group should be called OrderGroup. Update: Keep Order for now
- [x] The main iterator operates directly on the collection. The becomes confusing now there are several ways to request layers (all, or grouped). UPDATE: I kept this one for now because it will not be clear with which to replace it.
- [x] The same problem for the this[int index]. I think this should be a method. UPDATE: There is a Get(index) now, and the [index] was removed
- [x] The public methods call private methods and at some point the event handlers are called. It is not very clear which method is responsible for calling the event handlers. UPDATE: Removed the individual change events, now only using OnChanged (which posts in bulk). Calling OnChanged on all public method. Renamed all private counter parts to <MethodName>Internal. Never call OnChanged in the internal methods.
- [x] MoveUp(ILayer layerToMove); // Move one up within the group (increasing index)
- [x] MoveDown(ILayer layerToMove); // Move one down within the group (lowering the index)
- [x] Add api documentation.

### DIFF
--- a/.husky/task-runner.json
+++ b/.husky/task-runner.json
@@ -12,17 +12,19 @@
         "--include",
         "${staged}"
       ]
-    },
-    {
-      "name": "re-add-the-staged-files-after-formatting",
-      "group": "pre-commit",
-      "output": "always",
-      "command": "bash",
-      "args": [ "git", "add", "${staged}" ],
-      "windows": {
-        "command": "cmd",
-        "args": [ "git", "add", "${staged}" ]
-      }
     }
+    // This adds the changed files back to the staging after formatting, but it sometimes fails
+    //,
+    //{
+    //  "name": "re-add-the-staged-files-after-formatting",
+    //  "group": "pre-commit",
+    //  "output": "always",
+    //  "command": "bash",
+    //  "args": [ "git", "add", "${staged}" ],
+    //  "windows": {
+    //    "command": "cmd",
+    //    "args": [ "git", "add", "${staged}" ]
+    //  }
+    //}
   ]
 }
--- a/Benchmarks/Mapsui.Rendering.Benchmarks/RenderToBitmapPerformance.cs
+++ b/Benchmarks/Mapsui.Rendering.Benchmarks/RenderToBitmapPerformance.cs
@@ -99,7 +99,7 @@ public class RenderToBitmapPerformance
 
         map.Layers.Add(layer);
 
-        var extent = map.Layers[0].Extent!;
+        var extent = map.Layers.Get(0).Extent!;
         map.Navigator.ZoomToBox(extent);
 
         return map;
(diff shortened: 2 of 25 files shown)


pauldendulk:
There are four more methods I could add:

void InsertAbove(ILayer referenceLayer, ILayer layerToInsert); // Insert above relative to another layer
void InsertBelow(ILayer referenceLayer, ILayer layerToInsert); // Insert below relative to another layer
void MoveUp(ILayer layerToMove);  // Move one up within the group (increasing index)
void MoveDown(ILayer layerToMove); // Move one down within the group (lowering the index)

IOL0ol1:
I think a 
GroupLayer:ILayer,IEnumable\<ILayer\> 
is better than a 
LayerCollection with group

pauldendulk:
> I think a
> GroupLayer:ILayer,IEnumable<ILayer>
> is better than a
> LayerCollection with group

The idea of this setup is that in most cases users do not need to deal with groups other than group 0. So, in most cases you do not need to specify a group. But if the use cases get more complicated perhaps grouping like you describe become the best option. It is something to keep in mind the next time I work on the LayerCollection.

Issue #2758, opened 2024-10-10
After upgrade to Visual Studio to version 17.11.5 the app crashes with Mapsui control


willembos1969:
4.1.7

Winui

Windows

Upgrade VS and start the app without other modifications the app crashes.

**Expected behavior**
No crash

**Screenshots**

**Additional context**
comment out mapcontrol and app just starts without issues

willembos1969:
Please fix it quickly because this map control is important for my app

pauldendulk:
I have seen such issues in the past and then they could be fixed by deleting all bins and objs. Close VS and run `scripts\DeleteBinAndObjFolders.ps1` in the sln folder.

willembos1969:
I've deleted all obj and bin folders for all folders used by the solution. Then start VS again and run the app, but still crashes.

willembos1969:
Also I've created a new project and also crashes

inforithmics:
Could it be this Problem

WillemBos:
No i'm still on wasdk 1.5 because of that issue.

WillemBos:
Upgrade vs to 17.11.5, wasdk 1.6 latest version and build and SkiaSharp.Views.WinUI 3.0.0 preview 5.3 fixed the issue

erledzus:
I encountered a similar sounding issue, after upgrading to vs 17.11.5, Maui Windows project.  Commenting out the Mapsui control allowed the application to start up.

It seems not related to Mapsui.  The workaround that is successful for me is described in [Visual Studio 17.11.5 causes Windows MAUI to break build](https://github.com/dotnet/maui/issues/25160)

To resolve it I added these lines to the csproj:
```
<CsWinRTRcwFactoryFallbackGeneratorForceOptOut>true</CsWinRTRcwFactoryFallbackGeneratorForceOptOut>
<WindowsSdkPackageVersion>10.0.19041.54</WindowsSdkPackageVersion>

```

inforithmics:
The workaround is to Install the newest Preview
SkiaSharp 2 (2.88.9-preview.2.2) or 
SkiaSharp 3 nugets (3.0.0-preview.5.4 or greater).

pauldendulk:
Mapsui 4.1.8 was released to fix this: https://github.com/Mapsui/Mapsui/releases/tag/4.1.8

PR #2767, opened 2024-10-24
Fix: WinUI Targets (Maui,Uno,WinUI)

inforithmics:
With newer Visual Studio Versions the Default WindowsSdk that is used is per default higher or equal to 10.0.19041.38 which causes SkiaSharp to stop working on WinUI Targets, Like Mapsui.WinUI, Mapsui.Maui (Windows), Mapsui.Uno (WinUI)
This fixes this issue too. 

https://github.com/Mapsui/Mapsui/issues/2758

This happens too when the .NET 8 SDK is updated beyond Version 8.0.403 and greater. As an added Bonus Mapsui works now on .NET 9 Wasm targets too.


pauldendulk (approved):
Fantastic! I ran into this issue myself when upgrading my sdk to 8.0.403. I could not do anything until I downgraded again (by deleting the sdk folder). All my searches for a fix lead to the skiasharp issue that you posted https://github.com/mono/SkiaSharp/issues/2999.

Issue #2764, opened 2024-10-22
Location pin image not appearing in 5.0.0


smathersgit:
**Mapsui Version**
5.0.0-beta 3

**Mapsui Platform**
MAUI

**Device**
Android, MAC

**Describe the bug**
I have been successfully adding a red location pin image to my mapView using Xamarin, and Maui app builds with 4.x.x versions, including the last stable 4.1.7 release. However, the pin image doesn't appear when moving to 5.0.0-beta 3. I tried experimenting with the pin Callout feature as an alternative way of achieving the same, but although CalloutType.Detail works fine, I cannot find an example of how to use the CalloutType.Image with my own location image.

            mapView.Map = map;
            mapView.Pins.Add(new Pin()
            {
                Label = "Pin Label",
                Position = groundPosition,  // In my case this is a point within the map's view 
                Type = PinType.Pin,
                Scale = 0.5f,
                Color = Xamarin.Forms.Color.Red,   // or Colors.Red with MAUI
                Transparency = 0.0f,
                RotateWithMap = true,
                IsVisible = true,
            });

`

**To Reproduce**
In my case I am using Mapsui in a with dotnet 9 Preview Maui app build, and testing on Android, and IOS emulators. I haven't tried it on a real device yet.

1. Add a pin to a mapView (as in the bug description)
2. Build with Mapsui 4.1.7 and confirm pin is visible on device
3. Repeat 1. and 2. using Mapsui 5.0.0-beta 3 and confirm pin doesn't appear

**Expected behavior**
My expectation was that the location pin would appear on Mapsui 5.x.x builds as it did previously with 4.x.x.

inforithmics:
I could reproduce it. Will try to implement a fix

smathersgit:
Can confirm that your suggested workaround was successful (on Android an IOS emulators), thanks!

My slightly amended version of the relevant Category Demo and Add Pin extract:
`
        Mapsui.Styles.Color colour = new Mapsui.Styles.Color(255, 0, 0);
        var layer = new GenericCollectionLayer<List<IFeature>>
        {
            Style = SymbolStyles.CreatePinStyle(fillColor: colour, symbolScale:0.5f)
        };
        mapView.Map.Layers.Add(layer);
        layer?.Features.Add(new GeometryFeature
        {
            Geometry = groundLocation.ToPoint(),           
        });
        // To notify the map that a redraw is needed.
        layer?.DataHasChanged();
`

janusw:
[...]

Issue #2765, opened 2024-10-23
[5b3] setting Pin.Scale triggers sequence contains no element


softlion:
**Mapsui Version**
5.0.0-beta3

**Mapsui Platform**
Android

**Device**
Android simulator and devices

**Describe the bug**
When creating a new Pin, setting its Scale value throws an exception "sequence contains no element".

```
System.InvalidOperationException: Sequence contains no elements
   at System.Linq.ThrowHelper.ThrowNoElementsException()
   at System.Linq.Enumerable.First[IStyle](IEnumerable`1 source)
   at Mapsui.UI.Maui.Pin.OnPropertyChanged(String propertyName)
   at Mapsui.UI.Maui.Pin.set_Scale(Single value)
```

**To Reproduce**
Use that code to trigger the issue:
```c#
            var pin = new Pin
            {
                Tag = new(),
                Type = PinType.Pin,
                Position = new (0,0),
                Label = "label",
                Address = "address",
                Svg = "svg string",
                Color = Colors.Blue.ToMapsui(),
                Scale = .7f,
                Transparency = .45f, 
            };
```
Also, could you replace Transparency with the standard Opacity and reverse the values (1-n) ?

**Expected behavior**
regression: that does not crash in 4.x

It looks like it's trying to get a Style, which is not set.
```c#
((SymbolStyle) this.Feature.Styles.First<IStyle>()).SymbolScale = (double) this.Scale;
```

Also, that code does not work when a `VectorStyle` is added to Features first. It supports only a `LabelStyle`.

That also applies to `Pin.Transparency`. When a VectorStyle is set, you can not set `Pin.Transparency` (crash).

```c#
            var pin = new Pin
            {
                Tag = new(),
                Type = PinType.Pin,
                Position = new (0,0),
                Label = "label",
                Address = "address",
                Svg = "svg string",
                Color = Colors.Blue.ToMapsui(),
                Feature =
                {
                    // Styles = { new SymbolStyle { SymbolScale = .7f } }
                    Styles = { new VectorStyle
                    {
                        Fill = new Brush { Color = Colors.Blue.ToMapsui(), FillStyle = FillStyle.Svg, },
                        Outline = new Pen(Colors.Blue.ToMapsui()) { Width = 1, PenStyle = PenStyle.Solid },
                        //Opacity = 0.55f
                    } },
                },
                //Scale = .7f,
                Transparency = .45f, //this value is reversed. 0 = opaque. 1 = transparent.
            };
```

[...]

PR #2768, opened 2024-10-24
Fixing Pin in MapView

inforithmics:
Fixes this 
https://github.com/Mapsui/Mapsui/issues/2764

### DIFF
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -91,6 +91,11 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
     /// Occurs when selected pin changed
     /// </summary>
     public event EventHandler<SelectedPinChangedEventArgs>? SelectedPinChanged;
+    
+    /// <summary>
+    /// Occurs when map clicked
+    /// </summary>
+    public event EventHandler<MapClickedEventArgs>? MapClicked;
 
     #endregion
 
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -666,7 +666,7 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
                 return handled;
             }
         }
-        
+
         return handled;
     }
 
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -673,11 +673,11 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
         {
             try
             {
-                Map?.Layers.Remove([MyLocationLayer, _mapCalloutLayer, _mapPinLayer, _mapDrawableLayer]);    
+                Map?.Layers.Remove([MyLocationLayer, _mapCalloutLayer, _mapPinLayer, _mapDrawableLayer]);
             }
             catch (Exception ex)
             {
-                Logger.Log(LogLevel.Error, "Layers could not be removed", ex);   
+                Logger.Log(LogLevel.Error, "Layers could not be removed", ex);
             }
         }
     }
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -91,7 +91,7 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
     /// Occurs when selected pin changed
     /// </summary>
     public event EventHandler<SelectedPinChangedEventArgs>? SelectedPinChanged;
-    
+
     /// <summary>
     /// Occurs when map clicked
     /// </summary>
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -527,18 +527,6 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
 
     private void HandlerInfo(MapInfoEventArgs e)
     {
-        // Handle Map Clicked
-        if (MapClicked != null)
-        {
-            var position = e.MapInfo.WorldPosition.ToMaui();
-            var mapClicked = new MapClickedEventArgs(position, TapType.Single);
-            MapClicked.Invoke(this, mapClicked);
-            if (mapClicked.Handled)
-            {
-                return;
-            }
-        }
-        
         // Click on pin?
         if (e.MapInfo?.Layer == _mapPinLayer)
         {
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -40,9 +40,11 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
     private readonly ObservableRangeCollection<Pin> _pins = [];
     private readonly ObservableRangeCollection<Drawable> _drawables = [];
     private readonly ObservableRangeCollection<Callout> _callouts = [];
+    private readonly MapTappedWidget _mapTappedWidget;
 
     public MapView()
     {
+        _mapTappedWidget = new MapTappedWidget(HandlerTap);
         MyLocationFollow = false;
 
         IsClippedToBounds = true;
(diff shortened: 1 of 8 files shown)


pauldendulk (approved):
Thanks for fixing the MapView. By using the MapInfoEventArgs it is a bit more similar to the MapControl. I still have my doubts about supporting the MapView. In part because I don't want to spend my own time in maintaining it. That problem is solved if you (and/or others) maintain it. But the MapView is sometimes also blocking my refactoring, turning a change that is basically very simple into a puzzle if I want to keep the MapView working. But we will see how it evolves. It works for now.

pauldendulk:
This project is marked as WIP. What else needs to be done?

janusw:
> Fixes this #2764

Does it only fix #2764 or also #2765?

inforithmics:
The reason this is WIP is that:
1. I first fixed the Mapview Samples Click didn't work.
2. The actual Bug that the Pin isn't rendered I haven't fixed yet.
    Before the Pin was registered as an Image and then this Image was drawn. 
    So there are two possible Solutions one is to allow Image Data in ImageUrl like this https://github.com/Mapsui/Mapsui/pull/2725 or Create the Correct Feature so that it is rendered correctly (What I'm aiming at)

inforithmics:
@janusw I think both will be fixed, because on CreateFeature no Style was created for the Pin so that explains the Crash on `((SymbolStyle) this.Feature.Styles.First<IStyle>()).SymbolScale = (double) this.Scale;`

inforithmics:
@pauldendulk @janusw  Fixed the Pin now in MapView with https://github.com/Mapsui/Mapsui/pull/2768/commits/7a77a9d8325d0c02318cb52ef2cd2a6164c1cdba

What's missing is that a Double Tap needs to be Handled for the PinSample to work properly, currently only single tap is handled because I subscribed to the Info Event. But I think this should be another Pull Request.
Possible Follow Up Pull Requests:
1. Handle DoubleTap
2. Use Layer Groups in MapView.

Mapsui.UI.MapView/MapView.cs
```diff
@@ -709,7 +674,17 @@ private void AddLayers()
     private void RemoveLayers()
     {
         // Remove MapView layers
-        Map?.Layers.Remove([MyLocationLayer, _mapCalloutLayer, _mapPinLayer, _mapDrawableLayer]);
+        if (Map?.Layers.Count >= 4)
```
inforithmics, line 677: Fix Crash on Startup, needs to be rewritten to use Groups

inforithmics:
I Fixed now the Tap Event with the help of an MapTapWidget, so accessing the Tap Event from the Map and routing it to the TapHandler which is now connected to this.

janusw:
I can confirm that at this point the pin samples in the MapView category work again. 🥳 The error message is gone and the pins are rendered again. Thanks for that, @inforithmics!

However, it seems that the `Add mans Pins` sample takes forever to load. Apparently the performance is much worse than in v4. 🤷 Any idea what might be the reason for this?

inforithmics:
I'll have a look at it. Maybe some exceptions are thrown somewhere

inforithmics:
@janusw I had a look at the Many Pins sample. There are two reasons. 
1. in V4 Skia was used for changing the color and in V5 svg is used for changing the color.
2. Every Pin has a different color so for every different color a seperate Pin is created (around 900, changing svg, converting svg to pin).
3. So one possibility is not to have 900 different colors only 1-5 and then this is much faster, because the svg and pin are cached. and only 1-5 skia pins are created.

inforithmics:
This is the code that was used for changing the color on a pin with skia. Don't know If I want to port that.
```
 var colorInHex = Color.ToHex();
                    if (colorInHex == null)
                        return;

                    // Check, if it is already in cache
                    if (_bitmapIds.TryGetValue(colorInHex, out var id))
                    {
                        _bitmapId = id;
                        _bitmapIdKey = colorInHex;
                        break;
                    }

                    // Load the SVG document
                    Svg.Skia.SKSvg svg;
                    using (var stream = EmbeddedResourceLoader.Load("Images.Pin.svg", typeof(Pin)))
                    {
                        if (stream == null)
                            return;

                        // Create a new SVG object
                        svg = stream.LoadSvg();
                        if (svg.Picture == null)
                            return;
                    }

                    Width = svg.Picture.CullRect.Width * Scale;
                    Height = svg.Picture.CullRect.Height * Scale;
                    // Create bitmap to hold canvas
                    var info = new SKImageInfo((int)svg.Picture.CullRect.Width, (int)svg.Picture.CullRect.Height) { AlphaType = SKAlphaType.Premul };
                    using (var bitmap = new SKBitmap(info))
                    {
                        using var canvas = new SKCanvas(bitmap);
                        // Now draw Svg image to bitmap
                        using (var paint = new SKPaint() { IsAntialias = true })
                        {
                            // Replace color while drawing
                            paint.ColorFilter = SKColorFilter.CreateBlendMode(Color.ToSKColor(), SKBlendMode.SrcIn); // use the source color
                            canvas.Clear();
                            canvas.DrawPicture(svg.Picture, paint);
                        }
                        // Now convert canvas to bitmap
                        using (var image = SKImage.FromBitmap(bitmap))
                        using (var data = image.Encode(SKEncodedImageFormat.Png, 100))
                        {
                            _bitmapData = data.ToArray();
                        }
                        // Broke this with no plans to repair: _bitmapId = BitmapRegistry.Instance.Register(_bitmapData);
                        _bitmapIdKey = colorInHex;
                        _bitmapIds.Add(colorInHex, _bitmapId);
                    }
```

current code for changing color in Svg.

```
 if (symbolStyle.SvgFillColor.HasValue || symbolStyle.SvgStrokeColor.HasValue)
 {
     var drawableImage = renderService.DrawableImageCache.GetOrCreate(ToModifiedSvgKey(symbolStyle.ImageSource, symbolStyle.SvgFillColor, symbolStyle.SvgStrokeColor),
         () =>
         {
             var modifiedSvgStream = SvgColorModifier.GetModifiedSvg(svgImage.OriginalStream, symbolStyle.SvgFillColor, symbolStyle.SvgStrokeColor);
             var skSvg = new SKSvg();
             modifiedSvgStream.Position = 0;
             skSvg.Load(modifiedSvgStream);
             if (skSvg.Picture is null)
                 throw new Exception("Failed to load modified SVG picture.");
             return new SvgImage(skSvg.Picture);
         });

     PictureRenderer.Draw(canvas, ((SvgImage)drawableImage).Picture,
         (float)destinationX, (float)destinationY,
         rotation,
         (float)offset.X, (float)offset.Y,
         opacity: opacity, scale: (float)symbolStyle.SymbolScale, blendModeColor: symbolStyle.BlendModeColor);
 }
```

pauldendulk:
Using [SKSvg](https://www.nuget.org/packages/Svg.Skia) adds an important feature. It becomes possible to have a single SVG and use custom styles for fill and stroke. Unfortunately the only way to implement this was the create a separate instance of the SVG. I realized this problem when writing this code and added a remark about it somewhere. If you work with continuous colors the cache would keep on growing. 

This is an unfortunate necessity, but you only need it if you use custom styles for both the fill and the outline. In v4 we did not have this functionality, so in v5 it should not be needed for the same sample. So, it should be possible to think of a way to have both options (fast performance in some situations, flexibility in other situations).

In the SVG color there are currently these 3 fields to determine color. BlendModeColor is the old one. It draws the SVG in a single color and should need a single instance. SvgFillColor and SvgStrokeColor are for custom colors.

```csharp
    public Color? BlendModeColor { get; set; }
    public Color? SvgFillColor { get; set; }
    public Color? SvgStrokeColor { get; set; }
```

I wrote this comment without looking at the code involved in either the samples or the renderer.

pauldendulk (approved):
Thanks for fixing this

softlion:
> I wrote this comment without looking at the code involved in either the samples or the renderer.

Is this cache cleaning itself when images are not used anymore or does it keep growing indefinetly ?
Maybe provide a way to clean it, either completely, or based on a last used date associated with each entry.

pauldendulk:
> > I wrote this comment without looking at the code involved in either the samples or the renderer.
> 
> Is this cache cleaning itself when images are not used anymore or does it keep growing indefinetly ? Maybe provide a way to clean it, either completely, or based on a last used date associated with each entry.

This particular cache (which was added recently) does not have a way to cleanup. Before we add this we should straighten out all the other caches and their cleaning mechanism. So, I will not accept a PR to fix this. In general caches cause problems, and they cause problems of the kind that are hard to detect and show up late in production. In a mapping application caching has its own complications which are different from most other applications. You need to take into account iterations, and which items are show in an iteration, you can not simply cleanup based on time. 

Also, in the case of the 'Add many pins' sample cache cleanup would not solve the problem.

PR #2774, opened 2024-10-28
Use BlendModeColor instead of SvgFillColor in MapView Pin

pauldendulk:


inforithmics:
Thanks for fixing this.

PR #2781, opened 2024-10-28
Fixing: Maui Samples Build (Specify SkiaSharp PackageReferences)

inforithmics:
Things done: 
1.) Added SkiaSharp references to Mapsui.Samples.Maui so that it compiles (Rasterizer has SkiaSharp Dependencies and therefere I have to specifiy it explicitly so that it takes the 2.89 preview versions).
2. Revert Uno.sdk to 5.0.108 later uno versions Seen to have Problems

### DIFF
--- a/Mapsui.UI.WinUI/Mapsui.UI.WinUI.csproj
+++ b/Mapsui.UI.WinUI/Mapsui.UI.WinUI.csproj
@@ -18,7 +18,7 @@
   </PropertyGroup>	
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


inforithmics:
If going back to Uno 5.3 fixes the build this means Uno 5.4 has some Side effects in other targets which means I only should set Uno.sdk global json settings in Uno projects avoiding these side effects.

inforithmics:
Uno.Sdk needs to be set in the Solution Folder Level

pauldendulk (approved):

PR #2769, opened 2024-10-25
Fix: WinUI Targets (Maui,Uno,WinUI) [4.1]

inforithmics:
Fixes this
https://github.com/Mapsui/Mapsui/issues/2758

PR #2790, opened 2024-11-09
Use release version of SkiaSharp 2.88.9 [4.1]

inforithmics:
Use now the non Preview Version of SkiaSharp 2.88.9


janusw (approved):
Yes, please!

I know that the preview version of SkiaSharp was introduced in PR #2769, in order to fix #2758, but of course it is not a good idea to use a preview package in a stable release, and it's good that we have a final release of SkiaSharp 2.88.9 now.

PR #2796, opened 2024-11-17
Add way to add Pin with Callout through the MapBuilder

pauldendulk:
Investigating to allow a simpler way to add callout through the builder. Several questions and related issues come up while working on this:

- The IsInfoLayer field is annoying. Let's investigate to fetch the layer info actively in the Info event. So, it is the users call which determines which layers are fetched in MapInfo.
- The string indexer on IFeature to get data (`["bla"]`) becomes clumsy when building more advanced functionality like this. Perhaps it is time to look into the idea to add a Data field of type object which could be cast to a specific type in user code.
- We have to pass the Map object along several method calls because we need the (Map)Info event. I am thinking of an easier way. Perhaps we need some tap event on Style. Or pass in a style tap event in the builder methods.
- A callout's purpose is to shows feature specific information. The text if different for different features, that is why you want to see it. This is why you almost always need to nest the CalloutStyle in ThemeStyle. Perhaps the CalloutStyle (or a variant of the callout style) should have ThemeStyle method built in.
- A layer should have Styles instead of a Style so that you do not have to use a StyleCollection all the time. This is a big change.
- An alternative would be composed styles, with the CalloutStyle as a child style of the SymbolStyle. So, any feature would become a StyleCollection. We should make sure the VisualFeatureIterator works correctly with this. 
- The CalloutStyle defaults could be better so less fields need initialization (the title font color is white on a white background).

### DIFF
--- a/Mapsui.Tiling/Extensions/MapBuilderExtensions.cs
+++ b/Mapsui.Tiling/Extensions/MapBuilderExtensions.cs
@@ -7,9 +7,9 @@ public static class MapBuilderExtensions
 {
     public static MapBuilder WithOpenStreetMapLayer(this MapBuilder mapBuilder, ConfigureLayer configureLayer)
     {
-        return mapBuilder.WithLayer((map) => OpenStreetMap.CreateTileLayer(), (l) =>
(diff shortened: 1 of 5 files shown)


pauldendulk:
This is work in progress, but since the code is mostly in the samples I merge it.

PR #2798, opened 2024-11-17
Updating Avalonia for .Net 9 (Browser) Compatibility

inforithmics:
Avalonia is updated so that it works with .Net 9 Targets and Browser. Is taken from this pull request. https://github.com/Mapsui/Mapsui/pull/2744

### DIFF
--- a/.github/workflows/dotnet-docs.yml
+++ b/.github/workflows/dotnet-docs.yml
@@ -42,7 +42,7 @@ jobs:
       uses: actions/setup-dotnet@v4
       with:
          dotnet-version: |
-           8.0.x
+           9.0.x
     - name: install workloads
       run: dotnet workload install maui macos android ios maccatalyst wasm-tools wasm-tools-net8 
     - name: Run DocFx
--- a/.github/workflows/dotnet-docs.yml
+++ b/.github/workflows/dotnet-docs.yml
@@ -44,7 +44,7 @@ jobs:
          dotnet-version: |
            8.0.x
     - name: install workloads
-      run: dotnet workload install maui macos android ios maccatalyst wasm-tools wasm-tools-net8
+      run: dotnet workload install maui macos android ios maccatalyst wasm-tools 
     - name: Run DocFx
       shell: pwsh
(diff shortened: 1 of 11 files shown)


.github/workflows/dotnet-release-nugets.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         echo $VERSION_OF_RELEASE
 
     - name: install workloads
-      run: dotnet workload install maui macos android ios maccatalyst wasm-tools wasm-tools-net8
```
pauldendulk, line 45: The wasm-tools-net8 tools where only for Avalonia?

.github/workflows/dotnet-docs.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       uses: actions/setup-dotnet@v4
       with:
          dotnet-version: |
-           9.0.x
```
pauldendulk, line 45: Why back to 8 in the docs? This seems not related to Avalonia? Is it by mistake? Or does this revert an earlier error?

inforithmics: No this is an error, will revert that

.github/workflows/dotnet-release-nugets.yml
inforithmics: No this is an error, will revert that

.github/workflows/dotnet.yml
```diff
@@ -218,8 +218,8 @@ jobs:
       uses: NuGet/setup-nuget@v2
       with:
         nuget-version: latest
-    - name: install maui macos android ios maccatalyst wasm-tools wasm-tools-net8 wasm-tools wasm-tools-net8
-      run: dotnet workload install maui macos android ios maccatalyst wasm-tools wasm-tools-net8 wasm-tools wasm-tools-net8
+    - name: install maui macos android ios maccatalyst wasm-tools wasm-tools-net8
```
inforithmics, line 221: wasm-tools and wasm-tools-net8 was duplicated

pauldendulk (approved):
👍

PR #2802, opened 2024-11-18
Add IFeature.Data and remove ZOrder

pauldendulk:
- It is possible to store all kinds of data in the indexer fields but this is not always handy. A `Data` field is added and a `DataAs<T>` method to make casting easy. Look at the MapBuilderSample.cs to see how this can affect the code. I am not sure if this is a great solution, an improvement but there might be better ways. I like to hear suggestions.
- Also I made the BaseFeature inherit from IFeature so that I can use `/// <inheritdoc/>` in the BaseFeature so I don't have to copy the documentation from IFeature in BaseFeature. Currently there was no documentation in BaseFeature. This also affects the classes derived from BaseFeature because they need to override keyword for the newly added abstract methods/fields.
- Removed the ZOrder. This was introduced to fix a problem with randomly changing zorder of features in the MemoryLayer. This was because the order of features is not always the same in the collection class used. This problem is now resolved by using the Id to order. The ZOrder caused confusion because it existed for all features but only worked in the MemoryLayer. The SortFeatures could be used to sort features to your choosing. If users want to use a ZOrder field they could do this with such a field in the IFeature.Data object.


pauldendulk:
I removed the `DataAs<T>` method that was added in an earlier commit. It did not add a lot of value. Instead you need to cast it. Not pretty but fully understood by most dotnet developers.

I have investigated using a `<T>` on IFeature to add a type to the Data, but did not find a good solution. I don't want to force inheritors to use the `<T>` if they don't care about the Data field. There is a way to workaround this, but it would need special instances for all implementations (GeometryPolygon, PointPolygon). A general improvement would be to introduce a Geometry class, which has to implement the CoordinateVisitor and the Copy method and such, so that we can have just one type of Feature. This is a long term thing if we ever find time for this.

The WithPinWithCalloutLayer method needs more improvement. I am now considering a different type of callout which has readonly fields and a configure method with which to set the fields based on the feature, somewhat like the ThemeStyle. This will be in a different PR.

PR #2804, opened 2024-11-18
Update docfx to 2.78

inforithmics:
Trying to fix doc build


pauldendulk (approved):
Let's see. You could test it locally with `docfx --serve`. I wrote something about the setup here: https://mapsui.com/v5/documentation/

PR #2805, opened 2024-11-18
Remove .NET 6 Targets (End of Life)

inforithmics:
Removed .Net 6 Targets because they are end of life as of November 2024 (And could cause problems later on).


Mapsui.Nts/Mapsui.Nts.csproj.Backup.tmp
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
pauldendulk, line 1: I think this was added by accident.

pauldendulk (commented):
Great! Please look at my remark.

Mapsui.Nts/Mapsui.Nts.csproj.Backup.tmp
inforithmics: Yes you are right

inforithmics:
Is now fixed

pauldendulk (approved):

PR #2807, opened 2024-11-22
Get MapInfo through method

pauldendulk:
What I trying to achieve with this PR is that ILayer.IsInfoLayer can be removed. This is done by letting the caller retrieve the info through a new GetMapInfo method, which takes the layers to query as parameters. 

Things done so far:
- In this first commit we add GetMapInfo but still use IsInfoLayer.
- Since MapInfo can not be null and I removed a lot of checks for null.
- We sometimes use `a` for the EventArgs but `e` is more common, I changed the `a`s to `e`s.

An issue I ran into:
- The GetMapInfo method is added to the MapInfoEventArgs but there is already a GetMapInfoAsync on the MapInfo class, which does requests to the server. It is better to have them on the same level.
   - A downside of MapInfo.GetMapInfoAsync is that it turns the MapInfo in a a basic core component does not depend on requests to the server.
   - A downside of MapInfoEventArgs.GetMapInfo is that you need to call the method to get the ScreenPosition and WorldPosition.


pauldendulk:
I will merge this in its current form. Work is continue here: https://github.com/Mapsui/Mapsui/pull/2807

PR #2809, opened 2024-11-25
Update system text json to version 9.0.0

inforithmics:


pauldendulk (commented):
What led you to make this change?

IldarKhayrutdinov:
@pauldendulk I was asking https://github.com/Mapsui/Mapsui/discussions/2808

pauldendulk:
@IldarKhayrutdinov And why do you need that? Is that the only thing blocking the use of .net9?

Haukinger:
If I add a direct dependency in my app, it seems to work fine with net9, as far as I have tested by now.

IldarKhayrutdinov:
Yes, I have direct dependency. System.Text.Json v9.0.0 supports .net8, there are no obstacles to using v9.0.0 in mapsui (on .net8).

pauldendulk (approved):

PR #2816, opened 2024-11-30
Update Dot.Next Threading for .NET 8 support

inforithmics:
Updated dotnext.threading (for .net 8 support)


pauldendulk (approved):
Thanks for the improvement. At what point could we replace DotNext with 'the real thing'? Since DotNext is supposed to be a precursor to what is coming in dotnet.

Issue #2810, opened 2024-11-25
Panning does not work after the WPF quickstart


pauldendulk:
**Mapsui Version**
5.0.0-beta.5

**Mapsui Platform**
WPF

**Device**
Windows 11

**Describe the bug**
Following the WPF quickstart. Try to drag the map. Nothing changes. Mouse wheel works. Flinging the map works.

**Additional context**
I can not reproduce the problem in Visual Studio using the mapsui.sln.

pauldendulk:
I've spend quite some time on this today. I can not reproduce the bug using the mapsui.sln in VS. So, I've built a nuget package on disk and used that in the sample project (in which I could reproduce it using the online package) but could not reproduce it either. I did built Release packages. Right now I don't know how to proceed.

Possible causes:
- Something in SkiaSharp
- Some bug we have in our logic that only shows with some particular timing.
- some bug we have in out logic that only shows with a particular build process (I did a build in VS with Release selected and used this pack command: `dotnet pack /p:RestorePackages=true /p:Configuration=Release /p:Version=5.0.0-beta.15 Mapsui.sln --output Artifacts`.

inforithmics:
workaround is setting the Map from external.
like
`MapControl.Map = new Map(); // and then set the layers`

TAlecksen:
Thank you both for the workaround. This fixes the issue for me.

pauldendulk:
I assume this is fixed in 5.0.0-beta.6. Please reopen if it is not.

PR #2818, opened 2024-11-30
BugFix: Map not fully initialized when it is created in the Map Property

inforithmics:
Fixes this https://github.com/Mapsui/Mapsui/issues/2810


pauldendulk (approved):
Great, thanks for the fix. I assume the _imageSource thing is not related but just an improvement.

PR #2820, opened 2024-12-08
Workaround: Android Maui no Map displayed a second time when using MapControl.UseGPU = true

inforithmics:
Fixes this: https://github.com/Mapsui/Mapsui/discussions/2255

extracted from the SkiaSharp 3 pull request

The Skiasharp fix would be
https://github.com/mono/SkiaSharp/pull/3076


pauldendulk (approved):
Thanks a lot for debugging and fixing such a nasty issue. One thing I am worried about is that it should be clear what is part of the workaround and what not (so we can remove that later one). All added methods will become unused when you remove OnParentSet, so no problem there. Perhaps the _page and _element could have had a comment mentioning the workaround, but I will merge it as is.

PR #2737, opened 2024-08-26
Fix mistake in theme style sample

ndsvw:
Removed line that overwrote "style.Outline", if it was set in the switch-case block.

Also improved code comments that did not match the code.

Did not create an issue, because it's a very small fix...

PR #2822, opened 2024-12-12
Update to SkiaSharp 3.116.1 With Minimal Changes

inforithmics:
This pull request updates only SkiaSharp to 3 without Obsoletion Fixes. 
I'm doing this to unlock
https://github.com/Mapsui/Mapsui/pull/2744

And to Later Merge changes of 
without massive Render Changes


pauldendulk (commented):
Good approach to turn this into a separate step. 

I compared the ThemeStyleSample.Regression.png images and the new one has black outline on the colored geometries while the old one does not. What causes this change?

image

inforithmics:
> Good approach to turn this into a separate step.
> 
> I compared the ThemeStyleSample.Regression.png images and the new one has black outline on the colored geometries while the old one does not. What causes this change?
> 
> image

The changes came by this pull request 
https://github.com/Mapsui/Mapsui/pull/2737
So the Image was not changed enough to be detected by the unit tests. Since then the Countries have a border.

pauldendulk:
> The changes came by this pull request #2737 
> So the Image was not changed enough to be detected by the unit tests. Since then the Countries have a border.

ah, so I should have updated the image in my PR as well. 

At the start I was very strict with the image checks. Later I loosened the checks because there were difference between the images that I could not explain from Mapsui code. But nowadays a lot of real differences are accepted by the asserts. Perhaps with the new skia version we could try to make it more strict again. And perhaps be more clever in what we do accept, for instance, only small differences in color, and not one pixel that is an entirely different color.

pauldendulk (approved):

PR #2824, opened 2024-12-19
Change: Set UseGPU to true in Blazor

inforithmics:
Change the default of Blazor to UseGPU = true


pauldendulk (approved):
Great this is now possible.

PR #2832, opened 2025-01-02
Move GetMapInfoAsync to MapInfoEventArgs

pauldendulk:
Follow up of https://github.com/Mapsui/Mapsui/pull/2807

With this PR both GetMapInfo and GetMapInfoAsync become methods of MapInfo.

To allow this GetMapInfoAsync was turned into a separate method within the renderer. Splitting up a method that was quite large. It is now more clear that GetMapInfoAsync is not related to rendering at all and may have to move somewhere else altogether. 

More changes will probably follow.


pauldendulk:
Hey @inforithmics, I have some questions related to GetMapInfo and GetMapInfoAsync and since you also did a lot of work on this perhaps you can give me some feedback on this topic.

I will start with a practical question. The ILayerFeatureInfo is implemented in four classes. It seems they go in pairs.

The ImageLayer and the WmsProvider. The ImageLayer calls the WmsProvider. There is a sample that gets the feature info in "Wms info" under "Info". It works properly.

The RasterizingTileLayer and the RasterizingTileSource. The RasterizingTileLayer calls the RasterizingTileSource. I found one sample of the RasterizingTileLayer that has an InfoWidget, but it does not show the MapInfo: The "RasterizingTileLayer with LineString" under "Performance".

Do you know if there are any other samples that should work? Could it be it was broken along the way? If you think it is just broken, no problem, I will look into it further, but I just want to hear your feedback before I spend time on it.

pauldendulk:
Some other thoughts that I would like to have feedback on:
- In my branch I had GetMapInfo and GetMapInfoAsync at some point. This resulted in a compiler warning for GetMapInfo, saying that it was a blocking operation. This makes no sense in our scenario, but we got the warning because the similarity in name suggests it is doing the exact same thing. This is not the case. Both return MapInfo but it is a different operation, calling a local or remote source, so we should make a distinction in the name. In my current version I renamed GetMapInfoAsync to GetRemoteMapInfoAsync. What are other options?
- The term 'featureInfo' was introduced because of the WMS GetFeatureInfo. In Mapsui we may have other server side sources of feature info. We will have an abstraction over such services. Since that is the case we can come up with our own name and could use the MapInfo term which we are using for so many other things.
- The RasterizingTileLayer and RasterizingTileSource are both local services, so perhaps they should not have an async way to fetch the MapInfo, just a two step way.

pauldendulk:
@inforithmics I am going to merge this to make all the changes related fetching MapInfo more stepwise. I am still interested in feedback. 

Here is how I now think about my own remarks:
> Do you know if there are any other samples that should work? Could it be it was broken along the way? If you think it is just broken, no problem, I will look into it further, but I just want to hear your feedback before I spend time on it.

I assume it is broken and will investigate why "RasterizingTileLayer with LineString" returns no result.

> In my branch I had GetMapInfo and GetMapInfoAsync at some point. This resulted in a compiler warning for GetMapInfo, saying that it was a blocking operation. This makes no sense in our scenario, but we got the warning because the similarity in name suggests it is doing the exact same thing. This is not the case. Both return MapInfo but it is a different operation, calling a local or remote source, so we should make a distinction in the name. In my current version I renamed GetMapInfoAsync to GetRemoteMapInfoAsync. What are other options?

I will continue to work with GetRemoteMapInfoAsync as name as long as I don't have a better name.

> The term 'featureInfo' was introduced because of the WMS GetFeatureInfo. In Mapsui we may have other server side sources of feature info. We will have an abstraction over such services. Since that is the case we can come up with our own name and could use the MapInfo term which we are using for so many other things.

I will not rename this for now, because if I do rename it to something with MapInfo in the name the function should also return MapInfo and this is a bigger change, which I don't want to make with the limited overview I have now. So, this will not change in V5.

> The RasterizingTileLayer and RasterizingTileSource are both local services, so perhaps they should not have an async way to fetch the MapInfo, just a two step way.

This is not quite right. The IProvider.GetFeaturesAsync method is called to get the feature info, and this could be remote info. The odd thing is that the tiles are drawn using the features, so these features were local at some point, but if we do not preserve them we need to go back to the source to fetch them again. This may need to be organized differently. This is not something I want to look into in V5.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -223,7 +222,7 @@ private void Render(object canvas, Viewport viewport, IEnumerable<IWidget> widge
         WidgetRenderer.Render(canvas, viewport, widgets, WidgetRenders, _renderService, layerOpacity);
     }
 
-    public MapInfo GetMapInfo(double x, double y, Viewport viewport, IEnumerable<ILayer> layers, int margin = 0)
+    public MapInfo GetMapInfo(ScreenPosition screenPosition, Viewport viewport, IEnumerable<ILayer> layers, int margin = 0)
```
pauldendulk, line 225: More use of ScreenPosition, I think this is clearly better.

Mapsui.Rendering.Skia/MapRenderer.cs
```diff
@@ -270,77 +268,49 @@ public MapInfo GetMapInfo(double x, double y, Viewport viewport, IEnumerable<ILa
                 for (var index = 0; index < mapInfoLayers.Count; index++)
                 {
                     var mapList = new List<MapInfoRecord>();
+                    list.Enqueue(mapList);
                     var infoLayer = mapInfoLayers[index];
-                    if (infoLayer is ILayerFeatureInfo layerFeatureInfo)
-                    {
-                        tasks.Add(Task.Run(async () =>
```
pauldendulk, line 276: The diff is confusing, but here I removed a large chunk of code that is solely about calling a remote service, not related to rendering.

Mapsui/RemoteMapInfoFetcher.cs
```diff
@@ -0,0 +1,86 @@
+﻿using Mapsui.Extensions;
```
pauldendulk, line 1: This is a new class which does the remote fetching. It is good that this can be separated out of the rest. The class is not easy to find for users, but most of the time they don't need to, because they well call the wrapper in the MapInfoEventArgs.

pauldendulk (commented):
Add a few comments on my own code.

PR #2837, opened 2025-01-03
Get version from git tag

pauldendulk:
This adds a ps1 script and changes to the Directory.Build.props to get the git tag version (which we use for the version of our releases). The version is now available in our code and scripts. This way we can show the version in the logs:

image

In the Directory.Package.props we also specify a Mapsui version. Perhaps we can remove it there too.


pauldendulk:
Locally it works, on the build server the three platforms have three different errors 😕

inforithmics:
My 5 second guesses are.

1. Linux is missing Powershell or the path syntax is wrong / instead of \ on Windows.
2. The Mac build Fails because the version is somehow duplicated meaning either two times the version variable or something similar. Maybe the version is appended instead of overwriting the version.
3. The Windows build fails because more than one process writes the version string. Solution to make try catch around writing ( probably a retry isn’t needed because both processes write the same information.

inforithmics:
I merged main into the Branch and fixed 2 merge conflicts in the Directory.Build.props file lets see if it works better now.

inforithmics:
A delete of version_from-git_tag.txt and then a rebuild triggers the problem too.
I think this line is execucted in parallel to the writing of the file.
` <VersionFromGitFile>$(MSBuildThisFileDirectory)\version-from-git-tag.txt</VersionFromGitFile>`

Don't know if there is a target before prepareforbuild
` <Target Name="SetVersionFromFile" BeforeTargets="PrepareForBuild" Condition="">`

ChatGpt's opinion to that is.

In MSBuild, property evaluation happens before any targets are executed. In other words, the MSBuild engine must read and evaluate all properties (and items) before it can even decide which targets to run. Therefore, there is no built-in target that executes before properties are read—property evaluation is part of the MSBuild engine’s initial project evaluation phase, not part of the target execution phase.

So this means the first time the writing of the property is to late.
Possible Solutions.

1. Retry Logic in Build.
2. Explicitly execute the Script before msbuild than it works too.
3. Reading directly the Tag in Msbuild (don't know if this is possible).

inforithmics:
Chat Gpt thinks this can be easier achieved by this
```
<PropertyGroup>
     <Version>unknown</Version>
</PropertyGroup>

<Target Name="GetGitTagVersion" BeforeTargets="BeforeBuild"> 
  <Exec 
      Command="git describe --tags --abbrev=0"
      WorkingDirectory="$(MSBuildProjectDirectory)"
      IgnoreExitCode="true">
    <Output TaskParameter="ConsoleOutput" PropertyName="Version"/>
  </Exec>
</Target>
```

pauldendulk:
@inforithmics Thanks for you input. I tried your suggestions but had no luck. My last attempt was to write the version to an envvar, but did not get it to work yet. My idea was to quickly fix this issue and I don't want to spend my time on this. I will close this PR and perhaps later on an easier fix will come to mind.

PR #2838, opened 2025-01-03
Update: SQLitePCLRaw.bundle_green to 2.1.10

inforithmics:
Fixed this warning XA0141: NuGet package '<unknown>' version '<unknown>' contains a shared library 'libe_sqlite3.so' which is not correctly aligned (This warning is displayed in .NET 9 build)

Version 2.1.10 is aligned to 16Kb so this warning isn't displayed anymore

And .NET 9 WebAssembly support


pauldendulk (approved):
Simple fix, but a possibly nasty issue, great,

PR #2839, opened 2025-01-03
Warning Fixes: Uno\Maui Samples\Tests

inforithmics:


pauldendulk (approved):
Right, i did notice this problem. Instead of suppressing it I tried to fix it properly but failed, and never got around to suppressing it. The dispose issue are now the biggest problem we still have in V5.

PR #2840, opened 2025-01-04
Add BaseEventArgs to share code between WidgetEventArgs and MapInfoEventArgs

pauldendulk:
- The GetRemoteMapInfoAsync method is now available in WidgetEventArgs as well.
- These changes are also related to earlier changes to the widget events. For widget we support different events like Tapped, Release, Moved, Pressed. It is now logical to add similar events for basic map manipulation. So, the MapInfo event may have to be replaced by events for all the gestures we support. I makes sense to add this in V5. But this is a bit more work again.


pauldendulk:
Will merge this and continue work here: #2841

PR #2841, opened 2025-01-04
Remove ILayer.IsMapInfoLayer

pauldendulk:


Mapsui.Nts/Editing/EditManipulation.cs
```diff
@@ -10,7 +10,7 @@ public static class EditManipulation
 {
     public static bool OnPointerPressed(ScreenPosition screenPosition, EditManager editManager, IMapControl mapControl)
     {
-        var mapInfo = mapControl.GetMapInfo(screenPosition, editManager.VertexRadius);
```
pauldendulk, line 13: The EditManipulation class will get an update for the changes in a later PR

Mapsui.Nts/Widgets/EditingWidget.cs
```diff
@@ -18,14 +18,14 @@ public EditingWidget(IMapControl mapControl, EditManager editManager)
     }
 
     public override bool OnPointerPressed(Navigator navigator, WidgetEventArgs e)
-        => EditManipulation.OnPointerPressed(e.Position, _editManager, _mapControl);
+        => EditManipulation.OnPointerPressed(e.ScreenPosition, _editManager, _mapControl);
```
pauldendulk, line 21: Rename EventArgs.Position to ScreenPosition, better now we also have EventArgs.WorldPosition

Mapsui.UI.MapView/MapView.cs
```diff
@@ -29,9 +29,9 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
     private const string _calloutLayerName = "Callouts";
     private const string _pinLayerName = "Pins";
     private const string _drawableLayerName = "Drawables";
-    private readonly ObservableMemoryLayer<Callout> _mapCalloutLayer = new(f => f.Feature) { Name = _calloutLayerName, IsMapInfoLayer = true };
-    private readonly ObservableMemoryLayer<Pin> _mapPinLayer = new(f => f.Feature) { Name = _pinLayerName, IsMapInfoLayer = true };
-    private readonly ObservableMemoryLayer<Drawable> _mapDrawableLayer = new(f => f.Feature) { Name = _drawableLayerName, IsMapInfoLayer = true };
+    private readonly ObservableMemoryLayer<Callout> _mapCalloutLayer = new(f => f.Feature) { Name = _calloutLayerName };
```
pauldendulk, line 32: For as far is I can see there are four layers in the MapView with IsMapInfoLayer = true. I added them to MapInfoLayers and use that in the query. Not sure how that will work with user layers set to IsMapInfoLayers = true. This may need a more specific fix if I understand the problem.

Mapsui/BaseEventArgs.cs
```diff
@@ -1,16 +1,21 @@
...
+using Mapsui.Manipulations;
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace Mapsui;
 
-public class BaseEventArgs(ScreenPosition screenPosition, MPoint worldPosition, TapType tapType,
-    Func<MapInfo> getMapInfo, Func<Task<MapInfo>> getRemoteMapInfoAsync) : EventArgs
+public delegate Task<MapInfo> GetRemoteMapInfoAsyncDelegate(ScreenPosition screenPosition, Viewport viewport, IEnumerable<ILayer> layers);
```
pauldendulk, line 9: BaseEventArgs now contains a lot of the functionality that we want everywhere. This will make it easier to introduce the Map events.

Mapsui/Widgets/InfoWidgets/MapInfoWidget.cs
```diff
@@ -1,18 +1,50 @@
...
+    {
+    }
+
+    /// <summary>
+    /// Widget displaying information about the feature at the current mouse position
+    /// </summary>
+    /// <param name="map">The map that is queried.</param>
+    /// <param name="layersFilter">The filter to select the layers to query. The advantage of a filter is that 
+    /// it can handle changes to the layer list later on.</param>
+    public MapInfoWidget(Map map, Func<ILayer, bool> layersFilter)
```
pauldendulk, line 33: Now the MapInfoWidget needs to a way to specify the layers. There are three constructor overloads. This one where you can specify a filter, this is often just a little code and works better of the layer list changes later.

Samples/Mapsui.Samples.Common/Maps/Animations/ViewportCenterOnAnimationSample.cs
```diff
@@ -30,9 +30,8 @@ public static Map CreateMap()
 
         map.Info += (s, e) =>
         {
-            var mapInfo = e.GetMapInfo();
             // Animate to the new center.
-            map.Navigator.CenterOn(mapInfo.WorldPosition, 500, Easing.CubicOut);
+            map.Navigator.CenterOn(e.WorldPosition, 500, Easing.CubicOut);
```
pauldendulk, line 34: In many situations we just need to the Screen or WorldPosition and no feature info. This is now directly available on the EventArgs.

Samples/Mapsui.Samples.Common/Maps/Animations/ViewportFlyToAnimationSample.cs
```diff
@@ -26,9 +26,8 @@ public static Map CreateMap()
 
         map.Info += (s, e) =>
         {
-            var mapInfo = e.GetMapInfo();
             // 'FlyTo' is a specific navigation that moves to a new center while moving in and out.
-            map.Navigator.FlyTo(mapInfo.WorldPosition, mapInfo.Resolution * 1.5, 500);
+            map.Navigator.FlyTo(e.WorldPosition, e.Viewport.Resolution * 1.5, 500);
```
pauldendulk, line 30: The Resolution is another thing we retrieved from the MapInfo. Now the Viewport (immutable) is also part of the EventArgs.

Samples/Mapsui.Samples.Common/Maps/DataFormats/WfsSample.cs
```diff
@@ -32,7 +32,7 @@ public async Task<Map> CreateMapAsync()
             map.Layers.Add(CreateWfsLayer(provider));
             map.Layers.Add(CreateLabelLayer(provider));
 
-            map.Widgets.Add(new MapInfoWidget(map));
+            map.Widgets.Add(new MapInfoWidget(map, (l) => Name == "Districts"));
```
pauldendulk, line 35: There are many options to determine which layers need MapInfo. Perhaps we will develop some preferred habits.

Samples/Mapsui.Samples.Common/Maps/Info/CustomCalloutStyleSample.cs
```diff
@@ -37,17 +39,18 @@ public static Map CreateMap()
...
+        map.Info += (s, e) => MapOnInfo(s, e, calloutLayer);
 
         return map;
     }
 
-    private static void MapOnInfo(object? sender, MapInfoEventArgs e)
+    private static void MapOnInfo(object? sender, MapInfoEventArgs e, ILayer calloutLayer)
     {
-        var feature = e.GetMapInfo().Feature;
+        var feature = e.GetMapInfo([calloutLayer]).Feature;
```
pauldendulk, line 53: The PR contains a lot of changes in the samples where we now have to specify the MapInfo layers.

Samples/Mapsui.Samples.Common/Maps/Info/GeoJsonInfoSample.cs
```diff
@@ -45,12 +45,8 @@ public static Map CreateMap()
...
 
         map.Layers.Add(Tiling.OpenStreetMap.CreateTileLayer());
-        map.Layers.Add(new RasterizingTileLayer(CreateCityLabelLayer(dataSource))
-        {
-            IsMapInfoLayer = true,
-        });
-
-        map.Widgets.Add(new MapInfoWidget(map));
+        map.Layers.Add(new RasterizingTileLayer(CreateCityLabelLayer(dataSource)));
+        map.Widgets.Add(new MapInfoWidget(map, l => l is RasterizingTileLayer));
```
pauldendulk, line 49: Here the type is used to determine the MapInfo layers.

Samples/Mapsui.Samples.Common/Maps/Info/ImageCalloutSample.cs
```diff
@@ -26,23 +27,26 @@ public class ImageCalloutSample : ISample
...
         map.Layers.Add(OpenStreetMap.CreateTileLayer());
-        map.Layers.Add(CreatePointLayer());
+        var pointLayer = CreatePointLayer();
+        map.Layers.Add(pointLayer);
         map.Navigator.CenterOnAndZoomTo(map.Layers.Get(1).Extent!.Centroid, map.Navigator.Resolutions[5]);
 
-        map.Widgets.Add(new MapInfoWidget(map));
-        map.Info += MapOnInfo;
+        map.Widgets.Add(new MapInfoWidget(map, l => l.Name == _pointLayerName));
+        map.Info += (s, e) => MapOnInfo(s, e, pointLayer);
```
pauldendulk, line 42: Here the pointer to the layer is passed to the Info event. Not a lot of code, but if we could just use a filter (like in the MapInfoWidget) it may be easier. This is not possible currently because we have no access to the layer list at all. Should we add more to the EventArgs. Perhaps even the entire Map object?

Samples/Mapsui.Samples.Common/Maps/Performance/ManyMutatingLayersSample.cs
```diff
@@ -42,7 +43,7 @@ public Task<Map> CreateMapAsync()
         map.Layers.Add(CreatePointLayers(_random, features).ToArray());
         map.Navigator.ZoomToBox(map.Layers.Get(0).Extent);
 
-        map.Widgets.Add(new MapInfoWidget(map));
+        map.Widgets.Add(new MapInfoWidget(map, l => l is not TileLayer));
```
pauldendulk, line 46: You can also specify the MapInfo layers be excluding one type of layer.

pauldendulk (commented):
Added some remarks to my own changes.

pauldendulk:
This is main PR in a series of PRs where we use methods to get MapInfo through the methods GetMapInfo and GetRemoteMapInfoAsync, and those methods have a parameter to specify which layers to query. Why is this change important?  
- ILayer.IsMapInfoLayer is a general way to say that information should be returned, but in different scenarios we want different things. We use MapInfo for: Toggling a callout, Showing info in the MapInfoWidget, Editing (where we don't show anything but just need the information of the feature to make changes), when hovering with the mouse over. We want to be able to just retrieve information that we need for that task.
- Fetching MapInfo can impact performance, by letting the user do the call it is more clear what is happening.
- MapInfo was previously tightly coupled to the Info event, but in V5 different types of events are introduced (Pressed, Released, Tapped, Moved). We want all those events to have the option to GetMapInfo but we do not want to let Mapsui do a GetMapInfo on every time such an event occurs. Especially for Moved (which could be just hovering) this could become a performance problem.

PR #2843, opened 2025-01-05
Fix bug in RasterizingTileLayer MapInfo

pauldendulk:
When tapping the line in the RasterizingTileLayer sample no info was shown and now there is.

### DIFF
--- a/Mapsui.Tiling/Provider/RasterizingTileSource.cs
+++ b/Mapsui.Tiling/Provider/RasterizingTileSource.cs
@@ -250,17 +250,18 @@ public class RasterizingTileSource : ILocalTileSource, ILayerFeatureInfo
         var renderLayer = layer.RenderLayer;
         List<ILayer> renderLayers = [renderLayer];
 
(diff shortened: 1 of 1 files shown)


pauldendulk:
@inforithmics My changes may have introduced the bug, it is fixed now but I am still not quite sure how it should work. What happened is that the result from the regular MapInfo was ignored, only the async MapInfo was used. Perhaps in an earlier stage this result contained both sync and async results. Because of the changes to the MapInfo these are now completely separate things. 

One thing not fully clear to me was if or why we needed the async request in the RasterizingTileLayer. Now looking at how it behaves I see the actually values come from the sync MapInfo. So, I think we should remove the async MapInfo request. If we find a case where it is needed we could think about it again.

pauldendulk:
@inforithmics I am going to merge this to include it in the next release. In an another PR I intent to remove the ILayerFeatureInfo from the RasterizingTileLayer because I don't see samples that depend on it.

PR #2844, opened 2025-01-05
Reduce dispose suppressions

pauldendulk:


pauldendulk:
If I search the codebase for "idisp0" I get 255 results. Apart from the 26 in .editorconfig all seem to be suppressions (when glancing over the results). The idea is to reduce those. Possible ways to achieve this:
- Remove the unnecessary suppressions (I just pushed one, so 254 left)
- Small local fixes (I just pushed such a fix, so 253 left)
- Reorganizing the samples. There should be an IDisposable sample.
- Big reorganization in Mapsui. I am thinking about making Map disposable.

Mapsui/Extensions/CacheExtensions.cs
```diff
@@ -22,31 +22,19 @@ public static async Task<byte[]> UrlCachedArrayAsync(this IUrlPersistentCache? p
...
-                else
-                {
-                    var handler = new HttpClientHandler();
-                    using var httpClient = new HttpClient(handler);
-                    // https://github.com/xamarin/xamarin-android/issues/5264 use ConfigureAwait(false) for Network access
-                    response = await httpClient.GetStreamAsync(url).ConfigureAwait(false);
-                }
-#pragma warning restore IDISP001
 
+            if (loadUrl != null)
```
pauldendulk, line 26: @inforithmics I rewrote this part of the code to remove the suppressions. DisposeAsync is now called automatically because of the `await using` statement, see https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#the-disposeasync-method. Whether `.ConfigureAwait(false);` is used depends on the implementation of that method, to keep our own code simpler I think we should assume it does, and this will probably be the case for all Microsoft code.

Mapsui.Nts/Providers/Shapefile/DbaseReader.cs
```diff
@@ -309,6 +309,7 @@ public DataTable GetSchemaTable()
     {
         var tab = new DataTable();
         // all of common, non "base-table" fields implemented
+#pragma warning disable IDISP004 // Don't ignore created IDisposable. Justification: The DataTable is returned to the caller.
```
pauldendulk, line 312: Here is move the global suppressions to 'in code'. I think this is what I now prefer. Global suppression has the problem that refactoring code reverts it. And perhaps it is better to be aware of the suppression by having it in code.

Mapsui.Nts/Providers/Wfs/Utilities/HttpClientUtil.cs
```diff
@@ -113,45 +113,17 @@ public void AddHeader(string name, string value)
...
-            throw;
-        }
-        catch (NotSupportedException ex)
-        {
-            Logger.Log(LogLevel.Error, "An exception occurred while initializing a request to " + _url + ": " + ex.Message, ex);
-            throw;
-        }
-
-        httpClient.Timeout = new TimeSpan(0, 0, 1, 30);
+            using var httpClient = new HttpClient(httpClientHandler);
```
pauldendulk, line 118: Here the httpClient was not disposed. This was the case in a number of situations. It is better to use a single instance, but not disposing it is not the solution.

Mapsui.Nts/Providers/Wfs/Utilities/HttpClientUtil.cs
```diff
@@ -169,13 +141,37 @@ public void AddHeader(string name, string value)
...
+            throw;
+        }
         catch (Exception ex)
         {
             Logger.Log(LogLevel.Error, "An exception occurred during a HTTP request to " + _url + ": " + ex.Message, ex);
             throw;
         }
     }
 
+    private async Task<HttpResponseMessage> GetWebResponseAsync(HttpClient httpClient)
```
pauldendulk, line 162: The HttpResponseMessage needs to be disposed, but because it can be created in two scenarios we could not use 'using', by moving it to a method there is only one assignment in the calling code so that we can use 'using' again.

Mapsui.Nts/Providers/Wfs/Xml/XPathQueryManager.cs
```diff
@@ -269,7 +269,8 @@ public async Task SetDocumentToParseAsync(HttpClientUtil httpClientUtil)
     {
         try
         {
-            InitializeXPathObjects(await httpClientUtil.GetDataStreamAsync());
+            using var stream = await httpClientUtil.GetDataStreamAsync();
+            InitializeXPathObjects(stream);
```
pauldendulk, line 273: The suppressed warning was about not disposing injected. In this case it was not very hard to dispose it on the outside.

Mapsui.ArcGIS/DynamicProvider/ArcGISLegend.cs
```diff
@@ -72,19 +72,17 @@ public void GetLegendInfoRequest(string serviceUrl, ICredentials? credentials =
     {
         var uri = CreateRequestUrl(serviceUrl);
         var data = _urlPersistentCache?.Find(uri);
-        Stream stream;
         if (data == null)
         {
             using var httpClient = CreateRequest(credentials);
             using var response = await httpClient.GetAsync(uri).ConfigureAwait(false);
-            stream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
+            using var stream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
```
pauldendulk, line 79: This just needs to be disposed.

Mapsui.ArcGIS/DynamicProvider/ArcGISLegend.cs
```diff
@@ -72,19 +72,17 @@ public void GetLegendInfoRequest(string serviceUrl, ICredentials? credentials =
...
             data = StreamHelper.ReadFully(stream);
             _urlPersistentCache?.Add(uri, data);
             await stream.DisposeAsync();
         }
 
-        stream = new MemoryStream(data);
-        _legendResponse = GetLegendResponseFromWebResponse(stream);
+        _legendResponse = GetLegendResponseFromWebResponse(data);
```
pauldendulk, line 85: Moving the stream to the method to make it easier to read.

pauldendulk:
After my last commit there are 236 suppressions left, still a lot. The main purpose of this PR was going through the warnings and getting an overview. There are a couple of categories:

- Skia objects being cached. Sometime in a cache, in which case the ownership is the cache, and it should be good. In other cases we keep a skia object in a class. In those cases I would prefer to just dispose it, since we are not sure if there is a significant performance improvement and we need to reduce the disposable exceptions.
- Layers. These generally live as long as the app. So often not a problem. I have no good solution in mind yet. I think this will not be solved in V5.
- Providers. Same thing.
- MapControl related. I would like to make these non-disposable. This will require some detailed work.
- Samples. I think there should a IDisposableSample.

pauldendulk:
The disposal of HttpClient is an issue. The advise is to create one instance in a class that is instantiated only once during the apps lifetime. This instance would still need to be disposed. And since we are writing a library we do not really know how our classes are used, they could be re-instantiated many times. In aspnet the HttpClientFactory is used. We do not have access to this in our library. I am now thinking of creating our own simple static HttpClientFactory.

pauldendulk:
@inforithmics This commit is relevant for things that you worked on. It is a fairly big change (considering this PR is just to reduce the dispose suppressions). Instead of GetStreamAsync GetBytesAsync is used. This way the method can convert to bytes and can dispose the stream right away, making the calling code simpler. 

A potential downside is that it is slower because sometimes you may need to convert to byte array and back to MemoryStream for parsing. However, the cache was already stored as bytes, so we already had to convert to bytes and then return a MemoryStream version again. So, where caching was used it is now faster or the same. In the case of feature info (where we did not cache) we now do an extra conversion, but I am happy to accept this if it makes the disposing story simpler.

Mapsui.Nts/Providers/Shapefile/DbaseReader.cs
inforithmics: Good idea

Mapsui.Nts/Providers/Wfs/Utilities/HttpClientUtil.cs
inforithmics: Disposing here is problematic because the Stream is given outside the Method that means the Httpclient is disposed and the data is still read from the Stream, This could cause an already disposed exception or something similar.

I think the solution is to Create a WrappedStream or whatever it should be called, where the Stream and the httmplcient are added and disposed when the Wrapped Stream is disposed.

pauldendulk: You are right. I assumed the byte array was first read and added to the cache, but if the cache is null the http stream is returned directly. 

I've spend some time thinking about the way to go about this. The fix I pushed is to convert the stream to byte array and return the byte array in a memory stream. The downside is that this is not good for performance. The upside is that there is no dispose needed at all, making the code simpler.

Alternatives:
- Leave it like it was before. This is a real problem however, we create many HttpClient instances and never dispose it.
- Use a HttpClientFactory. I think we need that, but it would make the scope of this PR much bigger. 
- Make the HttpClient an instance of the HttpClientUtil and dispose it from the classes using it. De upside is that you would use less HttpClient instances, which is advised. The downside is that we would have another disposable object which would make all users disposable. A better solution is the HttpClientFactory. Also, this was quite some work. The best solution would be to make the HttpClientUtil immutable (remove the four properties and pass them in the request or constructor).

PR #2854, opened 2025-01-18
Use Metal on iOS

inforithmics:
Use Metal on iOS instead of OpenGL Because OpenGL is depreceted on iOS since version 12


pauldendulk (approved):
Great!

PR #2856, opened 2025-01-24
Make styleRenderers and widgetRenderers static 

pauldendulk:
You can now register anywhere with:
```csharp
MapRenderer.RegisterStyleRenderer(typeof(CustomStyle), new SkiaCustomStyleRenderer());
```
One of our aims is to make the samples copy-pastable. This is one step in that direction. You can do the registering in the sample class itself.


pauldendulk:
I also update the upgrade guide for the breaking change.

PR #2857, opened 2025-01-24
Add RulerWidget

pauldendulk:
Add RulerWidget with which you can drag a line to see the distance of the line.

image


pauldendulk:
Some future improvements that I am not going to work on right now:
- Add a button to activate or deactivate the ruler (optional). I think in most cases you need to use a platform specific button to enable it, and currently you have to.
- Always show the info on the far end of the ruler, so if you are dragging to the bottom left the info should be to the bottom left of the mouse/finger position.

PR #2862, opened 2025-02-01
Improve logging for logging widget

pauldendulk:
Some things I ran into while investigating an issue. Especially wrt a com exception which had no message.

### DIFF
--- a/Mapsui.Rendering.Skia/SkiaWidgets/LoggingWidgetRenderer.cs
+++ b/Mapsui.Rendering.Skia/SkiaWidgets/LoggingWidgetRenderer.cs
@@ -83,8 +83,7 @@ public class LoggingWidgetRenderer : ISkiaWidgetRenderer, IDisposable
                 _ => _informationTextPaint,
             };
 
-            canvas.DrawText(entry.LogLevel.ToString(), (float)(marginX + paddingX), (float)(marginY + (paddingX * line) + loggingWidget.TextSize * (line + 1)), _font, paint);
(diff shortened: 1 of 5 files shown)


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -190,7 +190,7 @@ private void InvalidateTimerCallback(object? state)
         }
         catch (Exception ex)
         {
-            Logger.Log(LogLevel.Error, ex.Message, ex);
+            Logger.Log(LogLevel.Error, $"Error in render loop", ex);
```
pauldendulk, line 193: If the exception has no message there will be no message at all in our logs. It is the responsibility of the logger implementation to log the exception message.

Mapsui.Rendering.Skia/SkiaWidgets/LoggingWidgetRenderer.cs
```diff
@@ -83,8 +83,7 @@ public void Draw(SKCanvas canvas, Viewport viewport, IWidget widget, RenderServi
                 _ => _informationTextPaint,
             };
 
-            canvas.DrawText(entry.LogLevel.ToString(), (float)(marginX + paddingX), (float)(marginY + (paddingX * line) + loggingWidget.TextSize * (line + 1)), _font, paint);
-            canvas.DrawText(entry.Description, (float)(marginX + paddingX + _levelWidth + 2 * paddingX), (float)(marginY + (paddingY * line) + loggingWidget.TextSize * (line + 1)), _font, paint);
+            canvas.DrawText(entry.FormattedLogLine, (float)(marginX + paddingX), (float)(marginY + (paddingX * line) + loggingWidget.TextSize * (line + 1)), _font, paint);
```
pauldendulk, line 86: Draw the text with just one line in the log renderer. The logic to create the logline is now in the widget itself.

PR #2871, opened 2025-02-09
Fix the 4.1 build

pauldendulk:


pauldendulk:
I used this branch to publish the docs and samples to mapsui.com. I had to remove all the net7 platforms to get this to compile this with the net9 sdk. I won't merge this into develop/4.1 for now. If we need to release a patch for 4.1 perhaps we will need to merge it, but I will think about that then.

Issue #2704, opened 2024-07-08
WPF: Shape layer performance issue


EmaDozer:
**Mapsui Version**
4.1.7

**Mapsui Platform**
WPF

**Device**
Windows

**Describe the bug**
Relative small shapes cause the map to be laggy. I use RasterizingTileLayer with SqlitePersistentCache and without projection between map and shape.

I attached a sample project: 
[ShapeLayer.zip](https://github.com/user-attachments/files/16129097/ShapeLayer.zip)

Method to create shape layer:

` private static ILayer CreateShapeLayer()
 {
     var shapeFile = new Mapsui.Nts.Providers.Shapefile.ShapeFile(
        Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "shapes", "EZG_KB_LM.shp"), false)
     { CRS = "EPSG:3857" };

     var sqlitePersistentCache = new SqlitePersistentCache("shapes");
     sqlitePersistentCache.Clear();

     var layer = new Layer
     {
         Name = "Test",
         DataSource = shapeFile,
         Style = CreateVectorThemeStyle(),
     };

     return new RasterizingTileLayer(layer, persistentCache: sqlitePersistentCache);
 }`

**To Reproduce**
Open the sample project and zoom to the shape.

EmaDozer:
I tried it also without the persistentCache, made not really a difference.

EmaDozer:
I added another sample where you can enable 2 different layers (size 1MB and 3.6MB). Currently its hard to use Mapsui with shapes on WPF or maybe i am missing something fundamental?

zip:
[Uploading shape perfromance issue.zip…]()

video:
https://github.com/user-attachments/assets/0122be6d-140d-4f18-9389-9c686ac62e6f

ChrisSC-dev:
Is there any news on this issue? I would be interested in the solution as well. I also note the map to be laggy when shapes (even really small <1mb ones) are loaded with Mapsui in WPF...

EmaDozer:
There is no update yet. Performance is still a problem. Can anyone put us on the right track? Maybe there is a starting point/method to investigate the problem. I would be honored if I could contribute to this issue. At the moment I haven't figured out where to start. I've used the Visual Studio Performance Profiler, but that hasn't been helpful so far.

pauldendulk:
@EmaDozer  Thanks a lot for the bug report. The minimal sample helps a lot. I will try to look into this next Friday.

How I would investigate this.
[...]

PR #2879, opened 2025-02-14
Added HugeShapeFileWithSimplifierAndClippingSample for #2704 v3

EmaDozer:
related to WPF: Shape layer performance issue https://github.com/Mapsui/Mapsui/issues/2704

I added the HugeShapeFileWithSimplifierAndClippingSample rendering the two shapes. Additionally i created the GeometrySimplifyAndClippingProvider to demontrate the effect of simplify and clip the geometry/shape. This is mainly important fo the EZG_KB_LM.shp which is more or over one huge polygon with 33k points. The RasterizingLayer does good work here as well, in my application however, applying the clip by the provider leads to better fps.

With RasterizingTileLayer the fps drops.

Maybe simplifying (and clipping?) should be added to the ShapeFile class by default without the need to apply the provider by the user. A flag could indicate if the simplification should be applied.

I think also creating a shapeCache in the ShapeFile class is beneficial. But it is not part of this merge request.


inforithmics:
I had a lock at the commiting Email Adresses and there are two EmaDozer Email Adresses a real one and a @norpley one, so for both email adresses the cla needs to be signed, I guess the Real Email Adress is not a Github User yet. The other solution would be to commit with only the github User Email Adress with @norepley ....

EmaDozer:
The noreply email is automatically created by GitHub which is connected to my real address. I was expecting, that the CLA recognize this relationship. I really dont know how i can sign the CLA with the no reply email. Can i explicitly specify the email when signing the CLA? 

'''Because you have email privacy enabled, e.doeser@posteo.de will be used for account-related notifications as well as password resets. 93271991+EmaDozer@users.noreply.github.com will be used for web-based Git operations, e.g., edits and merges.'''

I can just sign the CLA without speciying the email. I would expect that the primary email address selected in the sttings is used for this assignment and i cant select the no reply as my primary address or add it as an addidional address.

inforithmics:
I ask an AI and this is the reponse:
Solution
Link Emails in GitHub Settings
Ensure your real email and GitHub-provided noreply email are both added and verified in your GitHub account:

Go to GitHub Settings → Emails (https://github.com/settings/emails).

Add both your real email and the noreply email (if not already listed).

Verify all emails.

Enable "Keep my email address private" to retain the noreply address for public commits.

This allows GitHub to associate both emails with your account.

Configure Git Locally
Update your local Git configuration to use the same email used to sign the CLA (your real email):

bash
Copy
git config --global user.email "your-real-email@example.com"
If you prefer to keep using the noreply email, ensure it is added to your GitHub account (as in Step 1). Future commits will then link properly.

Re-sign the CLA (if needed)
If the CLA still isn’t recognized, sign it again using the same email configured in your Git client (either your real email or the verified noreply address).

Ask Maintainers for Help
If the issue persists, repository maintainers can:

Whitelist the noreply email in the CLA tool.

Use the CLA assistant’s "whitelist" feature for contributors.

Check logs for specific mismatched emails.

Why This Works
GitHub links commits to your account based on verified emails. By adding both your real email and noreply address to your GitHub account, the CLA assistant can recognize contributions from either address once the CLA is signed with one of them.

EmaDozer:
Is the license/cla check still needed? 

@inforithmics I did all the necessary steps but somehow it does not work.

pauldendulk:
The problem is that you used two different accounts in your commits.

image

To resolve this your could have squashed the commits into one and created a new PR, but never mind, it is not that important.

pauldendulk:
I checked out your branch to inspect the performance. It is not different from what I expected. In general performance can be improved, but that is on ongoing thing that will get more attention after the release of V5.

What surprised me in your comment was this:
> With RasterizingTileLayer the fps drops.

But I guess that is compared to clipping, not compared to no performance optimization?

pauldendulk:
Perhaps you are primarily referring to the lagginess that you see when zooming in deeper and deeper. Panning a bit while you are there is fast, but that is not what you are doing most of the time. 

At least one cause we know is that we generate the SKImage on the render thread. What we need to do is create the skia objects (like SKImage) on a background thread and in the render loop only draw those.

pauldendulk:
This will improve the RasterTileLayer performance https://github.com/Mapsui/Mapsui/pull/2947

pauldendulk:
Moved the code in this PR into this PR: https://github.com/Mapsui/Mapsui/pull/3228

PR #2886, opened 2025-02-23
CancellationToken Implementation for FetchMaschine

inforithmics:
Part of this PR https://github.com/Mapsui/Mapsui/pull/2564
Implment property the Stop for the Fetchmachine, by using a CancellationToken


pauldendulk (commented):
Thanks for the PR. It makes sense to add a cancellation token, but I am hesitant to make the FetchMachine IDisposable because I am currently trying to make less things IDisposable. 

There may be other solutions, but what I really would like to do is to remove the FetchMachine for each layer and replace it with a central fetch machine owned by the Map. The data fetch requests could pass that in a parameter. I don't have everything clear yet, so I have to play around with it. I hope to have time for it in the coming weeks.

PR #2889, opened 2025-02-23
fixing maui quickstart (.SkiaSharp() instead of .SkiaSharp(true))

inforithmics:


pauldendulk (approved):
ah yes, and the getting started mds do not exist anymore. 

Something else I wanted to change in the gettings started is to use a command line instead of the Nuget Manager commands.

PR #2891, opened 2025-02-23
Blazor Server Rendering

inforithmics:
I'm trying here to enable Server rendering for Blazor. This solves the Problem that on certain Platforms only Server Rendering is possible.
The idea is that when the control is rendered on the server I add an Image that is filled on the server and only updated when the image content is changed. I render it with webp because this as less data than png, I'll investigate if png is faster later.

When this works I'll test it how it works in the Hybrid mode.

- [ ] Renders the same as the Blazor Control (Widgets ...)
- [ ] Mouse Interaction works


pauldendulk:
Hmm, so actually rendering the image on the server and sending it over blazors signalr to the client? I worked on a project where we used leaflet.js like that (by accident) and it was super slow. But then if you optimize for this scenario it may be an good enough solution where there is no other option.

What I am personally interested in is to use it in the InteractiveWebAssembly render mode https://learn.microsoft.com/en-us/aspnet/core/blazor/components/render-modes?view=aspnetcore-9.0. But perhaps that is already possible with our current component.

inforithmics:
It is still a work in Progress. I have to provide the imageinfo in the snapshot function or adjust the Viewport area.
I know that it is Slow it is mainly for static Maps or an initial render in the hybrid Mode.

pauldendulk:
Closing this because it has been in draft for a long time now.

PR #2894, opened 2025-02-25
Add 'svg-content://' and 'base64-content://' schemes as options for Image.Source

pauldendulk:


pauldendulk:
Some follow ups on the things I said [here](https://github.com/Mapsui/Mapsui/pull/2725#issuecomment-2650018346) and [here](https://github.com/Mapsui/Mapsui/pull/2725#issuecomment-2679519040).

I ended up implementing a form of option 2 - generate an id for each imageSource. The solution has to guarantee that there is a key for all imageSources in a way to is not very error prone. The best solution that I could find was to create an Image class to replace the ImageSource. The id (Image.SourceId) is now set the in the constructor of the Image class. 

Introducing the Image class happens to be in line with what was suggested in option 3 and has additional benefits. 
All the fields that are related to the Image could now be moved to the new Image class (SvgFillColor, SvgStrokeColor, BitmapRegion, BlendModeColor. We currently have Images in four places:
- SymbolStyle
- Brush
- ImageWidget
- Callout

Those fields now become available on all four (although they are not yet used on all).

Currently there is one Image class but it may be better to work with different types of image classes. For instance, to distinguish between Svg and Bitmap. The SymbolTypes  Ellipse, Rectangle and Triangle should perhaps be in it's own image class. We may also add a specific class to support a future [CustomSymbolStyleRenderer](https://github.com/Mapsui/Mapsui/issues/2895).

Mapsui/Styles/ImageFetcher.cs
```diff
@@ -6,23 +6,34 @@
...
 using System.Threading.Tasks;
 
 namespace Mapsui.Styles;
 public static class ImageFetcher
 {
+    public const string EmbeddedScheme = "embedded";
+    public const string FileScheme = "file";
+    public const string HttpScheme = "http";
+    public const string HttpsScheme = "https";
+    public const string SvgContentScheme = "svg-content";
```
pauldendulk, line 19: Renamed to svg-content and base64-content. I think it is clearer like that.

Mapsui/Styles/Image.cs
```diff
@@ -0,0 +1,72 @@
...
+    /// source image. Note that each different color used will add an new object to the image cache.
+    /// </summary>
+    public Color? SvgStrokeColor { get; set; }
+
+    /// <summary>
+    /// This allows for the automatic conversion of a string to an Image object. This was added to make the creation 
+    /// code simpler.
+    /// </summary>
+    /// <param name="source"></param>
+    public static implicit operator Image(string source) => new() { Source = source };
```
pauldendulk, line 47: This implicit case operator allows for automatic conversion from a string (an image source) to an Image. This allows for more compact code and made it easy for me to convert the ImageSource string to the Image class. But if we will introduced different Image types this will probably have to go.

pauldendulk:
Follow ups:
- Introduce separate image classes.
- Make use of all Image field in all four IHasImage classes.
- Add CustomSymbolStyleRenderer.
- Remove Equals implementations for all styles, they have never been completely implemented.

pauldendulk:
Merging. Thanks @charlenni and @inforithmics for your contributions!

PR #2901, opened 2025-03-03
Work on solution to make Image rendering en SymbolType rendering more similar

pauldendulk:


pauldendulk:
I think I will not merge this PR and go another route. Dont' introduce an IImage interface. Introduce an ImageStyle with the Image class. Leave the SymbolTypes on the SymbolStyle but remove the SymbolType.Image option. Move all fields that can be applied to all point symbols to a separate base class that both ImageStyle and SymbolStyle can inherit from.

pauldendulk:
This is replaced with this: https://github.com/Mapsui/Mapsui/pull/2904

PR #2905, opened 2025-03-05
Add Mercator to Projection

Liklainy:
* Add `Mercator` to `Projection` mapping as `EPSG:3395` (ref https://gis.stackexchange.com/a/259334)
* Unify `Mercator` interface similar to `SphericalMercator`


Liklainy:
Thank you for quick and extensive reply!

I'm sure that `Mercator` usage was not intended in `ShapefileTileSample`, because crs for projection and point was different.
I understand your concern about updating the image, so in my new attempt I did it by adjusting coordinates for `SphericalMercator`

* Reverted the image, and adjusted coordinates for `SphericalMercator` in this test
* Added `Mercator` precision test similar to `SphericalMercator`
* Added `EPSG:3395` to `ProjectionTests`

Probably we should consider renaming `Mercator` class?

Tests/Mapsui.Tests/Projections/MercatorTests.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        // Act
+        var (x, y) = Mercator.FromLonLat(inLon, inLat);
+        var (outLon, outLat) = Mercator.ToLonLat(x, y);
+
+        // Assert
+        var distanceInKilometer = Haversine.Distance(inLon, inLat, outLon, outLat);
+        var distanceInCentimeter = distanceInKilometer * 100000;
+        ClassicAssert.Less(distanceInCentimeter, 1);
+    }
```
pauldendulk, line 26: 👍, btw, the ClassicAssert that we use almost everywhere needs to be replaced with the modern syntax, which in this case would be: Assert.That(distanceInCentimeter, Is.LessThan(1));

Tests/Mapsui.Tests/Projections/ProjectionTests.cs
```diff
@@ -44,8 +44,9 @@ public void MultiLineStringCoordinatesTest()
         ClassicAssert.AreEqual(expectedCoordinateCount, enumeration.Count());
     }
 
-    [Test]
-    public void CoordinateProjectionTest()
+    [TestCase("EPSG:3857")]
+    [TestCase("EPSG:3395")]
+    public void CoordinateProjectionTest(string crs)
```
pauldendulk, line 49: Good use of TestCase, the original test's assert is not great. I would have been better if it transformed back and compared the distance, like in the test you added.

pauldendulk (approved):
Great, thanks, I'll merge it.

pauldendulk:
> Probably we should consider renaming `Mercator` class?

Yes, perhaps WorldMercator like on epsg.io? https://epsg.io/3395

PR #2916, opened 2025-03-15
Disable widget logging events by default

pauldendulk:


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -583,14 +583,30 @@ private void CommonDispose(bool disposing)
         _invalidateTimer = null;
     }
 
+    private bool OnWidgetTapped(ScreenPosition screenPosition, MPoint worldPosition, GestureType gestureType, bool shiftPressed)
```
pauldendulk, line 586: I just decided that whenever I do something with all events the Tapped event should go first, it is the most used, then PointerPressed, PointerMoved and PointerReleased.

Mapsui/Logging/Logger.cs
```diff
@@ -51,7 +51,8 @@ private static void DefaultLogging(LogLevel level, string message, Exception? ex
 
     public class LoggerSettings
     {
-        public static bool LogMapEvents { get; set; }
+        public bool LogMapEvents { get; set; }
```
pauldendulk, line 54: Logging of Map events and logging of Widget events is now default off and can be turned on with settings on the static Logger:
```
Logger.Settings.LogMapEvents = true;
Logger.Settings.LogWidgetEvents = true;

```

Issue #518, opened 2018-09-25
World Wrap


CliffAgius:
QUESTION:

Is there a way to make the Map Tile (I am using the world.mbtiles) wrap so that if I scroll left I can scroll around the world just like in GoogleMaps for example?

Thanks

Cliff.

pauldendulk:
Unfortunately this is not possible yet. It is a feature we would like to have but there are a number of other important things that will get priority. 

Related:
- https://github.com/BruTile/BruTile/issues/92
- https://stackoverflow.com/questions/46189038/xamarin-mapsui-how-do-i-make-the-map-roll-over-on-end-of-world

pauldendulk:
Closing this. The feature will come up later.

Sumn3rd:
Hoping this comes up later, would be neat as am mainly working next to the date line and there's a lot of white space. Currently using a TileLayer from a HttpTileSource which is working well though

pauldendulk:
I understand this is a very important feature. But it will not be part of V5. I am hoping for a faster release schedule so v6 does not have to wait that long.

trampster:
I need this as well, I live in New Zealand and it is perilously close to the edge.

molloyin:
Bump. Did #2920 bring world wrap capabilities?

PR #2920, opened 2025-03-17
Add CustomLayerRenderer

pauldendulk:
This is an early version of the CustomLayerRenderer. The idea is that you can assign a custom renderer to a layer. This way you could optimize for performance, or implement [world wrap](https://github.com/Mapsui/Mapsui/issues/518). 

Rendering is what we will focus on in Mapsui V6, but by adding custom rendering to V5 we can start preparing for V6 without introducing breaking changes.

There is also a downsides to this. Because renderers can now have completely different implementation some generic functionality may not work. I am sure it is worth it, but something to keep in mind.

image


pauldendulk:
I am going to merge this version because it is functional even though will probably need more improvements later.

PR #2924, opened 2025-03-18
Always show PerformanceWidget in de map in debug mode

pauldendulk:


Mapsui.Rendering.Skia/SkiaWidgets/PerformanceWidgetRenderer.cs
```diff
@@ -8,17 +8,20 @@ namespace Mapsui.Rendering.Skia.SkiaWidgets;
 
 public class PerformanceWidgetRenderer : ISkiaWidgetRenderer
 {
-    private readonly string[] _textHeader = { "Last", "Mean", "Frames", "Min", "Max", "Count", "Dropped" };
+    private readonly string[] _textHeader = { "Frames", "Last", "Mean", "Min", "Max", "Count", "Dropped" };
```
pauldendulk, line 11: Show Frames first because that is what I am looking at most of the time. I would also like to see the actual FPS Other values that I might be interested in are, the actual frames (I usually have to explain it FPS is the theoretical possible FPS).

Mapsui.Rendering.Skia/SkiaWidgets/PerformanceWidgetRenderer.cs
```diff
@@ -8,17 +8,20 @@ namespace Mapsui.Rendering.Skia.SkiaWidgets;
...
     private readonly string[] _text = new string[7];
 
     public void Draw(SKCanvas canvas, Viewport viewport, IWidget widget, RenderService renderService, float layerOpacity)
     {
         var performanceWidget = (PerformanceWidget)widget;
+        if (!performanceWidget.Performance.GetIsActive())
```
pauldendulk, line 17: If not active then don't calculate to avoid a possible impact on performance (I don't think it will make a big difference).

Mapsui.Rendering.Skia/SkiaWidgets/PerformanceWidgetRenderer.cs
```diff
@@ -8,17 +8,20 @@ namespace Mapsui.Rendering.Skia.SkiaWidgets;
...
+            return;
+
         var textSize = performanceWidget.TextSize;
 
         using var font = new SKFont() { Size = (float)textSize };
         using var textPaint = new SKPaint { Color = performanceWidget.TextColor.ToSkia() };
-        using var backgroundPaint = new SKPaint { Color = performanceWidget.BackColor.ToSkia().WithAlpha((byte)(255.0f * performanceWidget.Opacity)), Style = SKPaintStyle.Fill, };
+        using var backgroundPaint = new SKPaint { Color = performanceWidget.BackColor.ToSkia().WithAlpha((byte)(performanceWidget.BackColor?.A ?? 255 * performanceWidget.Opacity)), Style = SKPaintStyle.Fill, };
```
pauldendulk, line 24: In the renderer of the performance widget also use the alpha value of the color. It should be multiplied with the opacity of the entire widget. I guess the text paint should take into account the alpha as well.

Mapsui.UI.Shared/MapControl.cs
```diff
@@ -136,7 +136,7 @@ private protected void CommonDrawControl(object canvas)
         _stopwatch.Stop();
 
         // If we are interested in performance measurements, we save the new drawing time
-        _performance?.Add(_stopwatch.Elapsed.TotalMilliseconds);
+        Map.Performance?.Add(_stopwatch.Elapsed.TotalMilliseconds);
```
pauldendulk, line 139: The Performance component is now not nullable and part of the map.

Mapsui/Map.cs
```diff
@@ -38,6 +39,7 @@ public Map()
         BackColor = Color.White;
         Layers = [];
         Widgets.Add(CreateLoggingWidget());
+        Widgets.Add(CreatePerformanceWidget(this));
```
pauldendulk, line 42: The PerformanceWidget is now always created as part of the Map just like the LoggingWidget. I would prefer to have no widget at all in case of ActiveMode.No, but the solutions I could think of have other downsided.

Mapsui/Widgets/ActiveMode.cs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Mapsui.Widgets;
+
+public enum ActiveMode
```
pauldendulk, line 3: The ShowLoggingInMap is now changed into a ActiveMode enum so that it can be used by the PerformanceWidget as well.

Samples/Mapsui.Samples.Common/Maps/Widgets/PerformanceWidgetSample.cs
```diff
@@ -1,48 +1,42 @@
...
-using Mapsui.UI;
-using Mapsui.Widgets;
 using Mapsui.Widgets.InfoWidgets;
+using System.Linq;
+using System.Threading.Tasks;
 
 namespace Mapsui.Samples.Common.Maps.Widgets;
 
-public class PerformanceWidgetSample : IMapControlSample
+public class PerformanceWidgetSample : ISample
```
pauldendulk, line 12: I prefer ISample ovr IMapControlSample, because I prefer to use platform independent code. The change to the PerformanceWidget now allow this.

Samples/Mapsui.Samples.Common/Maps/Widgets/PerformanceWidgetSample.cs
```diff
@@ -1,48 +1,42 @@
...
     {
         var map = new Map();
 
-        map.Layers.Add(CreateLayer());
+        map.Layers.Add(CreateBingTileLayer());
         map.Navigator.CenterOnAndZoomTo(new MPoint(1059114.80157058, 5179580.75916194), map.Navigator.Resolutions[14]);
         map.BackColor = Color.FromString("#000613");
 
+        // The PerformanceWidget is created as part of the map.
+        var performanceWidget = map.Widgets.OfType<PerformanceWidget>().First();
```
pauldendulk, line 31: Since the PerformanceWidget is now added by default I just change some properties to show how this would be possible.

Samples/Mapsui.Samples.MapView/IMapViewSample.cs
```diff
@@ -1,10 +1,11 @@
 ﻿using Mapsui.Samples.Common;
+using Mapsui.UI.Maui;
 
 namespace Mapsui.Samples.Maui;
 
 public interface IMapViewSample : IMapControlSample
 {
-    bool OnTap(object? sender, EventArgs args);
+    bool OnTap(object? s, MapClickedEventArgs e);
```
pauldendulk, line 8: There was an unnecessary cast in many implementations that I fixed by changing this interface. This change is not related to the core this PR, sorry.

pauldendulk:
@smathersgit like this:

```csharp
var performanceWidget = map.Widgets.OfType<PerformanceWidget>().First();
performanceWidget.Performance.IsActive = Mapsui.Widgets.ActiveMode.Yes;
```
And this could change in the future. I recently moved some things around and it usually takes some iterations before I am happy with it. It makes sense to make the logging widget and performance widget work similar.

In the next release it will be possible to change the default value through a static. This is what I needed to enable it for all sample maps in the release build of the Blazor sample website.

PR #2928, opened 2025-03-19
MapView: fix InvalidCastExceptions with Pin

### LINKED ISSUE #2923: "MapView / Add Pin Sample" fails with 5.0.0-beta.10
**Mapsui Version**
5.0.0-beta.10 / commit 241170b088e51c04cd05d38650140fbbacf0de58

**Mapsui Platform**
MAUI

**Device**
Android 14

**Describe the bug**
In the sample app `Mapsui.Samples.Maui.MapView`, the `Add Pin Sample` in the category `MapView` fails with the following exception:

```
Exception: System.InvalidCastException: Specified cast is not valid.
  at Mapsui.UI.Maui.Pin.OnPropertyChanged(String propertyName) in Mapsui/Mapsui.UI.MapView/Objects/Pin.cs:line 538
  at Mapsui.UI.Maui.Pin.set_Transparency(Single value) in Mapsui/Mapsui.UI.MapView/Objects/Pin.cs:line 270
  at Mapsui.Samples.Maui.PinSample.OnTap(Object sender, EventArgs args) in Mapsui/Samples/Mapsui.Samples.MapView/PinSample.cs:line 42
  at Mapsui.Samples.Maui.MapPage.OnMapClicked(Object sender, MapClickedEventArgs e) in Mapsui/Samples/Mapsui.Samples.Maui.MapView/MapPage.xaml.cs:line 95
```

The relevant piece of code is:
https://github.com/Mapsui/Mapsui/blob/241170b088e51c04cd05d38650140fbbacf0de58/Mapsui.UI.MapView/Objects/Pin.cs#L536-L539

**To Reproduce**
1. Build and run the `Mapsui.Samples.Maui.MapView` sample app.
2. Select the category `MapView`.
3. Select `Add Pin Sample`.
4. Tap on the map.

**Expected behavior**
A pin should appear (without an exception occurring).

janusw:
This fixes #2923 by casting to the correct style class (a regression that was caused by PR #2904).

Note: There is a remaining problem with the callouts not appearing in "MapView / Add Pin Sample", which I could not fix, see https://github.com/Mapsui/Mapsui/issues/2923#issuecomment-2734127389. It seems to be caused by #2914, but I could not figure out how and why. @pauldendulk could you possibly help with that, please?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@ packages/
 
 .vs/
 
+.mono/
+
 Release/
 
 *.ncrunchproject
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,9 @@
 
 *.userprefs
 
+*.orig
+*.rej
+
 Mapsui.Rendering.Android/obj/
 
 Samples/Mapsui.Samples.Silverlight.Web/ClientBin
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -653,7 +653,7 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
                 var eventArgs = new MapClickedEventArgs(worldPosition.ToNative(), e.GestureType);
                 MapClicked?.Invoke(this, eventArgs);
 
-                if (e.Handled)
+                if (eventArgs.Handled)
                 {
                     handled = true;
                     return handled;
--- a/Mapsui.UI.MapView/Objects/Pin.cs
+++ b/Mapsui.UI.MapView/Objects/Pin.cs
@@ -535,39 +535,39 @@ public class Pin : IFeatureProvider, INotifyPropertyChanged
                 break;
             case nameof(Transparency):
                 if (Feature != null)
-                    ((SymbolStyle)Feature.Styles.First()).Opacity = 1 - Transparency;
+                    ((IPointStyle)Feature.Styles.First()).Opacity = 1 - Transparency;
                 break;
             case nameof(Anchor):
                 if (Feature != null)
-                    ((SymbolStyle)Feature.Styles.First()).Offset = new Offset(Anchor.X, Anchor.Y);
+                    ((IPointStyle)Feature.Styles.First()).Offset = new Offset(Anchor.X, Anchor.Y);
                 break;
             case nameof(Rotation):
                 if (Feature != null)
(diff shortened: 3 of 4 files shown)


pauldendulk (changes requested):
Thanks for looking into this. It is better to change these in to casts to IPointStyle, all the fields that are modified here are on the IPointStyle. The SymbolStyle, ImageStyle and CustomPointStyle all inherit from IPointStyle.

pauldendulk:
> but I could not figure out how and why. @pauldendulk could you possibly help with that, please?

Right now I can think of two possible causes.
- **Failed casts**: In many places there is a cast using the `as` keyword, and if the type is not the expected type the method returns. This is rarely what you want. Only in situations where there could actually be different types during runtime this could be useful. It is better to do a hard cast using (MyExpectedType) so that you get an error when things are not the way it used to be. My advice would be to change those to a hard cast and check if you get an exception somewhere. There is also the Utilities.Caster which you could use for this. It is slightly easier to get a good error message, and makes it easier if the parameter is nullable.
- **Mistakes in e.Handled**: This one is more likely. I had to rewrite the return true/false into setting the e.Handled. It is easy to make a mistake that way.

janusw:
> Thanks for looking into this. It is better to change these in to casts to IPointStyle, all the fields that are modified here are on the IPointStyle. The SymbolStyle, ImageStyle and CustomPointStyle all inherit from IPointStyle.

Thanks for the review! And good point. I just fixed it.

janusw:
>     * **Mistakes in e.Handled**: This one is more likely. I had to rewrite the return true/false into setting the e.Handled. It is easy to make a mistake that way.

I believe I found one such case (and fixed it in 7fa0a9c964125d4ff1064e19e59e4a432f192d5f). It does not seem to have any impact on the callouts unfortunately.

pauldendulk:
@janusw Great you fixed this hard to find issue! I really like that you describe the steps you went through yourself, this makes it really easy to follow along. About preventing bugs in the future:
- The MapTappedWidget was added to fix a bug we had at a time we did not have the Map events yet. I was happy with the fix at that time, but perhaps it can be improved now by using the Map.Tapped event.
- I will add the line below to the .editorconfig. So, at least you get to see dots to suggest you should do something with the result. I would like to set it to warning, but we have warningsAsErrors, and there are 90 such cases, most of them are fine, but I think we should fix this. In many cases this will be by explicitly discarding the result, but at least you are forced to think about that when you do.
``` console
dotnet_diagnostic.IDE0058.severity = suggestion# IDE0058: Value is never used
```

PR #2934, opened 2025-03-21
MauiVersion to 9.0.50

pauldendulk:
Could only get this to build by removing the net8 target from Mapsui.UI.Maui.


pauldendulk:
Removing net8 from Mapsui.UI.Maui will force user that are on Maui 8 to upgrade, and I think there are a lot, especially since net8 is LTM. Perhaps this should not be merged. Options:
- Figure out how to build for net8 anyway
- Release a separate nuget for MAUI net8, like we did for Avalonia at some point.

pauldendulk:
Trying an alternative approach https://github.com/Mapsui/Mapsui/pull/2954

PR #2937, opened 2025-03-24
Add option to select sample through the route in the Blazor sample

pauldendulk:
Done:
- Add option to select sample through the route in the Blazor sample so that you can point to a specific sample through a url.
- Unrelated: Updated the selection sample to a V5 way of doing things
- Unrelated: Moved the events in the shared MapControl together.


pauldendulk:
Merging this in it's current form but should be improved:
- Perhaps work with Id's so that renaming won't change the url. Perhaps a guid. Also a guid for the category?
- Perhaps it should be possible to use just the sample id, the category could be found from the sample.
- Perhaps it should not be an guid but a more meaning full name which we should try not to change.

PR #2938, opened 2025-03-25
Start on a sample to demonstrate VexTile.TileSource.Mvt

pauldendulk:
It is not working yet. 

The y-axis is upside down. This looks like an old problem in BruTile which is called the Inverted Y-Axis. The way to solve it is to change the YAxis field from TMS to OSM or the other way around, but if I do that in this sample nothing shows up, so there is a bit more to the problem. Have to figure that out later.

I created a wrapper tile source class so that you could use it as just another raster tile provider. Perhaps something like this should be in VexTile.TileSource.Mvt.

image


memsom:
I think the code had the YAxis set to TMS in the working example I have. I bet moving the code to use the same GlobalMercator implementation as BruTile/Mapsui might actually help somewhat also.

I will try to add a small demo app when I get a chance. I have the rendering working properly in another codebase, but I'm not entirely sure what the difference is.

memsom:
Yes, the data has the Y axis inverted. I just managed to get the tiles rendering in my NTS demo at zoom level 1:

image

The code to do that is in the branch I mentioned on the thread this was linked from.

pauldendulk:
Fixed the inverted y-axis with a workaround

image

pauldendulk:
The sample show a map in the right location. There are issues which will be fixed over time. This is only sample code, no library code. I will merge it.

PR #2943, opened 2025-03-27
Issue 2926 - Allow Microsoft.Bcl.AsyncInterfaces >= 8.0.0 in Mapsui 4.x (.NET Framework)

dougwebb:
The only change made is to remove the version range on Microsoft.Bcl.AsyncInterfaces in Directory.Packages.Props. 
I'm not sure why it's showing 99 files changed. This should be a branch off of the 4.1.8 tag, but I couldn't select that in the branches dropdown when creating the pull request.

To test this I built Mapsui.sln locally without the restriction. Then in my application I removed the Mapsui packages listed below and updated my project to directly reference the assemblies I built. I verified that was working with Microsoft.Bcl.AsyncInterfaces 7.0.0. Then I upgraded to v9.0.3 using NuGet Manager, which updated both the references and assembly binding redirects. (Prior to my Directory.Packages.Props change this is where the failure happened; NuGet complained that Mapsui wasn't compatible with Microsoft.Bcl.AsyncInterfaces 9.0.3)
After upgrading I built and ran my app and tested my maps, and everything appeared to be ok.

Whenever you upgrade Microsoft.Bcl.AsyncInterfaces, the package goes through all of your projects and updates the assembly binding redirects. It seems like Microsoft is pretty confident that there are no forward or backward incompatibilities to worry about. I suspect my change to Mapsui is pretty safe.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -58,7 +58,7 @@
     <PackageVersion Include="Microsoft.AspNetCore.Components.Web" Version="7.0.5" />
     <PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly" Version="7.0.5" />
     <PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="7.0.5" />
-    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="[7.0.0,8.0.0)" />
+    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="7.0.0" />
     <PackageVersion Include="Microsoft.CodeCoverage" Version="[16.11.0,17.0.0)" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Console" Version="[7.0.0,8.0.0)" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="[16.9.4,17.0.0)" />


janusw:
> The only change made is to remove the version range on Microsoft.Bcl.AsyncInterfaces in Directory.Packages.Props. I'm not sure why it's showing 99 files changed. This should be a branch off of the 4.1.8 tag.

The merge target should be the branch `develop/4.1` instead of `main`, I guess.

pauldendulk:
Thanks for the merge request.

The last time I tried to get develop 4.1 to build on the build server I got a lot of errors because of older components that were not supported (and may have had vulnerabilities). Since at that point I was just trying to redeploy the documentation I removed those older component to get it to build, but this will probably a breaking change for users. There may alternative ways to get develop/4.1 to build on the build server, but in my experience that is a lot of hassle and may even be impossible (since the Microsoft tools simply refuse to build). I am not eager to spend a free day on this. 😳

This is the branch with the fix for the documentation https://github.com/Mapsui/Mapsui/tree/develop/4.1-9sdk

dougwebb:
I'm not sure how to help you debug your build server issues. I've got develop/4.1 checked out with my Directory.Packages.props change on top. I've got Visual Studio 17.11.5 installed, and in the VS Powershell I found and ran Scripts\buildpack.bat. It was mostly successful; Mapsui.UI.Forms.csproj failed because I don't have tooling for xamarinios1.0. My Artifacts folder ended up with the rest of the packages.

When I first opened Mapsui.sln Visual Studio complained that it didn't have everything it needed for all of the project types, and I let it auto-install some things. I guess I didn't get Xamarin, at least not the version Mapsui needs.

janusw:
> The last time I tried to get develop 4.1 to build on the build server I got a lot of errors because of older components that were not supported (and may have had vulnerabilities). Since at that point I was just trying to redeploy the documentation I removed those older component to get it to build, but this will probably a breaking change for users. There may alternative ways to get develop/4.1 to build on the build server, but in my experience that is a lot of hassle and may even be impossible (since the Microsoft tools simply refuse to build). I am not eager to spend a free day on this. 😳

Well, .NET 8 (incl MAUI) is still officially supported at this point, so I think it should be possible to make it work. Also, as long as Mapsui 5 is in beta, we need to be capable of acting on the 4.1 release branch. I can try to help with the build errors on GHA (but cannot promise to make it work, either). Will try to have a look over the weekend ...

janusw:
I have addressed the most obvious build issues in #2944, and it works much better already, but there are still some issues left ...

pauldendulk:
Thanks to @janusw but build has been fixed for develop/4.1. The build here should also go green now that the branch has been updated.

pauldendulk (approved):
Thanks for your contribution!

PR #2944, opened 2025-03-28
Try to fix the GHA builds on the 4.1 branch

janusw:


pauldendulk:
Fantastic you helped out with this not so much fun job! I guess the spinner on the windows build is just a one time thing. And yes, you can just remove the mac legacy, it is not needed for a release. If you do that I can merge it.

janusw:
> I just realized dotnet-release-nugets.yml will also need to be adapted in order to release, but I guess these will be similar to those here.

Yes, I just opened PR #2946 for this, but I cannot really test it (one probably needs to set a tag to trigger the release script, right?).

I guess it would be a good idea to publish a release 4.1.9 with the recent changes soon ...

PR #2947, opened 2025-03-29
Fix repeated cache misses for tiles 

pauldendulk:
Cache cleanup removes old tiles first, New tiles were added with an iterationNumber of 0 and were thus removed first, and had to be created again on the next iteration. 

When panning on the most zoomed in level of openstreetmap in a city I now get 500 FPS instead of 55.
When panning on the most zoomed in level of linestring with rastertilelayer I now get 600 FPS instead of 60.

By looking at the code it seems this bug does not exist in V4.


pauldendulk:
The lesson learned is that we should have more metrics. It should be possible to see how many items we have in all our caches and the proportion cache hits and misses.

PR #2951, opened 2025-03-30
Improve performance of the Many SVGs sample when maximally zoomed in 

pauldendulk:
The case: 
- When you zoom in to the lowest level with the default openstreetmap sample you get 500 fps with WPF.
- When the Many SVGs sample is selected and you zoom in to the lowest level where you don't see SVGs the FPS is 300 instead of the 500 you get with openstreetmap. Apparently 200 FPS gets lost on processing related to that layer even though it is not rendered. What is going on?

I managed to get this back to 450 with several improvements.  I will comment the code changes.


Mapsui/Features/PointFeature.cs
```diff
@@ -10,22 +10,20 @@ public class PointFeature : BaseFeature, IFeature
     public PointFeature(PointFeature pointFeature) : base(pointFeature)
     {
         Point = new MPoint(pointFeature.Point);
+        Extent = Point.GetExtent();
     }
 
     public PointFeature(MPoint point)
     {
         Point = point ?? throw new ArgumentNullException(nameof(point));
+        Extent = Point.GetExtent();
```
pauldendulk, line 19: GetExtent creates a new MRect so only call it ones in the constructor and not every time Extent get is called.

Mapsui/Layers/MemoryLayer.cs
```diff
@@ -15,26 +15,47 @@ namespace Mapsui.Layers;
...
+    {
+        get => _features;
+        set
+        {
+            _features = value;
+            FeaturesWereModified();
+        }
+    }
+
+    public void FeaturesWereModified()
```
pauldendulk, line 37: Create a local copy and calculate the extent when the features are assigned. If the source array changes FeaturesWereModified() needs to be called. Mutiple enumerations of the source features was one of the biggest performance problems.

Mapsui/Layers/MemoryLayer.cs
```diff
@@ -15,26 +15,47 @@ namespace Mapsui.Layers;
...
-        return Features.Where(f => f.Extent?.Intersects(biggerRect) == true);
+        foreach (var feature in _localFeatures)
+        {
+            if (feature?.Extent?.Intersects(biggerRect) == true)
+                yield return feature;
+        }
     }
 
-    public override Func<IEnumerable<IFeature>, IEnumerable<IFeature>> SortFeatures { get; set; } = (features) => features.OrderBy(f => f.Id);
+    public override Func<IEnumerable<IFeature>, IEnumerable<IFeature>> SortFeatures { get; set; } = (_localFeatures) => _localFeatures.OrderBy(f => f.Id);
```
pauldendulk, line 58: Perhaps this logic of SortFeatures should change. For the MemoryLayer you would only want to sort once. However we do not know the order in which Features and SortFeatures are set, so we should sort when Features is assigned and when SortFeature is set. Possible but better to think of a simpler solution.

Mapsui/MPoint.cs
```diff
@@ -29,9 +29,14 @@ public MPoint(MPoint point)
     }
 
     public double X { get; set; }
+
     public double Y { get; set; }
 
-    public MRect MRect => new MRect(X, Y, X, Y);
+    // Creates an MRect with the point as both the minimum and maximum
+    public MRect GetExtent()
```
pauldendulk, line 36: The MRect property creates a class. Make it a method instead. We should have as a rule that properties do not create objects.

Mapsui/MRect.cs
```diff
@@ -7,14 +7,13 @@ public class MRect : IEquatable<MRect>
 {
     public MPoint Max { get; }
     public MPoint Min { get; }
+    public MPoint Centroid { get; }
```
pauldendulk, line 10: Centroid is now calculated when the MRect is created.

Mapsui/MRect.cs
```diff
@@ -23,16 +22,21 @@ public class MRect : IEquatable<MRect>
     public double Top => Max.Y;
     public double Right => Max.X;
 
-    public MPoint TopLeft => new(Left, Top);
-    public MPoint TopRight => new(Right, Top);
-    public MPoint BottomLeft => new(Left, Bottom);
-    public MPoint BottomRight => new(Right, Bottom);
+    // Creates an MPoint with the coordinates of the top left corner of the rectangle
+    public MPoint GetTopLeft() => new(Left, Top);
```
pauldendulk, line 26: All these are now methods because they create an object. For better performance we should try to use properties.

Mapsui.Nts/Extensions/MRectExtensions.cs
```diff
@@ -11,11 +11,11 @@ public static Envelope ToEnvelope(this MRect rect)
 
     public static Polygon ToPolygon(this MRect rect)
     {
-        return new Polygon(new LinearRing(new[] {
-            rect.TopLeft.ToCoordinate(),
-            rect.TopRight.ToCoordinate(),
-            rect.BottomRight.ToCoordinate(),
-            rect.BottomLeft.ToCoordinate(),
-            rect.TopLeft.ToCoordinate() }));
+        return new Polygon(new LinearRing([
```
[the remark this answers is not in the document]
pauldendulk: It is valid

Mapsui/Layers/MemoryLayer.cs
```diff
@@ -15,26 +15,47 @@ namespace Mapsui.Layers;
 /// <param name="layerName">Name to use for layer</param>
 public class MemoryLayer(string layerName) : BaseLayer(layerName)
 {
+    private IEnumerable<IFeature> _features = [];
+    private IFeature[] _localFeatures = [];
```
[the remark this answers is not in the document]
pauldendulk: Copilot is wrong, It is valid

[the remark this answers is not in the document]
pauldendulk: Copilot is wrong, It is valid

Mapsui/Layers/MemoryLayer.cs
```diff
@@ -15,26 +15,47 @@ namespace Mapsui.Layers;
...
+        _localFeatures = _features.ToArray();
+        _extent = _localFeatures.GetExtent();
+    }
 
     public override IEnumerable<IFeature> GetFeatures(MRect? rect, double resolution)
     {
-        // Safeguard in case BoundingBox is null, most likely due to no features in layer
-        if (rect == null) { return []; }
+        if (rect == null)
+            yield break;
```
pauldendulk, line 46: Not completely sure yield is the best option. Pro: you do not have to create the array. Con: You could get multiple enumeration because it works with deferred execution.

PR #2957, opened 2025-04-01
Make Map nullable

pauldendulk:
This is big change which is likely to cause bugs.

The disposable map will now be owned by the user, this is clearer, the user is responsible for disposing. The MapControl.Map is null by default. 

This also fixes frequent confusions where both the default Map and the user created Map get initialized.


pauldendulk:
I won't implement this in v5.

PR #2961, opened 2025-04-07
Fix update of modified ObservableMemoryLayer

pauldendulk:


Mapsui/Layers/ObservableMemoryLayer.cs
```diff
@@ -101,5 +101,6 @@ private void DataSource_CollectionChanged(object? sender, NotifyCollectionChange
                 // do nothing
                 break;
         }
+        FeaturesWereModified();
```
pauldendulk, line 104: This was not updating as a consequence of this change https://github.com/Mapsui/Mapsui/pull/2951. Perhaps ObservableMemoryLayer should not be derived from MemoryLayer. There is little code to inherit and it is simpler on it's own.

PR #2966, opened 2025-04-10
Make add pins sample a bit simpler

pauldendulk:


Samples/Mapsui.Samples.Common/Maps/Demo/AddPinsSample.cs
```diff
@@ -4,47 +4,46 @@
...
     public Task<Map> CreateMapAsync() => Task.FromResult(CreateMap());
 
     private static Map CreateMap()
     {
+        var features = new List<IFeature>();
+
         var map = new Map();
         map.Layers.Add(OpenStreetMap.CreateTileLayer());
-        map.Layers.Add(CreateGenericCollectionLayer());
+        map.Layers.Add(CreatePinLayer(features));
```
pauldendulk, line 25: We don't use the GenericCollectionLayer anywhere but here. Perhaps it should be removed, but then it could break something for someone. By using an ordinary MemoryLayer it is clearer that you do not have to do anything special to add features.

Samples/Mapsui.Samples.Common/Maps/Demo/AddPinsSample.cs
```diff
@@ -4,47 +4,46 @@
...
-            var layer = (GenericCollectionLayer<List<IFeature>>)e.Map.Layers.First(l => l.Name == "GenericCollectionLayer");
+            var layer = e.Map.Layers.OfType<MemoryLayer>().First();
             // Add a point to the layer using the Info position
-            layer?.Features.Add(new GeometryFeature
+            features.Add(new GeometryFeature
             {
                 Geometry = new Point(e.WorldPosition.X, e.WorldPosition.Y)
             });
+            // The MemoryLayer needs to update the changed features.
+            layer.FeaturesWereModified();
```
pauldendulk, line 35: I recently change the MemoryLayer to be somewhat more performant, but now you have to call FeaturesWereModified if the source changes. Not sure if this the right decision.

PR #2969, opened 2025-04-11
Rewrite invalidate loop 

pauldendulk:
Rewrite the invalidate loop so that it does not drop frames but waits until the previous Draw is done before starting a new invalidate loop iteration.

- Also, always wait 8 ms before starting a new Draw even if the Draw takes long.
- When Draw takes less than 8 ms the next Draw will start 16 ms after the previous.


Mapsui.UI.Shared/MapControl.cs
```diff
@@ -126,121 +124,87 @@ private void SharedConstructor()
...
+                if (map.Navigator.UpdateAnimations()) // Are there animations running on the Navigator
+                    _needsRefresh = true;
+
+                if (map.Widgets.Any(w => w.NeedsRedraw)) // Do the widgets require a redraw
+                {
+                    var w = (map.Widgets.Where(w => w.NeedsRedraw).FirstOrDefault());
+                    Console.WriteLine(w);
+                    _needsRefresh = true;
+                }
+            }
```
pauldendulk, line 161: This part of the logic where we have to check for Animations and NeedsDraw is still not so nice. I think they are two separate problems. 

- Animations: I don't want is to have the animation send a sequence of Refresh calls to orchestrate the invalidate loop, because I don't want to animations to deal with update frequency. The animation defines a change of some value over time, that's it. How often the animation is updated should be up to the invalidate loop. The invalide loop only needs to know that the animation is running. That is close to how it is currently implemented, but now the loop has to check the status of the animation in the loop. A better solution may by that animation informs the the loop about the end time op the animation, and adjusts the logic of the loop accordingly. 
- NeedsRedraw: Perhaps the change in the widget should send a notification that ends up calling RefreshGraphics. Or perhaps the call to RefreshGraphics should be done in user code, not through a notification mechanism.

PR #2989, opened 2025-05-06
Fix RotationTo at map start

pauldendulk:


Mapsui/Navigator.cs
```diff
@@ -121,12 +121,9 @@ private void Initialize()
         {
             IsInitialized = true;
 
-            if (_initialization.Count == 0)
-            {
-                ZoomToPanBounds();
-                return;
-            }
+            ZoomToBox(PanBounds, MBoxFit.Fill);
```
pauldendulk, line 124: It was already clear there was a problem with this code before this issue. In some cases you want to override the default behavior, in other cases you want to slightly modify it (for instance if you just want rotate). Now this is solved by always calling the default. If you want to override the full extent then call that from your own code, the default call will be irrelevant.

Samples/Mapsui.Samples.Common/Maps/Navigation/KeepCenterInMapSample.cs
```diff
@@ -21,20 +21,20 @@ public Task<Map> CreateMapAsync()
...
+        // the extent of Madagascar is used. When the PanBounds are limited it usually makes
         // sense to also limit the ZoomBounds.
 
-        var panBounds = GetLimitsOfMadagaskar();
+        var panBounds = GetLimitsOfMadagascar();
         map.Layers.Add(CreatePanBoundsLayer(panBounds));
 
         map.Navigator.OverridePanBounds = panBounds;
-        map.Navigator.OverrideZoomBounds = new MMinMax(0.15, 2500);
+        map.Navigator.OverrideZoomBounds = new MMinMax(0.15, 3000);
```
pauldendulk, line 31: The original idea was that this would fit the whole extent, but this depends on the size of the map so does not work in general. I made it a bit bigger so that at least all our samples show the full extent.

charlenni:
Good solution for this problem. Thank you for looking into this.

pauldendulk:
ah, the mac build works again, my plan worked, just wait a week and see if they have fixed the build agents.

PR #2995, opened 2025-05-19
Add RasterStyle.Outline

pauldendulk:
This adds the option to add a Outline to tiles. This was added for debugging tile fetch and draw behavior. It also adds a sample with random outline colors.

image


charlenni:
This could be done by an own style. Then is independent from the RasterStyle. 

It would be helpful to have the tile information (TileIndex) displayed.

A more elaborated version would work without a real raster tile source by using a TileSource, that only delivers tiles with a special style. With this, the tile debug information hasn't to be on the same layer as the raster tiles. You could place the layer with the debug information on top of all other layers.

PR #2999, opened 2025-05-23
Improve invalidate loop

pauldendulk:
Improvements after recent changes to the invalidate loop: https://github.com/Mapsui/Mapsui/pull/2969

- Environment.TickCount is not precise enough in general and on my machine was chuncked into steps of ~15ms. That is not useful if you calculate within a 16ms iteration. I used Stopwatch instead. The difference was noticeable. 
- I also reduced the use of await because I feared scheduling of the threads would be a bit too much if it needs to be within such a small time scale. So, now i use a Regular AutoResetEvent to wait for the previous draw and Thread.Sleep to wait for the next iteration. There is only one await left in the loop, this is to make sure we do not block a thread when nothing is drawn. I did not verify that this actually improves things, it is just guessing.
- I reduced the interval van 16ms to 8ms. To me it looks calmer when I drag the map.


pauldendulk:
Worked on an outdated branch.

PR #3009, opened 2025-06-02
Avoid possible NRE in multi threading situation by assigning to var

pauldendulk:


Mapsui/Extensions/FeatureExtensions.cs
```diff
@@ -65,14 +65,13 @@ public static IEnumerable<IFeature> Project(this IEnumerable<IFeature> features,
...
         return result;
     }
-
     public static MRect? GetExtent(this IEnumerable<IFeature> features)
     {
         MRect? result = null;
         foreach (var feature in features)
         {
-            if (feature.Extent is null) continue;
-            result = result is null ? new MRect(feature.Extent) : result.Join(feature.Extent);
```
pauldendulk, line 75: In theory it is possible that feature.Extent is not null on the check, the feature is then modified on another thread and on the next line feature.Extent is null. By assigning it to `extent` that mutation will not cause an NRE.

Issue #3014, opened 2025-06-06
Use Outline for LineStrings


clintsinger:
**The problem**
It appears that the only way to create an "Outline" on a `LineString` is by rendering the line twice, once as a the outline shape which is slightly larger than the the actual line on top of the outline.   Since there is already the `Outline` property on the `VectorStyle` it would be great if the `LineStringRenderer` handled the logic for you.  The comments already say "Outline style for **line** and polygon geometries" but doesn't actually do anything for lines.

I did look briefly at the skia library to see if it has a better solution but I couldn't find one that worked with the painting.  It looks like you can set a stroke but the fill and stroke option doesn't appear to do anything for lines.  If it can be done directly in skia, then that would be a better option than what is proposed below.

**The proposed solution**
The `LineStringRenderer` could be simply modified as follows to support the outlines.

```
public static class LineStringRenderer
{
    public static void Draw(SKCanvas canvas, Viewport viewport, VectorStyle? vectorStyle,
        IFeature feature, LineString lineString, float opacity, RenderService renderService, int position)
    {
        if (vectorStyle == null)
            return;

        // lineString - relevant for GeometryCollection children
        SKPath ToPath((long featureId, int position, MRect extent, double rotation, float lineWidth) valueTuple)
        {
            var result = lineString.ToSkiaPath(viewport, viewport.ToSkiaRect(), valueTuple.lineWidth);
            _ = result.Bounds;
            _ = result.TightBounds;
            return result;
        }

        var extent = viewport.ToExtent();
        var rotation = viewport.Rotation;
        var lineWidth = (float)(vectorStyle.Line?.Width ?? 1f);

        if (vectorStyle.Line.IsVisible())
        {
            using var path = renderService.VectorCache.GetOrCreate((feature.Id, position, extent, rotation, lineWidth), ToPath);

            if (vectorStyle.Outline?.Width > 0)
            {
                var width = vectorStyle.Outline.Width + vectorStyle.Outline.Width + vectorStyle.Line?.Width ?? 1;
                using var paintOutline = renderService.VectorCache.GetOrCreate((vectorStyle.Outline, (float?)width, opacity), CreateSkPaint);
                canvas.DrawPath(path, paintOutline);
            }

            using var paintLine = renderService.VectorCache.GetOrCreate((vectorStyle.Line, (float?)null, opacity), CreateSkPaint);
            canvas.DrawPath(path, paintLine);
        }
    }

[...]

PR #3016, opened 2025-06-06
Adds outline support to LineString

clintsinger:
The documentation states that `Outline` on `VectorStyle` is used by lines and polygons to draw an outline around the shape.   In reality it does not.  This PR intends to fix it for `LineString` 

see: https://github.com/Mapsui/Mapsui/issues/3014 

The `LineString` sample has been updated to show black border around the line.

```cs
public static IStyle CreateLineStringStyle()
{
    return new VectorStyle
    {
        Fill = null,
        Outline = { Color = Color.Black, Width = 1 },
        Line = { Color = Color.YellowGreen, Width = 4 }
    };
}
```

Image


pauldendulk (changes requested):
oh, forgot that the build needs to be fixed. One thing you need to do is to update the reference image of the line sample. 

What you need to do is run PS> .\scripts\CopyGeneratedImagesOverOriginalImages.ps1. It will overwrite all the existing reference images. There will be a lot of changes but if the tests do not fail we accept the difference, so revert all changed (and delete some files that have no reference image yet) and only commit the change line sample image.

I will update the outdated docs today: https://mapsui.com/v5/rendering-tests/?h=test

pauldendulk:
I see the geometry collection sample is also failing. This is a different kind of problem. It shows both polygons and linestring. The polygons have an outline, but since the style is shared the linestring now also show an outline. I suggest to update the image (instead of modifying the test).

pauldendulk:
There about six such failed tests. What you should do it inspect them individually and decide if you could update the image or need to alter the test.

pauldendulk:
I updated the documentation: https://mapsui.com/v5/rendering-tests/?h=test Tell me if this does not work for you.

clintsinger:
I'm going through the tests and updating the images and for the most part it makes sense to update the "original" image with the new outline.  I have run into an issue with the `PenStrokeCapSample` due to the default setting for the `Outline` property of the `VectorStyle`.   With the default outline set, the resulting image looks off.

Currently the defaults are as follows:

```
public VectorStyle()
{
    Outline = new Pen { Color = Color.Gray, Width = 1 };
    Line = new Pen { Color = Color.Black, Width = 1 };
    Fill = new Brush { Color = Color.White };
}
```

I would like to remove the default `Outline` configuration.  The `Outline` should only be intentionally set by the developer. 

The proposed change would be:
```
public VectorStyle()
{    
    Line = new Pen { Color = Color.Black, Width = 1 };
    Fill = new Brush { Color = Color.White };
}
```

 I'm also wondering if any of the properties should have any kind of default at all?

clintsinger:
Turns out that the polygon examples relied heavily on that default `Outline`.   One option is to fix the `PenStrokeSample` and `LineSample` style tests so that the `Outline` property is explicitly nulled or clean up the other polygon samples to not expect `Outline` to to be set.

The issue with the second option, is that people who didn't set the `Outline` property themselves and are expecting a gray 1 pixel width border will lose that.  That being said, people who are drawing lines and didn't set the `Outline` to null will now have a gray 1 pixel border on their `LineString`s.

I personally prefer to explicitly set values to something rather than having to explicitly null them.

pauldendulk (changes requested):
I am sorry you had to run into this mess while trying to add such a basic feature. This is how it goes in bigger projects with a lot of old code. The VectorStyle and its default values was implemented this way in SharpMap and was imported into Mapsui in January 2012.

The problem we are now facing is that Outline has a default value and it changes existing functionality. 

Your solution is to remove all the default values and fix it in the samples. This is too much of a breaking change. It would change the way things are rendered for all users of the VectorStyle who did not override the default. Since this is old code there will be quite a few. Also it is not really necessary. You mention in your comment that users do not expect default values, they may be (but I am not sure), but that is not something I want to change after thirteen years. More fundamental improvements could be applied by adding new styles at some point (and those should perhaps not have shared fields between polygons, lines and points).

To minimize the impact I suggest to keep the defaults, and the change in rendering will only be for the LineString. This change will also be less surprising. It makes sense that things change now that Outline for LineString is implemented. 

So, my request for change is to minimize the impact, just go through the failing tests and fix those by either updating the image or changing the styles.

clintsinger:
I rolled back the VectoSyle changes.   The tests were already updated to go either way so it was just a matter of putting back the styles in VectorStyle and then making the VectorStyle test work again.

The one issue is that we are in a situation where no matter which direction we go, the results won't be the same.  The `Outline` property used to be ignored but still had a default value for `LineStyle`.   Now that value won't be ignored so people will have to manually `null` the `Outline` to get the old behavior.

I figured it was probably better for none of the styles to be set by default and then people would just manually opt-in instead of having to opt-out.

pauldendulk:
> The one issue is that we are in a situation where no matter which direction we go, the results won't be the same.

Yes, I did realize that. It seemed like a nice isolated feature but then you get into this hairy thing. That is annoying but this is what we have a lot in an old project like this.

> I figured it was probably better for none of the styles to be set by default and then people would just manually opt-in instead of having to opt-out.

It may be, but my assessment was that this was a bit too big of a breaking change and not really necessary. I have been breaking quite a lot of things in the past because I did not want to maintain the old stuff. In future releases I will try to reduce that by introducing new functionality alongside existing functionality. So, you could imagine a vector style where there are separate pens and brushes for polygon, line and point, or perhaps styles that only apply only to one of those. This has no priority and there are plenty of other things to work on, but just as an example.

pauldendulk (approved):
Great! This is a feature a lot of people will be using.

PR #3018, opened 2025-06-07
Cleanup related to Styles

pauldendulk:


Mapsui/Styles/LabelStyle.cs
```diff
@@ -99,7 +99,7 @@ public LabelStyle()
     public LabelStyle(LabelStyle labelStyle)
     {
         Font = new Font(labelStyle.Font);
-        Offset = new Offset(labelStyle.Offset);
+        Offset = labelStyle.Offset;
```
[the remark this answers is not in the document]
pauldendulk: The advantage of immutability is that it does not matter if you point to the original or a copy they will always be the same. For performance it is better to not make a copy so lets take advantage of that.

Mapsui/Styles/BaseStyle.cs
```diff
@@ -1,8 +1,6 @@
-﻿namespace Mapsui.Styles;
-public abstract class BaseStyle : IStyle
-{
-    public double MinVisible { get; set; } = 0;
-    public double MaxVisible { get; set; } = double.MaxValue;
-    public bool Enabled { get; set; } = true;
-    public float Opacity { get; set; } = 0.7f;
-}
+﻿using System;
```
pauldendulk, line 1: The BaseStyle was introduce by mistake, Style should used instead.

Mapsui/Styles/Font.cs
```diff
@@ -67,12 +66,12 @@ public bool Bold
...
             {
                 _bold = value;
-                Invalidated = true;
             }
 
         }
     }
 
+    [Obsolete("There is no need to indicate invalidation", true)]
     public bool Invalidated { get; set; }
```
pauldendulk, line 75: Invalidated was used nowhere, just confusing.

Mapsui/Styles/Offset.cs
```diff
@@ -1,76 +1,14 @@
-// ReSharper disable NonReadonlyMemberInGetHashCode // todo: Fix this real issue
 namespace Mapsui.Styles;
 
-public class Offset
+public record Offset(double X, double Y)
```
pauldendulk, line 3: Relative Offset did not implement Equals while almost all other style primitives do. I implemented it by making it a record, then I turned Offset also into a record, so now you can not mutate X and Y anymore.

Mapsui/Styles/Style.cs
```diff
@@ -4,13 +4,12 @@
...
 // This file was originally created by Morten Nielsen (www.iter.dk) as part of SharpMap
 
-// ReSharper disable NonReadonlyMemberInGetHashCode // todo: Fix this real issue
 namespace Mapsui.Styles;
 
 /// <summary>
 ///     Defines a style used for rendering vector data
 /// </summary>
-public class Style : IStyle
+public abstract class Style : IStyle
```
pauldendulk, line 12: Style is a base class, it should always have been abstract.

PR #3034, opened 2025-06-16
Use SkiaSharp.Skottie Version 3.119.0 in the Uno Sample instead of 3.119.0-preview.1.2

inforithmics:
Small Uno WinUI Sample change. So that the referenced SkiaSharp.Skottie Version is 3.119.0 instead of 3.119.0-preview.1.2 reducing the used versions


pauldendulk (commented):
🤔

pauldendulk (approved):
It is weird how it was green when merging the original change.

PR #3039, opened 2025-06-17
Adds current Viewport to ViewportChangedEventArgs

clintsinger:
In addition to the previous viewport values, the current ones are included as part of the `ViewportChangedEventArgs`.

The `OldViewport` property has been marked `Obsolete` and has been replaced with `PreviousViewport`


pauldendulk (approved):
Perfect!

PR #3056, opened 2025-07-14
WIP Get Uno Skia rendering to work

pauldendulk:
Figuring out Uno skia rendering. The first commit uses only the net9-desktop TFM. I am happy at least something works. Next I have to back track which of my changes were really needed.

Let's see if this builds on the build server.

This looks like the fastest rendering platform we have had so far 🙀


mikernet:
> This looks like the fastest rendering platform we have had so far

It's quite fabulous, isn't it? 😊

Issue #3065, opened 2025-07-24
[Regression] Map.Tapped triggered when tapping Widget


janusw:
**Mapsui Version**
5.0.0-beta.19

**Mapsui Platform**
MAUI

**Device**
Android 16

**Describe the bug**
When tapping a widget on the map, the widget action is triggered, but also a `Map.Tapped` event is triggered on the underlying map.

**To Reproduce**
1. Build and run the sample app `Mapsui.Samples.Maui.MapView`.
2. Select the category 'Demo'.
3. Select the sample 'Add Pins'.
4. Tap one of the buttons/widgets (e.g. the northing button).

**Expected behavior**
The button action should be triggered, but no Pin should be set on the map. The `Map.Tapped` event should be blocked by the widget (as it was the case with Mapsui 4).

**Additional context**
The problem occurs at least since 5.0.0-beta.11, while 5.0.0-beta.6 still seems to work well. The versions in between (beta.7 to beta.10) cannot directly be tested, since the sample app crashes.

janusw:
> **To Reproduce** Steps to reproduce the behavior:
> 
>     1. Build and run the sample app `Mapsui.Samples.Maui.MapView`.
>     2. Select the category 'Demo'.
>     3. Select the sample 'Add Pins'.
>     4. Tap one of the buttons/widgets (e.g. the northing button).

Actually the problem seems to occur with the northing button only, not with the others.

janusw:
Fixed via #3066.

PR #3066, opened 2025-07-26
Mapsui.UI.MapView: properly handle the Tapped event in the northing button

janusw:
- in order to avoid the map from receiving the event in addition
- the Handled flag must not be set on the main thread!
- this fixes issue #3065


pauldendulk (approved):
ah, e.Handled was not set in the proper scope. Thanks for this fix!

PR #3071, opened 2025-08-04
[4.1] Mapsui.UI.Maui: allow usage with .NET MAUI 9

janusw:
This PR is motivated by the discussion at https://github.com/Mapsui/Mapsui/discussions/3063 (which for now is largely a monologue 😆): Although MAUI 9 is potentially a breaking change wrt MAUI 8, it seems that Mapsui4 works well with .NET MAUI 9 in common scenarios (although only targeting .NET 8).

Consequently, the PR loosens the package-version restrictions of Mapsui.UI.Maui to allow Maui 9.x, fixing warnings like this:

```
warning NU1608: Detected package version outside of dependency constraint:
Mapsui.Maui 4.1.9 requires Microsoft.Maui.Controls (>= 8.0.100 && < 9.0.0) but version Microsoft.Maui.Controls 9.0.51 was resolved.
warning NU1608: Detected package version outside of dependency constraint:
Mapsui.Maui 4.1.9 requires Microsoft.Maui.Controls.Compatibility (>= 8.0.100 && < 9.0.0) but version Microsoft.Maui.Controls.Compatibility 9.0.51 was resolved.
```

### DIFF
--- a/Mapsui.UI.Maui/Mapsui.UI.Maui.csproj
+++ b/Mapsui.UI.Maui/Mapsui.UI.Maui.csproj
@@ -48,8 +48,8 @@
 	</ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)'=='net8.0'">
-    <PackageReference Include="Microsoft.Maui.Controls" VersionOverride="[$(MauiVersion),9.0.0)" />
-    <PackageReference Include="Microsoft.Maui.Controls.Compatibility" VersionOverride="[$(MauiVersion),9.0.0)" />
+    <PackageReference Include="Microsoft.Maui.Controls" VersionOverride="[$(MauiVersion),10.0.0)" />
+    <PackageReference Include="Microsoft.Maui.Controls.Compatibility" VersionOverride="[$(MauiVersion),10.0.0)" />
   </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


pauldendulk (approved):
Hard for me to judge the impact. Some things might not work properly with v10. If so this is not something I want to spend time on. But let's see what happens.

pauldendulk:
hmm, perhaps I should have fixed the CI builds first. It seems to be failing because of this change. https://github.com/Mapsui/Mapsui/actions/runs/16986808593/job/48157371924

janusw:
> hmm, perhaps I should have fixed the CI builds first. It seems to be failing because of this change.

I was surprised that the CI builds did not even seem to get triggered for the PR. Judging from the GHA config, I exepected it should be running for PRs to any branch, but apparently it didn't. 🤔 


> hmm, perhaps I should have fixed the CI builds first. It seems to be failing because of this change. https://github.com/Mapsui/Mapsui/actions/runs/16986808593/job/48157371924

I doubt the failure is due to my changes. It seems that a test is being tried to run with .NET 6 (which is apparently not installed any more on the GHA Windows host, and anyway of out support for a while). Probably one should update that test to .NET 8?

Issue #3030, opened 2025-06-15
Maps do not render for Uno 6 apps using Skia Rendering


kazo0:
**Mapsui Version**
5.0.0-beta.17

**Mapsui Platform**
Uno Desktop/WASM/Android/iOS using Uno 6.0.130 and Skia Rendering

**Describe the bug**
As of Uno 6, you can use the [unified skia rendering engine](https://platform.uno/docs/articles/migrating-to-uno-6.html#optional-use-of-skia-rendering-for-ios-android-and-webassembly) for all supported platforms. 

The maps render on the net9.0-desktop target but with the following error:

> mapsuiuno6 Error: 0 : No canvas was assigned. This is unexpected. 
>  Exception: System.Exception: No canvas was assigned. This is unexpected.
>    at Mapsui.UI.WinUI.MapControl.InvalidateCanvas()
>    at Mapsui.Rendering.RenderController.InvalidateLoopAsync()
>    at Mapsui.Extensions.Catch.<>c__DisplayClass3_0.<<TaskRun>b__0>d.MoveNext()

For all other platforms, the maps do not display and fail with the following error:

`System.NotSupportedException: SKSwapChainPanel is not supported for Skia based platforms`

Full stacktrace:

> fail: Uno.UI.Dispatching.NativeDispatcher[0]
>       NativeDispatcher unhandled exception
> System.NotSupportedException: SKSwapChainPanel is not supported for Skia based platforms
>    at SkiaSharp.Views.Windows.SKSwapChainPanel..ctor() in /_/source/SkiaSharp.Views.Uno/SkiaSharp.Views.Uno.WinUI.Skia/SKSwapChainPanel.Skia.cs:line 22
>    at Mapsui.UI.WinUI.MapControl.CreateGpuRenderTarget()
>    at Mapsui.UI.WinUI.MapControl..ctor()
>    at mapsuiuno6.MainPage.InitializeComponent() in X:\src\TestApps\mapsuiuno6\mapsuiuno6\obj\Debug\net9.0-browserwasm\Uno.UI.SourceGenerators\Uno.UI.SourceGenerators.XamlGenerator.XamlCodeGenerator\MainPage_cd9fc8bdd233f377a45a8704c1af1a88.cs:line 57
>    at mapsuiuno6.MainPage..ctor() in X:\src\TestApps\mapsuiuno6\mapsuiuno6\MainPage.xaml.cs:line 13
[...]

PR #3092, opened 2025-08-27
Update Mapsui.UI.Uno.WinUI to Uno 6.2 and adjust RenderControl to be condition compilation-free

ramezgerges:
This is related to https://github.com/Mapsui/Mapsui/issues/3030 and follows up https://github.com/Mapsui/Mapsui/pull/3054. It moves to the just-released Uno 6.2 which includes adjustments to the SKCanvasElement API to make it available at compile time on all targets (but will crash at runtime if unsupported on the current target). This removes all the conditional compilation logic that was added in the previous PR.

More importantly, the Mapsui.UI.Uno.WinUI no longer declares `SkiaRenderer`. This allows it to support both native rendering and skia rendering depending on the end application declaration of `SkiaRenderer` instead of the current situation where only skia rendering is supported. How this works is through the Uno.Sdk.
* If the user app referencing Mapsui.UI.Uno.WinUI doesn't declare `SkiaRenderer`, the used TFM of Mapsui.UI.Uno.WinUI will match that of the app. Meaning that on Android, net9.0-android will be used, in which case SKSwapChainPanel will be used.
* If the user app referecing Mapsui.UI.Uno.WinUI declares `SkiaRenderer`, Uno.Sdk will force Mapsui.UI.Uno.WinUI to switch to the net9.0 TFM (even though by default MSBuild would grab the net9.0-<platform> TFM assemblies) which will use SKCanvasElement.

### DIFF
--- a/Mapsui.UI.WinUI/RenderControl.cs
+++ b/Mapsui.UI.WinUI/RenderControl.cs
@@ -1,10 +1,7 @@
+using Windows.Foundation;
 using Microsoft.UI;
 using SkiaSharp;
-
-#if __UNO_SKIA__
-using Windows.Foundation;
 using Uno.WinUI.Graphics2DSK;
-#endif
 
 namespace Mapsui.UI.WinUI;
 
@@ -21,15 +18,18 @@ abstract partial class RenderControl : Microsoft.UI.Xaml.Controls.UserControl
 
     public static RenderControl CreateControl(MapControl owner, System.Action<SKCanvas> renderCallback)
     {
-#if __UNO_SKIA__
-        return new SKCanvasElementRenderControl(owner, renderCallback);
-#else
-        // GPU does not work currently on Windows
-        bool useGPU = System.OperatingSystem.IsBrowser() || System.OperatingSystem.IsAndroid(); // Works not on iPhone Mini;
-        return useGPU
-            ? new SKSwapChainPanelRenderControl(owner, renderCallback)
-            : new SKXamlCanvasRenderControl(owner, renderCallback);
-#endif
+        if (SKCanvasElement.IsSupportedOnCurrentPlatform())
+        {
+            return new SKCanvasElementRenderControl(owner, renderCallback);
+        }
+        else
+        {
+            // GPU does not work currently on Windows
+            bool useGPU = System.OperatingSystem.IsBrowser() || System.OperatingSystem.IsAndroid(); // Works not on iPhone Mini;
+            return useGPU
+                ? new SKSwapChainPanelRenderControl(owner, renderCallback)
+                : new SKXamlCanvasRenderControl(owner, renderCallback);
+        }
     }
 
     public abstract void Invalidate();
@@ -119,7 +119,6 @@ partial class SKSwapChainPanelRenderControl : RenderControl
     }
 }
 
-#if __UNO_SKIA__
 partial class SKCanvasElementRenderControl : RenderControl
 {
 #pragma warning disable IDISP006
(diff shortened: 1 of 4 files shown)


ramezgerges:
@pauldendulk It seems like the lint error is actually a compile error. The issue is that the RenderControl file is shared between WinUI (i.e. Mapsui.UI.WinUI) and Uno (i.e. Mapsui.UI.Uno.WinUI), but WinUI doesn't have a reference to Uno.WinUI.Graphics2DSK (which includes the SKCanvasElement control). There are two solutions
1. The easy and clean one is to reference Uno.WinUI.Graphics2DSK in Mapsui.UI.WinUI. The reference is basically just a compile-time dependency for RenderControl, but at runtime, SKCanvasElement.IsSupportedOnCurrentPlatform will always be false and no further code from that package will be executed. The drawback is that you're adding a dependency from a non-WinUI project to the WinUI project (even though the dependency is super small and just depends on SkiaSharp).
2. Add an `#if` condition that excludes the portions of the file that use SKCanvasElement and replace the SKCanvasElement.IsSupportedOnCurrentPlatform check with a hardcoded `false` for WinUI.

I went with solution 1 in the last commit I pushed but I will let you decide which one is better. Thanks!

pauldendulk (changes requested):
>  error : The given key 'IDisposableAnalyzers' was not present in the dictionary

That is indeed a weird error, I would think it is something with the tooling itself.

> It seems like the lint error is actually a compile error. 

I am gradually making the linting warnings stricter by configurating them as errors in the .editorconfig. This way I can enforce correct formatting. In many cases those are easy to fix, but this one is trickier because of the conditional compile.

> I went with solution 1 in the last commit I pushed but I will let you decide which one is better. Thanks!

Thanks, I prefer option 1 in this case, although I don't like the `#if`s. There just really is a difference in how it is compiled, and I don't like to add a reference that is not needed to fix a linting error.

pauldendulk (changes requested):
I again the the IDisposableAnalysers error on windows. If this is introduced by this PR than that it is a problem that blocks the merge. We just have to built. The IDisposableAnalysers are in use for years. It is possible that some build server side change introduces this, it should not but happens regularly.

global.json
```diff
@@ -10,6 +10,6 @@
     "rollForward": "disable"
   },
   "msbuild-sdks": {
-    "Uno.Sdk": "6.0.130"
+    "Uno.Sdk": "6.2.29"
```
pauldendulk, line 13: Could the IDisposableAnalyzers be related to this upgrade?

Mapsui.UI.WinUI/RenderControl.cs
```diff
@@ -131,20 +136,27 @@ public SKCanvasElementRenderControl(MapControl owner, Action<SKCanvas> renderCal
         Content = _skCanvasElement = new MapControlSKCanvasElement(this);
     }
 
-    public override void Invalidate()
+    public override void InvalidateRender()
```
pauldendulk, line 139: Yep! Good solution.

pauldendulk:
@ramezgerges Perhaps relevant, there is a specific version of Uno.Check used in the dotnet.yml: `run: dotnet tool install -g Uno.Check --version 1.30.1`.

In a situation like this where you have no clue what causes the problem you could revert to 'systematic elimination'. You could create a new branch of your current branch, and mercilessly remove everything that might have introduced the problem until it compiles again and then add part back step by step to see where it fails. It may seem like a lot of work but if you if you set your mind to it it can go fairly fast, and often faster than just staring at a problem you do not understand. 

Also it will pay off to simplify the dotnet.yml in that branch to do only the Uno build on Windows to make it a LOT faster.

Samples/Mapsui.Samples.Uno.WinUI/Mapsui.Samples.Uno.WinUI/Mapsui.Samples.Uno.WinUI.csproj
```diff
@@ -67,10 +67,35 @@
...
-  </ItemGroup>
+  <Choose>
+    <When Condition="!$(TargetFramework.Contains('windows')) and $(UnoFeatures.Contains('SkiaRenderer'))">
+      <ItemGroup>
+        <!-- The UndefineProperties and SetTargetFramework properties below are necessary to force
+           this TFM of the projects below to be net9.0 even when the TFM of this project is net9.0-<platform>.
+           This is only necessary when SkiaRenderer is set and only for this sample app that uses ProjectReferences.
+           For user apps that use the Mapsui.UI.Uno.WinUI package (with PackageReferences), the logic below is unnecessary
+           and selecting the correct TFM is handled by Uno.Sdk  -->
+        <ProjectReference Include="..\..\..\Mapsui.UI.Uno.WinUI\Mapsui.UI.Uno.WinUI.csproj">
```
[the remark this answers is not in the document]
ramezgerges: It is. Uno.Sdk is designed to do all this voodoo so that library authors don't have to think about it. Unfortunately, the logic inside Uno.Sdk only works for PackageReferences, not for ProjectReferences as done here. This hack is therefore necessary to get the sample app inside the repo working but is not relevant for MapsUI users.

pauldendulk:
@ramezgerges I would have time this Monday to discuss this. I am not sure if I can help, but at least I like to learn a bit more about how the new Uno works. You could contact me at info.mapsui@gmail.com.

PR #3095, opened 2025-08-30
MapView: prevent PinClicked event from being triggered multiple times

### LINKED ISSUE #3091: MapView: HandlerInfo and PinClicked are triggered multiple times
**Mapsui Version**
5.0.0-beta.20

**Mapsui Platform**
MAUI

**Device**
Android 16

**Describe the bug**
The `HandlerInfo` method of the `MapView` class is triggered multiple times for every click (e.g. four times with the sample app). This means that the `PinClicked` events is also triggered the same amount of times.

**To Reproduce**
1. Build and run the sample app Mapsui.Samples.Maui.MapView.
2. Select the category "MapView" and the sample "Add Pin Sample".
3. Add a breakpoint (or debug output) in the `HandlerInfo` method.
4. Tap the map to add a pin, then tap the pin.
5. Observe that `HandlerInfo` is called four times upon tapping the pin.

**Expected behavior**
The `HandlerInfo` method and the `PinClicked` event should only be triggered once.

janusw:
This is supposed to fix #3091 (by removing the use of the `MapControl.Info` event, which is apparently "deprecated", from the `MapView` class).

In addition it fixes a minor problem with the PinSample (in the `Mapsui.Samples.Maui.MapView` app), which has been noticed by @pauldendulk:

> BTW, I notices something odd. If I remove a callout by clicking on the callout I can not enable the callout by clicking on the pin. If I remove the callout by clicking on the pin I can toggle it back on. Perhaps this is just an unrelated problem.

This is not a general `MapView` issue, but only occurs in the sample app. It is fixed by the second commit.

### DIFF
--- a/Mapsui.UI.MapView/MapView.cs
+++ b/Mapsui.UI.MapView/MapView.cs
@@ -56,7 +56,6 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
 
         // Add some events to _mapControl
         Map.Navigator.ViewportChanged += HandlerViewportChanged;
-        Info += (s, e) => HandlerInfo(e);
         SizeChanged += HandlerSizeChanged;
 
         // Add MapView layers to Map
@@ -430,7 +429,6 @@ public class MapView : MapControl, INotifyPropertyChanged, IEnumerable<Pin>
 
                 // Add event handlers
                 Map.Navigator.ViewportChanged += HandlerViewportChanged;
-                Info += (s, e) => HandlerInfo(e);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


pauldendulk (changes requested):
The MapView_Info is commented out. I guess it could just be removed entirely, right?

janusw:
> Also I just noticed that the same `MapView_Info` also exists on the `MainPageLarge` (used with tablets). I'll remove that one too, in order to fix the sample on tablets as well ...

Note that for the tablet case, the `MapView_Info` method does a bit more (i.e. it displays some info about the clicked object in a label). Therefore I did not remove the whole method there, but only the part that makes the callouts disappear forever ...

pauldendulk (approved):
Thanks for the fixes!

PR #3098, opened 2025-09-01
Remove Skottie and Uno.Resizetizer workaround 

pauldendulk:
Perhaps not needed. Let's see what happens.


pauldendulk:
This does not fix the Resizetizer intermittent build fail, but apparently this is not necessary either.

PR #3100, opened 2025-09-02
Fix Blazor touch on iOS

pauldendulk:


Mapsui.UI.Blazor/MapControl.cs
```diff
@@ -287,14 +289,16 @@ public void OnTouchMove(TouchEventArgs e)
...
-            if (_lastMovePosition is null)
-                return;
-            var position = _lastMovePosition.Value;
-            OnPointerReleased([position]);
+            var positions = e.TargetTouches.ToScreenPositions(_clientRect);
+
+            if (positions.Length > 1)
+                OnPointerReleased(positions);
+            else if (_lastTouchPosition is ScreenPosition screenPosition) // Workaround for missing TouchEnd position.
+                OnPointerReleased([screenPosition]);
```
p2pbsh, line 301: 'OnPointerReleased' (and pressed, moved etc) naming is a little confusing when called by both pointer and touch handlers, perhaps rename to something more generic like 'OnUserInputReleased'?

Mapsui.UI.Blazor/MapControl.cs
```diff
@@ -266,6 +266,8 @@ public void OnTouchStart(TouchEventArgs e)
             var positions = e.TargetTouches.ToScreenPositions(_clientRect);
             _manipulationTracker.Restart(positions);
 
+            _lastTouchPosition = positions.Length > 1 ? positions[0] : null; // Workaround for missing TouchEnd position
```
p2pbsh, line 269: This should resolve the iOS issues I encountered, get her in the next beta and I'll give it a good test.

Mapsui.UI.Blazor/MapControl.cs
p2pbsh: Ah, I see, that's a fair point (er).

PR #3124, opened 2025-09-16
Move parts of the vector tile sample code to a new Mapsui.Experimental.VectorTiles project

pauldendulk:


charlenni:
VexTile doesn't deliver a vector tile. It does deliver a normal bitmap, which is created from vector data on the fly. If you now use the name VectorTileSource for this, you couldn't use it for real vector tiles later.

Issue #3102, opened 2025-09-03
Mapsui support 16KB memory page sizes?


JORGEGO:
Google is notifying all developers that by November 1st their apps must support 16KB memory page sizes, one of the errors I'm getting when compiling at 16KB is Android 16 will require 16KB page sizes, shared library 'libSkiaSharp.so' does not have a 16KB page size. Please inform the authors of the NuGet package 'SkiaSharp.NativeAssets.Android' version '2.88.9', I'm wondering if Mapsui.Maui is ready for 16KB and Android 16 will require 16 KB page sizes, shared library 'libHarfBuzzSharp.so' does not have a 16 KB page size. Please inform the authors of the NuGet package 'HarfBuzzSharp.NativeAssets.Android' version '7.3.0.3'

pauldendulk:
Thanks for reporting this problem. I was not aware of this, and this could become quite a serious problem.

You mention HarfBuzzSharp 7.3.0.3, so you are working with V4. In V5 we upgraded to the latest SkiaSharp which does support 16KB pages according to this release https://github.com/mono/SkiaSharp/releases/tag/v3.119.0-preview.1.2. 

The upgrade to SkiaSharp 3.119.0 was quite a substantial breaking change and is not something we are going to do in V4. So, my suggestion is to start working on an upgrade to Mapsui v5. It is not released yet but we are now in the bug fixing phase.

pauldendulk:
In the meantime 5.0.0-rc.1 was released. This should support 16KB memory pages.

PR #3128, opened 2025-09-17
Use SourceGear Sqlite for Android 16KB support

inforithmics:
SQLitePCLRaw.bundle_green is deprecated and does not support Android 16KB pages. So I changed the dependencies to Reference SourceGear.Sqlite which supports Android 16KB pages.


pauldendulk:
Great to see you here again @inforithmics!!

The 16KB is pretty serious, thanks for looking into this https://github.com/Mapsui/Mapsui/issues/3102

PR #3138, opened 2025-09-25
Update SkiaSharp to 3.119.1

inforithmics:
Update to newest SkiaSharp Version. With some important Bugfixes


pauldendulk:
Here are the release notes: https://github.com/mono/SkiaSharp/releases

This one in particular caused us a lot of trouble. https://github.com/mono/SkiaSharp/pull/3338. When users did the getting started of MAUI .NET 9 this resulted in an immediate crash. This is why we start it MAUI on .NET 9 in CPU mode with this workaround code:

```cs
public static bool UseGPU = !IsMaui9();
```
```cs
private static bool IsMaui9()
{
    var frameworkDescription = System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription;
    return frameworkDescription.Contains(".NET 9");
}
```

PR #3139, opened 2025-09-25
Update Microsoft.WindowsAppSDK to a1.6.250602001

inforithmics:
Upgrade Microsoft.WindowsAppSDK to a non vulnerable Version (according to Mend.io\Rider)


pauldendulk:
One of the errors in the build is a known problem with VexTile. A fix was merged but not yet released.

PR #3140, opened 2025-09-25
Evaluation if mapsui works on .net10 

inforithmics:
.NET 10 is now RC 1 with a go live license this pull request is currently mainly for investigating things that break or could be improved. In November when .NET 10 is release this pull request leaves WIP

Findings:
.net8-android; net8-ios; are end of life


pauldendulk:
I am not a big fan of supporting multiple versions at the same time because in the past I have found this can cause obscure bugs which at some point took a significant part of the time to diagnose. It also results in more non-deterministic build errors. Now supporting multiple versions may not be very problematic in all scenarios, so we should consider supporting 10, but we should weigh the pros and cons. For MAUI I introduced a separate nuget package to support [MAUI 8](https://www.nuget.org/packages/Mapsui.Maui8/5.0.0-beta.24). Not sure if that is the best option, but at least it makes diagnosing problems a bit easier.

pauldendulk:
Closing this because it has been in draft for a long time now.

PR #3141, opened 2025-09-25
Update uno and set specifiy maui.resizetizer version

inforithmics:
Trying to make build more reliable.

see here

https://github.com/unoplatform/uno/issues/21344


pauldendulk:
We had this workaround but it did not fix the problem, but the a difference is that [it was only in Uno](https://github.com/Mapsui/Mapsui/pull/3098/files), so perhaps this works. I restarted the build. Perhaps I should do that a few more time today.

inforithmics:
The problem the workaround tries to solve is that there are two native skia sharp versions 2.88 and 3.116 und unlike dotnet only one can be loaded into a process. Maui resizetizer uses currently 2.88 and Maui Resizetizer uses 3.116 an akternative solution would be to make for example Mapsui without Maui or Uno slnf and tehn build uno or maui slnf project

pauldendulk:
@inforithmics You closed and then reopened this. I think this was not intended?

PR #3146, opened 2025-09-26
Compile Maui and Uno Samples separate on Windows

inforithmics:
This should fix the Problems that the Maui and Uno Resizetizer use different SkiaSharp Versions, which causes sporadic crashes


inforithmics:
Close in favor of https://github.com/Mapsui/Mapsui/pull/3146

pauldendulk (approved):
Great thanks, this was an annoying issue. I will wait a few weeks before closing the Uno issue just to be sure it is actually fixed.

PR #3166, opened 2025-10-05
Add ObservableCollectionLayer

pauldendulk:
Identical to ObservableMemoryLayer except it is not inherited from MemoryLayer which has the Features field which is essentially replaced by the ObservableCollection field. The code itself was confusing and the Features field could be confusing for users.

image


charlenni:
Sorry to say this, but for me, it isn't very clear. Why are there different Layer types? With this ObservableCollectionLayer you have even more layers. The difference is the provider of the data: in one case it is an array (MemoryLayer), which could be writable (WritableLayer). In the other case, you have an eye on the data (ObservableCollectionLayer). Then you have a layer with only one feature (MyLocationLayer). And a layer specially for images (ImageLayer).

So why not use one type of Layer, holding the visibility, zoom min and max, the provider for the data and the renderer to use for rendering this type of data. Then the Layer don't need any knowledge about the type of data and how it is displayed.

charlenni:
I fully understand your intention, but what I will say is, that there is no plan in which direction you are heading. Instead of thinking, what is the easiest way to understand Mapsui for newbies you add another layer. Wouldn't it be much better to make a clear structure for this? Layers are responsible to hold a Provider, which delivers all the features, and a renderer, which describes, how this features are drawn on screen. And why is the layer responsible for fetching the data? Isn't that something the provider is responsible? Wouldn't it be more useful to create an ObservableCollectionProvider than an ObservableCollectionLayer? And a MemoryProvider instead of a MemoryLayer? Why is it possible that some layers contains their features and some not? So the question is, why some data has to be provided by providers and some data is stored directly in the layer?

pauldendulk:
> I fully understand your intention, but what I will say is, that there is no plan in which direction you are heading. Instead of thinking, what is the easiest way to understand Mapsui for newbies you add another layer. 

The `ObservableMemoryLayer` is not another layer, it is a refactoring of the `ObservableMemoryLayer`. In the current situation (right before the release) it is better to do this in an experimental package. I think, having the experimental packages developed against the stable v5 is a nice way to move forward in small steps. 

Currently, there is no single `Layer` type, because a `Layer` with a `MemoryProvider` does not behave the same way as a `MemoryLayer`, and I have to work with this until this is solved. I cannot just make it so. Perhaps, at some point, we could turn the `ObservableMemoryLayer` into an `ObservableMemoryProvider`, but not now.

The idea of just a single layer type is nice, but it is the kind of idea you could have while biking to the train station, but the real work is in actually realizing it. You can only make such changes when you have a clear picture of all the constraints. To make it work, you have to look into the details of all the current layer types, you have to experiment with code changes, and you need to get feedback from users building applications with Mapsui. And perhaps you will find out it is better to work with a few different layer types. 

It is possible to make big changes though. We did so in the past: 
- Centralizing the `Navigator` in v4. 
- Harmonizing the pointer events in v5. 
- Realize centralized data fetching in v5. 

For v6, I will work on bigger changes in the rendering.

> Wouldn't it be much better to make a clear structure for this? Layers are responsible to hold a Provider, which delivers all the features, and a renderer, which describes, how this features are drawn on screen.

This way, the layer would need knowledge of the specific rendering technology. Currently, the renderer has no knowledge of the layer implementations, it just works with the `ILayer` interface. The specifics of rendering are within the `IStyleRenderer` implementations. It is quite a clear separation of concerns. 

The OO design where the type implements the rendering was discussed in my very first OO programming book. There were squares, triangles and circles, which had their own rendering implementation. The nice thing was that you could add new types, while the main renderer would not have to change. The type would need to know about the rendering, though. In our case, we have a more fine-grained design. The renderer does not need to know about the layer types, and layers do not need to know about the render technology. That responsibility is pushed further down into the `IStyleRenderer`.

> And why is the layer responsible for fetching the data? Isn't that something the provider is responsible?

Data fetching is now centralized. It is fetched through the `IFetchableSource` interface. It has no knowledge of the implementation. The requests are executed by the central data fetcher. The layer forwards this to child components, which would be the provider for the `Layer`, and the `TileFetchPlanner` for the `TileLayer`. The `TileLayer` is quite different in nature, and it would not be easy to change this into a `Layer`/`Provider` model. Perhaps this is not impossible, but the real design work begins when you start thinking about these specific problems. And even if it is possible, it might not make things simpler. Without knowing the details, it is easy to think ‘why don’t we do everything in the same way’, but when you are working on this, you might have to conclude it would be simpler to have different implementations.

> Wouldn't it be more useful to create an ObservableCollectionProvider than an ObservableCollectionLayer? And a MemoryProvider instead of a MemoryLayer? Why is it possible that some layers contains their features and some not? So the question is, why some data has to be provided by providers and some data is stored directly in the layer?

All layers have the features in memory before they get rendered. The difference is that some fetch the data from disk or the web. This is because we support two different scenarios:  
- **Mapsui fetches the data**: This is the classical GIS data provider. There is geodata, like a shapefile or a WMS, and you configure the source, and the map component fetches and shows the data.  
- **The app fetches the data**: This may be a more common scenario. The user is building an app. Entity Framework is used to fetch the data. Perhaps it is stored in a `ViewModel`, and shown in a list in the `View`. At some point, the user wants to show the data in the map because it happens to have a coordinate. They download Mapsui and just want to show the data they already have in memory. No async/await fetch logic is needed here. 

Writing it like this, I now think it may be better to keep the distinction.

PR #3167, opened 2025-10-06
Dependencies: Update NTS to Version 2.6.0

inforithmics:
Update NTS NetTopologySuite to 2.6.0

Main changes is the removal of the System.Memory Dependency for .NET Core which often caused dependency Problems.
(.Net Standard 2.1 Target).


pauldendulk (approved):
Great! We had a lot of trouble with that System.Memory nuget.

PR #3193, opened 2025-10-16
Add Map dependency property for WPF MapControl (enables data binding)

AhmadovMahammad:
### Summary

This PR introduces support for **XAML data binding of the `Map` property in WPF’s `MapControl`**.
Previously, WPF developers could not bind a `Map` instance directly via MVVM; they had to assign it in code-behind.

### The problem

WPF developers could not bind a `Map` instance to the `MapControl` via XAML.
Currently, the `MapControl.Map` property is not a `DependencyProperty`, which makes MVVM or binding scenarios impossible without code-behind initialization.

### The proposed solution

Added a `DependencyProperty` named `MapProperty` to `MapControl` (in `Mapsui.UI.Wpf`).
This allows developers to use the following pattern in XAML:

```xml
<wpf1:MapControl Map="{Binding Map}" />
```

The property automatically updates `MapControl.Map` when the binding context changes.

### Alternative solutions

Previously, developers had to use attached behaviors (custom dependency properties) or manually set `MapControl.Map` in code-behind.
This change removes that need by providing direct binding support within the library itself.

### How to test

1. Create a simple WPF window with a `MapControl` bound to a view model:

   ```xml
   <wpf1:MapControl Map="{Binding Map}" />
   ```
2. In the view model:

   ```csharp
   public class MainViewModel
   {
       public Map Map { get; } = new Map { Layers = { OpenStreetMap.CreateTileLayer() } };
   }
   ```
3. Run the application — the map will display automatically without any code-behind assignments.

### Additional context

This approach is consistent with existing binding support in **Avalonia** and **MAUI**, ensuring API parity across all UI platforms.
The change is fully isolated to the WPF control and introduces no breaking changes.

### DIFF
--- a/.github/workflows/dotnet-docs.yml
+++ b/.github/workflows/dotnet-docs.yml
@@ -42,7 +42,7 @@ jobs:
       uses: actions/setup-dotnet@v4
       with:
          dotnet-version: |
-           9.0.304
+           9.0.305
            8.0.407
     - name: install workloads
       run: dotnet workload install maui macos android ios maccatalyst wasm-tools wasm-tools-net8 
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -56,7 +56,7 @@ jobs:
       uses: actions/setup-dotnet@v4
       with:
          dotnet-version: |
-           9.0.304
+           9.0.305
            8.0.407
     # Open JDK
     - name: Install Open JDK 11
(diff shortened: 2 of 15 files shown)


pauldendulk (approved):
Great! That looks good. And thanks for adding a sample of the View/ViewModel. I guess this ViewModel is not now default active? I expect I need to change some code for that?Anyway. I will merge that now and look at the sample afterwards.

pauldendulk:
The build fails because of a missing dispose in the samples. I am going to merge this anyway because I want to have this in rc.3, and I will fix it in another PR.

AhmadovMahammad:
Thanks for reviewing and merging. I’m glad this improves WPF parity and simplifies MVVM usage for MapControl bindings.

pauldendulk:
Thanks @AhmadovMahammad. I created a small cleanup PR. I used the OsmView to test the binding, but then removed it, because as a rule we don't keep unused code. https://github.com/Mapsui/Mapsui/pull/3202

A better sample would be the update the primary WPF sample to use binding. This currently has no ViewModel at all and is messy (and really old code). It would be some work because there are also the lists of the sample categories and names. Instead of the list I would prefer two dropdown, like in the new Blazor sample: https://mapsui.com/v5/samples/

PR #3196, opened 2025-10-18
Add option to inject a custom HttpClient into the TileLayer

### LINKED ISSUE #3194: Porting to Mapsui5/BruTile6: WMS layer with auth token in http header
With Mapsui 4.x / BruTile 5.x, I was using code like below to create WMS layers that require an authentication token (by using a custom `HttpClient` that contains the auth token in its header):

```
        private readonly HttpClient http = new();

        public ILayer CreateWMSLayer(Attribution attr, string name, IPersistentCache<byte[]> cache)
        {
            var uri = new Uri("...");
            var schema = new GlobalSphericalMercator()
            {
                Format = "image/png"
            };
            var layers = new[] { "..." };
            var styles = new List<string> { "..." };
            Dictionary<string, string> customParams = new Dictionary<string, string>()
            {
                { "...", "..." }
            };
            var request = new WmscRequest(uri, schema, layers, styles, customParams);
            var ts = new HttpTileSource(schema, request, persistentCache: cache, attribution: attr, tileFetcher: http.GetByteArrayAsync);
            return new TileLayer(ts) { Name = name };
        }
```

With the Mapsui 5 / BruTile 6 RCs, this does not work any more (in particular due to https://github.com/BruTile/BruTile/pull/234).

@pauldendulk How is this use case supposed to be handled with the new library versions?

AFAICS, the `TileLayer` in Mapsui 5 is always using an internal `HttpClient` instance. Do we need to extend the `TileLayer` constructor, to add the possibility of using a custom `HttpClient`? Or is there another way of implementing the above use case?

janusw:
* in order to use custom header parameters (e.g. authorization or user agent) etc
* fixes #3194
* also avoid a warning being thrown by mistake
  - https://github.com/DotNetAnalyzers/IDisposableAnalyzers/blob/master/documentation/IDISP007.md
  - we already make sure that the HttpClient client is only disposed if it was not injected from the outside

### DIFF
--- a/Mapsui.UI.Uno.WinUI/Mapsui.UI.Uno.WinUI.csproj
+++ b/Mapsui.UI.Uno.WinUI/Mapsui.UI.Uno.WinUI.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Uno.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net8.0-ios;net8.0-maccatalyst;net8.0-android;net8.0-windows10.0.19041;net9.0;net9.0-ios;net9.0-maccatalyst;net9.0-android;net9.0-windows10.0.19041</TargetFrameworks>
+    <TargetFrameworks>net8.0;net8.0-android;net8.0-windows10.0.19041;net9.0;net9.0-ios;net9.0-maccatalyst;net9.0-android;net9.0-windows10.0.19041</TargetFrameworks>
     <UnoSingleProject>true</UnoSingleProject>
     <OutputType>Library</OutputType>
     <!-- Ensures the .xr.xml files are generated in a proper layout folder -->
--- a/Mapsui.UI.iOS/Mapsui.UI.iOS.csproj
+++ b/Mapsui.UI.iOS/Mapsui.UI.iOS.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <OutputType>Library</OutputType>
     <IPhoneResourcePrefix>Resources</IPhoneResourcePrefix>
-    <TargetFrameworks>net8.0-ios;net9.0-ios</TargetFrameworks>
+    <TargetFrameworks>net9.0-ios</TargetFrameworks>
     <ImplicitUsings>true</ImplicitUsings>
     <PackageId>Mapsui.iOS</PackageId>
     <!--<RuntimeIdentifiers>iossimulator-x64</RuntimeIdentifiers>-->


janusw:
Thanks for merging! 👍

Mapsui.Tiling/Layers/TileLayer.cs
```diff
@@ -140,7 +154,7 @@ private void TileFetchPlanner_OnDataChanged(object? sender, Exception? ex)
     {
         if (_tileSource is IHttpTileSource httpTileSource)
         {
-            var tileData = await httpTileSource.GetTileAsync(_httpClient, tileInfo).ConfigureAwait(false);
+            var tileData = await httpTileSource.GetTileAsync(HttpClient, tileInfo).ConfigureAwait(false);
```
janusw: Btw, to follow up on the discussion from https://github.com/BruTile/BruTile/pull/234: I still find it weird that the `HttpClient` enters the `HttpTileSource` via `GetTileAsync` and not via the constructor (as proposed in https://github.com/BruTile/BruTile/pull/226, which was rejected). This would actually remove the handling of the `HttpClient` from the `TileLayer` completely (and thus help to move this complexity from the core to the surface, as you put it).

pauldendulk: The current solution in BruTile is better. No pointers to the HttpClient at all, so no need to think about who is responsible for disposing. That is the way to go. We should remove it from the TileLayer in a similar way. See my comment here: https://github.com/Mapsui/Mapsui/issues/3194#issuecomment-3415039385

janusw: > The current solution in BruTile is better. No pointers to the HttpClient at all, so no need to think about who is responsible for disposing. That is the way to go. We should remove it from the TileLayer in a similar way. See my comment here: [#3194 (comment)](https://github.com/Mapsui/Mapsui/issues/3194#issuecomment-3415039385)

That'e exactly what I'm saying:
* Pass the `HttpClient` to the constructor of the `HttpTileSource`. Typically the header etc will not change anyway after construction, i.e. there is no need to pass a (potentially differing) client to every invocation of `GetTileAsync`.
* The fully-constructed `HttpTileSource` is passed to the `TileLayer` constructor (as an `ITileSource` argument).
* Therefore, the `TileLayer` will not have to handle any `HttpClient` business, and the corresponding member (or rather: both of them) can be removed.
* The object which creates the `HttpTileSource` (outside of BruTile, and outside of the Mapsui `TileLayer`) is now the owner of the `HttpClient`. It takes care of creating and disposing the client, and decides which header parameters to set etc. The library is fully transparent to all of that.

That's what you're aiming for, right?

pauldendulk: > Pass the HttpClient to the constructor of the HttpTileSource

This means that the HttpTileSource has a pointer to the HttpClient. So the object that has to do the disposing of the HttpClient has to take into account the lifetime of the HttpTileSource. This is more state to think about. The BruTile v6 HttpTileSource is just a configuration of the tile source. No thinking at all.

The pointer to the HttpClient should also be removed from the TileLayer. Also because I don't want the layers to be disposable. I currently see two options:
- Provide a way to do the HTTP calls from the FetchInfo. This is similar to how the new HttpTileSource works, where the HttpClient is in the call.
- Inject a HttpClientFactory into the TileLayer. This is a pattern used a lot in ASP.NET Web API. It is pretty standard nowadays, which is in itself an advantage. In ASP NET there is Dependency Injection (DI), so you inject an IHttpClientFactory. In Mapsui we cannot assume DI is used, so we have to do it a bit differently. Perhaps we should work with a default static HttpClientFactory that could be overridden. The [IHttpClientFactory.CreateClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.ihttpclientfactory.createclient?view=net-9.0-pp&utm_source=chatgpt.com) (which does not actually create a HttpClient most of the time) takes a `string namedHttpClient` so that it can return an instance with a specific configuration.

janusw: > btw. This is not something I want to change in the short term.

Yeah, I understand that this discussion may be a bit too heavy for the current RC stage of BruTile / Mapsui. And I think it would be good for both to come to a stable release rather soon, without changing even more interfaces.

And after all I can live well with the current state of things, after this PR has been merged.

So we can easily just defer this whole discussion ...

PR #3197, opened 2025-10-18
Remove net8.0-ios and net8.0-maccatalyst

pauldendulk:
We are forced to remove net8.0-ios on Mapsui.iOS because the windows agents refuses to build because they are end of life.

### DIFF
--- a/Mapsui.UI.Uno.WinUI/Mapsui.UI.Uno.WinUI.csproj
+++ b/Mapsui.UI.Uno.WinUI/Mapsui.UI.Uno.WinUI.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Uno.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net8.0-ios;net8.0-maccatalyst;net8.0-android;net8.0-windows10.0.19041;net9.0;net9.0-ios;net9.0-maccatalyst;net9.0-android;net9.0-windows10.0.19041</TargetFrameworks>
+    <TargetFrameworks>net8.0;net8.0-android;net8.0-windows10.0.19041;net9.0;net9.0-ios;net9.0-maccatalyst;net9.0-android;net9.0-windows10.0.19041</TargetFrameworks>
     <UnoSingleProject>true</UnoSingleProject>
     <OutputType>Library</OutputType>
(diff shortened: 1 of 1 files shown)


inforithmics:
.net8.0-ios is end of life since spring 2025 current supported version is .net9-ios. For the mobile targets there are different lifetimes than the regular .net versions. Mobile Versions like .net8-android or net8-ios are only supportet until 6 months after the last major release.
.net 8 is supported still until next year but not net8-ios or net8-android

pauldendulk:
@inforithmics Yes, I understand that this is what underlies this build error. This is also something we can use to explain why we removed it. So, not a big problem.

### Pin to a specific agent works better
Yet, I don't like the way this now works on GitHub (or perhaps it is our setup). I also work with GitLab where everything is built with specific Docker images. There are no moving parts there. Once the build works it will always work unless the infrastructure that you build on stops working. When the build does break you know that you have made a mistake and you can be quick to fix it. On GitHub I often assume I made a mistake and after an investigation it turns out to be the agent. Sometimes this takes a lot of time.

### Get informed about needing an update
With the GitLab setup you still have the problem of unsupported dependencies, or worse, dependencies with vulnerabilities, that need to be fixed. This is solved with a dependency tracker that can warn the developer as soon as the vulnerability is found. This is better than that you get informed when you build it again, because the software that you had built a year ago could have this vulnerability. So I think you can have both ways: You are quickly informed of things that need to be upgraded and the build is always building.

### The GitHub policy is not a 'better safe than sorry' policy.
Not being able to build with unsupported package may seem very secure, or proper. But it also means you cannot release updates of software that may already be running on your customers’ machines. The running software may have more serious problems, like an exploitable vulnerability. This is a more serious issue than the fact that some of its components are out of support.

pauldendulk:
@BobbyCannon Thanks for you suggestion but I think this would not help for our scenario. On our windows build agent we do not just build an app for windows but build all of our NuGets. So an iOS developer would use a Mapsui.iOS package build on windows. So, we can not skip building it on windows.

A possible solution for the current situation is if we would build Mapsui.iOS from our Linux agent. This would work today because for some random reason you can still build net8.0-ios on Linux, but maybe not tomorrow, because net8.0-ios is not supported on any platform, so the Linux agent could stop working any day too.

PR #3198, opened 2025-10-19
WIP Update To Uno.Sdk 6.3.28

inforithmics:
Update to Uno.Sdk 6.3.28

- [x] .Net 8 Uno targets removed (are not anymore supported).
- [x] Crashes currently with strange exception on Samples startup (Runtime exception)
- [ ] Not all Samples start maybe uno samples template needs to be updated

Wait for the next uno.sdk if this is fixed


pauldendulk:
In the meantime we upgraded the SDK in this PR: https://github.com/Mapsui/Mapsui/pull/3345

PR #3199, opened 2025-10-19
Add missing Experimental projects to slnfs

inforithmics:
…to slnf

Openin Mapsui.Maui.Slnf I disocvered that Mapsui.Experimental was not in the slnf, then I fixed this for all slnf files

### DIFF
--- a/Mapsui.Android.slnf
+++ b/Mapsui.Android.slnf
@@ -3,6 +3,8 @@
     "path": "Mapsui.slnx",
     "projects": [
       "Mapsui.ArcGIS\\Mapsui.ArcGIS.csproj",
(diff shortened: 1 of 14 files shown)


pauldendulk (approved):
Thanks for taking care of this! 

It is easy to to forget about this. Perhaps we should add it to the GHAs. However, the build is currently already taking too long. Perhaps we should move this to a scheduled daily build, so it would run one time a day at most. I would see the build fail notification. We could also move other stuff there. Like building the samples with the nuget reference instead of the project reference.

I wondered if we need this many slnfs. I personally always work with the main one. But then if you are focusing on one UI framework it will speed up development when you have just this one UI framework in the slnf. Also, if you are an app developer checking out the source code you don't want to have to install all the other UI frameworks to get it to build. So I just convinced myself it is a good thing to have them. 😄

PR #3228, opened 2025-11-10
Add HugeShapeFileWithSimplifierAndClippingSample

pauldendulk:
This is about #2704

I update the [this PR](https://github.com/Mapsui/Mapsui/pull/2879) for v5 and turned it into a new PR, which I will merge after some investigation.


pauldendulk:
We have build server errors introduced by agents changes that are out of our control so I merge it anyway.

PR #3273, opened 2026-01-16
Fix MAUI MapControl memory leak (SKGLView not released after navigation)

### LINKED ISSUE #3271: [MAUI 5.0.2] MapControl is not garbage collected after page navigation (SKGLView memory leak)
### Mapsui Version
- Mapsui.Maui: **5.0.2**
- Mapsui: **5.0.2**

### Platform(s)
- WinUI (MAUI)
- Android (MAUI)
- iOS (MAUI)

### Describe the bug
After navigating to a page containing `Mapsui.Maui.MapControl` and then leaving the page,
the page and its visual tree are correctly garbage collected, but `MapControl` itself and
the underlying `SKGLView` are **not collected**.

This results in a memory leak that accumulates when navigating multiple times to/from a page
containing a `MapControl`.

The issue is reproducible on **WinUI, Android and iOS**.

---

### Steps to reproduce
1. Clone the repro project:  
   https://github.com/MereDoku/Mapsui.Maui.MemoryLeak.Repro
2. Build and run the app (WinUI / Android / iOS)
3. From the first page, click **"Open Map Page"**
4. Navigate back
5. Trigger `VisualLeakCheckQueue.Monitor()` (already wired in the repro)
6. Observe that `MapControl` and `SKGLView` are not garbage collected

---

### Expected behavior
After leaving the page and running GC, `MapControl` and `SKGLView` should be garbage collected.

---

### Actual behavior
`MapControl` and its underlying `SKGLView` remain alive after page navigation, causing
a memory leak.

---

#### WinUI

MapPage => ✅ Collected

    Grid => ✅ Collected

        MapControl => 💦 Leak

            SKGLView => 💦 Leak

#### Android

[0:] MapPage => ✅ Collected
[0:] - Grid => ✅ Collected
[0:] - MapControl => 💦 Leak
[0:] - SKGLView => 💦 Leak

#### iOS

[0:] MapPage => ✅ Collected
[0:] - Grid => ✅ Collected
[0:] - MapControl => 💦 Leak
[0:] - SKGLView => 💦 Leak

---

### Environment
- .NET: MAUI (.NET version used in repro)
- MAUI: (same as repro)
- SkiaSharp / SkiaSharp.Views.Maui: (same as repro)

---

### Additional context
- The repro uses a custom weak-reference based leak detector (`VisualLeakCheckQueue`)
- The leak occurs even with simple navigation and without specific user interaction
- I created a fork and implemented fixes that resolve the leak on all three platforms  
  (WinUI / Android / iOS)
- I can open a Pull Request once we agree on the approach

Please let me know if you would like the repro simplified further (e.g. map without layers).

MereDoku:
Fixes #3271

### Summary
This PR fixes a memory leak in `Mapsui.Maui.MapControl` where `MapControl` and the underlying
`SKGLView` were not garbage collected after navigating away from a page.

### Repro
https://github.com/MereDoku/Mapsui.Maui.MemoryLeak.Repro

### Changes
- Disconnect MAUI handlers / clear native view references to allow GC
- Stop `MyLocationLayer` animation on cleanup
- Ensure `MapView` no longer keeps strong references after view is removed

### How to test
1. Run the repro project
2. Navigate to MapPage and back multiple times
3. Run `VisualLeakCheckQueue.Monitor()`
4. Verify `MapControl` and `SKGLView` are collected

### Platforms tested
- WinUI ✅
- Android ✅
- iOS ✅


pauldendulk (approved):
Thanks for the PR — the problem is real and the fixes are correct.

**What's good:**
- `RenderController.Dispose()` now unblocks the render loop (the `.Set()` calls are a genuine bug fix)
- `DataFetcher` getting `IDisposable` with cancellation is the right approach
- Android `FixInvisible()` infinite loop getting a `CancellationToken` is a clear improvement
- MAUI handler disconnection via `DisconnectHandler()` is the correct MAUI pattern

**Minor notes:**
- `OpenInBrowserStatic` is added in the constructor but `SharedConstructor()` already calls `SetOpenInBrowserFunc` — results in a redundant double call
- `TryStopMyLocationAnimations()` — the "Try" prefix implies error tolerance that isn't there; `StopMyLocationAnimations()` would be more accurate
- `CleanupMapViewResources()` calls `Map.Layers.Clear()` before `base.Dispose()` also clears layers — harmless but redundant

**Architectural note:**
The disposal code is now correct, but users still need to call `Dispose()` explicitly on navigation. A follow-up worth considering: overriding `OnHandlerChanged()` on the MAUI `MapControl` to call `Dispose()` automatically when `Handler == null` — that's the idiomatic MAUI lifecycle pattern and would make this fully transparent to users. This is something for a future PR.

pauldendulk:
@MereDoku Sorry this review took so long!

Issue #3281, opened 2026-01-23
HTTP Headers for WFS / WMS requests


MLipfert1:
**The problem**
For our application, we need to pass additional HTTP headers to the WFS and WMS web requests. Right now, the HTTPClient is encapuslated inside the WFS and WMS providers, with the ability to pass the headers to it.

**The proposed solution**
I would propose to add additional properties to the providers to hand these HTTP headers over to the HTTP client instances. I have prepared a commit here: https://github.com/MLipfert1/Mapsui/commit/284e538696a2a23a5551cf34ec27753269d17da6

If you are fine with my proposal, I would like to create the pull request.

pauldendulk:
@gpsaliola 

> This would be very helpful because some maps need to indicate User Agent and Referer HTTP headers.

Was there no option for this at all in the current version?

PR #3282, opened 2026-01-26
Add properties for passing additional HTTP headers to requests.

MLipfert1:
This commit adds the ability to add custom HTTP Headers to WFS and WMS web requests (e.g. for authorization).

I tested this implementation via Mapsui WPF.

This PR is related to this Feature Request: https://github.com/Mapsui/Mapsui/issues/3281

### DIFF
--- a/Mapsui.Nts/Providers/Wfs/Utilities/HttpClientUtil.cs
+++ b/Mapsui.Nts/Providers/Wfs/Utilities/HttpClientUtil.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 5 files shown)


MLipfert1:
Hi @pauldendulk and thanks for your review.

The distinction between permanent and temporary headers in HttpClientUtil is added, because the method "ConfigureForWfsGetFeatureRequest" in the class WfsProvider adds a temporary header for post requests, which might be removed by the reset method, when the HttpClientUtil instance gets re-used for get requests. The manually added HTTP headers of course should be permanent. As you already mentioned, I have done it this way, to not change any existing behavior in an unintended way.

PR #3295, opened 2026-02-09
☢️ Improve vector tile as PNG performance

pauldendulk:
This introduces two step rendering initially used only for SymbolStyle. The core change is that the SymbolStyleRenderer is replaced by a SymbolStyleDrawableRenderer which implement a new IDrawableStyleRenderer interface. It has two methods, one to create the drawable and one to draw the drawable.

```cs
IReadOnlyList<IDrawable> CreateDrawables(Viewport viewport, ILayer layer, IFeature feature, IStyle style, RenderService renderService);
void DrawDrawable(object canvas, Viewport viewport, IDrawable drawable, ILayer layer);
```

For this initial version of two-step rendering many design choices are made to make it fit in the existing design without too many changes. Bigger changes could be made later on.

## Two-Step Rendering Design

This PR introduces a two-step drawable architecture for the experimental renderer that separates heavy Skia object creation from the fast draw step:

**Step 1 – `CreateDrawables` (background thread):** When layer data changes, `DrawableRenderer.UpdateDrawables` iterates features and calls each `IDrawableStyleRenderer.CreateDrawables`. This builds lightweight drawable objects containing pre-prepared data (world-coordinate `SKPath`s, style parameters, or pre-rendered tile images). This runs on the thread that raised `DataChanged` — typically a thread-pool thread after a data fetch completes.

**Step 2 – `DrawDrawable` (UI thread):** During rendering, `MapRenderer.RenderFeature` retrieves cached drawables and calls `DrawDrawable`, which only applies a viewport transform and draws. For vector geometry this uses `canvas.Concat(matrix)` to let the GPU transform world-coordinate paths. For vector tiles it's a simple image blit. No heavy computation happens here.

### Currently implemented renderers

- **`VectorStyleDrawableRenderer`** – Polygons and linestrings store an `SKPath` in world coordinates. At draw time a world-to-screen affine matrix is concatenated onto the canvas, and `SKPaint` is created locally from stored style parameters (Brush/Pen/opacity) to avoid native object lifecycle issues across threads.
- **`SymbolStyleDrawableRenderer`** – Point symbols store an `SKPath` plus paint objects, positioned via world-to-screen transform at draw time.

### Current design decisions

- **So far no cached `SKPaint` in drawables** (for VectorStyle): Using matrix scale of the SKPath is much faster than modifying the vertices. When you scale the matrix you need to counter scale the width of the outline path. This turns out to be a problem. If we store the `SKPaint` paint and mutate the line width this cause an access violation because of our multi threading design. Currently we create a new SKPaint, there may be better ways to deal with this. 
- **Unified style renderer dictionary**: `Dictionary<Type, IStyleRenderer>` holds both `IDrawableStyleRenderer` and legacy `ISkiaStyleRenderer` implementations; `RenderFeature` dispatches based on type.


Mapsui.Experimental.Rendering.Skia/DrawableRenderers/VectorStyleDrawableRenderer.cs
```diff
@@ -0,0 +1,396 @@
...
+                outlinePaint.StrokeWidth *= res;
+                canvas.DrawPath(drawable.WorldPath, outlinePaint);
+            }
+
+            // Draw line (linestring only — drawn on top of outline)
+            if (drawable.LinePen is not null)
+            {
+                using var linePaint = CreateStrokePaint(drawable.LinePen, null, drawable.LineOpacity);
+                linePaint.StrokeWidth *= res;
+                canvas.DrawPath(drawable.WorldPath, linePaint);
```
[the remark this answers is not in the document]
pauldendulk: Yes, I think this is a bug.

Mapsui.Experimental.Rendering.Skia/MapRenderer.cs
```diff
@@ -303,11 +304,26 @@ private void Render(SKCanvas canvas, Viewport viewport, IEnumerable<ILayer> laye
...
+            // 1. Layers that already have features but missed the DataChanged event
+            //    (e.g. MemoryLayer features set before the layer was added to the map).
+            // 2. Features that entered the viewport due to panning/zooming but are
+            //    already in the layer's memory cache, so no DataChanged fires
+            //    (e.g. coarser tile levels when zooming out).
+            foreach (var layer in layers)
+            {
+                if (layer.Enabled)
+                {
+                    DrawableRenderer.UpdateDrawables(viewport, layer, _styleRenderers, renderService);
```
charlenni: If I understand it correct, then this is the point, where Mapsui checks, if the Drawable is ok. But what happens, if the Drawable needs more time to update? Think about a symbol layer, which determines the symbols, that should be drawn, on the fly? It takes some time to update the layer. Should it in this case start a task in the background?

You could see this behavior in Google Maps when you zoom in. The streets are drawn immediately with the old symbols and a short time later this old symbols fade out and the new ones fade in. So the Drawables for this layer change in the background.

Mapsui/Rendering/IDrawable.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace Mapsui.Rendering;
+
+/// <summary>
+/// Represents a pre-created rendering object that can be drawn quickly on the render thread.
+/// Drawables store world coordinates and pre-built platform-specific objects (e.g. SKPath, SKPaint)
+/// so that the draw step on the UI thread is minimal.
+/// </summary>
+public interface IDrawable : IDisposable
```
charlenni: If I understand it correct, then all drawables are stored in world coordinates an later drawn with a Skia transformation. If correct, then you would get in trouble with lines and polygons, because Skia only works with floats and the transformation has rounding problems.

PR #3313, opened 2026-03-22
Sparse mbtile fix

inforithmics:
I use mbtiles that are sparse meaning for certain areas only higher level resolutions are provided.

Without this fix the tile Layer returns a feature on lower resolution without images causing to render an empty tile. 
With this fix the high resolution tile is displayed.

An alternative implementation would be to ignore Raster images with empty Image Data, and not to render tehme.


pauldendulk:
I am going to close this one to replace it with this other one because to avoid repeated tile requests to non-existing tiles: https://github.com/Mapsui/Mapsui/pull/3320

pauldendulk:
@inforithmics Release 5.1.0-beta.5 should contain the fix, please tell me if this solved your problem.

PR #3324, opened 2026-04-03
Create Claude.md

inforithmics:
Created Claude.md file with claude /init .
This file is autogenerated and helps the Claude Agent to generate Code. 
If its not wanted this pull request can be closed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,106 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Common Commands
+
+### Build
+```bash
+# Restore packages and build the entire solution in Debug configuration
+# (the Release configuration is identical except for the --configuration flag)
+#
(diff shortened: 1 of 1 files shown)


pauldendulk:
Thanks, this is welcome. I actually wanted to start using Claude as well, not because I am not happy with Copilot but I want to know Claude as well. I did not find the time for it yet. I had already been looking into sharing the copilot instructions.md and Claude.md but there is no clear standard for that yet. So, let's duplicate it for now. 

About the use of AI. Of course it is tricky, and I have already pushed some flawed code because I was not careful enough. Still there is no way around it. I feel I get more control over it with more experience. I try to update the instructions regularly, add skills when I think it could help, and I use regular documentation pages a lot as a focal point of the conversation with the AI. To get a better overview myself and to summarize things, and to use it discuss changes in the code again (did this not that much in Mapsui but in other projects).

pauldendulk (approved):

Issue #3255, opened 2025-12-04
Map doesn't display on iOS with .NET 10


nilobaz:
**Mapsui Version**
5.0.1

**Mapsui Platform**
Mapsui.MAUI

**Device**
iOS iPhone (simulator and physical device)

**Describe the bug**
Using the latest version of Mapsui.MAUI, worked for .NET 9, but when upgrading my POC project to .NET 10, iOS has stopped showing the map. It only displays a blank screen area, no tiles, no pins, nothing is being rendered on the canvas. Android and Windows are working fine.

**To Reproduce**
Create a MapView with a Map, configure and add the TileLayer, add some pins to the map, and put it into a simple XAML page.

**Expected behavior**
Expected the MapView Map to display the tiles, pins, etc., accordingly.

**Screenshots**
None.

**Additional context**
Some error logs I got from the debug console window. I'm not sure if this still applies, but it seems similar to this bug #2855 
````
2025-12-04 18:00:52.520025-0300 MapPOC[2726:29846] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x600001767a40> to GL_RENDERBUFFER 1
2025-12-04 18:00:52.520110-0300 MapPOC[2726:29846] Failed to make complete multisample framebuffer object 8cd6
2025-12-04 18:00:52.520177-0300 MapPOC[2726:29846] Failed to make complete framebuffer object 8cd6
2025-12-04 18:00:52.526591-0300 MapPOC[2726:29846] Failed to bind EAGLDrawable: <CAEAGLLayer: 0x600001767a40> to GL_RENDERBUFFER 1
2025-12-04 18:00:52.526659-0300 MapPOC[2726:29846] Failed to make complete multisample framebuffer object 8cd6
2025-12-04 18:00:52.526720-0300 MapPOC[2726:29846] Failed to make complete framebuffer object 8cd6
Navigator: SetSize 390 x 643,3333129882812
Navigator: Zooming to PanBounds at startup.
````

pauldendulk:
I created a bug report at SkiaSharp: https://github.com/mono/SkiaSharp/issues/3428

smathersgit:
[...]

Issue #2855, opened 2025-01-23
Map doesn't display on IOS emulator


smathersgit:
**Mapsui Version**
5.0.0-beta.7

**Mapsui Platform**
IOS

**Device**
IOS iPhone 16 Plus iOS 18.2 emulator on MacBook Air.

**Describe the bug**
The map, and pin, are are not displayed, with the screen area remaining white.

**To Reproduce**
At this time I shall assume that this can be repeated using any sample app build with any IOS emulator.  
Using an app build with 5.0.0-beta.6 confirm map displays on IOS emulator.
Using an app build with 5.0.0-beta.7 confirm map no longer displays on IOS emulator.
This 5.0.0.-beta.7 issue is not prevalent on Android (API 33, 34 & 35 emulators and Samsung Galaxy 7) 

**Expected behaviour**
Map displays successfully on IOS as per its operation with 5.0.0-beta.6, and Android devices.

inforithmics:
Which CPU architecture does the MacBook air have arm  or intel

inforithmics:
https://github.com/Mapsui/Mapsui/pull/2854
Here on iOS metal is used because opengl is deprecated I tested it on an iOS simulator on an Mac mini m1 and it worked. So maybe this will be fixed in the next Mapsui Version

smathersgit:
The MacBook Air uses the ARM based M3.

pauldendulk:
Wait for the next release and test again.

smathersgit:
Re-tested with Mapsui 5.0.0-beta.8 and the issue still exists with the same symptoms.

evgenyvalavin:
Mapsui 5.0.0-beta.8 works fine on iPhone 16 simulator iOS 18.2 on my Mac mini M4

smathersgit:
Performed a further sanity test. Rolled back to 5.0.0-beta.6 and confirmed it worked fine on the iPhone 16 simulator iOS 18.2 (on MacBook Air - ARM M3), and didn't work when upgrading back to 5.0.0-beta.8.

evgenyvalavin:
> Performed a further sanity test. Rolled back to 5.0.0-beta.6 and confirmed it worked fine on the iPhone 16 simulator iOS 18.2 (on MacBook Air - ARM M3), and didn't work when upgrading back to 5.0.0-beta.8.

Do you use MAUI or native iOS project?

smathersgit:
[...]

PR #3349, opened 2026-05-11
fix: unblock invalidation loop when paint fires before view is laid out

pauldendulk:
**iOS blank map — root cause diagnosis and fix**

**The bug:** On iOS (and potentially other platforms), the map would render blank at startup. Reported in issues #3255 and #2855, and more reproducible on .NET 10 which initializes faster.

**Root cause:** `RenderController.InvalidateLoopAsync` gates each render cycle on two signals: `_needsRefresh` (something changed) and `_isDrawingDone` (the last paint completed). `_isDrawingDone` was only set inside `Render()`. Platform paint callbacks had early-return guards — if `GetPixelDensity()` returned null (view not yet laid out) or `GRContext` was null (GL context not ready), they returned before ever calling `Render()`. This permanently stalled the loop: `_invalidateCanvas()` would fire, the platform would call the paint handler, the handler would return early, and the loop would wait forever for a signal that never came.

**Fix:** Moved the pixel density parameter into `Render()` itself and wrapped the entire body in `try/finally { _isDrawingDone.Set(); }`. Platform callbacks no longer return early — they always call `Render()`, passing `null` for density when the view isn't ready. `Render()` skips actual painting in that case but still unblocks the loop. Applied across iOS, Android, MAUI, WPF, WindowsForms, and WinUI/Uno. The old `Render(object canvas)` one-argument overload was kept as a backward-compatible delegate (`=> Render(canvas, 1f)`) for Avalonia and WinUI which manage DPI scaling differently.

**Validation:** 265 unit tests passed. 20 regression test failures were all pixel-comparison mismatches pre-existing from the SkiaSharp upgrade — none related to the `RenderController` change.

### DIFF
--- a/Mapsui.UI.Android/MapControl.cs
+++ b/Mapsui.UI.Android/MapControl.cs
@@ -57,16 +57,7 @@ public partial class MapControl : ViewGroup, IMapControl
     }
 
     private void CanvasOnPaintSurface(object? sender, SKPaintSurfaceEventArgs args)
-    {
-        if (GetPixelDensity() is not float pixelDensity)
-            return;
-
-        var canvas = args.Surface.Canvas;
-
-        canvas.Scale(pixelDensity, pixelDensity);
-
-        _renderController?.Render(canvas);
-    }
+        => _renderController?.Render(args.Surface.Canvas, GetPixelDensity());
 
     public SkiaRenderMode RenderMode
     {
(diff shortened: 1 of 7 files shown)


smathersgit:
Ported the changes over and can't get it to fail on IOS simulator (and Android Emulator)! I'll await your official version before posting on the relevant bugs!

smathersgit:
Reloaded Mapsui main into my Maui .NET 10 based Android/IOS app build and retested. In my Scottish Football app's case, this was 10 displays of Aberdeen's Pittodrie ground location map using static map tiles.

 - Android Emulator API 33 - PASS

Visual Studio controlled remote IOS simulator testing on MacBook Air M3 using macOS Tahoe 26.4.1 and Xcode 26.3.

 - iPhone 16 iOS 18.2  - PASS
 - iPhone 16 iOS 18.4 - PASS
 - iPhone 16 Pro iOS 18.4 - PASS

Have now closed my original https://github.com/Mapsui/Mapsui/issues/2855 issue as fixed.

PR #3369, opened 2026-05-31
feat: add VisibilityMargin to BaseLayer

pauldendulk:
## What

Adds a `VisibilityMargin` property to `BaseLayer` that controls how far (in pixels) the query extent is expanded beyond the viewport when fetching and culling features.

## Why

Several `GetFeatures` and `FetchAsync` implementations were hardcoding `SymbolStyle.DefaultWidth * 2 * resolution` as a growth margin. This was not configurable, not discoverable, and asymmetric (width and height were treated separately even though the values were always identical). Layers using large symbols, wide strokes, or labels had no way to increase this margin without subclassing.

## Changes

- `BaseLayer` — new `VisibilityMargin` auto-property (default: `SymbolStyle.DefaultWidth * 2` = 64 px) with XML doc comment
- `MemoryLayer`, `WritableLayer`, `GenericCollectionLayer` — `GetFeatures` uses `VisibilityMargin * resolution` instead of the hardcoded expression
- `ObservableCollectionLayer` (Experimental) — same fix
- `Layer.FetchAsync` — replaces `fetchInfo.Grow(SymbolStyle.DefaultWidth)` (which had a hidden x2 multiplier) with an explicit pixel to world expansion using `VisibilityMargin * fetchInfo.Resolution`, keeping behaviour identical but making the semantics consistent
- `using Mapsui.Styles;` removed from all 5 changed layer files (was only used for the now-removed `SymbolStyle` reference)
- `docs/general/markdown/working-with-layers.md` — `VisibilityMargin` documented in the 'Layer properties that affect drawing' section

## Behaviour

Default behaviour is **unchanged**: all locations still produce 64 px of margin per edge. Users can now increase the margin per layer when needed:

`csharp
var layer = new MemoryLayer { Features = myFeatures };
layer.VisibilityMargin = 128; // e.g. for a layer using large custom symbols
`

## Potential breaking changes

- A custom `BaseLayer` subclass that already defines its own `VisibilityMargin` property will silently shadow the base class property (CS0108).
- The grow was previously asymmetric (separate width/height). It is now uniform. Anyone who had set `SymbolStyle.DefaultWidth` and `SymbolStyle.DefaultHeight` to different values will see a behaviour change.

### DIFF
--- a/Mapsui.Experimental/Layers/ObservableCollectionLayer.cs
+++ b/Mapsui.Experimental/Layers/ObservableCollectionLayer.cs
@@ -3,7 +3,6 @@ using System.Collections.ObjectModel;
 using System.Collections.Specialized;
 using System.Linq;
 using Mapsui.Extensions;
-using Mapsui.Styles;
 using Mapsui.Utilities;
 using System.Collections.Generic;
 using Mapsui.Layers;
@@ -135,9 +134,7 @@ public class ObservableCollectionLayer<T> : BaseLayer
         if (rect == null)
             yield break;
 
-        var biggerRect = rect.Grow(
-                SymbolStyle.DefaultWidth * 2 * resolution,
-                SymbolStyle.DefaultHeight * 2 * resolution);
+        var biggerRect = rect.Grow(VisibilityMargin * resolution);
         foreach (var feature in _shadowCollection.Select(i => i.Feature).ToArray())
         {
             if (feature?.Extent?.Intersects(biggerRect) == true)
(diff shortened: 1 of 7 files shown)


charlenni:
Perhaps it would be helpful to have the possibility to ask the renderer the size of the ready rendered feature. Or better size and position.

pauldendulk:
> Perhaps it would be helpful to have the possibility to ask the renderer the size of the ready rendered feature. Or better size and position.

This would not solve the fetching problem. If you fetch data you need to get a bigger extent, and how much bigger depends on how big the style used for that feature is. And that again could depend on the type of feature. So the source data influences the maximum VisibilityMargin you need. So measuring the rendered feature is too late.

PR #3371, opened 2026-06-13
fix: guard EndEdit against polygons with fewer than three points

### LINKED ISSUE #3370: EditManager.EndEdit() throws unhandled ArgumentException when finishing AddPolygon with fewer than 3 points
**Mapsui Version**
5.0.2

**Mapsui Platform**
Avalonia

**Device**
Windows 11 Enterprise

**Describe the bug**

When EditMode is set to AddPolygon and the user double-clicks to finish the polygon with fewer than 3 points placed, EditManager.EndEdit() throws an unhandled System.ArgumentException (Number of points must be 0 or >=3) originating from NetTopologySuite.Geometries.LinearRing. This crashes the application entirely because there is no event or hook on EditingWidget / EditManager to catch or suppress such exceptions before they propagate.

  To Reproduce
  1. Create a Map with an EditingWidget and EditManager
  2. Set EditManager.EditMode = EditMode.AddPolygon
  3. Click once or twice on the map (add fewer than 3 points)
  4. Double-click to finish drawing

**Expected behavior**

EndEdit() should guard against invalid geometry before constructing the LinearRing. If the point count is insufficient (< 3), the edit should either be silently cancelled or the in-progress geometry should be discarded -
  without throwing an unhandled exception. 

Alternatively, an OnEditError / EditFailed event on EditManager or EditingWidget would allow the caller to handle the case gracefully.

  Stack trace
  Unhandled exception. System.ArgumentException: Number of points must be 0 or >=3
     at NetTopologySuite.Geometries.LinearRing.ValidateConstruction()
     at NetTopologySuite.Geometries.LinearRing..ctor(Coordinate[] points)
     at Mapsui.Nts.Editing.EditManager.EndEdit()
     at Mapsui.Nts.Editing.EditManipulation.OnTapped(WidgetEventArgs e, EditManager editManager)
     at Mapsui.Nts.Widgets.EditingWidget.OnTapped(WidgetEventArgs e)
     at Mapsui.UI.Avalonia.MapControl.OnWidgetTapped(ScreenPosition screenPosition, MPoint worldPosition, GestureType gestureType, Boolean shiftPressed)
     at Mapsui.UI.Avalonia.MapControl.OnPointerReleased(ReadOnlySpan`1 screenPositions)
     at Mapsui.UI.Avalonia.MapControl.MapControl_PointerReleased(Object sender, PointerReleasedEventArgs e)

SAY-5:
## What

`EditManager.EndEdit()` now discards the in-progress polygon and resets the edit instead of constructing a `LinearRing` when fewer than three points were placed.

## Why

Finishing an `AddPolygon` edit with one or two points (for example double-clicking right after the first click) reached `new LinearRing(...)` with too few coordinates, which throws `System.ArgumentException: Number of points must be 0 or >=3` from NetTopologySuite and crashed the application. The `DrawingLine` branch already guards against this with a `Count < 2` check, so this mirrors that behaviour for polygons. Fixes #3370.

### DIFF
--- a/Mapsui.Nts/Editing/EditManager.cs
+++ b/Mapsui.Nts/Editing/EditManager.cs
@@ -64,6 +64,16 @@ public class EditManager
             if (polygon == null) return false;
 
             _addInfo.Vertices.RemoveAt(_addInfo.Vertices.Count - 1); // Remove the last vertex, because it is the hover vertex
+
+            if (_addInfo.Vertices.Count < 3) // A polygon ring needs at least 3 distinct points, otherwise LinearRing throws.
+            {
+                // And reset it back to the previous state.
+                Layer?.TryRemove(_addInfo.Feature);
+                _addInfo.Reset();
+                EditMode = EditMode.AddPolygon;
+                return false;
+            }
+
             var linearRing = _addInfo.Vertices.ToList();
             linearRing.Add(linearRing[0].Copy()); // Add first coordinate at end to close the ring.
             _addInfo.Feature.Geometry = new Polygon(new LinearRing(linearRing.ToArray()));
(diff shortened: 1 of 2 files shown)


pauldendulk (approved):
Great, thanks! I even see some additional approvements. I will make them in a follow up PR.

Issue #3376, opened 2026-07-01
Upgrade SkiaSharp to 4.X


Cairn23:
**The problem**
SkiaSharp has updated the version and the frequency of updates to match Google and Skia to approx. fortnightly.
The latest version of SkiaSharp is 4.148.
There is a large performance update with version 4, and this will improve once Graphite work has completed.

**The proposed solution**
Update the nugget packages to the latest versions of SkiaSharp. 

**Alternative solutions**

**Additional context**

Sebastian1989101:
@pauldendulk is there any ETA for the SkiaSharp 4.x upgrade? Because we have another component in our app that requires SkiaSharp 4.x by now while Mapsui still binds to 3.x. So far it seems to just work if we directly reference the latest SkiaSharp (currently in dev branch) but it always feels a lot safer if all libraries just target at least the same major version.

follesoe:
> **Updated 2026-07-28.** My original comment concluded that Mapsui could adopt SkiaSharp 4 now, with the Avalonia projects staying on 3.x behind a `VersionOverride`. **That conclusion was wrong**, and I've rewritten this comment rather than leave it standing. Two things I had missed: no Avalonia release supports SkiaSharp 4 at all, and the per-project pin is a runtime trap rather than a safe carve-out. Details below. The original also didn't mention that net8.0 support ends at **SkiaSharp 3.119.4**, which turns out to be the most practically important constraint here.
>
> **Further updated 2026-07-29:** Svg.Skia's SkiaSharp 4 support turned out to already be merged ([#544](https://github.com/wieslawsoltes/Svg.Skia/pull/544)), just unreleased — corrected below. That PR also notes SkiaSharp 4 *removes* the `SKPaint` text surface (`Typeface`, `TextSize`, `MeasureText`, `GetTextPath`, …) on net6.0+ rather than merely obsoleting it. Mapsui is unaffected: `LabelStyleRenderer` already calls those on `SKFont`, not `SKPaint`.

I spent some time on this on a branch. Summary: **the SkiaSharp 4 API surface is not the problem — Avalonia is.**

## The blocker: no Avalonia release supports SkiaSharp 4

| Avalonia.Skia | SkiaSharp |
|---|---|
| 11.0.13 | 2.88.7 |
| 11.3.18 (latest 11.x) | 2.88.9 |
| 12.0.5 | 3.119.4 |
| 12.1.0 (latest) | 3.119.4 |

[...]

PR #3380, opened 2026-07-29
update: Move to SkiaSharp 3.119.4 and prepare for SkiaSharp 4

follesoe:
Relates to #3376.

## Approach

Mapsui can't adopt SkiaSharp 4 yet: **no Avalonia release supports it**. The 11.x line is on SkiaSharp 2.88.x, the whole 12.x line on 3.119.4, and [AvaloniaUI/Avalonia#21696](https://github.com/AvaloniaUI/Avalonia/issues/21696) is open, unassigned and unanswered.

So rather than upgrade, this makes Mapsui **loadable on either major**. `Mapsui.Rendering.Skia` is built against SkiaSharp 3.119.4 using only APIs present in both 3.x and 4.x, so it binds to `SkiaSharp 3.119.0.0` and .NET rolls that reference forward when a consumer supplies SkiaSharp 4. Anyone needing SkiaSharp 4 for another component can reference it directly today, and the eventual v4 adoption becomes a mechanical follow-up once Avalonia lands.

Concretely that means keeping the `SKPath` mutators rather than moving to `SKPathBuilder`, which exists in no 3.x release. Two changes are safe now because they exist in 3.x and are the non-obsolete form in 4.x: the `DrawImage` overloads taking `SKSamplingOptions`, and `SKFont.CountGlyphs` instead of `SKTypeface.CountGlyphs`.

**Verified, not assumed:** SkiaSharp 4.150.1 — managed *and* native — swapped into the built test output and the suites re-run. All rendering and core tests pass with the renderer still compiled against 3.x.

## Changes

- SkiaSharp `3.119.2` → `3.119.4`; HarfBuzzSharp `8.3.1.3` → `8.3.1.5` (required by `SkiaSharp.HarfBuzz 3.119.4`)
- Avalonia 12 off its prerelease Skia dependency: `12.0.2` → `12.1.0`, which wants stable `3.119.4` rather than `3.119.4-preview.1.1`
- Avalonia 11 `11.3.1` → `11.3.9` — not `11.3.18`, because `Avalonia.ReactiveUI` stops at `11.3.9`; this keeps all Avalonia 11 packages in lockstep
- **net8.0 dropped from the UI heads.** Not a v4 requirement — `SkiaSharp.Views.*` drops its net8.0 assets at **3.119.4**, so any move past 3.119.2 forces it. `Mapsui.UI.Maui8` is net8.0-only and stays on `3.119.2` via `$(Maui8SkiaSharpVersion)`; both share assembly version `3.119.0.0`, so the shared renderer loads against either
- `TreatWarningsAsErrors` for `Mapsui.Experimental.VectorTiles`, with a folder-scoped `.editorconfig` exempting the vendored `VexTileCopies` files. Fixing the nullable warnings in `VexTileRenderer.cs` included **two behaviour changes**: tile layers with a null `Name` are skipped rather than throwing, and null attribute values are no longer inserted into a non-nullable dictionary
- `RasterizingLayerTests` asserted an exact encoded PNG byte length, which shifts whenever the bundled Skia encoder does; it now decodes the raster and asserts dimensions and non-empty content
- Removes the dead `SkiaSharp.Views.Uno` `PackageVersion` (unreferenced; no release above `2.88.9`)
- CI: pack steps fall back to a placeholder version when no tags are reachable, so forks aren't reported as failing purely because of packaging

## Verification

Green on all three legs. The Windows leg runs the full rendering regression suite — **154 tests, 0 failures** — which Linux and macOS silently skip, since those fixtures are `[Apartment(ApartmentState.STA)]`.

### DIFF
--- a/Mapsui.Experimental.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
+++ b/Mapsui.Experimental.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
@@ -73,7 +73,7 @@ public class SymbolStyleRenderer : ISkiaStyleRenderer, IFeatureSize
                 TrianglePath(skPath, 0, 0, width);
                 break;
             default: // Invalid value
-                throw new ArgumentException($"Unknown {nameof(SymbolType)} '{nameof(symbolType)}'");
+                throw new ArgumentException($"Unknown {nameof(SymbolType)} '{symbolType}'");
         }
 
         return skPath;
--- a/.github/workflows/dotnet-validate-nugets.yml
+++ b/.github/workflows/dotnet-validate-nugets.yml
@@ -46,15 +46,26 @@ jobs:
     - name: Restore dependencies (dotnet)
       run: dotnet restore Mapsui.slnx  
     # Release Build
+    # Forks have no release tags, so `git describe --tags` yields an empty version and the pack step
+    # fails with MSB4044. Fall back to a placeholder so this is runnable on a fork; upstream, where
+    # the tags exist, it resolves to the real version exactly as before. This job runs on Windows,
+    # so the step is PowerShell rather than the bash form used in dotnet.yml.
+    - name: Determine package version
+      shell: pwsh
+      run: |
+        $version = git describe --tags 2>$null
+        if ([string]::IsNullOrWhiteSpace($version)) { $version = '0.0.1-fork' }
+        "PACKAGE_VERSION=$version" | Out-File -FilePath $env:GITHUB_ENV -Append
+        $LASTEXITCODE = 0
     - name: Build nuget packages
-      run: dotnet pack Mapsui.slnx --configuration Release /p:Version=$(git describe --tags) -o Artifacts
+      run: dotnet pack Mapsui.slnx --configuration Release /p:Version=$env:PACKAGE_VERSION -o Artifacts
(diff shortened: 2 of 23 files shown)


Mapsui.Experimental.Rendering.Skia/SkiaStyles/SymbolStyleRenderer.cs
```diff
@@ -73,7 +73,7 @@ internal static SKPath CreatePath(SymbolType symbolType)
                 TrianglePath(skPath, 0, 0, width);
                 break;
             default: // Invalid value
-                throw new ArgumentException($"Unknown {nameof(SymbolType)} '{nameof(symbolType)}'");
+                throw new ArgumentException($"Unknown {nameof(SymbolType)} '{symbolType}'");
```
inforithmics, line 76: good catch

inforithmics:
In general good improvements. But because .Net Targets are removed maybe the major version needs to be incremented to 6.0. But @pauldendulk has to decide this.

pauldendulk (approved):
Thanks, Jonas. This is an complicated issue. The useful compatibility bridge you created is a good solution: Mapsui stays on SkiaSharp 3.119.4, while MAUI applications can use SkiaSharp 4. I’ll merge this and handle newer dependency updates in follow-up PRs.

